PR #6, opened 2019-04-02
Update vendor

djwhatle:


djwhatle:
@jmontleon and @jortel were able to build this on their machines, so think it's safe to merge.

jortel:
Agreed.

PR #8, opened 2019-04-02
Add more validations.

jortel:
Another round of validation work.

- Moved `MigPlan` update predicate to its own module and applied this pattern to the `MigStorage` and `MigAssetCollection` controllers.
- Defined `message` constants and added to _raised_ conditions.  Having the message seems helpful users and likely to be helpful for UI.
- Adds validations to `MigAssetCollection` controller.


pkg/controller/migassetcollection/predicate.go
```diff
@@ -0,0 +1,25 @@
...
+	"reflect"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+type UpdatedPredicate struct {
+	predicate.Funcs
+}
+
+func (r UpdatedPredicate) Update(e event.UpdateEvent) bool {
```
djwhatle: +1 on sharing this across controllers

pkg/controller/migplan/validation.go
```diff
@@ -33,17 +33,44 @@ const (
 	False = "False"
 )
 
+// Message
+const (
+	ReadyMessage                        = "The migration plan is ready."
+	InvalidSourceClusterRefMessage      = "A valid source cluster resource has not been associated with this plan."
```
djwhatle: Are these intended for the UI? I'm wondering, from a CLI user POV, if a message explicitly mentioning the field needing attention (sourceClusterRef in this case) would be easier to rectify.

djwhatle: Could read something like "A valid source cluster reference (srcClusterRef) has not been provided with this plan."

jortel: It's mainly intended for a human that is editing yaml but thought it _may_ be used by the UI.  I considered including field name causing the problem but was a little concerned about keeping it in sync with the field _tags_ (yaml).  Also, if used by the UI, the field names would have little meaning.  By not including the field name but describing the data instead, it correlates more loosely to the yaml fileds which is less helpful but also won't get out of sync and may be more broadly useful.
It's a trade.  Not sure what the right approach is.

jortel: One possible way to mitigate the problem of keeping the field name in the API (yaml) and the field name in the condition message the same is to use _reflect_ to get the struct field tag, parse it to get the name and use it in the message.

jortel: Using reflect may not be a good idea either.  I looked into using reflect and the code would need to pass the field name (string) which could also get out of sync.

jortel: I reworded the `message` to include the json field name.  I agree this is better.

PR #10, opened 2019-04-09
Add another round of validations.

jortel:
Add another round of validations.
Also add `Mig` prefix to Cluster refs in a few places that got missed.
Add watch on referenced resources.  This is a little crude and my be enhanced by handler using the shared map.


pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -23,8 +23,8 @@ import (
 
 // MigPlanSpec defines the desired state of MigPlan
 type MigPlanSpec struct {
-	SrcClusterRef         *kapi.ObjectReference `json:"srcClusterRef,omitempty"`
-	DestClusterRef        *kapi.ObjectReference `json:"destClusterRef,omitempty"`
+	MigSrcClusterRef      *kapi.ObjectReference `json:"migSrcClusterRef,omitempty"`
```
djwhatle: Guessing this is here to disambiguate between a reference to a cluster-registry `Cluster` and a `MigCluster`, but I think we'll want to keep `MigCluster` in there as one word, otherwise it looks like we're just prepending `Mig` to everything in our spec (_in my opinion_), and ambiguity remains of whether we are referring to a `Cluster` or `MigCluster`.

Ideas
 - `SrcMigClusterRef`
 - `MigClusterSrcRef`

pkg/controller/migplan/migplan_controller.go
```diff
@@ -66,6 +66,33 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to MigCluster
+	err = c.Watch(&source.Kind{
```
djwhatle: We already spoke about this, but this kind of `Watch` will generate events for `MigCluster` nsNames rather than `MigPlan` nsNames that this controller is able to reconcile.

See https://book.kubebuilder.io/beyond_basics/controller_watches.html

pkg/apis/migration/v1alpha1/migplan_types.go
jortel: Oh, yeah, completely agree.  I was not paying close enough attention when I changed that.  I think `SrcMigClusterRef` etc would be most consistent.

pkg/controller/migplan/migplan_controller.go
jortel: Agreed.  This was replaced when rebased.

pkg/apis/migration/v1alpha1/migplan_types.go
jortel: Fixed

pkg/controller/migstorage/validation.go
```diff
@@ -21,12 +30,103 @@ const (
...
+
 	return nil, totalSet
 }
+
+func (r ReconcileMigStorage) validateStorageProvider(storage *migapi.MigStorage) (error, int) {
+	provider := storage.Spec.BackupStorageProvider
+	var err error
+	nSet := 0
+
+	switch provider {
```
djwhatle: +1, I was wondering if you were narrowing down checks to only look at the supplied BackupStorageProvider relevant settings.

pkg/controller/migcluster/validation.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// Validate the asset collection resource.
+// Returns error and the total error conditions set.
+func (r ReconcileMigCluster) validate(assetCollection *migapi.MigCluster) (error, int) {
+	totalSet := 0
+	var err error
+
+	// Apply changes
+	err = r.Update(context.TODO(), assetCollection)
```
djwhatle: Curious about these... If the `r.Update()` fails because the resource has been modified by someone else, do we requeue the item so that we can attempt to run the validation again later?

pkg/controller/migstorage/validation.go
```diff
@@ -21,12 +30,103 @@ const (
...
+
+	// Backup location provider.
+	err, nSet := r.validateStorageProvider(storage)
+	if err != nil {
+		return err, 0
+	}
+	totalSet += nSet
+
+	// Apply changes.
+	err = r.Update(context.TODO(), storage)
```
djwhatle: Same question here, curious how this handles a situation where the update fails. Does it retry the reconcile / validation?

pkg/controller/migcluster/validation.go
jortel: The error should be passed all the way up to the reconcile main loop and result in a failed reconcile.  It's my understanding that this results in a requeue.

PR #17, opened 2019-04-10
Tweak remote watch log messages

djwhatle:
Small changes brought over from the branch I was working on before we spun yesterday's PRs.

### DIFF
--- a/pkg/controller/remotewatcher/remotewatcher_controller.go
+++ b/pkg/controller/remotewatcher/remotewatcher_controller.go
@@ -71,8 +71,8 @@ type ReconcileRemoteWatcher struct {
 
 // Reconcile reads that state of the cluster for a RemoteWatcher object and makes changes
 func (r *ReconcileRemoteWatcher) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	log.Info(fmt.Sprintf("[rWatch] FORWARDING reconcile [%s/%s] to MigCluster [nsName=%s/%s]",
-		request.Namespace, request.Name, r.ForwardEvent.Meta.GetNamespace(), r.ForwardEvent.Meta.GetName()))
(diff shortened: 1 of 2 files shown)


pkg/controller/remotewatcher/remotewatcher_controller.go
```diff
@@ -71,8 +71,8 @@ type ReconcileRemoteWatcher struct {
 
 // Reconcile reads that state of the cluster for a RemoteWatcher object and makes changes
 func (r *ReconcileRemoteWatcher) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	log.Info(fmt.Sprintf("[rWatch] FORWARDING reconcile [%s/%s] to MigCluster [nsName=%s/%s]",
-		request.Namespace, request.Name, r.ForwardEvent.Meta.GetNamespace(), r.ForwardEvent.Meta.GetName()))
+  log.Info(fmt.Sprintf("[rWatch] Forward reconcile to MigCluster: [%s/%s] <= [%s/%s]",
+    r.ForwardEvent.Meta.GetNamespace(), r.ForwardEvent.Meta.GetName(), request.Namespace, request.Name))
```
jortel: I don't think you need the `fmt.Sprintf` here.  I think you can just do:
```
log.Info("[rWatch] Forward reconcile to MigCluster: [%s/%s] <= [%s/%s]",
    r.ForwardEvent.Meta.GetNamespace(),
    r.ForwardEvent.Meta.GetName(),
    request.Namespace, 
    request.Name)
```

djwhatle: Oh neat! I didn't know that. Will try that out.

djwhatle: Will go through later and do a pass where I convert all of the sprintf calls to what you're describing.

PR #19, opened 2019-04-11
Revised ref mapping managed by predicates.

jortel:
Manage reference mapping to support:
```
handler.EnqueueRequestsFromMapFunc{
	ToRequests: handler.ToRequestsFunc()
}
```
in predicates.  

Each predicate `Create`, `Update` & `Delete` updates the ref mapping as needed.  The `ToRequestsFunc` functions in the _handler.go_ file were reduced to 1 function so I removed _handler.go_ and used anonymous functions in the controller.

The parent-child map itself was reworked a bit:
- With _ref owner/ref-target_ semantics.  This seemed like an improvement.
- Removed constants for `Kind` using a _string_ representation of resource _type_.  This seems simpler and less to maintain over time.  Also, in kube, the type seems to be the canonical source for _kind_ and passing an instance of _type_ seems to be a pattern.  Eg: `Watch()`.
- Replaces `KubeResource` with existing `ObjectReference` which has all the same fields.
- Moved everything _ref_ related from _util_ to _reference_ package.


pkg/controller/migstage/migstage_controller.go
```diff
@@ -62,7 +66,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	err = c.Watch(
 		&source.Kind{Type: &migapi.MigPlan{}},
 		&handler.EnqueueRequestsFromMapFunc{
-			ToRequests: handler.ToRequestsFunc(MigPlanToMigStage),
+			ToRequests: handler.ToRequestsFunc(
+				func(a handler.MapObject) []reconcile.Request {
+					return migref.GetRequests(a, migapi.MigStage{})
```
djwhatle: This is a nice enhancement! Like you say allows us to avoid having a function for each type combo.

PR #20, opened 2019-04-12
Validation additions and fixes.

jortel:
Add new validations.
Fixes index out-of-bounds SEGV in `refmap.Delete()`.
Standardized management of `Ready` condition.  Add `Conditions.SetReady()` to better support validations & `Conditions.IsReady()` to support controllers.

Moves the determination of `Ready` from the `Reconcile()` to the `validate()`.  After working with this for a bit it seems more appropriate and prevents cases where the `Reconcile()` returns without validating.  Also, I could not think of anything other than _validation_ that would factor into determining when the resource is `Ready`.

### DIFF
--- a/pkg/controller/migassetcollection/validation.go
+++ b/pkg/controller/migassetcollection/validation.go
@@ -12,7 +12,6 @@ import (
 
 // Types
 const (
-	Ready              = "Ready"
 	EmptyCollection    = "EmptyCollection"
 	NamespacesNotFound = "NamespacesNotFound"
 )
@@ -24,8 +23,8 @@ const (
 
 // Statuses
 const (
-	True  = "True"
-	False = "False"
+	True  = migapi.True
+	False = migapi.False
 )
 
 // Messages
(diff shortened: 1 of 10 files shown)


pkg/controller/migassetcollection/validation.go
```diff
@@ -24,8 +23,8 @@ const (
 
 // Statuses
 const (
-	True  = "True"
-	False = "False"
+	True  = migapi.True
```
djwhatle: +1

pkg/controller/migcluster/validation.go
```diff
@@ -3,38 +3,208 @@ package migcluster
...
+			Message: InvalidSaSecretRefMessage,
+		})
+		cluster.Status.DeleteCondition(InvalidSaToken)
+		return nil, 1
+	} else {
+		cluster.Status.DeleteCondition(InvalidSaSecretRef)
+	}
+
+	// saToken
+	token, found := secret.Data["saToken"]
```
djwhatle: I wonder if we should put this "magic string" in a var somewhere since we have it in multiple places now. +1 for checking if the data exists

pkg/reference/map.go
```diff
@@ -57,11 +57,15 @@ func (r *RefMap) Delete(refOwner RefOwner, refTarget RefTarget) {
 	if !found {
 		return
 	}
+	newList := []RefOwner{}
 	for i := range list {
 		if list[i] == refOwner {
-			list = append(list[:i], list[i+1:]...)
```
djwhatle: Nice catch

pkg/controller/migcluster/validation.go
jortel: Yeah a shared _const_ is a very good idea.

PR #30, opened 2019-04-22
Break out reconcile functions for MigMigration

djwhatle:
- Reduce migmigration_controller.go reconcile to a sequence of simple calls

```
// Validate MigMigration contents
r.validate(migMigration)

// Perform prechecks and gather resources needed for reconcile
r.initReconcile(migMigration)

// Mark the MigMigration as started once prechecks are passed
r.startMigMigration(migMigration)

// Ensure source cluster has a Velero Backup
r.ensureSourceClusterBackup(migMigration)

// Ensure destination cluster has a Velero Backup + Restore
r.ensureDestinationClusterRestore(migMigration)

// Mark MigMigration as complete if Velero Restore has completed
r.finishMigMigration(migMigration)
```

 - Create `migrate.go` to contain core functionality needed for Migration (Backup + Restore). 
   - Auxiliary functions may fit nicely in other similar go files, e.g. `migrate_registry.go` for Migration tasks related to movement of images between clusters

 - Update comments in `backup.go` and `restore.go` to put resource name in consistent location

### DIFF
--- a/pkg/controller/migmigration/migmigration_controller.go
+++ b/pkg/controller/migmigration/migmigration_controller.go
@@ -42,11 +42,6 @@ const logPrefix = "mMigration"
 // TODO: don't hard-code veleroNs
 const veleroNs = "velero"
 
-/**
-* USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
-* business logic.  Delete these comments after modifying this file.*
- */
-
 // Add creates a new MigMigration Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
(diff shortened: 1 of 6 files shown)


pkg/controller/migmigration/migmigration_controller.go
```diff
@@ -97,7 +91,8 @@ var _ reconcile.Reconciler = &ReconcileMigMigration{}
 // ReconcileMigMigration reconciles a MigMigration object
 type ReconcileMigMigration struct {
 	client.Client
-	scheme *runtime.Scheme
+	scheme    *runtime.Scheme
+	resources *reconcileResources
```
djwhatle: I'm having some second thoughts about whether this is a safe place to put the reconcileResources related to 1 particular reconciliation. Was originally thinking that each reconcile request gets a unique instance of this object, but now realizing that is probably not the case.

djwhatle: Changed this, `reconcileResources` are no longer stored on `ReconcileMigMigration`

PR #31, opened 2019-04-22
Add velero backup-storage-location management.

jortel:
Create a velero `BackupStorageLocation ` (BSL) on both the source and destination clusters referenced by a plan.  The plan controller is responsible for ensuring that both the BSL & VSL exists on both the source and destination clusters as defined by the referenced `MigStorage`.  The BSLs & VSLs created for a `MigStorage` will deleted on all clusters when the `MigStorage` is deleted. 

Limitations:
- Only supports AWS.
- Does not set velero `cloud-credentials` secret.
- Does not delete BSL when storage is deleted.
- Does not bounce velero pod.

Also:
- Test connection in remote cluster validation.
- Fixes `MigStorage` not clearing `credsSecretRef` error conditions.
- Makes `volumeSnapshotConfig` optional.


pkg/apis/migration/v1alpha1/migstorage_types.go
```diff
@@ -81,3 +84,73 @@ type BackupStorageConfig struct {
...
+		reflect.DeepEqual(
+			a.Spec.ObjectStorage,
+			b.Spec.ObjectStorage)
+}
+
+// Build a velero backup storage location.
+func (r *MigStorage) BuildBSL() *velero.BackupStorageLocation {
+	location := &velero.BackupStorageLocation{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:       CorrelationLabels(r, r.Namespace, r.Name),
```
djwhatle: So this will label our `BackupStorageLocation` with what is essentially an object reference to its owner, just in label format.

`'kind': 'namespace:name'`
`'MigStorage' : 'mig:sample-migstorage'`

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -25,6 +28,19 @@ func ListPlans(client k8sclient.Client, ns string) ([]MigPlan, error) {
 	return list.Items, err
 }
 
+// List MigCluster
+// Returns and empty list when none found.
+func ListClusters(client k8sclient.Client, ns string) ([]MigCluster, error) {
+	list := MigClusterList{}
+	options := k8sclient.InNamespace(ns)
```
djwhatle: Interesting, hadn't noticed this option before.

pkg/controller/migcluster/validation.go
```diff
@@ -193,3 +204,25 @@ func (r ReconcileMigCluster) validateSaSecret(cluster *migapi.MigCluster) (int,
 
 	return 0, nil
 }
+
+// Test the connection.
+func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) (int, error) {
+	if cluster.Spec.IsHostCluster {
+		return 0, nil
+	}
+	_, err := cluster.BuildControllerRuntimeClient(r)
```
djwhatle: I don't recall how this function behaves if you give it an invalid set of credentials. I assume it will error out and cause a `TestConnectFailed` status condition to be set?

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -82,6 +82,8 @@ type ReconcileMigStorage struct {
 // +kubebuilder:rbac:groups=migration.openshift.io,resources=migstorages,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=migration.openshift.io,resources=migstorages/status,verbs=get;update;patch
 func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log.Info("[mStorage] RECONCILE() ", "request", request)
```
djwhatle: Ah interesting, so does this show the request ns/name as a key/value in the log message?

pkg/controller/migstorage/validation.go
```diff
@@ -221,7 +227,7 @@ func (r ReconcileMigStorage) validateBSLCredsSecret(storage *migapi.MigStorage)
 		storage.Status.DeleteCondition(InvalidBSLCredsSecret)
 		return 1, nil
 	} else {
-		storage.Status.DeleteCondition(InvalidBSLCredsSecret)
+		storage.Status.DeleteCondition(InvalidBSLCredsSecretRef)
```
djwhatle: Was this the bug I saw yesterday?

pkg/reference/handler.go
```diff
@@ -13,17 +13,14 @@ func GetRequests(a handler.MapObject, source interface{}) []reconcile.Request {
...
 		Namespace: a.Meta.GetNamespace(),
 	}
-	refOwner := RefOwner{
-		Kind: ToKind(source),
-	}
 	requests := []reconcile.Request{}
-	sources := refMap.Find(refTarget, refOwner)
-	for i := range sources {
-		refSource := sources[i]
+	owners := refMap.Find(refTarget, RefOwner{Kind: ToKind(source)})
```
djwhatle: I think this is easier to grok, +1

djwhatle:
Want to bring this in, but we've identified a few things that we'll continue to look into re: this PR:

- Investigation into `IsHostCluster` behavior
- Adding log messages when creating BackupStorageLocations
- Removing BackupStorageLocations when we are done with them

PR #32, opened 2019-04-24
Fix 'oc create' example for creating cluster CRD

sseago:
Reference the CRD yaml rather than the api go file.


djwhatle:
Thanks @sseago!

PR #34, opened 2019-04-24
Lables ref: ObjectMeta.UID

jortel:
Using the `ObjectMeta.UID` as the _correlation_ label value (instead of namespaced name) resolves the problem created when a CR is re-created with the same name.  The `UID` is auto-generated and designed to identify unique instances of a resource.

[1] https://kubernetes.io/docs/reference/federation/v1/definitions/#_v1_objectmeta

### DIFF
--- a/pkg/apis/migration/v1alpha1/migstorage_types.go
+++ b/pkg/apis/migration/v1alpha1/migstorage_types.go
@@ -99,7 +99,7 @@ func (r *MigStorage) Equals(a, b *velero.BackupStorageLocation) bool {
 func (r *MigStorage) BuildBSL() *velero.BackupStorageLocation {
 	location := &velero.BackupStorageLocation{
 		ObjectMeta: metav1.ObjectMeta{
-			Labels:       CorrelationLabels(r, r.Namespace, r.Name),
+			Labels:       CorrelationLabels(r, r.UID),
 			Namespace:    "velero",
 			GenerateName: r.Name + "-",
 		},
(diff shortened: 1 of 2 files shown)


pkg/apis/migration/v1alpha1/migstorage_types.go
```diff
@@ -99,7 +99,7 @@ func (r *MigStorage) Equals(a, b *velero.BackupStorageLocation) bool {
 func (r *MigStorage) BuildBSL() *velero.BackupStorageLocation {
 	location := &velero.BackupStorageLocation{
 		ObjectMeta: metav1.ObjectMeta{
-			Labels:       CorrelationLabels(r, r.Namespace, r.Name),
+			Labels:       CorrelationLabels(r, r.UID),
```
djwhatle: Am I correct in understanding that this will change the format of the CorrelationLabels from:

`"owner_kind": "owner_ns/owner_name"`
to...
`"owner_kind": "owner_uid"`

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -167,11 +166,10 @@ func GetSecret(client k8sclient.Client, ref *kapi.ObjectReference) (*kapi.Secret
 }
 
 // Build  labels used to correlate CRs.
-// Format: <kind>: <namespace>:<name>
-func CorrelationLabels(kind interface{}, namespace, name string) map[string]string {
-	ref := fmt.Sprintf("%s.%s", namespace, name)
+// Format: <kind>: <uid>.  The <uid> should be the ObjectMeta.UID
```
djwhatle: I guess so, given that it's written right here 🥇

pkg/apis/migration/v1alpha1/migstorage_types.go
djwhatle: Answered my own question by looking a little further in the PR.

jortel:
> Mechanics of change look good. Are we able to query by UID?

Yes.  Done here: https://github.com/fusor/mig-controller/pull/34/files#diff-326f9f1213591a108556c82f691008e0R143

PR #39, opened 2019-04-24
Registry cluster ref not required for local cluster.

jortel:
Make the MigCluster `ClusterRef` and `ServiceAccountSecretRef` optional when `IsHostCluster` is true.

### DIFF
--- a/pkg/controller/migcluster/migcluster_controller.go
+++ b/pkg/controller/migcluster/migcluster_controller.go
@@ -127,31 +127,36 @@ func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 
 	// Validations.
-	_, err = r.validate(migCluster)
+	nSet, err := r.validate(migCluster)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+	if nSet > 0 {
+		return reconcile.Result{}, nil
+	}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -127,31 +127,36 @@ func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Re
...
-
-	if remoteWatchCluster == nil {
-		log.Info(fmt.Sprintf("[mCluster] Starting RemoteWatch for MigCluster [%s/%s]", request.Namespace, request.Name))
-
-		// restCfg := util.BuildRestConfig(remoteClusterURL, saToken)
-		restCfg, err := migCluster.BuildRestConfig(r.Client)
-		if err != nil {
-			log.Error(err, fmt.Sprintf("[mCluster] Error during BuildRestConfig for RemoteWatch on MigCluster [%s/%s]", request.Namespace, request.Name))
-			return reconcile.Result{}, nil // don't requeue
+	if !migCluster.Spec.IsHostCluster {
```
djwhatle: This actually won't work. We need the "Remote Watch" even if we're watching events against the local cluster, otherwise reconciles won't get triggered when Velero Backups / Restores / [...] get changed.

PR #43, opened 2019-04-29
Move shared functionality between MigStage and MigMigration to 'migshared', add MigStage fields / methods

djwhatle:
- Add fields to MigStageStatus
  - StageRunning 
  - StageCompleted 
  - StartTimestamp
  - CompletionTimestamp
  - SrcBackupRef
  - DestRestoreRef

- Copy helper functions from MigMigration to MigStage
  - GetPlan
  - MarkAsRunning
  - MarkAsCompleted

---

- Within MigMigration controller, split `initMigration` into `precheck` and `getReconcileResources`

---

- Add velerorunner/ensure.go, with shared `EnsureBackup` and `EnsureRestore` functions


pkg/controller/migmigration/migrate.go
```diff
@@ -85,21 +82,25 @@ func (r *ReconcileMigMigration) ensureSourceClusterBackup(migMigration *migapi.M
 		}
 	}
 
-	vBackup := vrunner.BuildVeleroBackup(backupNsName, rres.migAssets.Spec.Namespaces, false)
+	vBackup := vrunner.BuildVeleroBackup(backupNsName, rres.MigAssets.Spec.Namespaces, false)
 	srcBackup, err := vrunner.RunBackup(srcClusterK8sClient, vBackup, backupNsName, logPrefix)
```
dymurray: Is this right? Won't we be calling `RunBackup` twice in this case? Once when you call `getSrcBackupName` and once again on `EnsureBackup`?

djwhatle: Messed this up while rebasing... Good catch!

djwhatle:
@dymurray  @jortel updated, was able to run a Migration successfully with latest.

PR #44, opened 2019-04-29
Specify includedResources if stageMigration is specified

dymurray:
This is needed so that when the stage controller runs `RunBackup` which then calls `buildVeleroBackup` we specify the `includedResources` section to only backup PV data and Imagestreams

### DIFF
--- a/pkg/velerorunner/backup.go
+++ b/pkg/velerorunner/backup.go
@@ -32,10 +32,10 @@ import (
 var log = logf.Log.WithName("controller")
 
 // RunBackup runs a Velero Backup if it hasn't been run already
-func RunBackup(c client.Client, backupNsName types.NamespacedName, assets *migapi.MigAssetCollection, logPrefix string) (*velerov1.Backup, error) {
+func RunBackup(c client.Client, backupNsName types.NamespacedName, assets *migapi.MigAssetCollection, logPrefix string, stageBackup bool) (*velerov1.Backup, error) {
 
 	vBackupExisting := &velerov1.Backup{}
-	vBackupNew := buildVeleroBackup(backupNsName.Namespace, backupNsName.Name, assets.Spec.Namespaces)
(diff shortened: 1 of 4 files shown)


pkg/velerorunner/backup.go
```diff
@@ -32,10 +32,10 @@ import (
 var log = logf.Log.WithName("controller")
 
 // RunBackup runs a Velero Backup if it hasn't been run already
-func RunBackup(c client.Client, backupNsName types.NamespacedName, assets *migapi.MigAssetCollection, logPrefix string) (*velerov1.Backup, error) {
+func RunBackup(c client.Client, backupNsName types.NamespacedName, assets *migapi.MigAssetCollection, logPrefix string, stageBackup bool) (*velerov1.Backup, error) {
 
 	vBackupExisting := &velerov1.Backup{}
-	vBackupNew := buildVeleroBackup(backupNsName.Namespace, backupNsName.Name, assets.Spec.Namespaces)
+	vBackupNew := buildVeleroBackup(backupNsName.Namespace, backupNsName.Name, assets.Spec.Namespaces, stageBackup)
```
djwhatle: I wonder if it would make sense for us to pass an entire Velero Backup into `RunBackup` instead of adding this `stageBackup` bool. If we went this route, we could drop `backupNsName`, `assets`, and `stageBackup` from the param list.

djwhatle: This would also make RunBackup more general in a way. You'd hand it an object definition of a Velero Backup you want to run, and it goes and does it for you.

djwhatle: Maybe this would require making buildVeleroBackup public.

dymurray: +1 yeah I like that approach. I will refactor so that `buildVeleroBackup` is public and we just pass the backup object to `RunBackup`

PR #48, opened 2019-05-01
Manage velero cloud-credentials secret.  AWS only.

jortel:
Manages the velero cloud-credentials secret based on the migration plan and BSL configuration in the related storage.  After velero is updated to support multiple profiles, this logic will be updated to manage the sections in the file contained in the secret with more precision.

Adds watch on referenced `CredsSecretRef`.
Adds remote watch on `Secret` with a predicate used to permit only resources containing our _correlation_ label.  This is a general approach that should be expanded to include all of the remote watched resources.

### DIFF
--- a/config/samples/cloud-creds.yaml
+++ b/config/samples/cloud-creds.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: mig
+  name: cloud-creds
+type: Opaque
+data:
+  aws-access-key-id: aGVsbG8K
+  aws-secret-access-key: aGVsbG8K
+  azure-client-secret: aGVsbG8K
+  azure-subscription-id: aGVsbG8K
+  azure-tenant-id: aGVsbG8K
+  azure-client-id: aGVsbG8K
--- a/pkg/apis/migration/v1alpha1/migstorage_types.go
+++ b/pkg/apis/migration/v1alpha1/migstorage_types.go
@@ -20,11 +20,28 @@ import (
 	"context"
 	"fmt"
 	velero "github.com/heptio/velero/pkg/apis/velero/v1"
+	"github.com/pkg/errors"
 	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"reflect"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+// Cloud Secret Format
+const AwsCloudSecretContent = `
+[default]
+aws_access_key_id=%s
+aws_secret_access_key=%s
+`
+
+// Cred Secret Fields
+const (
+	AwsAccessKeyId     = "aws-access-key-id"
+	AwsSecretAccessKey = "aws-secret-access-key"
+)
+
+// Error
+var CredSecretNotFound = errors.New("Cred secret not found.")
 
 // MigStorageSpec defines the desired state of MigStorage
 type MigStorageSpec struct {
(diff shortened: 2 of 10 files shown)


pkg/apis/migration/v1alpha1/migstorage_types.go
```diff
@@ -232,3 +249,62 @@ func (r *MigStorage) GetVSL(client k8sclient.Client) (*velero.VolumeSnapshotLoca
...
+			Name:      "cloud-credentials",
+		})
+}
+
+// Build the cloud credentials secret.
+func (r *MigStorage) BuildCloudSecret(client k8sclient.Client) (*kapi.Secret, error) {
+	secret := &kapi.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:    CorrelationLabels(r, r.ObjectMeta.UID),
+			Namespace: "velero",
```
sseago: How does this interact with the relevant section in `backupStorageConfig` of `mig-storage.yaml`:
```
    credsSecretRef:
      namespace: velero
      name: cloud-credentials
```
Also, where in the yaml files does the actual secret with embedded keys go? Will there be a separate creds-secret-aws.yaml added to the samples in config/samples to fill in?

pkg/controller/migplan/storage.go
```diff
@@ -124,6 +132,36 @@ func (r ReconcileMigPlan) ensureVSL(client k8sclient.Client, storage *migapi.Mig
 	return true, nil
 }
 
+// Create the velero BSL cloud secret has been created.
+// Returns `true` when ensured.
+func (r ReconcileMigPlan) ensureCloudSecret(client k8sclient.Client, storage *migapi.MigStorage) (bool, error) {
+	newSecret, err := storage.BuildCloudSecret(r)
```
sseago: I'm a little unclear on the workflow here. From above, `BuildCloudSecret` will return an error `CredSecretNotFound` if it doesn't exist yet (as a result of the embedded `UpdateCloudSecret` call). If the secret already exists, all of this will return successfully, but on first run, it looks like this will error out rather than creating the secret.

pkg/apis/migration/v1alpha1/migstorage_types.go
```diff
@@ -20,12 +20,29 @@ import (
...
 	velero "github.com/heptio/velero/pkg/apis/velero/v1"
+	"github.com/pkg/errors"
 	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"reflect"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// Cloud Secret Format
+const AwsCloudSecretContent = `
```
djwhatle: Nice! This is super simple, I like it.

pkg/apis/migration/v1alpha1/migstorage_types.go
djwhatle:
```
"How does this interact with the relevant section in backupStorageConfig of mig-storage.yaml"
```
@jortel can correct me if I'm wrong, but I don't think this directly interacts with the credsSecretRef. This is a reference in the other direction: it puts the UID of the `MigStorage` that owns this secret on the secret as a label.

````
secret:
   labels: { 'MigStorage': '<owning_migstorage_UID>' }
   [... secret content ...]

Comment (djwhatle) [reply pkg/apis/migration/v1alpha1/migstorage_types.go]:
```
"Also, where in the yaml files does the actual secret with embedded keys go? Will there be a separate creds-secret-aws.yaml added to the samples in config/samples to fill in?"
```
Yes, we'll need to add a new sample. @jortel it would be good to add a sample along with this PR, or at least shortly after.

Comment (jortel) [reply pkg/apis/migration/v1alpha1/migstorage_types.go]: The credsSecretRef references a MIG secret (see: `cloud-creds.yaml`) which is used to generate the velero `cloud-credentials` secret.  I corrected the mig-storage.yaml.
@djwhatle, right.  The controller creates the velero cloud-credentials secret with the correlation label back to the mig-storage that created it.

Comment (sseago) [reply pkg/apis/migration/v1alpha1/migstorage_types.go]: @djwhatle "can correct me if I'm wrong, but I don't think this directly interacts with the credsSecretRef" -- yeah, that's clearer now with the updated yaml, since it's now referencing a mig secret. With the old yaml, both secrets were velero/cloud-credentials. The update resolves my question here.

File: pkg/controller/migplan/storage.go (hunk 2 above)
Comment (jortel) [reply pkg/controller/migplan/storage.go]: The `UpdateCloudSecret()` will return an error when the _source_ `CredsSecretRef` secret does not exist or invalid.  In practice this error will rarely be raised because `ensureCloudSecret()` is only called by `Reconcile()` after validation has passed.  This error handling is for robustness and to handle race conditions whereby the `CredsSecretRef` secret is deleted after validation is performed and the `ensureCloudSecret()` is called.

**Comment by jortel** [comment]
I am testing this end-to-end with a custom bucket.  This needs to be completed before merge.

File: pkg/controller/remotewatcher/predicate.go [hunk 4]
```diff
@@ -0,0 +1,51 @@
...
+
+type SecretPredicate struct {
+	predicate.Funcs
+}
+
+// Watched resource has been updated.
+func (r SecretPredicate) Update(e event.UpdateEvent) bool {
+	secret, cast := e.ObjectOld.(*kapi.Secret)
+	if cast {
+		return hasCorrelationLabel(secret)
```
djwhatle: Filtering so that we only reconcile `Secrets` which we own seems like a really good move here :+1: 

Thinking through the potential benefits of such a predicate for other parts of the system... I don't think we will get the same advantage anywhere that we currently uses `migRef`  to narrow down events that we reconcile based on. 

https://github.com/fusor/mig-controller/blob/master/pkg/controller/migcluster/migcluster_controller.go#L88-L98

pkg/controller/migplan/storage.go
sseago: As we discussed earlier today, it's much clearer now with the renaming.

pkg/apis/migration/v1alpha1/migstorage_types.go
```diff
@@ -232,3 +249,62 @@ func (r *MigStorage) GetVSL(client k8sclient.Client) (*velero.VolumeSnapshotLoca
...
+func (r *MigStorage) EqualsCloudSecret(a, b *kapi.Secret) bool {
+	return reflect.DeepEqual(a.Data, b.Data)
+}
+
+// Get the cloud credentials secret by labels.
+func (r *MigStorage) GetCloudSecret(client k8sclient.Client) (*kapi.Secret, error) {
+	return GetSecret(
+		client,
+		&kapi.ObjectReference{
+			Namespace: "velero",
```
sseago: Oh, one other thing I forgot. We should probably use the `veleroNs` const rather than hard-coding "velero" here.

jortel:
Tested.

config/samples/cloud-creds.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: mig
+  name: cloud-creds
```
djwhatle: Thanks for adding this sample

PR #49, opened 2019-05-02
Add validation to migstage reconcile loop

dymurray:
* Add Conditions to MigStage Status field
* Adds validation of MigPlan Ref

Derek and I talked about down the line refactoring the validation work to take in a generic interface since MigStage and MigMigration share so many fields. I figured that was lower priority so will come back to address this.


dymurray:
Issue to come back to refactoring validation code to have a common plan validation function: https://github.com/fusor/mig-controller/issues/50

PR #56, opened 2019-05-08
Set the ensured VSL on Backup.

jortel:
Default the VSL settings using the BSL settings when not defined.
Set the controller _ensured_ VSL on the Backup.

### DIFF
--- a/pkg/apis/migration/v1alpha1/migstorage_types.go
+++ b/pkg/apis/migration/v1alpha1/migstorage_types.go
@@ -195,6 +195,22 @@ func (r *MigStorage) GetBSL(client k8sclient.Client) (*velero.BackupStorageLocat
 	return nil, nil
 }
 
+// Default the VSL settings (when not defined) using the BSL settings.
+func (r *MigStorage) DefaultVSLSettings() {
+	if r.Spec.VolumeSnapshotProvider != "" {
+		return
+	}
+	r.Spec.VolumeSnapshotProvider = r.Spec.BackupStorageProvider
+	r.Spec.VolumeSnapshotConfig.CredsSecretRef = r.Spec.BackupStorageConfig.CredsSecretRef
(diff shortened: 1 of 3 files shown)


pkg/apis/migration/v1alpha1/migstorage_types.go
```diff
@@ -195,6 +195,21 @@ func (r *MigStorage) GetBSL(client k8sclient.Client) (*velero.BackupStorageLocat
...
 }
 
+// Default the VSL settings (when not defined) using the BSL settings.
+func (r *MigStorage) DefaultVSLSettings() {
+	if r.Spec.VolumeSnapshotProvider != "" {
+		return
+	}
+	r.Spec.VolumeSnapshotProvider = r.Spec.BackupStorageProvider
+	switch r.Spec.BackupStorageProvider {
+	case "aws":
```
djwhatle: @jortel this seems really handy! 

A few things I'm thinking about around this:

 - Would the relationship go the other way as well? Would undefined BSL settings be filled in by whatever the user supplied for VSL? If not, why only make this work one way?

 - Is it ever dangerous to assume that the user _wants_ us to use the same AWS credentials for Volume Snapshotting? I can't think of a good reason why someone wouldn't want to use the same credentials... Maybe just something to keep in mind.

jortel: Since the BSL is required there really isn't a use case for defaulting the BSL settings using the VSL settings.  The decision to use the BSL settings for the VSL (when not defined) is the result of a slack discussion with Dylan and Scott.  Is it safe?  The consensus was yes.

djwhatle: ACK

jortel:
> Looks good! Wonder what you think about having the secret ref propagate the other way as well (from VSL -> BSL)?
Yes, I think we probably should.

PR #57, opened 2019-05-08
Add support for stage/commit of conditions.

jortel:
Add support for for _stage_ and _commit_ of conditions.  This supports a new workflow for _validations_ as follows:
1. Unstage conditions (unstaged by default when resource is fetched).
2. Set desired conditions.
3. Commit _set_ conditions.  (unstaged are removed).
4. Store the resource.

Reduces complexity and code needing to explicitly _delete_ conditions not being set.  Also, the approach handles cases where a condition _type_ chances.  Currently we can get into a situation whereby a condition is renamed while raised and never get's replaced.

### DIFF
--- a/pkg/apis/migration/v1alpha1/condition.go
+++ b/pkg/apis/migration/v1alpha1/condition.go
@@ -17,16 +17,23 @@ const (
 )
 
 // Condition
+// Type - The condition type.
+// Status - The condition status.
+// Reason - The reason for the condition.
+// Message - The human readable description of the condition.
+// staged - The condition is staged for commit.
 type Condition struct {
 	Type               string      `json:"type"`
 	Status             string      `json:"status"`
 	Reason             string      `json:"reason,omitempty"`
 	Message            string      `json:"message,omitempty"`
 	LastTransitionTime metav1.Time `json:"lastTransitionTime"`
+	staged             bool
 }
 
 // Update this condition with another's fields.
 func (r *Condition) Update(other Condition) {
+	r.staged = true
 	if r.Equal(other) {
 		return
 	}
@@ -69,6 +76,7 @@ func (r *Conditions) SetCondition(condition Condition) {
 	if r.Conditions == nil {
 		r.Conditions = []Condition{}
 	}
+	condition.staged = true
 	_, found := r.FindCondition(condition.Type)
 	if found == nil {
 		condition.LastTransitionTime = metav1.NewTime(time.Now())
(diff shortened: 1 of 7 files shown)


pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -17,16 +17,23 @@ const (
...
 	Type               string      `json:"type"`
 	Status             string      `json:"status"`
 	Reason             string      `json:"reason,omitempty"`
 	Message            string      `json:"message,omitempty"`
 	LastTransitionTime metav1.Time `json:"lastTransitionTime"`
+	staged             bool
 }
 
 // Update this condition with another's fields.
 func (r *Condition) Update(other Condition) {
```
djwhatle: I'm trying to understand the intent here based on the comment on this method. 
```
Update this condition with another's fields.
```
Why would you want to update a condition with another conditions fields?

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -17,16 +17,23 @@ const (
...
+	r.staged = true
```
djwhatle: So now each condition has a `staged` bool attached to it, and `staged` gets set to true if a Condition changes.
I noticed you didn't include JSON serialization info and made `staged` lowercase. Does this keep it from showing up on the resource?

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -83,12 +91,42 @@ func (r *Conditions) DeleteCondition(cndTypes ...string) {
 	if r.Conditions == nil {
 		return
 	}
+	kept := []Condition{}
```
djwhatle: What's the purpose of the `DeleteCondition` method now? I thought we were clearing our the entire set when reconcile starts?

pkg/apis/migration/v1alpha1/condition.go
jortel: This is mainly used by `SetCondition()` when a condition _Type_ already exists.  The _other_ condition is a new one passed to SetCondition().  This updates an existing condition with a new one excluding the LastTransitionTime.  It supports SetCondition() being idempotent while preserving the LastTransitionTime.

pkg/apis/migration/v1alpha1/condition.go
jortel: I modeled this after `Git`.  All conditions are un-staged by default.  A condition is _staged_  when explicitly set (or updated) using `SetCondition()` like `git add`.  Then, calling `CommitConditions()` removes the un-staged conditions before the resource is stored.  This way, the validation code does not need to explicitly delete condition but rather only set the onces that still exist.

Correct, `staged` is not included in the JSON serialization.

pkg/apis/migration/v1alpha1/condition.go
jortel: The staged/commit work-flow is optional and mainly used by the _validation.go_ code.  There are still cases where the explicit set/delete work-flow is used.  For example: the condition related to ensuring the BSL/DSL in the plan controller.  And, in some cases, the final `Ready` condition.

djwhatle: ACK

PR #58, opened 2019-05-09
Refactor stage controller to use new velerorunner changes

dymurray:


pkg/velerorunner/task.go
```diff
@@ -230,11 +230,11 @@ func (t *Task) updateBackup(backup *velero.Backup) error {
 	backup.Spec = velero.BackupSpec{
 		StorageLocation:         backupLocation.Name,
 		VolumeSnapshotLocations: []string{"aws-default"},
-		TTL:                     metav1.Duration{Duration: 720 * time.Hour},
-		IncludedNamespaces:      namespaces,
-		ExcludedNamespaces:      []string{},
-		IncludedResources:       t.BackupResources,
-		ExcludedResources:       []string{},
+		TTL:                metav1.Duration{Duration: 720 * time.Hour},
```
dymurray: gofmt automatically moved this whitespace... not sure why it felt that this was needed since it doesn't line up with `VolumeSnapshotLocations` but it complained regardless.

PR #60, opened 2019-05-09
Validate asset-collection namespaces on the source cluster.

jortel:
Validate asset-collection namespaces on the source cluster.  Since an AssetCollection is shared between plans, it's more appropriate for the plan to perform the validation.

Depends on: https://github.com/fusor/mig-controller/pull/57 which needs to be merged first.  There should only be 2 affected files.

### DIFF
--- a/pkg/apis/migration/v1alpha1/condition.go
+++ b/pkg/apis/migration/v1alpha1/condition.go
@@ -17,16 +17,23 @@ const (
 )
 
 // Condition
+// Type - The condition type.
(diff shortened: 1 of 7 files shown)


jortel:
> ACK. I have to ask, what were we doing before? I have definitely seen the MigAssetCollection validation identify when I was trying to migrate a namespace that didn't exist on my src cluster.

The validation was using the local client so if you ran the controller on the source cluster, it would _seem_ to be working.

PR #65, opened 2019-05-10
Clearer name for CommitConditions.

jortel:
After explaining the concept in a previous PR, I became concerned that the term _commit_ suggested something persistent.  I think `DeleteUnstagedConditions()` would be clearer.


pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -21,7 +21,7 @@ const (
 // Status - The condition status.
 // Reason - The reason for the condition.
 // Message - The human readable description of the condition.
-// staged - The condition is staged for commit.
+// staged - A condition has been explicitly set/updated.
```
djwhatle: This is helpful

PR #69, opened 2019-05-10
Validate required namespaces.

jortel:
Validate required namespaces starting with the `velero` namespace.

### DIFF
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -3,6 +3,7 @@ package migplan
 import (
 	"context"
 	"fmt"
+	"github.com/fusor/mig-controller/pkg/velerorunner"
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


pkg/controller/migplan/validation.go
```diff
@@ -43,18 +46,20 @@ const (
...
+	InvalidStorageRefMessage              = "The `migStorageRef` must reference a `migstorage`."
+	InvalidAssetCollectionRefMessage      = "The `migAssetCollectionRef` must reference a `migassetcollection`."
+	SourceClusterNotReadyMessage          = "The referenced `srcMigClusterRef` does not have a `Ready` condition."
+	DestinationClusterNotReadyMessage     = "The referenced `dstMigClusterRef` does not have a `Ready` condition."
+	StorageNotReadyMessage                = "The referenced `migStorageRef` does not have a `Ready` condition."
+	AssetCollectionNotReadyMessage        = "The referenced `migAssetCollectionRef` does not have a `Ready` condition."
+	AssetNamespaceNotFoundMessage         = "The following asset `namespaces` [%s] not found on the source cluster."
+	InvalidDestinationClusterMessage      = "The `srcMigClusterRef` and `dstMigClusterRef` cannot be the same."
+	EnsureStorageFailedMessage            = "Failed to create/validate backup and volume snapshot storage."
+	NsNotFoundOnSourceClusterMessage      = "The follow namespaces [%s] not found on source cluster."
```
djwhatle: Assume this was meant to read:

`The following namespaces`

I think we could reword this to 

`Namespaces [%s] not found on source cluster`

pkg/controller/migplan/validation.go
```diff
@@ -43,18 +46,20 @@ const (
...
+	NsNotFoundOnDestinationClusterMessage = "The follow namespaces [%s] not found on destination cluster."
```
djwhatle: Same thing here

pkg/controller/migplan/validation.go
```diff
@@ -329,3 +341,111 @@ func (r ReconcileMigPlan) validateDestinationCluster(plan *migapi.MigPlan) (int,
...
+		})
+		return 1, nil
+	}
+
+	return 0, nil
+}
+
+// Validate required namespaces on the destination cluster.
+// Returns error and the total error conditions set.
+func (r ReconcileMigPlan) validateDestinationNamespaces(plan *migapi.MigPlan) (int, error) {
```
djwhatle: What are the "required namespaces" here? Velero will create any namespaces we are migrating as part of the "Restore" process.

jortel: For both the source and destination, the required namespaces are not related to asset collection.  Instead, the list (to start with) includes the namespace the controller will create velero resources in.  Since it's intended to be configurable, It may not be `velero`.

pkg/controller/migplan/validation.go
```diff
@@ -329,3 +341,111 @@ func (r ReconcileMigPlan) validateDestinationCluster(plan *migapi.MigPlan) (int,
...
+	}
+	totalSet += nSet
+
+	return totalSet, nil
+}
+
+// Validate required namespaces on the source cluster.
+// Returns error and the total error conditions set.
+func (r ReconcileMigPlan) validateSourceNamespaces(plan *migapi.MigPlan) (int, error) {
+	namespaces := []string{velerorunner.VeleroNamespace}
```
djwhatle: Ah, I see now

pkg/controller/migplan/validation.go
djwhatle: I see that now, thanks for the explanation.

pkg/controller/migplan/validation.go
jortel: Fixed.

PR #70, opened 2019-05-13
Bounce restic pod prior to running Velero backup

dymurray:
This is a workaround to solve the issue where 3.7-3.9 does not support the mount propagation feature needed for restic. To get around this, we bounce the pod prior to creating a backup.

### DIFF
--- a/config/crds/migration_v1alpha1_migmigration.yaml
+++ b/config/crds/migration_v1alpha1_migmigration.yaml
@@ -36,16 +36,14 @@ spec:
             completionTimestamp:
               format: date-time
               type: string
-            destRestoreRef:
-              type: object
             migrationCompleted:
               type: boolean
             migrationStarted:
               type: boolean
-            srcBackupRef:
-              type: object
             startTimestamp:
               format: date-time
+              type: string
+            taskPhase:
               type: string
           type: object
   version: v1alpha1
--- a/pkg/controller/migstage/stage.go
+++ b/pkg/controller/migstage/stage.go
@@ -18,6 +18,7 @@ package migstage
 
 import (
 	"context"
+
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
 	vrunner "github.com/fusor/mig-controller/pkg/velerorunner"
 )
--- a/pkg/velerorunner/backup.go
+++ b/pkg/velerorunner/backup.go
@@ -2,13 +2,18 @@ package velerorunner
 
 import (
 	"context"
+	"reflect"
+	"strings"
+	"time"
+
 	velero "github.com/heptio/velero/pkg/apis/velero/v1"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"reflect"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
-	"time"
 )
+
+var resticPodPrefix = "restic-"
 
 // Ensure the backup on the source cluster has been created
 // and has the proper settings.
@@ -144,11 +149,11 @@ func (t *Task) updateBackup(backup *velero.Backup) error {
 	backup.Spec = velero.BackupSpec{
 		StorageLocation:         backupLocation.Name,
 		VolumeSnapshotLocations: []string{snapshotLocation.Name},
-		TTL:                     metav1.Duration{Duration: 720 * time.Hour},
-		IncludedNamespaces:      namespaces,
-		ExcludedNamespaces:      []string{},
-		IncludedResources:       t.BackupResources,
-		ExcludedResources:       []string{},
+		TTL:                metav1.Duration{Duration: 720 * time.Hour},
+		IncludedNamespaces: namespaces,
+		ExcludedNamespaces: []string{},
+		IncludedResources:  t.BackupResources,
+		ExcludedResources:  []string{},
 		Hooks: velero.BackupHooks{
 			Resources: []velero.BackupResourceHookSpec{},
 		},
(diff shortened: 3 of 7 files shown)


pkg/velerorunner/backup.go
```diff
@@ -178,3 +190,60 @@ func (t *Task) getReplicatedBackup() (*velero.Backup, error) {
...
+	if err != nil {
+		return err
+	}
+
+	// Set two vars to determine if we have bounced a pod and it is has become
+	// ready or if we need to delete the pod
+	var podDeleted = false
+	var podBounced = false
+	// Loop until pod has been deleted and is `Running`
+	for podBounced == false {
```
djwhatle: Rather than having the entire pod bounce happen inside of a single reconcile like this, I think we need to figure out how to make the controller kick off a "bounce" action here, then exit the reconcile to wait for another action, then come back and check if the bounce was completed successfully before taking further action.

I think we can store some state in the `status` of the MigMigration to help with this, e.g. the podBounced and podDeleted vars could be moved there (although they would likely be renamed to be a bit more descriptive out of context, e.g. resticPodDeleted, resticPodBounced).

The challenging part here will be detecting when the restic pod state changes. I'm not sure if it's possible to detect this using a watch. We can look into it. Another option would be to exit with a requeue set for 10 seconds in the future or so.

What I'm trying to avoid here is the controller getting stuck inside of this loop, since MigStage / MigMigration only have one worker each. No other work can be done by these controller while they are waiting on this process to finish.

pkg/velerorunner/backup.go
```diff
@@ -178,3 +190,60 @@ func (t *Task) getReplicatedBackup() (*velero.Backup, error) {
...
+				Namespace: VeleroNamespace,
+			},
+			&list)
+		if err != nil {
+			return err
+		}
+
+		// Loop through all pods in velero namespace
+		for _, pod := range list.Items {
+			if strings.HasPrefix(pod.Name, resticPodPrefix) {
```
jortel: Recommend using `continue` to avoid the deep nesting.
```
if !strings.HasPrefix(pod.Name, resticPodPrefix) {
    continue
}
```
and
```
if pod.Status.Phase != corev1.PodRunning {
    continue
}

File: pkg/velerorunner/backup.go (hunk 1 above)
Comment (jortel) [reply pkg/velerorunner/backup.go]: Agreed with @djwhatle.

File: pkg/velerorunner/backup.go (hunk 2 above)
Comment (djwhatle) [reply pkg/velerorunner/backup.go]: +1

File: pkg/velerorunner/backup.go [hunk 3]
```diff
@@ -22,6 +27,14 @@ func (t *Task) ensureBackup() error {
 		return err
 	}
 	if foundBackup == nil {
+		// Mount propagation workaround
+		// TODO: Consider only bouncing restic pod if cluster version is 3.7-3.9
+		// This would require passing in cluster version to the controller, but might
+		// not be necessary
+		/*err := t.bounceResticPod()
```
djwhatle: We can probably leave this commented section out

pkg/velerorunner/backup.go
dymurray: Done, thanks!

pkg/velerorunner/backup.go
dymurray: Done, now we are using the status field to monitor the restic restart state

pkg/controller/migstage/stage.go
```diff
@@ -61,6 +61,22 @@ func (r *ReconcileMigStage) stage(stageMigration *migapi.MigStage) error {
...
+	// This would require passing in cluster version to the controller, but might
+	// not be necessary
+	err, restartStatus := task.BounceResticPod()
+	if err != nil {
+		return err
+	}
+	// Update status if restart status has changed
+	if restartStatus != stageMigration.Status.ResticRestartStatus {
+		stageMigration.Status.ResticRestartStatus = restartStatus
+		err = r.Update(context.TODO(), stageMigration)
```
jortel: Best I can tell, the result of this update failing is:
- The restic pods could get deleted more than necessary.
- The controller will list the pods more than necessary.

Right?

dymurray: Yes, because `BounceResticPod` will get called again with a stale status so it will essentially repeat the previous run.

pkg/velerorunner/task.go
```diff
@@ -42,8 +48,23 @@ type Task struct {
 // Run the task.
 // Return `true` when run to completion.
 func (t *Task) Run() (bool, error) {
+	// Mount propagation workaround
```
djwhatle: @dymurray @jortel moved the workaround code section into `Task.Run()`

djwhatle:
Tested this, worked for me 👍

Nice work Dylan, I'm good to merge when you guys are.

pkg/velerorunner/backup.go
```diff
@@ -178,3 +189,64 @@ func (t *Task) getReplicatedBackup() (*velero.Backup, error) {
...
+		},
+		&list)
+	if err != nil {
+		return err, ""
+	}
+
+	// Loop through all pods in velero namespace
+	for _, pod := range list.Items {
+		if !strings.HasPrefix(pod.Name, resticPodPrefix) {
+			continue
```
djwhatle: 👍

config/crds/migration_v1alpha1_migmigration.yaml
```diff
@@ -36,14 +36,12 @@ spec:
...
               type: string
-            destRestoreRef:
-              type: object
             migrationCompleted:
               type: boolean
             migrationStarted:
               type: boolean
-            srcBackupRef:
-              type: object
+            resticRestartStatus:
```
djwhatle: Thanks for committing updated CRD contents ⚡️

pkg/velerorunner/backup.go
```diff
@@ -178,3 +183,64 @@ func (t *Task) getReplicatedBackup() (*velero.Backup, error) {
...
+	if err != nil {
+		return err
+	}
+
+	// List all pods in velero namespace
+	list := corev1.PodList{}
+	err = client.List(
+		context.TODO(),
+		&k8sclient.ListOptions{
+			Namespace: VeleroNamespace,
```
djwhatle: I'm not sure if it would make any performance difference, but if we're really trying to clamp down on apiserver traffic at some point in the future we could perhaps query with a labelselector for `name=restic`. Don't think there's a good reason to switch this around at this point.

djwhatle: ^ This kind of thing would be _very_ low on the priority list though... we are generating expontentially more apiserver traffic through validations compared to this one time query. Was really just curious if there was a label here.

djwhatle:
@jortel just tested this, worked for me. I'm good to merge whenever.

pkg/velerorunner/backup.go
dymurray: Ah yeah I actually don't know if restic has a label, I can check.

PR #72, opened 2019-05-14
Task phases

jortel:
Introduces `Task` phases.  This is needed to support resuming a task by reconcile() trigged by watched resources and polling.  The polling is needed to support Restic pod restarts.  The `Task.Phase` provides Task.Run() caller with simple information used to determine if a reconcile requeue is needed.  It also provides for other conditional behavior without introducing complex state.

### DIFF
--- a/pkg/apis/migration/v1alpha1/migmigration_types.go
+++ b/pkg/apis/migration/v1alpha1/migmigration_types.go
@@ -39,6 +39,7 @@ type MigMigrationStatus struct {
 	MigrationCompleted  bool         `json:"migrationCompleted,omitempty"`
 	StartTimestamp      *metav1.Time `json:"startTimestamp,omitempty"`
 	CompletionTimestamp *metav1.Time `json:"completionTimestamp,omitempty"`
+	TaskPhase           string       `json:"taskPhase,omitempty"`
 }
 
 // +genclient
--- a/pkg/apis/migration/v1alpha1/migstage_types.go
+++ b/pkg/apis/migration/v1alpha1/migstage_types.go
@@ -32,15 +32,11 @@ type MigStageSpec struct {
 // MigStageStatus defines the observed state of MigStage
 type MigStageStatus struct {
 	Conditions
-
-	StageRunning   bool `json:"stageStarted,omitempty"`
-	StageCompleted bool `json:"stageCompleted,omitempty"`
-
+	StageRunning        bool         `json:"stageStarted,omitempty"`
+	StageCompleted      bool         `json:"stageCompleted,omitempty"`
 	StartTimestamp      *metav1.Time `json:"startTimestamp,omitempty"`
 	CompletionTimestamp *metav1.Time `json:"completionTimestamp,omitempty"`
-
-	SrcBackupRef   *kapi.ObjectReference `json:"srcBackupRef,omitempty"`
-	DestRestoreRef *kapi.ObjectReference `json:"destRestoreRef,omitempty"`
+	TaskPhase           string       `json:"taskPhase,omitempty"`
 }
 
 // +genclient
--- a/pkg/controller/migmigration/migmigration_controller.go
+++ b/pkg/controller/migmigration/migmigration_controller.go
@@ -21,6 +21,7 @@ import (
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
 	migref "github.com/fusor/mig-controller/pkg/reference"
 	"k8s.io/apiserver/pkg/storage/names"
+	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -120,13 +121,17 @@ func (r *ReconcileMigMigration) Reconcile(request reconcile.Request) (reconcile.
 	}
 
 	// Do the migration.
-	err = r.migrate(migration)
+	requeue, err := r.migrate(migration)
 	if err != nil {
 		if errors.IsConflict(err) {
 			return reconcile.Result{Requeue: true}, nil
 		} else {
 			return reconcile.Result{}, err
 		}
+	}
+	if requeue {
+		delay := time.Second * 5
+		return reconcile.Result{RequeueAfter: delay}, nil
 	}
 
 	return reconcile.Result{}, nil
--- a/pkg/controller/migmigration/migrate.go
+++ b/pkg/controller/migmigration/migrate.go
@@ -22,19 +22,19 @@ import (
 	vrunner "github.com/fusor/mig-controller/pkg/velerorunner"
 )
 
-func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) error {
+func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (bool, error) {
 	if migration.IsCompleted() {
-		return nil
+		return false, nil
 	}
 
 	// Ready
 	plan, err := migration.GetPlan(r)
 	if err != nil {
-		return err
+		return false, err
 	}
 	if !plan.Status.IsReady() {
 		log.Info("Plan not ready.", "name", migration.Name)
-		return err
+		return false, err
 	}
 
 	// Started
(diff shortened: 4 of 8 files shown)


pkg/velerorunner/task.go
```diff
@@ -109,3 +123,91 @@ func (t *Task) getSourceClient() (k8sclient.Client, error) {
...
+	log := t.Log
+	switch t.Phase {
+	case "":
+		log.Info(
+			"Task has started.",
+			"owner",
+			t.Owner.GetName())
+	case RestartRestic:
+		log.Info(
+			"Restart Restic pods.",
```
dymurray: NIT: Change this to "Restarting Restic pod"

pkg/velerorunner/task.go
```diff
@@ -109,3 +123,91 @@ func (t *Task) getSourceClient() (k8sclient.Client, error) {
...
+			"owner",
+			t.Owner.GetName())
+	case WaitOnResticRestart:
+		log.Info(
+			"Waiting on Restic pod restart to complete.",
+			"owner",
+			t.Owner.GetName())
+	case ResticRestartCompleted:
+		log.Info(
+			"Restic pod has restart completed.",
```
dymurray: NIT: "Restic pod successfully restarted"

pkg/velerorunner/task.go
```diff
@@ -109,3 +123,91 @@ func (t *Task) getSourceClient() (k8sclient.Client, error) {
...
+	case BackupFailed:
+		log.Info(
+			"Backup has failed.",
+			"owner",
+			t.Owner.GetName(),
+			"backup",
+			t.Backup.Name)
+	case WaitOnBackupReplicated:
+		log.Info(
+			"Waiting Backup replication to destination.",
```
dymurray: NIT: "Waiting for Backup to replicate to destination", I don't like mine much better but grammar here is off

pkg/velerorunner/task.go
```diff
@@ -109,3 +123,91 @@ func (t *Task) getSourceClient() (k8sclient.Client, error) {
...
+	case RestoreCompleted:
+		log.Info(
+			"Restore has completed.",
+			"owner",
+			t.Owner.GetName(),
+			"restore",
+			t.Restore.Name)
+	case RestoreFailed:
+		log.Info(
+			"Restore as failed.",
```
dymurray: NIT: "has failed"

pkg/velerorunner/task.go
```diff
@@ -36,68 +55,63 @@ type Task struct {
...
-			t.Owner.GetName(),
-			"backup",
-			t.Backup.Name)
-		return false, nil
+		t.Phase = BackupStarted
+		t.logPhase()
+		return nil
+	} else {
+		t.Phase = BackupCompleted
+		t.logPhase()
```
dymurray: Do you want to return here? Otherwise the phase will get changed at line 83

pkg/apis/migration/v1alpha1/migmigration_types.go
```diff
@@ -39,6 +39,7 @@ type MigMigrationStatus struct {
 	MigrationCompleted  bool         `json:"migrationCompleted,omitempty"`
 	StartTimestamp      *metav1.Time `json:"startTimestamp,omitempty"`
 	CompletionTimestamp *metav1.Time `json:"completionTimestamp,omitempty"`
+	TaskPhase           string       `json:"taskPhase,omitempty"`
```
djwhatle: Should we use `Phase` here? Seems more standard, unless you were purposely differentiating.

pkg/apis/migration/v1alpha1/migstage_types.go
```diff
@@ -32,15 +32,11 @@ type MigStageSpec struct {
...
-	StageCompleted bool `json:"stageCompleted,omitempty"`
-
+	StageRunning        bool         `json:"stageStarted,omitempty"`
+	StageCompleted      bool         `json:"stageCompleted,omitempty"`
 	StartTimestamp      *metav1.Time `json:"startTimestamp,omitempty"`
 	CompletionTimestamp *metav1.Time `json:"completionTimestamp,omitempty"`
-
-	SrcBackupRef   *kapi.ObjectReference `json:"srcBackupRef,omitempty"`
-	DestRestoreRef *kapi.ObjectReference `json:"destRestoreRef,omitempty"`
+	TaskPhase           string       `json:"taskPhase,omitempty"`
```
djwhatle: Same

pkg/controller/migmigration/migmigration_controller.go
```diff
@@ -120,14 +121,18 @@ func (r *ReconcileMigMigration) Reconcile(request reconcile.Request) (reconcile.
 	}
 
 	// Do the migration.
-	err = r.migrate(migration)
+	requeue, err := r.migrate(migration)
```
djwhatle: Maybe at some point we'll want to pass a delay in seconds, but I think defaulting to 5 meets our needs well for now.

pkg/controller/migmigration/migrate.go
```diff
@@ -22,19 +22,19 @@ import (
 	vrunner "github.com/fusor/mig-controller/pkg/velerorunner"
 )
 
-func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) error {
+func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (bool, error) {
```
djwhatle: I might consider a named return value for `requeue bool` here. It doesn't feel obvious that true/false here refers to "requeue" rather than something like "success".

pkg/controller/migmigration/migrate.go
```diff
@@ -43,35 +43,43 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) error {
...
 		if err != nil {
-			return err
+			return false, err
 		}
+	case vrunner.WaitOnResticRestart:
+		requeue = true
 	}
 
-	return nil
+	return requeue, nil
```
djwhatle: It does become more obvious down here

pkg/controller/migstage/stage.go
```diff
@@ -24,19 +24,19 @@ import (
 
 var stageResources = []string{"pods", "persistentvolumes", "persistentvolumeclaims", "imagestreams", "imagestreamtags"}
 
-func (r *ReconcileMigStage) stage(stageMigration *migapi.MigStage) error {
+func (r *ReconcileMigStage) stage(stageMigration *migapi.MigStage) (bool, error) {
```
djwhatle: I'd also consider a named return value here.

pkg/controller/migmigration/migrate.go
```diff
@@ -43,35 +43,42 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) error {
...
 	if err != nil {
-		return err
+		return false, err
 	}
 	task := vrunner.Task{
 		Log:           log,
 		Client:        r,
 		Owner:         migration,
 		PlanResources: planResources,
+		Phase:         migration.Status.TaskPhase,
```
djwhatle: Note to self: we are passing Phase into `task.Run()` now.

pkg/apis/migration/v1alpha1/migmigration_types.go
jortel: Perhaps.  Since we were not convinced about exposing the `phase` I thought it best to make it clear that it was different.  That said, I can be easily convinced to name it `Phase`.  @dymurray?

pkg/velerorunner/task.go
jortel: True.  The phase will never be saved in the `Status` but it's more efficient to continue here.

pkg/apis/migration/v1alpha1/migmigration_types.go
dymurray: I'm a little indifferent... but I think since we were hesitant about using the phase field I would keep it different for now

pkg/velerorunner/task.go
```diff
@@ -9,12 +9,30 @@ import (
 
 var VeleroNamespace = "velero"
 
+// Phases
+const (
+	Started                = ""
+	RestartRestic          = "RestartRestic"
```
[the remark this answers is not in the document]
dymurray: I think you are right. Can remove `RestartRestic` if you want or I can do it in my PR

pkg/velerorunner/task.go
```diff
@@ -9,12 +9,30 @@ import (
...
+	WaitOnResticRestart    = "WaitOnResticRestart"
+	ResticRestartCompleted = "ResticRestartCompleted"
+	BackupStarted          = "BackupStarted"
+	BackupCompleted        = "BackupCompleted"
+	BackupFailed           = "BackupFailed"
+	WaitOnBackupReplicated = "WaitOnBackupReplicated"
```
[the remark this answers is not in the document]
jortel: Agreed.  That is a better.

pkg/velerorunner/task.go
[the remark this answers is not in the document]
jortel: True, It's mainly just a breakcrumb for @dymurray.  I assume he'll adjust things as needed.

PR #75, opened 2019-05-15
Create migration-registry before backup/restore

sseago:
Create the migration registry deploymentconfig and related
resources (secret, imagestream, service) in source and destination
clusters before starting backup or restore.

### DIFF
--- a/pkg/controller/migcluster/remote_watch.go
+++ b/pkg/controller/migcluster/remote_watch.go
@@ -18,8 +18,10 @@ package migcluster
 
 import (
 	"github.com/fusor/mig-controller/pkg/controller/remotewatcher"
+	"github.com/fusor/mig-controller/pkg/imagescheme"
 	"github.com/fusor/mig-controller/pkg/remote"
 	velerov1 "github.com/heptio/velero/pkg/apis/velero/v1"
+	appsv1 "github.com/openshift/api/apps/v1"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -41,6 +43,16 @@ func StartRemoteWatch(r *ReconcileMigCluster, config remote.ManagerConfig) error
 	log.Info("[rWatch] Adding Velero to scheme")
 	if err := velerov1.AddToScheme(mgr.GetScheme()); err != nil {
 		log.Error(err, "[rWatch] Unable add Velero APIs to scheme")
+		return err
+	}
+	log.Info("[rWatch] Adding OpenShift imagestream to scheme")
+	if err := imagescheme.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "[rWatch] Unable add OpenShift image APIs to scheme")
+		return err
+	}
+	log.Info("[rWatch] Adding OpenShift deploymentconfig to scheme")
+	if err := appsv1.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "[rWatch] Unable add OpenShift apps APIs to scheme")
 		return err
 	}
 
--- a/pkg/velerorunner/registry.go
+++ b/pkg/velerorunner/registry.go
@@ -1,0 +1,551 @@
+package velerorunner
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"reflect"
+
+	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
+	appsv1 "github.com/openshift/api/apps/v1"
+	imagev1 "github.com/openshift/api/image/v1"
+	kapi "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Ensure the migration registry on the source cluster has been created
+// and has the proper settings.
+func (t *Task) ensureSrcMigRegistry() error {
+	t.SrcRegistryResources = &MigRegistryResources{}
+	client, err := t.getSourceClient()
+	if err != nil {
+		return err
+	}
+	return t.ensureMigRegistry(client, t.SrcRegistryResources)
+}
+
+// Ensure the migration registry on the destination cluster has been created
+// and has the proper settings.
+func (t *Task) ensureDestMigRegistry() error {
+	t.DestRegistryResources = &MigRegistryResources{}
+	client, err := t.getDestinationClient()
+	if err != nil {
+		return err
+	}
+	return t.ensureMigRegistry(client, t.DestRegistryResources)
+}
+
+// Ensure the migration registry on the specified cluster has been created
+// and has the proper settings.
+func (t *Task) ensureMigRegistry(client k8sclient.Client, registryResources *MigRegistryResources) error {
+	err := t.ensureRegistrySecret(client, registryResources)
+	if err != nil {
+		return err
+	}
+
+	err = t.ensureRegistryImageStream(client, registryResources)
+	if err != nil {
+		return err
+	}
+
+	err = t.ensureRegistryDC(client, registryResources)
+	if err != nil {
+		return err
+	}
+	err = t.ensureRegistryService(client, registryResources)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Ensure the credentials secret for the migration registry on the specified cluster has been created
+func (t *Task) ensureRegistrySecret(client k8sclient.Client, registryResources *MigRegistryResources) error {
+	newSecret, err := t.buildRegistrySecret()
+	if err != nil {
+		return err
+	}
+	foundSecret, err := t.getRegistrySecret(client)
+	if err != nil {
+		return err
+	}
+	if foundSecret == nil {
+		registryResources.CredSecret = newSecret
+		err = client.Create(context.TODO(), newSecret)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+	registryResources.CredSecret = foundSecret
+	if !t.equalsRegistrySecret(newSecret, foundSecret) {
+		t.updateRegistrySecret(foundSecret)
+		err = client.Update(context.TODO(), foundSecret)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Build a credentials Secret as desired for the source cluster.
+func (t *Task) buildRegistrySecret() (*kapi.Secret, error) {
+	secret := &kapi.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:       t.PlanResources.MigPlan.GetCorrelationLabels(),
+			GenerateName: t.PlanResources.MigPlan.GetName() + "-registry-",
+			Namespace:    VeleroNamespace,
+		},
+	}
+	err := t.updateRegistrySecret(secret)
+	return secret, err
+}
+
+// Update a Registry credentials secret as desired for the specified cluster.
+func (t *Task) updateRegistrySecret(secret *kapi.Secret) error {
+	credSecret, err := t.PlanResources.MigStorage.Spec.BackupStorageConfig.GetCredsSecret(t.Client)
+	if err != nil {
+		return err
+	}
+	if credSecret == nil {
+		return migapi.CredSecretNotFound
+	}
+	secret.Data = map[string][]byte{
+		"access_key": []byte(credSecret.Data[migapi.AwsAccessKeyId]),
+		"secret_key": []byte(credSecret.Data[migapi.AwsSecretAccessKey]),
+	}
+	return nil
+}
+
+// Get an existing credentials Secret on the source cluster.
+func (t *Task) getRegistrySecret(client k8sclient.Client) (*kapi.Secret, error) {
+	list := kapi.SecretList{}
+	labels := t.PlanResources.MigPlan.GetCorrelationLabels()
+	err := client.List(
+		context.TODO(),
+		k8sclient.MatchingLabels(labels),
+		&list)
+	if err != nil {
+		return nil, err
+	}
+	if len(list.Items) > 0 {
+		return &list.Items[0], nil
+	}
+
+	return nil, nil
+}
+
+// Determine if two registry credentials secrets are equal.
+// Returns `true` when equal.
+func (t *Task) equalsRegistrySecret(a, b *kapi.Secret) bool {
+	return reflect.DeepEqual(a.Data, b.Data)
+}
+
+// Ensure the imagestream for the migration registry on the specified cluster has been created
+func (t *Task) ensureRegistryImageStream(client k8sclient.Client, registryResources *MigRegistryResources) error {
+	registrySecret := registryResources.CredSecret
+	if registrySecret == nil {
+		return errors.New("migration registry credentials not found")
+	}
+	newImageStream, err := t.buildRegistryImageStream(registrySecret.GetName())
+	if err != nil {
+		return err
+	}
+	foundImageStream, err := t.getRegistryImageStream(client)
+	if err != nil {
+		return err
+	}
+	if foundImageStream == nil {
+		registryResources.ImageStream = newImageStream
+		err = client.Create(context.TODO(), newImageStream)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+	registryResources.ImageStream = foundImageStream
+	if !t.equalsRegistryImageStream(newImageStream, foundImageStream) {
+		t.updateRegistryImageStream(foundImageStream)
+		err = client.Update(context.TODO(), foundImageStream)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Build a Registry ImageStream as desired for the source cluster.
+func (t *Task) buildRegistryImageStream(name string) (*imagev1.ImageStream, error) {
+	labels := t.PlanResources.MigPlan.GetCorrelationLabels()
+	labels["app"] = name
+	imagestream := &imagev1.ImageStream{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:    labels,
+			Name:      name,
+			Namespace: VeleroNamespace,
+		},
+	}
+	err := t.updateRegistryImageStream(imagestream)
+	return imagestream, err
+}
+
+// Update a Registry ImageStream as desired for the specified cluster.
+func (t *Task) updateRegistryImageStream(imagestream *imagev1.ImageStream) error {
+	imagestream.Spec = imagev1.ImageStreamSpec{
+		LookupPolicy: imagev1.ImageLookupPolicy{Local: false},
+		Tags: []imagev1.TagReference{
+			imagev1.TagReference{
+				Name: "2",
+				Annotations: map[string]string{
+					"openshift.io/imported-from": "registry:2",
+				},
+				From: &kapi.ObjectReference{
+					Kind: "DockerImage",
+					Name: "registry:2",
+				},
+				Generation:      nil,
+				ImportPolicy:    imagev1.TagImportPolicy{},
+				ReferencePolicy: imagev1.TagReferencePolicy{Type: ""},
+			},
+		},
+	}
+	return nil
+}
+
+// Get an existing registry ImageStream on the specifiedcluster.
+func (t *Task) getRegistryImageStream(client k8sclient.Client) (*imagev1.ImageStream, error) {
+	list := imagev1.ImageStreamList{}
+	labels := t.PlanResources.MigPlan.GetCorrelationLabels()
+	err := client.List(
+		context.TODO(),
+		k8sclient.MatchingLabels(labels),
+		&list)
+	if err != nil {
+		return nil, err
+	}
+	if len(list.Items) > 0 {
+		return &list.Items[0], nil
+	}
+
+	return nil, nil
+}
+
+// Determine if two imagestreams are equal.
+// Returns `true` when equal.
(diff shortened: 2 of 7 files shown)


pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+	}
+	secret.Data = map[string][]byte{
+		"access_key": []byte(credSecret.Data[migapi.AwsAccessKeyId]),
+		"secret_key": []byte(credSecret.Data[migapi.AwsSecretAccessKey]),
+	}
+	return nil
+}
+
+// Get an existing credentials Secret on the source cluster.
+func (t *Task) getRegistryCredentialsSecret(client k8sclient.Client) (*kapi.Secret, error) {
```
djwhatle: Suggest `getRegistrySecret()`

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+	if len(list.Items) > 0 {
+		return &list.Items[0], nil
+	}
+
+	return nil, nil
+}
+
+// Determine if two registry credentials secrets are equal.
+// Returns `true` when equal.
+func (t *Task) equalsRegistryCredentialsSecret(a, b *kapi.Secret) bool {
```
djwhatle: Suggest `equalsRegistrySecret()`

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+	return reflect.DeepEqual(a.Data, b.Data)
+}
+
+// Ensure the imagestream for the migration registry on the specified cluster has been created
+func (t *Task) ensureRegistryImageStream(client k8sclient.Client, registryResources *MigrationRegistryResources) error {
```
djwhatle: On the fence, possible rename `ensureRegistryIS()`. Almost looks like a method that checks for existence of a registry with that name though.

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+	}
+	if foundImageStream == nil {
+		registryResources.ImageStream = newImageStream
+		err = client.Create(context.TODO(), newImageStream)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+	registryResources.ImageStream = foundImageStream
```
djwhatle: Wondering if this is intended. it looks like `registryResources.ImageStream` gets set to newImageStream above, but that will always be rewritten by this value unless the create errors out.

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+		err = client.Update(context.TODO(), foundImageStream)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Build a Registry ImageStream as desired for the source cluster.
+func (t *Task) buildRegistryImageStream(name string) (*imagev1API.ImageStream, error) {
```
djwhatle: Again on the fence, could be `buildRegistryIS()`

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+			Name:      name,
+			Namespace: VeleroNamespace,
+		},
+	}
+	err := t.updateRegistryImageStream(imagestream)
+	return imagestream, err
+}
+
+// Update a Registry ImageStream as desired for the specified cluster.
+func (t *Task) updateRegistryImageStream(imagestream *imagev1API.ImageStream) error {
```
djwhatle: On fence, could be `updateRegistryIS()`

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+	imagestream.Spec = imagev1API.ImageStreamSpec{
+		LookupPolicy: imagev1API.ImageLookupPolicy{Local: false},
+		Tags: []imagev1API.TagReference{
+			imagev1API.TagReference{
+				Name: "2",
+				Annotations: map[string]string{
+					"openshift.io/imported-from": "registry:2",
```
djwhatle: Looks like we are hard-coding the "imported-from" registry name.  In what ways, if any, does this limit us?

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+				ImportPolicy:    imagev1API.TagImportPolicy{},
+				ReferencePolicy: imagev1API.TagReferencePolicy{Type: ""},
+			},
+		},
+	}
+	return nil
+}
+
+// Get an existing registry ImageStream on the specifiedcluster.
+func (t *Task) getRegistryImageStream(client k8sclient.Client) (*imagev1API.ImageStream, error) {
```
djwhatle: On fence, `getRegistryIS()`

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+
+// Determine if two imagestreams are equal.
+// Returns `true` when equal.
+func (t *Task) equalsRegistryImageStream(a, b *imagev1API.ImageStream) bool {
+	return reflect.DeepEqual(a.Spec.Tags[0].Name, b.Spec.Tags[0].Name) &&
+		reflect.DeepEqual(a.Spec.Tags[0].From, b.Spec.Tags[0].From)
+}
+
+// Ensure the deploymentconfig for the migration registry on the specified cluster has been created
+func (t *Task) ensureRegistryDeploymentConfig(client k8sclient.Client, registryResources *MigrationRegistryResources) error {
```
djwhatle: Could be `ensureRegistryDC()`

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+		err = client.Update(context.TODO(), foundDeploymentConfig)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Build a Registry DeploymentConfig as desired for the source cluster.
+func (t *Task) buildRegistryDeploymentConfig(name, srcName string) (*appsv1API.DeploymentConfig, error) {
```
djwhatle: Could be `buildRegistryDC()`

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+			Name:      name,
+			Namespace: VeleroNamespace,
+		},
+	}
+	err := t.updateRegistryDeploymentConfig(deploymentconfig, name, srcName)
+	return deploymentconfig, err
+}
+
+// Update a Registry DeploymentConfig as desired for the specified cluster.
+func (t *Task) updateRegistryDeploymentConfig(deploymentconfig *appsv1API.DeploymentConfig, name, srcName string) error {
```
djwhatle: Could be `updateRegistryDC()`

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+								Name: "REGISTRY_STORAGE_S3_SECRETKEY",
+								ValueFrom: &kapi.EnvVarSource{
+									SecretKeyRef: &kapi.SecretKeySelector{
+										LocalObjectReference: kapi.LocalObjectReference{Name: name},
+										Key:                  "secret_key",
+									},
+								},
+							},
+						},
+						Image: "registry:2",
```
djwhatle: Is this going to start a container running an image called `registry` with tag `2`? Can we expect `image:2` to exist on our clusters?

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+				Volumes: []kapi.Volume{
+					kapi.Volume{
+						Name:         name + "-volume-1",
+						VolumeSource: kapi.VolumeSource{EmptyDir: &kapi.EmptyDirVolumeSource{}},
+					},
+				},
+			},
+		},
+		Test: false,
+		Triggers: appsv1API.DeploymentTriggerPolicies{
```
djwhatle: What's going on with this DeploymentTrigger?

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+					},
+				},
+			},
+		},
+	}
+	return nil
+}
+
+// Get an existing registry DeploymentConfig on the specifiedcluster.
+func (t *Task) getRegistryDeploymentConfig(client k8sclient.Client) (*appsv1API.DeploymentConfig, error) {
```
djwhatle: Suggest `getRegistryDC()`

pkg/velerorunner/task.go
```diff
@@ -4,6 +4,9 @@ import (
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
 	"github.com/go-logr/logr"
 	velero "github.com/heptio/velero/pkg/apis/velero/v1"
+	appsv1API "github.com/openshift/api/apps/v1"
```
djwhatle: Suggest switch to `appsv1`

pkg/velerorunner/task.go
```diff
@@ -4,6 +4,9 @@ import (
...
+	imagev1API "github.com/openshift/api/image/v1"
```
djwhatle: Suggest switch to `imagev1`

pkg/velerorunner/task.go
```diff
@@ -35,14 +38,29 @@ const (
...
+	DestRegistryResources *MigrationRegistryResources
+}
+
+// The set of resources associated with a source or destination
+// Migration Registry.
+// CredSecret -       A secret containing the credentials for the registry storage
+// ImageStream -      The ImageStream resource pointing to the image used by the Registry DeploymentConfig
+// DeploymentConfig - The DeploymentConfig that runs the Migration Registry
+// Service          - The Service resource for the Migration Registry
+type MigrationRegistryResources struct {
```
djwhatle: Suggest rename `RegistryResources`

djwhatle: Suggest moving struct definition for this to `velerorunner/registry.go`

pkg/controller/migcluster/remote_watch.go
```diff
@@ -29,6 +30,7 @@ import (
 
 	migrationv1alpha1 "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
 	velerov1 "github.com/heptio/velero/pkg/apis/velero/v1"
+	appsv1API "github.com/openshift/api/apps/v1"
```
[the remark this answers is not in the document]
sseago: That works for me. I just took what we were using in the plugin, but there's no need for the two imports in separate projects to match.

pkg/controller/migcluster/remote_watch.go
```diff
@@ -64,6 +66,16 @@ func StartRemoteWatch(r *ReconcileMigCluster, config RemoteManagerConfig) error
 		log.Error(err, "[rWatch] Unable add Velero APIs to scheme")
 		return err
 	}
+	log.Info("[rWatch] Adding OpenShift imagestream to scheme")
+	if err := imagestreamregister.AddToScheme(mgr.GetScheme()); err != nil {
```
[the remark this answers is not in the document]
sseago: I suspect adding all OpenShift APIs would fail in the same place that adding just the image APIs would fail -- because that class is re-adding SecretList, which is causing errors for us.

pkg/imagestreamregister/imagestream_register.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/openshift/api/image/docker10"
+	"github.com/openshift/api/image/dockerpre012"
+	imagev1API "github.com/openshift/api/image/v1"
+)
+
+var (
+	GroupName     = "image.openshift.io"
+	GroupVersion  = schema.GroupVersion{Group: GroupName, Version: "v1"}
+	schemeBuilder = runtime.NewSchemeBuilder(addKnownTypes, docker10.AddToScheme, dockerpre012.AddToScheme, corev1.AddToScheme)
```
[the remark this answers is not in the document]
sseago: I could take the minimalist approach -- and include the minimum amount that runs without error. My initial approach was to do the opposite -- make as few changes from what the standard AddToScheme would do as possible. Since only one line was failing us, I just removed that line, realizing that if that line wasn't there upstream we wouldn't be repeating this code here.

pkg/imagestreamregister/imagestream_register.go
```diff
@@ -0,0 +1,52 @@
+package imagestreamregister
```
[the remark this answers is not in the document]
sseago: The only reason this is included is because I needed to change the file. The line below is removed, otherwise this is identical to the upstream file. Without this problematic line upstream, I'd just reference the remote/vendor code like we're doing with openshift apps.
https://github.com/openshift/api/blob/master/image/v1/register.go#L48

pkg/velerorunner/registry.go
sseago: I'm basically trying to recreate, as closely as possible, the deployment config definition returned by `oc new-app`. For the registry creation that worked in my dev environment, took the output produced by `oc new-app <...> --dry-run -o yaml` and reproduced it here. If there are things you think I should remove that would leave this fully-functional, I could do that, but it seemed safer as a starting point to replicate what the oc tool did.

pkg/velerorunner/registry.go
sseago: Works for me.

pkg/velerorunner/task.go
sseago: Most of my registry-related names have Migration prepended to make it clear we're not talking about the internal registry. What about `MigRegistryResources`?

pkg/velerorunner/registry.go
sseago: But to answer your question more specifically, if I'm understanding dc triggers properly, this should trigger a new pod any time the dc config changes or the associated imagestream gets a new image.

pkg/imagestreamregister/imagestream_register.go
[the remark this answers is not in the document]
djwhatle: That's right. I remember our conversation on this now. Thoughts on moving to `pkg/apis/addtoscheme_imagev1.go`? @jortel do you have any input on a good place to put this?

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+}
+
+// Determine if two deploymentconfigs are equal.
+// Returns `true` when equal.
+func (t *Task) equalsRegistryDeploymentConfig(a, b *appsv1API.DeploymentConfig) bool {
+	return reflect.DeepEqual(a.Spec.Replicas, b.Spec.Replicas) &&
+		reflect.DeepEqual(a.Spec.Selector, b.Spec.Selector) &&
+		reflect.DeepEqual(a.Spec.Template.ObjectMeta, b.Spec.Template.ObjectMeta) &&
+		reflect.DeepEqual(a.Spec.Template.Spec.Volumes, b.Spec.Template.Spec.Volumes) &&
+		reflect.DeepEqual(a.Spec.Template.Spec.Containers[0].Env, b.Spec.Template.Spec.Containers[0].Env) &&
```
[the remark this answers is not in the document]
sseago: Yeah, we probably should -- len >=1 and len is equal on both.

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+	imagev1API "github.com/openshift/api/image/v1"
+	kapi "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Ensure the migration registry on the source cluster has been created
+// and has the proper settings.
+func (t *Task) ensureSourceMigrationRegistry() error {
```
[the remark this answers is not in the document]
sseago: Yeah, I started with src and dest, but then I noticed the rest of the code spelled it out. I can shorten it, though.

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+	}
+	err = t.ensureRegistryService(client, registryResources)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Ensure the credentials secret for the migration registry on the specified cluster has been created
+func (t *Task) ensureRegistryCredentialsSecret(client k8sclient.Client, registryResources *MigrationRegistryResources) error {
```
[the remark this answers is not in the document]
sseago: Works for me.

pkg/velerorunner/registry.go
```diff
@@ -0,0 +1,526 @@
...
+			Namespace:    VeleroNamespace,
+		},
+	}
+	err := t.updateRegistryCredentialsSecret(client, secret)
+	return secret, err
+}
+
+// Update a Registry credentials secret as desired for the specified cluster.
+func (t *Task) updateRegistryCredentialsSecret(client k8sclient.Client, secret *kapi.Secret) error {
+	credSecret, err := t.PlanResources.MigStorage.Spec.BackupStorageConfig.GetCredsSecret(t.Client)
```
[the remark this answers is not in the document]
sseago: Two different clients. passed-in client is what we use to set/create src and destination resources, but t.Client here is specifically referencing the *controller* client, since that's where the MigStorage secret will be found. Otherwise this fails for remote clusters.

pkg/velerorunner/registry.go
sseago: Note the `return nil` at the end of that block. If it doesn't yet exist, it will return after setting it to newImageStream or it will error out. Otherwise, it found one already and sets it to that found value here. The backup creation follows the same logic.

pkg/velerorunner/registry.go
sseago: I'm not sure. I"m following the example of `oc new-app` functionality. I'm open to removing it, but I was trying to stick with what I know already works.

pkg/velerorunner/registry.go
sseago: We can because of the ImageStream we just created above which matches this.

jortel:
LGTM.
Nothing to add beyond what @djwhatle has noted.  Nice Job!

pkg/imagestreamregister/imagestream_register.go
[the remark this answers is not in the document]
sseago: Anywhere works for me, as long as it's a separate package. With all of the package-exported variables there, I'd be afraid that putting it in an existing package could clash (or potentially clash in the future).

pkg/velerorunner/registry.go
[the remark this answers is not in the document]
djwhatle: Right, but I don't think we are actually using the client passed in as a param here (unless I'm missing something).

pkg/velerorunner/registry.go
djwhatle: Ah, missed that. Thanks for pointing that out.

pkg/velerorunner/registry.go
djwhatle: So the `registry:2` part is from oc new-app?

pkg/velerorunner/registry.go
djwhatle: Ok. I imagine we won't be publishing new stuff to the imagestream that would trigger a redeploy. Was thinking we don't want the registry to go down during a migration.

pkg/velerorunner/task.go
djwhatle: I feel that since everything in our controller is aimed at performing migrations, it might be clear on its own. Do we expect to be interfacing with the internal OpenShift registry at any point?

pkg/velerorunner/registry.go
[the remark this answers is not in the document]
sseago: Oh, right -- good catch. I initially added the client param before I realized I needed to use t.Client, and then when I updated the other code I forgot that it wasn't in the 'update' methods that the client param was normally used (the other update methods don't have a client param at all). I wonder why Go didn't complain about this (non-used method param).

pkg/velerorunner/registry.go
sseago: We could pull it out, since I imagine this should be a pretty stable codebase anyway -- no reason we should be updating the underlying image for the registry, although we should be careful about the change. The registry will be up from the first stage migration until the final migrate. Will we eventually support controller upgrades with in-progress plans? Hopefully we could limit that -- basically if we can assume that we will never update the controller code between stage and migrate, then we can probably eliminate the image trigger. We probably want to keep the config trigger though.

pkg/velerorunner/task.go
djwhatle: Also wonder if putting `Mig` in front of it makes it sound like that's one of our CRDs, which it isn't.

sseago: The reason I hesitate on dropping the prefix completely is that we are still interacting with the internal registry in the plugin (which is where this registry gets used) -- basically the point of the migration registry is for the backup plugin to copy images from internal to migration registry, and for the restore plugin to copy images from migration registry to internal registry. Also, the annotation that we set on the velero CRs as a result of all of this is "openshift.io/migration-registry". What about MRegistry? It looks odd, but it won't be confused with "Mig-" names. I still like "Mig" or "MIgration" better than "M", but any of the above is better than omitting it. Also, we could go back to the longer name -- maybe that's OK if I make all the other name-shortening changes suggested.

djwhatle: Oh I see. What if we went back to using the full name then?

pkg/velerorunner/registry.go
dymurray: Yeah I think you both bring up good points. I would leave the config trigger and remove the build trigger since config changes are not likely to occur during the migration run whereas a build trigger I could see the problem arising. But honestly, even if you leave the build trigger I don't think it would cause any problems, just more to be safe. OCP should scale up a pod before taking the old one down.

pkg/velerorunner/registry.go
sseago: I guess I'll keep the ImageStream component for now.

sseago:
Updated with changes. Haven't tested the updated PR yet since I'm hitting rebase-related problems right now.

sseago:
OK, looks like the updated PR works, but now I have to merge/rebase and test again with the annotation PR merged.

PR #77, opened 2019-05-15
Manage PV list on plan.

jortel:
Add PV discovery.

Manage a list of `PersistentVolume` on `MigPlan`.  The list is generated based on the PVC on Pods within the asset-collection namespaces on the source cluster.  PVs referenced by the PVCs are included in the list.  The `action` of _copy_ or _move_ must be set by the User.

The `PvsDiscovered` condition is set to indicate that the PV discovery has been performed.
The `InvalidPvAction` condition is a _Warn_ because it should should prevent the reconcile from progressing.  However, it does block the `Ready` condition.

Needs: https://github.com/fusor/mig-controller/pull/82


pkg/controller/migplan/pvlist.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"context"
+	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
+	migref "github.com/fusor/mig-controller/pkg/reference"
+	core "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type PvTable map[types.NamespacedName]core.PersistentVolume
```
djwhatle: Thoughts on PvMap vs PvTable here? Seems like a map

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -144,3 +145,89 @@ func (r *MigPlan) GetRefResources(client k8sclient.Client) (*PlanResources, erro
...
+type PersistentVolume struct {
+	Name         string `json:"name,omitempty"`
+	StorageClass string `json:"storageClass,omitempty"`
+	Action       string `json:"action,omitempty"`
+	staged       bool
+}
+
+// Collection of PVs
+type PersistentVolumes struct {
+	Content []PersistentVolume `json:"persistentVolumes"`
```
dymurray: Any thoughts on making this a map? Since everytime you are adding a new PV you have to loop through content to see if it already exists. 

Then `FindPv` becomes:
```
return r.Content[pv.Name]
```

Probably not a huge deal but if we are doing a migration with 100 PVs this could affect performance

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -25,6 +25,7 @@ import (
 
 // MigPlanSpec defines the desired state of MigPlan
 type MigPlanSpec struct {
+	PersistentVolumes
```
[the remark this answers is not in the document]
jortel: Yes, so it could be a receiver and provide specific behavior for managing the collection.

pkg/apis/migration/v1alpha1/migplan_types.go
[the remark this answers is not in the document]
jortel: Mainly to avoid the awkward redundant semantics of PersistentVolumes.PersistentVolumes.  By name, the struct is a collection.  The purpose of the struct field tags is specifically to support the names being different.

jortel: It's a `slice` primarily because of how it will serialize in the API.  Agreed that a map would be more efficient.  I'll play with it and see what I can do.

pkg/controller/migplan/pvlist.go
jortel: Agreed. `Map` naming and semantics is better.

pkg/apis/migration/v1alpha1/migplan_types.go
jortel: Actually, I have an idea about how to do this easily.  I'll play with it.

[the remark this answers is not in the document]
djwhatle: Fair enough, I hadn't thought about that (re: purpose of struct field tags)

dymurray: np!

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -144,3 +145,121 @@ func (r *MigPlan) GetRefResources(client k8sclient.Client) (*PlanResources, erro
...
+	if r.index == nil {
+		r.buildIndex()
+	}
+}
+
+// Build the index.
+func (r *PersistentVolumes) buildIndex() {
+	r.index = map[string]int{}
+	for i, pv := range r.List {
+		r.index[pv.Name] = i
```
djwhatle: Looks like @dymurray's idea on using a map got implemented after all! Guess you figured out how to do it without switching the user-provided data type to a map.

pkg/apis/migration/v1alpha1/migplan_types_test.go
```diff
@@ -56,3 +57,200 @@ func TestStorageMigPlan(t *testing.T) {
 	g.Expect(c.Delete(context.TODO(), fetched)).NotTo(gomega.HaveOccurred())
 	g.Expect(c.Get(context.TODO(), key, fetched)).To(gomega.HaveOccurred())
 }
+
+func TestPersistentVolume_Update(t *testing.T) {
```
djwhatle: Is that... a test? :+1:

pkg/controller/migplan/validation.go
```diff
@@ -28,14 +29,20 @@ const (
 	InvalidDestinationCluster      = "InvalidDestinationCluster"
 	NsNotFoundOnSourceCluster      = "NamespaceNotFoundOnSourceCluster"
 	NsNotFoundOnDestinationCluster = "NamespaceNotFoundOnDestinationCluster"
-	StorageEnsured                 = "StorageEnsured"
+	// Warn
+	PvInvalidAction = "PvInvalidAction"
```
djwhatle: I see that this is a warning. Do we select a default action in this case?

pkg/apis/migration/v1alpha1/migplan_types.go
jortel: @dymurray I added an _index_ which seemed less complex than a custom field tag that would be needed to json serialize a `map` as an array.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -144,3 +145,121 @@ func (r *MigPlan) GetRefResources(client k8sclient.Client) (*PlanResources, erro
...
+	staged       bool
+}
+
+// Update the PV with another.
+func (r *PersistentVolume) Update(pv PersistentVolume) {
+	r.StorageClass = pv.StorageClass
+	r.staged = true
+}
+
+// Collection of PVs
```
djwhatle: Thanks for the comment explaining the process here.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -144,3 +145,121 @@ func (r *MigPlan) GetRefResources(client k8sclient.Client) (*PlanResources, erro
...
+// List - The collection of PVs.
+// index - List index.
+// --------
+// Example:
+// plan.Spec.BeginPvStaging(false)
```
djwhatle: What does the "false" mean here?

pkg/controller/migplan/validation.go
```diff
@@ -59,7 +67,11 @@ const (
 	InvalidDestinationClusterMessage      = "The `srcMigClusterRef` and `dstMigClusterRef` cannot be the same."
 	NsNotFoundOnSourceClusterMessage      = "Namespaces [%s] not found on the source cluster."
 	NsNotFoundOnDestinationClusterMessage = "Namespaces [%s] not found on the destination cluster."
-	StorageEnsuredMessage                 = "The storage resources have been created."
+	// Warn
+	PvInvalidActionMessage = "PV in `persistentVolumes` [%s] has invalid `action`. Must be (%s)."
+	// Required.
+	StorageEnsuredMessage = "The storage resources have been created."
+	PvsDiscoveredMesssage = "The `persistentVolumes` list has been updated with discoverd PVs."
```
djwhatle: Typo on spelling of "discovered"

pkg/controller/migplan/validation.go
```diff
@@ -448,3 +466,38 @@ func (r ReconcileMigPlan) validateDestinationNamespaces(plan *migapi.MigPlan) er
...
+		}
+		message := fmt.Sprintf(
+			PvInvalidActionMessage,
+			strings.Join(list, ", "),
+			strings.Join(choices, "|"))
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     PvInvalidAction,
+			Status:   True,
+			Reason:   NotDone,
+			Category: migapi.Warn,
```
djwhatle: Still looking for where we set the default action...

pkg/controller/migplan/validation.go
```diff
@@ -448,3 +466,38 @@ func (r ReconcileMigPlan) validateDestinationNamespaces(plan *migapi.MigPlan) er
...
+func (r ReconcileMigPlan) validatePvAction(plan *migapi.MigPlan) error {
+	list := []string{}
+	actions := map[string]bool{
+		migapi.PvCopyAction: true,
+		migapi.PvMoveAction: true,
+	}
+	for _, pv := range plan.Spec.PersistentVolumes.List {
+		_, found := actions[pv.Action]
+		if !found {
+			list = append(list, pv.Name)
```
djwhatle: Looks like `list` will contain all PVs with _invalid_ actions set

pkg/controller/migplan/validation.go
```diff
@@ -448,3 +466,38 @@ func (r ReconcileMigPlan) validateDestinationNamespaces(plan *migapi.MigPlan) er
...
+		}
+	}
+	if len(list) > 0 {
+		choices := []string{}
+		for key := range actions {
+			choices = append(choices, key)
```
djwhatle: Here it looks like we're transforming our map of actions into a list containing the keys from the map: `migapi.PvCopyAction` and `migapi.PvMoveAction`

pkg/controller/migplan/validation.go
```diff
@@ -448,3 +466,38 @@ func (r ReconcileMigPlan) validateDestinationNamespaces(plan *migapi.MigPlan) er
...
+		}
+		message := fmt.Sprintf(
+			PvInvalidActionMessage,
+			strings.Join(list, ", "),
+			strings.Join(choices, "|"))
+		plan.Status.SetCondition(migapi.Condition{
```
djwhatle: Are we setting one condition for each invalid PV choice here?

pkg/controller/migplan/validation.go
```diff
@@ -448,3 +466,38 @@ func (r ReconcileMigPlan) validateDestinationNamespaces(plan *migapi.MigPlan) er
 
 	return nil
 }
+
+// Validate PV actions.
+func (r ReconcileMigPlan) validatePvAction(plan *migapi.MigPlan) error {
+	list := []string{}
```
djwhatle: This is a list of PVs with invalid actions selected.

pkg/controller/migplan/pvlist.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+			table[key] = pv
+		}
+	}
+
+	return table, nil
+}
+
+// Get a list of PVCs found on pods with the specified namespaces.
+func (r *ReconcileMigPlan) getClaims(client k8sclient.Client, namespaces []string) (Claims, error) {
```
djwhatle: So to get the `PersistentVolumeClaims`, we have to look at all of the pods in a namespace. Good to see this written out, helped me understand.

pkg/controller/migplan/pvlist.go
```diff
@@ -0,0 +1,117 @@
...
+	claims := Claims{}
```
djwhatle: @jortel Does this work? I would expect that you would need to initialize this map before using it.

pkg/apis/migration/v1alpha1/migplan_types.go
jortel: That is a _typo_.  Fixing.

pkg/controller/migplan/validation.go
jortel: No, it's a warning because it requires the user to resolve (choose an action).  It prevents the `Ready` condition but isn't something badly broken.  We _may_ be able to choose another category to make this clearer but nothing came to mind.

pkg/controller/migplan/pvlist.go
jortel: Which map?

pkg/controller/migplan/validation.go
jortel: No, it's a single condition listing all the PVs with invalid actions.

pkg/controller/migplan/validation.go
jortel: There is no default.  The user needs to set the action.

pkg/controller/migplan/validation.go
djwhatle: Got it. To me, `Warning` in the context of software is something that can be optionally ignored if a user / dev chooses to do so. 

If this stops the Migration from being able to run it seems like more than a warning, at least by my definition.

pkg/controller/migplan/validation.go
jortel: Renamed to `invalid`.

pkg/controller/migplan/validation.go
jortel: Fair enough.  Perhaps Category=`Incomplete` would be better.  Other ideas?

pkg/controller/migplan/pvlist.go
djwhatle: I was thinking `Claims{}` was declaring a map of claims but after looking closer I think this is just declaring a `Claims` instance with no fields set.

pkg/controller/migplan/validation.go
jortel: Adding a `Critical` condition in another PR which will be used like `Error` is now for validation errors.  Then will change this to use `Error`.

pkg/apis/migration/v1alpha1/migplan_types.go
dymurray: +1 Looks great!

PR #78, opened 2019-05-16
Add stage/migrate annotations to backup/restore

### LINKED ISSUE #74: Controller should be setting annotations on backup/restore
The plugin depends on annotations to function properly. The list of annotations to set are here: https://github.com/fusor/ocp-velero-plugin/blob/master/velero-plugins/common/types.go#L28

To solve this I think it makes the most sense to add a field to the `Task` struct `Annotations` that the stage and migration controllers can set before running `Task.Run()`.

dymurray:
https://github.com/fusor/ocp-velero-plugin/blob/master/velero-plugins/common/types.go#L31

This annotation is required for the plugin to work properly.

Fixes #74

### DIFF
--- a/pkg/controller/migmigration/migrate.go
+++ b/pkg/controller/migmigration/migrate.go
@@ -23,6 +23,7 @@ import (
 )
 
 var migrateAnnotationKey = "openshift.io/migrate-copy-phase"
+var pvAnnotationKey = "openshift.io/migrate-type"
 var migrateAnnotationValue = "final"
 
 func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (bool, error) {
@@ -53,6 +54,11 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (bool, e
 	// Build annotations
 	annotations := make(map[string]string)
 	annotations[migrateAnnotationKey] = migrateAnnotationValue
+	// TODO: Revisit this. We are hardcoding this for now until 2 things occur.
+	// 1. We are properly setting this annotation from user input to the UI
+	// 2. We fix the plugin to operate migration specific behavior on the
+	// migrateAnnnotationKey
+	annotations[pvAnnotationKey] = "swing"
 
 	// Run
 	planResources, err := plan.GetRefResources(r)
--- a/pkg/controller/migmigration/migrate.go
+++ b/pkg/controller/migmigration/migrate.go
@@ -58,7 +58,7 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (bool, e
 	// 1. We are properly setting this annotation from user input to the UI
 	// 2. We fix the plugin to operate migration specific behavior on the
 	// migrateAnnnotationKey
-	annotations[pvAnnotationKey] = "swing"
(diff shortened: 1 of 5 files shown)


pkg/controller/migmigration/migrate.go
```diff
@@ -22,6 +22,9 @@ import (
 	vrunner "github.com/fusor/mig-controller/pkg/velerorunner"
 )
 
+var migrateAnnotationKey = "openshift.io/migrate-copy-phase"
```
dymurray: If you guys have any suggestions on a way to share this as a constant between the two controllers please let me know. I don't think putting it in the `migapi` package makes sense so wasn't sure where else to put it.

sseago: How about a "common" package (or something similar) like we do with the velero plugin?

sseago:
It looks like you're missing "openshift.io/migrate-type" -- if this isn't set, the imagestream plugin won't copy images.

dymurray:
In reference to Scott's comment, we discussed in slack and I am going to set `migrate-type` to `swing` for now just so that the plugin doesn't break. We need to revisit this in the plugin to operate on the proper annotations because these annotations have changed meaning over time.

pkg/controller/migstage/stage.go
```diff
@@ -23,6 +23,9 @@ import (
 )
 
 var stageResources = []string{"pods", "persistentvolumes", "persistentvolumeclaims", "imagestreams", "imagestreamtags"}
+var stageAnnotationKey = "openshift.io/migrate-copy-phase"
+var pvAnnotationKey = "openshift.io/migrate-type"
+var stageAnnotationValue = "stage"
```
jortel: Perhaps a `const` would be better?

pkg/velerorunner/task.go
```diff
@@ -43,6 +43,7 @@ type Task struct {
 	Phase           string
 	Backup          *velero.Backup
 	Restore         *velero.Restore
+	Annotations     map[string]string
```
jortel: Please document the `Annotations` field.

Also, (I doubt this was obvious and me being OCD but ..) I have been trying to group the required fields passed in on construction together and putting the _transient_ fields at the end.

pkg/controller/migmigration/migrate.go
jortel: We've deliberately stayed away from packages and .go files named _common_, _util_, _helper_ etc thinking it's better to use meaningful names.  Also, to avoid them being a catch for things that have appropriate homes.

If this annotation is common for both stage and migration.  how about having the `Task` inject it?

sseago: So the existence of the annotation is common, but the value is different in stage and migration.

pkg/controller/migmigration/migrate.go
```diff
@@ -22,6 +22,10 @@ import (
...
+var pvAnnotationKey = "openshift.io/migrate-type"
```
djwhatle: Feel all of these should be constants unless I'm missing something.

pkg/velerorunner/task.go
djwhatle: I did not realize you were doing this, good to know.

PR #82, opened 2019-05-17
Add condition categories.

jortel:
In the controller, there is no formal distinction between error conditions and non-error conditions even though all of the validation conditions are _error_ conditions and the `Ready` condition is a non-error condition.  By adding the `Category` field to conditions, the controller, the UI and CLI users can distinguish between conditions that are _good_ and those that are _bad_.  For example, In the controller, the `Ready` condition can be set by checking to see if required _good_ conditions exist AND that no _bad_ (or Error) conditions exist.  This reduces complexity in method return values.  Further, code can understand the condition of a CR (beyond IsReady) with more detail to make more informed decisions.  To support this, the `Conditions` API has been augmented with `HasCondition()` like methods.  As a result, it's no longer useful for validation methods to return the number-of-conditions set which reduced complexity.

This PR also improves both the semantics and functionality of the condition _staging_ concept with explicit `BeginStagedConditions()` and `EndStagedConditions()`.

Here is a example of how _good_ conditions can be more appropriate than all conditions being _bad_. The plan controller use to set an error condition when the BSL/VSLs have not been created (or failed to create) on the source and destination clusters.  However, rather than this being a _error_ condition, having ensured (created) the BSL/VSLs is really a _good_ condition that is _required_ for the `Ready` condition.  This PR makes that change and it's a good example of non-error conditions.

Another use case for _good_ conditions is the plan controller signalling that the PV discovery has been performed and the Plan CR has been updated with the list of discovered PVs.  Another related condition is that the _user_ has selected the (copy|move) action on each PV.

This PR is needed for PV discovery.


pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -16,6 +16,16 @@ const (
 	False = "False"
 )
 
+// Category
+// Required - Required for the `Ready` condition.
+// Error - Errors that block the `Ready` condition.
+// Warning - Warnings that block the `Ready` condition.
+const (
+	Error    = "Error"
```
djwhatle: This introduces condition categories:
- `Error`
- `Required`
- `Warn`

@jortel if these are our only conditions, what category do we assign to good conditions? Do all good conditions fall under `Required`?

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -40,110 +51,193 @@ func (r *Condition) Update(other Condition) {
...
 		r.Status == other.Status &&
+		r.Category == other.Category &&
 		r.Reason == other.Reason &&
 		r.Message == other.Message
 }
 
 // Managed collection of conditions.
 // Intended to be included in resource Status.
+// List - The list of conditions.
+// staged - In `staged` mode, the search methods like
```
djwhatle: Refreshing my memory of why we have the concept of "staged" conditions... If I recall correctly, the crux is that it allows us to keep our old `lastTransitionTime` if nothing changes.

Going to keep reading and try to understand why we have the additional `BeginStagedConditions()` and `EndStagedConditions()` now.

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -40,110 +51,193 @@ func (r *Condition) Update(other Condition) {
...
+//     "Resource Ready.")
+//
 type Conditions struct {
-	Conditions []Condition `json:"conditions"`
+	List   []Condition `json:"conditions"`
+	staged bool
+}
+
+// Begin staged conditions.
+func (r *Conditions) BeginStagedConditions() {
```
djwhatle: So this sets an entire group of conditions as `staged`. What does that mean?

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -40,110 +51,193 @@ func (r *Condition) Update(other Condition) {
...
+	}
+	for index := range r.List {
+		condition := &r.List[index]
+		condition.staged = false
+	}
+}
+
+// End staged conditions.
+func (r *Conditions) EndStagedConditions() {
+	r.staged = false
```
djwhatle: This sets an entire group of conditions as `staged = false`.  Not sure what that means.

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -40,110 +51,193 @@ func (r *Condition) Update(other Condition) {
...
+		return
+	}
+	for index := range r.List {
+		condition := &r.List[index]
+		condition.staged = false
+	}
+}
+
+// End staged conditions.
+func (r *Conditions) EndStagedConditions() {
```
djwhatle: This method appears to discard all conditions in r.List that are not marked as `staged`. I wonder if the comment could be more informative.

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -40,110 +51,193 @@ func (r *Condition) Update(other Condition) {
...
+//     !thing.Status.HasBlockerCondition(),
+//     "Resource Ready.")
+//
 type Conditions struct {
-	Conditions []Condition `json:"conditions"`
+	List   []Condition `json:"conditions"`
+	staged bool
+}
+
+// Begin staged conditions.
```
djwhatle: This method appears to set each individual `condition` in a group of Conditions `r` to be `staged = false`, but sets the overall var `Conditions` to `staged = true`. Maybe a more detailed comment would help explain what's going on here?

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -40,110 +51,193 @@ func (r *Condition) Update(other Condition) {
...
 		}
+		return true
 	}
-	r.Conditions = kept
+
+	return false
+}
+
+// The collection contains an `Error` condition.
+func (r *Conditions) HasErrorCondition(category ...string) bool {
```
djwhatle: These are nice to have!

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -40,110 +51,193 @@ func (r *Condition) Update(other Condition) {
...
 // Set `Ready` condition.
 func (r *Conditions) SetReady(ready bool, message string) {
 	if ready {
 		r.SetCondition(Condition{
-			Type:    Ready,
-			Status:  True,
-			Message: message,
+			Type:     Ready,
+			Status:   True,
+			Category: Required,
```
djwhatle: So do we check that at least one condition with `Category: Required` is set on a resource before proceeding with Migration? I haven't seen this in your PR but will look further.

djwhatle: I'm conflicted on our "good" category being called `Required`. Would we ever want to show a "good" condition that isn't "required"?

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -40,110 +51,193 @@ func (r *Condition) Update(other Condition) {
...
 		r.DeleteCondition(Ready)
 	}
 }
 
-// Get if `Ready` condition is `True`.
+// The collection contains the `Ready` condition.
 func (r *Conditions) IsReady() bool {
-	_, condition := r.FindCondition(Ready)
-	return condition != nil && condition.Status == True
+	condition := r.FindCondition(Ready)
```
djwhatle: It doesn't look like we use the new "Required" category in any logic, I suppose it's just there to provide parity with the other conditions that also have a category.

pkg/controller/migassetcollection/migassetcollection_controller.go
```diff
@@ -111,15 +111,15 @@ func (r *ReconcileMigAssetCollection) Reconcile(request reconcile.Request) (reco
 	}
 
 	// Validations
-	_, err = r.validate(assetCollection)
+	err = r.validate(assetCollection)
 	if err != nil {
-		return reconcile.Result{}, err
+		if errors.IsConflict(err) {
+			return reconcile.Result{Requeue: true}, nil
```
djwhatle: +1 clean logs

pkg/controller/migassetcollection/validation.go
```diff
@@ -28,39 +28,48 @@ const (
...
 	// Empty collection
-	nSet, err := r.validateEmpty(assetCollection)
+	err := r.validateEmpty(assetCollection)
 	if err != nil {
-		return 0, err
+		return err
+	}
+
+	// Apply changes
+	assetCollection.Status.EndStagedConditions()
```
djwhatle: Maybe a better name would be `WriteStagedConditions()` or `ApplyStagedConditions()` or `CommitStagedConditions()`. `End` doesn't seem obvious to me.

pkg/controller/migassetcollection/validation.go
```diff
@@ -28,39 +28,48 @@ const (
 )
 
 // Validate the asset collection resource.
-// Returns error and the total error conditions set.
-func (r ReconcileMigAssetCollection) validate(assetCollection *migapi.MigAssetCollection) (int, error) {
-	totalSet := 0
+func (r ReconcileMigAssetCollection) validate(assetCollection *migapi.MigAssetCollection) error {
+	assetCollection.Status.BeginStagedConditions()
```
djwhatle: Thoughts on `EmptyStagedConditions()`? Is this <<< an accurate depiction of what `BeginStagedConditions()` does?

pkg/controller/migplan/storage.go
```diff
@@ -9,20 +9,20 @@ import (
 // Create the velero BackupStorageLocation(s) and VolumeSnapshotLocation(s)
 // have been created on both the source and destination clusters associated
 // with the migration plan.
-// Returns `true` when ensured.
```
djwhatle: +1

pkg/apis/migration/v1alpha1/condition.go
djwhatle: Would a better name be "staging"? If this is meant to communicate that the process of staging conditions is in progress.

djwhatle: Would `staging` perhaps be a better choice for the overall bool?

pkg/apis/migration/v1alpha1/condition.go
djwhatle: See above, wonder if `staging` would make this clearer if it's doing what I think it is.

pkg/apis/migration/v1alpha1/condition.go
jortel: Yes, I thought they would.

pkg/apis/migration/v1alpha1/condition.go
jortel: Not exactly.  I keeps the controller from having to explicitly deleting conditions that not longer exist.  The condition is marked as staged=true by SetCondition().  Then, EndStagedConditions() removes the conditions that were not marked as staged.

pkg/apis/migration/v1alpha1/condition.go
jortel: This puts the `Conditions` collection in _staging_ mode and _un-stages_ all the conditions. (Condition.staged=false).

pkg/apis/migration/v1alpha1/condition.go
jortel: No, This puts the `Conditions` collection in _non-staging_ mode and deletes any of the _un-staged_ conditions (Condition.staged=false).

pkg/apis/migration/v1alpha1/condition.go
jortel: Good suggestion.

pkg/apis/migration/v1alpha1/condition.go
jortel: Yes.  Since the controller will ensure that specific _required_ exceptions are set, a `HasRequiredCondition()` did not seem useful.

pkg/controller/migassetcollection/validation.go
jortel: Using `End` because it _ends_ `Conditions` staging mode.

Perhaps instead of `Conditions.staging` _bool_, `Conditons.mode` _string_ (Normal|Staging) would be clearer?  This seems to be causing confusion.

pkg/controller/migassetcollection/validation.go
jortel: This used to be `DeleteUnstagedConditions()` but since it also changes the _mode_, I thought this was clearer.  Not thinking that `Empty` fits.

pkg/controller/migassetcollection/validation.go
djwhatle: "Using End because it ends Conditions staging mode."
^ This makes sense to me. Would prefer we call the func `EndStaging` rather than `EndStaged`.

I like `Conditions.mode (normal|staging)` as you suggest!

pkg/controller/migassetcollection/validation.go
djwhatle: Ah I remember `DeleteUnstaged`. Yeah empty doesn't seem like a good fit knowing that.

pkg/controller/migassetcollection/validation.go
jortel: On 2nd thought, a binary mode may be overkill.  I like the _staging_ naming suggestion though.  How about we go with `Conditions.staging` and `BeginStagingConditions`, `EndStagingConditions`?

djwhatle: @jortel sounds good

PR #97, opened 2019-05-21
Consolidated MigMigration and MigStage.

jortel:
Adds MigMigration.staging.
Deletes MigStage.
Deletes velerorunner package.


jortel:
> Looks good to me. I had a minor question about terminology -- it looks like some things have shifted from "stage" to "staging" with the move here. Is that intentional?

Yes.  _stage_ seemed  more like a noun than an adjective.  That said, I don't feel strongly and a little on the fence myself.

sseago:
Yeah, I'm fine either way too. Whatever will cause the least amount of confusion when looking at UI and API together probably makes the most sense.

jortel:
After discussion, on forum-mig, it seems like consensus is `MigMigration.Stage` but make it _required_.
Pushed.

config/samples/mig-migration.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   name: migmigration-sample
   namespace: mig
 spec:
-
+  staging: false
```
dymurray: This needs to be changed to `stage`

jortel: Fixed.

PR #101, opened 2019-05-22
Remove migration registry after final migration

sseago:
This commit removes the migration registry on src and
dest clusters after final migration has completed.
Registry is not removed after stage migration.

### DIFF
--- a/pkg/controller/migmigration/registry.go
+++ b/pkg/controller/migmigration/registry.go
@@ -539,6 +539,10 @@ func (t *Task) equalsRegistryService(a, b *kapi.Service) bool {
 // Returns the right backup/restore annotations including registry-specific ones
 func (t *Task) getAnnotations(registryResources *MigRegistryResources) (map[string]string, error) {
 	annotations := t.Annotations
+	// registryResources will only be nil when reconcile is called after restore is complete
+	if registryResources == nil || registryResources.Service == nil {
+		return annotations, nil
+	}
 	if len(registryResources.Service.Spec.Ports) == 0 {
 		return nil, errors.New("Migration Registry service port not found")
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/migmigration/task.go
```diff
@@ -98,11 +98,16 @@ func (t *Task) Run() error {
 	if t.Phase == Started || t.Phase == WaitOnResticRestart {
 		return nil
 	}
+	// If restore is already done, don't ensure migration registry creation, since
+	// the next step is to remove it.
+	skipMigRegistry := t.Phase == RestoreCompleted || t.Phase == Completed
```
djwhatle: Looking at our phase list

```
	Started                 = ""
	WaitOnResticRestart     = "WaitOnResticRestart"
	ResticRestartCompleted  = "ResticRestartCompleted"
	BackupStarted           = "BackupStarted"
	BackupCompleted         = "BackupCompleted"
	BackupFailed            = "BackupFailed"
	WaitOnBackupReplication = "WaitOnBackupReplication"
	BackupReplicated        = "BackupReplicated"
	RestoreStarted          = "RestoreStarted"
	RestoreCompleted        = "RestoreCompleted"
	RestoreFailed           = "RestoreFailed"
	Completed               = "Completed"
```

Is it intended that we would _not_ skip creation of the migration registry if we were in `RestoreFailed` state?

pkg/controller/migmigration/registry.go
```diff
@@ -539,6 +539,10 @@ func (t *Task) equalsRegistryService(a, b *kapi.Service) bool {
 // Returns the right backup/restore annotations including registry-specific ones
 func (t *Task) getAnnotations(registryResources *MigRegistryResources) (map[string]string, error) {
 	annotations := t.Annotations
+	// registryResources will only be nil when reconcile is called after restore is complete
+	if registryResources == nil {
+		return annotations, nil
+	}
 	if len(registryResources.Service.Spec.Ports) == 0 {
```
djwhatle: I know this isn't part of the current PR, but curious if it's safe to assume that `registryResources.Service.Spec` will exist here... Any possibility this will eval to nil and crash the controller?

pkg/controller/migmigration/registry.go
```diff
@@ -553,3 +557,121 @@ func (t *Task) getAnnotations(registryResources *MigRegistryResources) (map[stri
...
+	// Delete src Registry
+	err = t.ensureSrcMigRegistryDelete()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Ensure that the migration registry is deleted on the src cluster after restore
+func (t *Task) ensureSrcMigRegistryDelete() error {
```
jortel: The _ensure_ naming conventoin used in the controller implies create or update in the declarative sense.  IMHO both ensure & delete is a little confusing.  Also, I don't think there is a `MigRegistry` resource.  Is there?  Suggest something like: `deleteSrcRegistry()`

pkg/controller/migmigration/task.go
djwhatle: Also important to note that if we ever add more phases between `RestoreCompleted` and `Completed`, we will need to keep adding them to this list of phases in the conditional otherwise a registry will get recreated late in the migration process.

```
skipMigRegistry := t.Phase == RestoreCompleted || t.Phase == Completed
```

Sub-optimal IMO, guessing this is what the k8s API convention doc was trying to help us avoid when they told us not to use phases.

_But_ since we're using phases now, what do you think about defining a phase or two for creation and deletion of the MigRegistry? There may be problems with this idea, maybe you can poke some holes in it.

pkg/controller/migmigration/registry.go
djwhatle: Perhaps a check for existence of `registryResources.Service != nil` would be enough to prevent this.

pkg/controller/migmigration/registry.go
sseago: Doesn't "delete" have the same problems as "create" -- what I really mean is "ensure that it's deleted" -- "delete" unless it's not there, in which case do nothing. In other words <the action of this method> is to "delete" as "ensure" is to "create". I'm not sure what the right word is. Ensure isn't quite right, as you point out. Delete doesn't seem to work either. In plain English, the action is "ensure deletion of", as opposed to "delete" because it is a declarative deletion.

sseago: So we @djwhatle and I had the "MigRegistry" vs "MigrationRegistry" vs "Registry" naming discussion around my last PR. I'm trying to avoid just "Registry" to avoid any confusion with the internal registry. I originally had "MigrationRegistry" but @djwhatle was concerned that it was too long, so we ultimately agreed on "MigRegistry" -- in any case that's what is used throughout the creation code, so if we want to change it, it would need to be changed there too, if there's some other naming that makes more sense.

pkg/controller/migmigration/task.go
sseago: I didn't worry about the failure situation since as of right now it's never set. Since I wasn't entirely sure around the dynamics of when it would be set and if/how the Run method would be re-entered after setting it, I left it out for now. However, you're right that we will need to deal with failed cases here once failure is being set. Handling it here is secondary to figuring out under which circumstances we would delete the registry on failure. The `skipMigRegistry` bit exists only to prevent us from recreating the registry after deleting it if for some reason we exit a reconcile after we start deleting registry components but before the migration is marked done.

sseago: Adding registry-related phases is probably a good idea, but it doesn't really change the nature of the problem -- it really just changes the `RestoreCompleted` comparison above into a `DeleteRegistryStarted` comparison. Basically we need to make sure we don't run any of the registry ensure methods once the task state has moved as far as starting registry deletion -- so for now this means `RestoreCompleted` or later (or, possibly, `RestoreFailed`, but again, that depends on whether we want to delete registry on restore failure or whether we want to allow another restore attempt. That's why I'm not touching failure cases yet), but if we add registry-related states, then `RegistryDeleteStarted` or later)

pkg/controller/migmigration/registry.go
sseago: So the way the code is now, if the registry ensure methods have run, then we will have `registryResources.Service.Spec`. I had to add the nil check because if reconcile is restarted after registry deletion (and after setting `skipMigRegistry`), then registryResources will be nil. It might be safer to check for nil here too, but, in general, it seems that we're only checking for nil in cases where there's code that allows it to be nil.

sseago: But yeah, I guess I could replace `if registryResources == nil` with `if registryResources == nil || registryResources.Service == nil` and just return without adding a registry hostname annotation if there is no Service here.

pkg/controller/migmigration/task.go
jortel: :+1:  adding explicit registry phases.

Between restic pod management and registry management, it looks being able to do >,<,= comparison on ordered phases would be handy.

Had this on the back burner for a bit: https://github.com/fusor/mig-controller/pull/109  Thoughts?

sseago: I'll look at that PR. Ordered phases is an improvement over the raw string comparison we have now, although things might break down a bit when the workflow ceases to be completely linear. For example where do BackupFailed and RestoreFailed fit into the comparison scheme?

sseago: Looks like we may need some combination of a comparison (i.e. `t.Phase >= RegistryDeleteStarted`) and explicit comparison for failed cases. Depending on what we do about failed tasks and the registry, we may or may not want to avoid creating the registry in one of those failure states, which may or may not match what we're doing in the non-failure comparison.

pkg/controller/migmigration/registry.go
jortel: > Doesn't "delete" have the same problems as "create" -- what I really mean is "ensure that it's deleted" -- "delete" unless it's not there, in which case do nothing.

Thanks for the explanation, @sseago.  You have convinced me.

jortel: > So @djwhatle and I had the "MigRegistry" vs "MigrationRegistry" vs "Registry" naming discussion around my last PR. I'm trying to avoid just "Registry" to avoid any confusion with the internal registry.

I don't see any way this could be confused with the internal registry but find the `MigRegistry` confusing because it follows our CR naming convention.  That said, if you settled this with @djwhatle already, let's leave it for now.  It's not worth holding up this PR.

sseago:
Updated to add the additional `registryResources.Service == nil` check and to add task phases for src and dest registry creation and registry deletion.

sseago:
This is going to have to wait on the underlying migration registry redesign -- all of this is moving to the MigPlan controller and reconcile functionality. Once the already merged code is moved over, an alternate version of this functionality can be re-submitted.

PR #103, opened 2019-05-22
Add PV supported-actions.

jortel:
Provide a list of supported PV actions based on the how the PV is configured.
This is the first iteration.  Future iterations will:
- take other storage types into account.
- validate that the target cluster can access network (shared) storage such as NFS.


pkg/controller/migplan/pvlist.go
```diff
@@ -115,3 +119,23 @@ func (r *ReconcileMigPlan) getClaims(client k8sclient.Client, namespaces []strin
...
+	if pv.Spec.HostPath != nil {
+		return []string{}
+	}
+	if pv.Spec.NFS != nil ||
+		pv.Spec.Glusterfs != nil ||
+		pv.Spec.AzureDisk != nil ||
+		pv.Spec.AzureFile != nil ||
+		pv.Spec.AWSElasticBlockStore != nil {
+		return []string{
+			migapi.PvCopyAction,
```
dymurray: This looks correct. Move and Copy are supported for NFS, Gluster, Azure, and EBS. What about GCP? Or do we only care about these for now?

jortel: Only care about these for now.

PR #108, opened 2019-05-22
Annotate MigCluster with UUID to avoid stall on OCP4

djwhatle:
OCP4 seems to be smarter about not bumping the `resourceVersion` when we don't actually change anything in an `update` call. Event propagation relies on the `resourceVersion`, so to get around this, I've started annotating a reconcile-unique UUID to the MigCluster object. 

This fixes stalled Migration jobs on OCP 4 clusters.

### DIFF
--- a/pkg/controller/migcluster/migcluster_controller.go
+++ b/pkg/controller/migcluster/migcluster_controller.go
@@ -24,10 +24,12 @@ import (
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
 	migref "github.com/fusor/mig-controller/pkg/reference"
 	"github.com/fusor/mig-controller/pkg/remote"
+	"github.com/google/uuid"
 	kapi "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -130,6 +132,9 @@ func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{}, err // requeue
 	}
 
+	// Annotate with reconcile-unique UUID to trigger MigMigration reconcile
```
jortel: Perhaps it would be better to describe this as needed for all remote watch reconcile propagation.

jortel:
Nice Fix!

pkg/controller/migcluster/migcluster_controller.go
djwhatle: Are you thinking we should add the UUID update in other places as well?

PR #109, opened 2019-05-22
Integer task phases [POC].

jortel:
Looks like being able to do integer comparison or ordered phases would be useful.

For example:
```
if t.Phase < ResticRestartCompleted {
    return nil
}
```


jortel:
No longer relevant.

PR #112, opened 2019-05-23
Delete stage pods after restore is completed

dymurray:
Plugin PR: https://github.com/fusor/ocp-velero-plugin/pull/44

### DIFF
--- a/pkg/controller/migmigration/task.go
+++ b/pkg/controller/migmigration/task.go
@@ -145,6 +145,13 @@ func (t *Task) Run() error {
 		t.Phase = RestoreCompleted
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/migmigration/task.go
```diff
@@ -145,6 +145,13 @@ func (t *Task) Run() error {
 		t.Phase = RestoreCompleted
 	}
 
+	if t.Phase == RestoreCompleted && t.stage() {
```
jortel: You don't need (or want) to check the phase here.  If the run() get this far, the restore has completed.

dymurray: Ah okay thanks Jeff

PR #113, opened 2019-05-23
Fix validation of PVs with no supported actions.

jortel:
Fix validation of PVs with no supported actions.  PVs with Action = "" and SupportedActions = [] to be valid.

### DIFF
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -27,6 +27,7 @@ const (
 	NsNotFoundOnSourceCluster      = "NamespaceNotFoundOnSourceCluster"
 	NsNotFoundOnDestinationCluster = "NamespaceNotFoundOnDestinationCluster"
 	PvInvalidAction                = "PvInvalidAction"
+	PvNoSupportedAction            = "PvNoSupportedAction"
 	StorageEnsured                 = "StorageEnsured"
 	PvsDiscovered                  = "PvsDiscovered"
 )
@@ -35,6 +36,7 @@ const (
 const (
 	Critical = migapi.Critical
 	Error    = migapi.Error
+	Warn     = migapi.Warn
 )
 
 // Reasons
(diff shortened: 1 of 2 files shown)


jwmatthews:
cc @dymurray @djwhatle @eriknelson @sseago @jmontleon @ibolton336 

I think we will __want__ to include PVs with no supported actions.

Reasoning is to provide information to the user that we are unable to do anything with the specific PV(s).

I see 2 potential examples:
 - Hostpath used to store generated cache data, if it's not migrated the application will still work, will just regenerate the data 

 - Hostpath used to store true persistent data, lack of migration causing application to fail

I think it's best to surface the information of "We found a PV but can't migrate it", this way the user is aware and can decide if they want to proceed or not.  Spoke to @eriknelson about this, plan is in the UI if a PV has no supported actions we will inform the user and when we validate the migrationplan in UI we will warn user a PV has no supported action for migration and ask if they want to continue.

jortel:
Thanks @jwmatthews!
I revised the PR. The controller will set a _Warn_ condition listing the PVs without supported actions.  Warning conditions do not block the `Ready` condition with means users will be informed but can still run migrations.

Ex:
```
    Type:                  PvNoSupportedAction
    Category:              Warn
    Last Transition Time:  2019-05-24T13:01:55Z
    Message:               PV in `persistentVolumes` [local] with no `SupportedActions`.
    Status:                True
```

pkg/controller/migplan/validation.go
```diff
@@ -35,6 +36,7 @@ const (
 const (
 	Critical = migapi.Critical
 	Error    = migapi.Error
+	Warn     = migapi.Warn
```
djwhatle: +1

pkg/controller/migplan/validation.go
```diff
@@ -398,11 +401,17 @@ func (r ReconcileMigPlan) validateDestinationNamespaces(plan *migapi.MigPlan) er
...
 		actions := map[string]bool{}
-		for _, a := range pv.SupportedActions {
-			actions[a] = true
+		if len(pv.SupportedActions) > 0 {
+			for _, a := range pv.SupportedActions {
+				actions[a] = true
+			}
+		} else {
+			unsupported = append(unsupported, pv.Name)
+			actions[""] = true
```
djwhatle: "" represents the "nil", or "do nothing" action here from what I gather.

PR #115, opened 2019-05-24
[WIP] Docs migration scenarios

djwhatle:
@jwmatthews @sseago wanted to see what feedback you guys might have before I go onto the other scenarios. Just completed the first scenario for stateless nginx migration.

### DIFF
--- a/docs/scenarios/ebs-pv/README.md
+++ b/docs/scenarios/ebs-pv/README.md
@@ -1,0 +1,5 @@
+## Migrating a **stateful** OpenShift app with **Amazon EBS PVs**
+
+This scenario walks through Migration of a stateful OpenShift app whose Persistent Volume (PV) claims tie into [Amazon EBS (Elastic Block Storage)](https://aws.amazon.com/ebs/).
+
+
--- a/docs/scenarios/ebs-pv/README.md
+++ b/docs/scenarios/ebs-pv/README.md
@@ -1,5 +1,13 @@
-## Migrating a **stateful** OpenShift app with **Amazon EBS PVs**
+## Migrating a *stateful* OpenShift app with *Amazon EBS PVs*
 
 This scenario walks through Migration of a stateful OpenShift app whose Persistent Volume (PV) claims tie into [Amazon EBS (Elastic Block Storage)](https://aws.amazon.com/ebs/).
+
+### Supported PV Actions - EBS to EBS
+
+| Action | Supported | Description |
+|-----------|------------|-------------|
+| Copy | Yes | EBS snapshots will be created from source cluster PVs. New EBS PVs for destination cluster will be created from snapshots |
+| Move | No  | ... |
+
 
 
--- a/docs/scenarios/nfs-pv/README.md
+++ b/docs/scenarios/nfs-pv/README.md
@@ -1,0 +1,7 @@
+## Migrating a **stateful** OpenShift app with **NFS PVs**
+
+This scenario walks through Migration of a stateful OpenShift app whose Persistent Volume (PV) claims tie into [NFS (Network File System)](https://en.wikipedia.org/wiki/Network_File_System) storage.
+
+
+
+
--- a/docs/scenarios/nfs-pv/README.md
+++ b/docs/scenarios/nfs-pv/README.md
@@ -1,7 +1,13 @@
-## Migrating a **stateful** OpenShift app with **NFS PVs**
+## Migrating a *stateful* OpenShift app with *NFS PVs*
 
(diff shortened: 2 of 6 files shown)


djwhatle:
@sseago perhaps that would make it clearer, I was trying to show only a focused snippet and added this comment, not sure if you saw it in there.

```
[... snipped, see config/samples/mig-plan.yaml for full version ...]
```

docs/scenarios/ebs-pv/README.md
```diff
@@ -0,0 +1,5 @@
+## Migrating a **stateful** OpenShift app with **Amazon EBS PVs**
+
+This scenario walks through Migration of a stateful OpenShift app whose Persistent Volume (PV) claims tie into [Amazon EBS (Elastic Block Storage)](https://aws.amazon.com/ebs/).
+
+
```
jwmatthews: Let's add a section to each scenario overview stating what PV actions are supported.

For example for ebs-pv we support
 'Copy', usage of ebs snapshots will result in a new PV being created from a snapshot of the PV.
'Move' is not supported.

docs/scenarios/nfs-pv/README.md
```diff
@@ -0,0 +1,7 @@
+## Migrating a **stateful** OpenShift app with **NFS PVs**
+
+This scenario walks through Migration of a stateful OpenShift app whose Persistent Volume (PV) claims tie into [NFS (Network File System)](https://en.wikipedia.org/wiki/Network_File_System) storage.
+
+
```
jwmatthews: For NFS we support 2 actions:

Copy, a new PV will be used on the destination cluster and data from the source PV will be copied to the destination PV.

Move, the existing PV on the source cluster will be detached from source and attached to destination.  We call this 'Swinging the PV'.

docs/scenarios/nfs-to-ebs-pv/README.md
```diff
@@ -0,0 +1,4 @@
+## Migrating a **stateful** OpenShift app with **NFS PVs** to **Amazon EBS PVs**
+
+This scenario walks through Migration of a stateful OpenShift app whose Persistent Volume (PV) claims tie into an NFS server. Through the Migration process, the data stored on NFS will be moved onto [Amazon EBS (Elastic Block Storage)](https://aws.amazon.com/ebs/). 
+
```
jwmatthews: For nfs to ebs we support:
 - Copy, we will create a new PV backed by EBS on the destination side and copy data from the source NFS PV to the destination EBS PV.

vice-versa is supported, i.e. EBS to NFS backed PV.

docs/scenarios/stateless/README.md
```diff
@@ -0,0 +1,192 @@
...
+To migrate our `nginx-example` namespace, we'll ensure that the `namespaces` field of our MigPlan includes nginx-example. Luckily [config/samples/mig-plan.yaml](https://github.com/fusor/mig-controller/blob/master/config/samples/mig-plan.yaml) does exactly this.
+
+```yaml
+apiVersion: migration.openshift.io/v1alpha1
+kind: MigPlan
+metadata:
+  name: migplan-sample
+  namespace: mig
+
+[... snipped, see config/samples/mig-plan.yaml for full version ...]
```
jwmatthews: +1

goal is to ensure CI is testing as much of the scenarios as possible, so we do not want to include a full yaml in the docs, more we want to point to where the yaml is so we can be sure CI is testing and it's still accurate.

sseago: @djwhatle Regarding this (and my comment above). I agree we shouldn't be including the full yaml in the docs, but I wonder whether we should at least mention (without including the yaml for it) that the clusters created in the earlier step should be referenced here. Otherwise it might seem like the "see elsewhere for full yaml" part is boilerplate rather than a necessary part of the config since it's an explicit object reference rather than default configuration.

docs/scenarios/ebs-pv/README.md
djwhatle: Will add this.

Sidenote:  are we tabling the concept of disconnecting an EBS volume from one instance and connecting it to another?

docs/scenarios/stateless/README.md
djwhatle: Updated the comment to hopefully be more clear that people should refer to sample to see other "required" fields.

```
[... snipped, see config/samples/mig-plan.yaml for other required fields ...]
```

docs/scenarios/nfs-pv/README.md
djwhatle: Added

PR #119, opened 2019-05-28
Moved migration registry creation to migplan controller

sseago:
Since the migration registries are tied to the MigPlan rather than
the MigMigration, the right place to create (and delete) these
is in the context of managing the MigPlan. This removes some of the
unnecessary complexity from the migration Task resource.


djwhatle:
@sseago looks good! Am I correct in thinking that this doesn't currently handle deletion of registries?

sseago:
@djwhatle It does not handle deletion since the deletion PR from last week was never merged. This just moves existing functionality over. Deletion will be a separate PR.

sseago:
updated PR to resolve merge conflicts.

PR #124, opened 2019-05-29
Add NFS stateful "move" migration scenario

djwhatle:
Documents a migration scenario involving movement of NFS PVs between two clusters that have shared access to a central NFS server.


docs/scenarios/nfs-pv/README.md
```diff
@@ -9,5 +9,320 @@ This scenario walks through Migration of a stateful OpenShift app with Persisten
...
+spec:
+  # [!] Change namespaces to adjust which OpenShift namespaces should be migrated 
+  #     from source to destination cluster.
+  namespaces:
+  - mysql-persistent
+
+[... snipped, see config/samples/mig-plan.yaml for other required fields ...]
+```
+
+We also need to create a MigMigration with `quiescePods: true` since only one MySQL Pod can hold the lock at once, and we this scenario covers the 'move' PV action that re-mounts the *source cluster* NFS PVs on the *destination cluster*
```
dymurray: `and we this scenario` -> `and this scenario`

docs/scenarios/nfs-pv/README.md
```diff
@@ -9,5 +9,320 @@ This scenario walks through Migration of a stateful OpenShift app with Persisten
...
+  Migration Completed:     true
+  Start Timestamp:         2019-05-31T14:22:08Z
+  Task Phase:              Completed
+```
+
+---
+
+### 7. Inspect migration results on destination cluster
+
+Logging into the destination cluster, we should now see a running mysql pod in the 'mysql-persistent' namespace, matching what existed on the source cluster.
```
dymurray: NIT: put mysql-persistent in backticks: `mysql-persistent`

djwhatle: Thanks, I meant to put all of the mentions of `mysql-persistent` in backticks for their first mention in each paragraph, missed this one.

docs/scenarios/nfs-pv/README.md
djwhatle: +1 I noticed I did this but lost track of it.

PR #125, opened 2019-05-30
Add the 'Closed' concept to MigPlan and remove registry for closed plans

sseago:
When a final migration completes, close the MigPlan.
Upon closing, a MigPlan removes migration registry resources.
Still to do:
1) Clean up storage resources as part of closing as well
2) Figure out how to deal with resource cleanup for clusters
   that are no longer referenced by the MigPlan but have
   registry or storage resources that have matching correlation labels.

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -35,6 +35,7 @@ type MigPlanSpec struct {
 	SrcMigClusterRef  *kapi.ObjectReference `json:"srcMigClusterRef,omitempty"`
 	DestMigClusterRef *kapi.ObjectReference `json:"destMigClusterRef,omitempty"`
 	MigStorageRef     *kapi.ObjectReference `json:"migStorageRef,omitempty"`
+	Closed            bool                  `json:"closed,omitempty"`
 }
 
 // MigPlanStatus defines the observed state of MigPlan
@@ -132,6 +133,11 @@ func (r *MigPlan) GetRefResources(client k8sclient.Client) (*PlanResources, erro
 	}
 
 	return resources, nil
+}
+
+// Set the MigPlan Status to closed
+func (r *MigPlan) SetClosed() {
+	r.Spec.Closed = true
 }
 
 // Build a credentials Secret as desired for the source cluster.
--- a/pkg/controller/migmigration/migrate.go
+++ b/pkg/controller/migmigration/migrate.go
@@ -105,6 +105,13 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (bool, e
 		if err != nil {
 			return false, err
 		}
+		if !migration.Spec.Stage {
+			plan.SetClosed()
+			err = r.Update(context.TODO(), plan)
+			if err != nil {
+				return false, err
+			}
+		}
 	}
 
 	return false, nil
--- a/pkg/controller/migplan/close.go
+++ b/pkg/controller/migplan/close.go
@@ -1,0 +1,63 @@
+/*
+Copyright 2019 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package migplan
+
+import (
+	"context"
+
+	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
+)
+
+func (r ReconcileMigPlan) handleClosed(plan *migapi.MigPlan) (bool, error) {
+	closed := plan.Spec.Closed
+	if !closed || plan.Status.HasCondition(Closed) {
+		return closed, nil
+	}
+
+	plan.Status.SetReady(false, ReadyMessage)
+	err := r.Update(context.TODO(), plan)
+	if err != nil {
+		return closed, err
+	}
+
+	err = r.ensureClosed(plan)
+	return closed, err
+}
+
+// Ensure that resources managed by the MigPlan have been cleaned up
+func (r ReconcileMigPlan) ensureClosed(plan *migapi.MigPlan) error {
+	// Migration Registry
+	err := r.ensureMigRegistriesDelete(plan)
+	if err != nil {
+		return err
+	}
+
+	plan.Status.DeleteCondition(RegistriesEnsured)
+	plan.Status.SetCondition(migapi.Condition{
+		Type:     Closed,
+		Status:   True,
+		Category: Critical,
+		Message:  ClosedMessage,
+	})
+	// Apply changes.
+	err = r.Update(context.TODO(), plan)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/controller/migplan/migplan_controller.go
+++ b/pkg/controller/migplan/migplan_controller.go
@@ -127,6 +127,18 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 
+	// If plan is closed, ensure resource cleanup
+	closed, err := r.handleClosed(plan)
+	if err != nil {
+		if errors.IsConflict(err) {
+			return reconcile.Result{Requeue: true}, nil
+		}
+		return reconcile.Result{}, err
+	}
+	if closed {
+		return reconcile.Result{}, nil
+	}
+
 	// Validations.
 	err = r.validate(plan)
 	if err != nil {
--- a/pkg/controller/migplan/registry.go
+++ b/pkg/controller/migplan/registry.go
@@ -221,3 +221,116 @@ func (r ReconcileMigPlan) ensureRegistryService(client k8sclient.Client, plan *m
 
 	return nil
 }
+
+// Ensure the migration registries on both source and dest clusters have been removed
+func (r ReconcileMigPlan) ensureMigRegistriesDelete(plan *migapi.MigPlan) error {
+	var client k8sclient.Client
+
+	clusters, err := migapi.ListClusters(r, "")
+	if err != nil {
+		return err
+	}
+
+	for _, cluster := range clusters {
+		if !cluster.Status.IsReady() {
+			continue
+		}
+		client, err = cluster.GetClient(r)
+		if err != nil {
+			return err
+		}
+
+		// Migration Registry Service
+		err = r.ensureRegistryServiceDelete(client, plan)
+		if err != nil {
+			return err
+		}
+
+		// Migration Registry DeploymentConfig
+		err = r.ensureRegistryDCDelete(client, plan)
+		if err != nil {
+			return err
+		}
+		// Migration Registry ImageStream
+		err = r.ensureRegistryImageStreamDelete(client, plan)
+		if err != nil {
+			return err
+		}
+
+		// Migration Registry Secret
+		err := r.ensureRegistrySecretDelete(client, plan)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Ensure the service for the migration registry on the specified cluster has been created
+func (r ReconcileMigPlan) ensureRegistryServiceDelete(client k8sclient.Client, plan *migapi.MigPlan) error {
+	foundService, err := plan.GetRegistryService(client)
+	if err != nil {
+		return err
+	}
+	if foundService == nil {
+		return nil
+	}
+	err = client.Delete(context.TODO(), foundService)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Ensure the deploymentconfig for the migration registry on the specified cluster has been created
+func (r ReconcileMigPlan) ensureRegistryDCDelete(client k8sclient.Client, plan *migapi.MigPlan) error {
+	foundDC, err := plan.GetRegistryDC(client)
+	if err != nil {
+		return err
+	}
+	if foundDC == nil {
+		return nil
+	}
+	err = client.Delete(context.TODO(), foundDC)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Ensure the imagestream for the migration registry on the specified cluster has been created
+func (r ReconcileMigPlan) ensureRegistryImageStreamDelete(client k8sclient.Client, plan *migapi.MigPlan) error {
+	foundImageStream, err := plan.GetRegistryImageStream(client)
+	if err != nil {
+		return err
+	}
+	if foundImageStream == nil {
+		return nil
+	}
+	err = client.Delete(context.TODO(), foundImageStream)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Ensure the credentials secret for the migration registry on the specified cluster has been created
+func (r ReconcileMigPlan) ensureRegistrySecretDelete(client k8sclient.Client, plan *migapi.MigPlan) error {
+	foundSecret, err := plan.GetRegistrySecret(client)
+	if err != nil {
+		return err
+	}
+	if foundSecret == nil {
+		return nil
+	}
+	err = client.Delete(context.TODO(), foundSecret)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -31,6 +31,7 @@ const (
 	StorageEnsured                 = "StorageEnsured"
 	RegistriesEnsured              = "RegistriesEnsured"
 	PvsDiscovered                  = "PvsDiscovered"
+	Closed                         = "Closed"
 )
 
 // Categories
@@ -73,6 +74,7 @@ const (
 	StorageEnsuredMessage                 = "The storage resources have been created."
 	RegistriesEnsuredMessage              = "The migration registry resources have been created."
 	PvsDiscoveredMessage                  = "The `persistentVolumes` list has been updated with discovered PVs."
+	ClosedMessage                         = "The migration plan is closed."
 )
 
 // Validate the plan resource.
--- a/pkg/apis/migration/v1alpha1/model.go
+++ b/pkg/apis/migration/v1alpha1/model.go
@@ -29,7 +29,10 @@ func ListPlans(client k8sclient.Client, ns string) ([]MigPlan, error) {
 // Returns and empty list when none found.
 func ListClusters(client k8sclient.Client, ns string) ([]MigCluster, error) {
 	list := MigClusterList{}
-	options := k8sclient.InNamespace(ns)
+	var options *k8sclient.ListOptions
+	if ns != "" {
+		options = k8sclient.InNamespace(ns)
+	}
 	err := client.List(context.TODO(), options, &list)
 	if err != nil {
 		return nil, err


pkg/controller/migmigration/migrate.go
```diff
@@ -96,6 +96,13 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (bool, e
 		if err != nil {
 			return false, err
 		}
+		if !migration.Spec.Stage {
+			plan.SetClosed()
+			err = r.Update(context.TODO(), plan)
+			if err != nil {
+				return false, err
```
djwhatle: For a second I thought this would fail to requeue since we're returning `false, err`, but I checked the calling block and looks like it will do some further error processing for us. Recalled that returning `true, err` would have triggered a time delayed requeue for cases where we won't get another event.

But think this will work :+1:

pkg/controller/migplan/migplan_controller.go
```diff
@@ -136,6 +136,24 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
 			return reconcile.Result{}, err
 		}
 	}
+
+	// plan is closed, ensure resource cleanup
+	if plan.Status.HasCondition(Closed) {
```
djwhatle: I think it will make the code easier to grok if we're only dealing with one err at a time where possible. I think the below is accomplishes something similar to this block, assuming it's OK to mark a plan as not ready (which seems sane to me) when we know it is closed and are deleting its registries.

```
// Plan is closed, ensure resource cleanup
if plan.Status.HasCondition(Closed) {
    
    // Set plan as not ready since it is marked as closed
    plan.Status.SetReady(false, ReadyMessage)
    err := r.Update(context.TODO(), plan)
    if err != nil {
        if errors.IsConflict(err) {
        }
        return reconcile.Result{}, err
    }


    // Migration Registry cleanup
    err = r.ensureMigRegistriesDelete(plan)
    if err != nil {
        if errors.IsConflict(err) {
        }
        return reconcile.Result{}, err
    }
}
```

pkg/controller/migplan/registry.go
```diff
@@ -221,3 +221,116 @@ func (r ReconcileMigPlan) ensureRegistryService(client k8sclient.Client, plan *m
...
+func (r ReconcileMigPlan) ensureMigRegistriesDelete(plan *migapi.MigPlan) error {
+	var client k8sclient.Client
+
+	clusters, err := r.planClusters(plan)
+	if err != nil {
+		return err
+	}
+
+	for _, cluster := range clusters {
+		if !cluster.Status.IsReady() {
```
djwhatle: Trying to understand big picture of how the concept of MigCluster readiness comes into our decision to delete registries with this PR. Can help me understand?

It looks like we're using readiness as a way to track whether we've already attempted deletion once. < Assuming this is true, I think we may get more predicatable outcomes by always checking if registry resources owned by our MigCluster currently exist anywhere, and deleting them if they do exist.

djwhatle: I may be misinterpreting what the logic is intended to do, so please correct me if this is wrong.

pkg/controller/migplan/migplan_controller.go
sseago: The thing we need to confirm is that since, with this modification, we may leave this reconcile loop without deleting its registries, we need to make sure that the `SetReady(false)`  doesn't somehow clear out the `Closed` condition. One thing the current version of this code does is to ensure that it attempts to both `SetReady(false)` and delete the registres. If possibly splitting them up into separate reconcile calls is OK, then this modification should be fine.

pkg/controller/migplan/registry.go
sseago: So this particular check was pulled from the registry and storage create code. I took the `IsReady` check to mean that we only want to attempt to execute actions on a cluster if the cluster is ready. I didn't actually look into what causes a cluster to be not ready, so that may not be the right check here. Is there some better way of saying "make sure this cluster is in a state to make calls on?"

sseago: Related to this -- is there a better way of choosing clusters to check? The code as written won't catch registry resources allocated for this MigPlan on a cluster no longer referenced by this MigPlan -- which would happen if a user created the plan referencing ClusterA and ClusterB but later changed the ClusterB reference to ClusterC.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -134,6 +135,12 @@ func (r *MigPlan) GetRefResources(client k8sclient.Client) (*PlanResources, erro
 	return resources, nil
 }
 
+// Set the MigPlan Status to closed
+func (r *MigPlan) SetClosed() error {
```
jortel: Why does `SetClosed()` return an _error_?

sseago: I wondered about whether to do this or have no return. I left the error return for consistency with similar methods -- we have others that return an error but there's no actual error returns in the current implementation. I can get rid of the return, though, if there's no need to keep the method signatures consistent for this sort of thing.

pkg/controller/migplan/registry.go
jortel: I'd imagined that all of the clusters would be checked.  Any resources with the plan label would be cleaned up.

pkg/apis/migration/v1alpha1/migplan_types.go
jortel: I appreciate consistency.  
The _other_ methods returning an error are also passed a client.  The _setter_ provides a nice abstraction from the Spec but don't think it's sufficiently complex to be passed a client.  I recommend removing the _error_ return value here.

sseago: OK. That works for me.

pkg/controller/migplan/registry.go
sseago: Is there an existing method to return a list of all clusters we know about? Right now, all of the cluster-getting code we have takes a client passed in. There's a GetClient method on MigCluster, but is there anything needed to get a list of all clusters to iterate over? I think I'm getting confused by the migration vs. other clusters here -- the single GetCluster method calls `client.Get` on the passed-in client, but if I want to get all clusters, each will have its own client.

sseago: Also, while this PR doesn't do it yet, we'll need to add a finalized that also cleans up resources on MigPlan delete. Thinking through the "remove from all clusters" bit, though, it looks like we'll also want a finalizer on MigCluster delete, since if we remove the reference to a MigCluster with plan resources allocated, there will be no way to remove these on MigPlan closing/deleting.

jortel:
I think there is a lot of really good stuff going here.  

However, I'd imagined that the `Spec.Closed` would trigger the _closure_ (resource clean up) and that the `Closed` condition would be set after everything is (ensured) cleaned up.  This supports the declarative, _make-the-world-as-it-should-be_ with a resumable Reconcile().

So, the reconcile() would:
```
# pseudo code
#
if plan.Spec.Closed {
  if plan.Status.HasCondition(Closed) {
    return
  err := r.ensureClosed()
  ...
}
```

The `ensureClosed()` would set the `Closed` condition on successful completion.

Let me know if I'm misunderstanding something.

pkg/controller/migplan/migplan_controller.go
jortel: I recommend moving this block into a top level `ensureClosed()` in a spearte .go file.  This well keep the main reconcile() cleaner especially when the storage clean up is added.

sseago:
Regarding the final comment -- one of the concerns that we had was making sure that closing a plan was a one-time change. Setting closed to 'false' after initiating closure shouldn't do anything. The reason for this is that re-opening a plan after closing it would result in unpredictable behavior -- since closing a plan should cause any in-progress migrations to cease processing (since the plan would no longer be ready), and storage/registry resources will be cleaned up. Subsequent re-opening of the plan could result in migrations resuming, but they wouldn't be in a resumable state, since storage/registry resources would be re-allocated in ways that might be inconsistent with the prior state (and possible in-progress migrations). The way this is currently implemented, setting closed will immediately set the 'closed' condition, which won't be reversed by later removing Spec.closed. I wonder if the right answer would be to have two conditions: Closed (acts as currently handled in this PR) and ClosedComplete (does what you're suggesting above for Closed). I think this would preserve the "can't re-open" behavior while also allowing us to skip going back into the closure logic for already-closed plans.

pkg/controller/migplan/registry.go
jortel: > Also, while this PR doesn't do it yet, we'll need to add a finalized that also cleans up resources on MigPlan delete. Thinking through the "remove from all clusters" bit, though, it looks like we'll also want a finalizer on MigCluster delete, since if we remove the reference to a MigCluster with plan resources allocated, there will be no way to remove these on MigPlan closing/deleting.

Which resources (that won't be cleaned up on close)?

It's my understanding that finalizers that can delay/fail are frowned upon.  Perhaps someone else can comment on this?

jortel:
You're probably right in that explicitly supporting a re-open does not seem valuable.  However, I'm not convinced that preventing it would be worth while either.  When I think through a re-open scenario on a plan with running or pending migrations, it seems the reconcile() would re-build (ensure) everything before becoming `Ready` again.  Then, the migrations would resume.  What am I missing?

pkg/controller/migplan/registry.go
sseago: The concern isn't not cleaning up on close -- the concern is if a MigPlan is deleted before it's closed. We'll need a finalizer on the MigPlan that does the closing cleanup before deleting. Also, we'll need a finalizer on MigCluster that removes storage/registry resources from the cluster before removing the MigCluster CR, since once we remove the MigCluster, the MigPlan close functionality will have no way of knowing about the resources in use there.

pkg/controller/migplan/validation.go
```diff
@@ -114,6 +116,12 @@ func (r ReconcileMigPlan) validate(plan *migapi.MigPlan) error {
 		return err
 	}
 
+	// PV list.
+	err = r.validateClosed(plan)
```
jortel: Fix cut-n-paste comment.

Perhaps the closed should be validated first and the remaining validations skipped (when closed).

sseago:
"When I think through a re-open scenario on a plan with running or pending migrations, it seems the reconcile() would re-build (ensure) everything before becoming Ready again. Then, the migrations would resume. What am I missing?" -- the newly-created storage/registry resources won't be the same as the deleted ones. This may not actually matter that much -- while the registry IP address will be different, the actual storage location (bucket, subdirectory, credentials) will be the same on the recreated cluster. Where it will fail is if a backup or restore is in progress, since the registry location passed into the Velero CR will no longer be valid. This will most likely result in the Backup or Restore failing.

sseago:
Thinking back to the "two conditions" proposal -- perhaps it would be clearer if we set a "Closing" condition where we currently set a "Closed" condition, and set "Closed" when the close-related functions are done? I think this will be clearer than having a single condition that's either set before (as done in the PR) or after (as you had proposed).

pkg/controller/migplan/validation.go
sseago: Sure. We could modify validateClosed to return a bool and an error, and if it returns `true`, we could return right there.

jortel:
> Thinking back to the "two conditions" proposal -- perhaps it would be clearer if we set a "Closing" condition where we currently set a "Closed" condition, and set "Closed" when the close-related functions are done? I think this will be clearer than having a single condition that's either set before (as done in the PR) or after (as you had proposed).

Yes. If going with 2 conditions, the names `Closing` and `Closed` sound good to me.

pkg/controller/migplan/validation.go
```diff
@@ -446,3 +454,17 @@ func (r ReconcileMigPlan) validatePvAction(plan *migapi.MigPlan) error {
...
 	return nil
 }
+
+// Validate Spec.Closed and set status accordingly.
+func (r ReconcileMigPlan) validateClosed(plan *migapi.MigPlan) error {
+	if plan.Spec.Closed {
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     Closed,
+			Status:   True,
+			Category: Critical,
```
jortel: The `Critical` category is intended to indicate a critical error.  Let's either omit the category here or add a new category.

pkg/controller/migplan/registry.go
jortel: Agreed with all the concerns and perhaps the finalizer is just a necessary evil.

jortel:
My concern with the 2 condition approach is that the Spec is supposed to be declarative.  What would it mean for a plan to be Spec.Closed=true, the 2 conditions eventually added.  Then Spec.Closed=false.  Based on reconciliation principal, The `reconcile()` _should_ make the plan open again which includes ensuring that all the cleaned up resources are created again and the 2 conditions cleared.  The plan would eventually become `Ready` again.  Right?  Am I applying this in too strict a sense?

djwhatle:
@jortel our controller has to take a specific sequence of actions to perform a migration, and these actions need to happen in order. 

If we close down a plan during an ongoing migration, the migration registries and all contained images will be destroyed, but the MigMigration phase will be kept intact. When we reopen the plan, our MigMigration will attempt to proceed but encounter fatal errors due to loss of state caused by the prior setting of `closed: true`.

I think the only way we could safely re-open a plan would be if we reset the progress on any connected MigMigration to start from the beginning. The alternative would be to require the user to recreate a new MigPlan with similar content in the spec if they want to use it again.

jortel:
> @jortel as you know, the MigMigration controller has to take a specific sequence of actions to perform a migration, and these actions need to happen in order.
> 
> If we close down a plan during an ongoing migration, the migration registries and all contained images will be destroyed, but the MigMigration phase will be kept intact. When we reopen the plan, our MigMigration will attempt to proceed but encounter fatal errors due to loss of state caused by the prior setting of `closed: true`.
> 
> I think the only way we could safely re-open a plan would be if we reset the progress on any connected in-progress MigMigration to start from the beginning. The alternative would be to require the user to recreate a new MigPlan with similar content in the spec if they want to use it again.
> 
> _Edit:_ Another option I can see would be to automatically send any connected MigMigrations to `Completed: True` with an error state if we close down their associated MigPlan in the middle of a migration. This might be closer to honoring the declarative nature of spec.

I would expect the plan.Spec.Closed would cause associated running or pending migrations to marked completed and an error condition added.

djwhatle:
> 
> I would expect the plan.Spec.Closed would cause associated running or pending migrations to marked completed and an error condition added.
> 
> Re-opening a plan would not be recommended but thinking it should behave intuitively.

:+1: I'm good with this approach

sseago:
Updated in response to feedback.

pkg/controller/migplan/close.go
```diff
@@ -0,0 +1,61 @@
...
+
+	return nil
+}
+
+// Remove the closed condition
+func (r ReconcileMigPlan) deleteClosedCondition(plan *migapi.MigPlan) error {
+	plan.Status.DeleteCondition(Closed)
+
+	// Apply changes.
+	plan.Status.EndStagingConditions()
```
jortel: Directly deleting the condition instead of staging.  Don't need/want:  `EndStagingConditions()` here.

pkg/controller/migplan/close.go
```diff
@@ -0,0 +1,61 @@
...
+	err = r.Update(context.TODO(), plan)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Remove the closed condition
+func (r ReconcileMigPlan) deleteClosedCondition(plan *migapi.MigPlan) error {
```
jortel: I think something like `planReopened()` would be more symetrical with `ensureClosed()`.  Although the behavior would (initially) be to remove the `Closed` condition, the more generic name would support anything related to _un-closing_ the plan.

pkg/controller/migplan/migplan_controller.go
```diff
@@ -145,7 +145,27 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
 		}
 	}
 
-	if plan.Status.HasCriticalCondition() {
+	// plan is closed, ensure resource cleanup
+	if plan.Spec.Closed {
```
jortel: Seems like the `closed` behavior should happen before everything else in the reconcile().  Even before the validation.  If the plan is closed, I don't think we want to reconcile anything else.  also, even if `ensureClosed()` succeeds, we should return.

djwhatle:
Tested out after the changes you made, these are my observations.

- Manually setting `closed: true` on a MigPlan prevents a migration from happening, and cleans up the Migration registries


- Removing `closed: true` on a MigPlan allowed me to re-open the plan and (after a few seconds of registry re-creation) led to a `Ready: true` condition on the plan


 - Running a Migration against a MigPlan completed successfully, and added `closed: true` to the plan afterwards.

pkg/controller/migplan/close.go
sseago: OK. I'll remove it. I was following the closest example I had, which followed this pattern:
https://github.com/fusor/mig-controller/blob/master/pkg/controller/migplan/storage.go#L69
Perhaps this code shouldn't be calling this either.

pkg/controller/migplan/close.go
sseago: My thinking wasn't so much keeping to the same go file but that we're triggering the condition changes from the same place. This is called from the `else` clause of the `if` statement that calls ensureClosed on the other side. Seemed cleaner to keep it together "if closed, close things, else open them"

pkg/controller/migplan/migplan_controller.go
sseago: We don't return on success because immediately following this block, the next block calls SetReady(false) on the plan and returns. If we return here, we'll have to add another SetReady(false) call -- which will work fine, but might be harder to follow. If we decide to add the SetReady here, along with the extra return, we could call it before validate. I guess there's no harm, for closed plans, to skip validating src/dest clusters, etc., especially since we're no longer explicitly referencing src/dest clusters in the delete anymore, since we delete from all clusters.

pkg/controller/migplan/close.go
jortel: No, it shouldn't.

pkg/controller/migplan/migplan_controller.go
jortel: I just noticed that you added the additional `plan.Spec.Closed` to the `plan.Status.HasCriticalCondition()` below.  This address my primary concern (and motivation for suggesting the return) that the reconcile would continue into the methods for ensuring the repository and storage.  So not returning will be fine.

_snip_
```
if plan.Spec.Closed || plan.Status.HasCriticalCondition() {
```
As for validating a closed plan - I guess it depends on how may closed plans will exist at any given time.  I'd imagined 40-50 (closed) plans all doing needless validation at potentially high rates given the local and remote watches will still be wired up.  Perhaps I'm overestimating the performance impact or the number of closed plans but this seems fairly undesirable.  Another issue with running this after the validations is that the condition _staging_ used by `validate()` will always clear the `Closed` condition.  If need, I have thoughts on how this can be dealt with but should probably be done in a separate PR.

I agree that if it's behavior is moved to be beginning of the reconcile() and the return added, the `Ready` condition will need to be explicitly cleared.

I also noticed that nothing is checking for the `Closed` condition so the resource clean up will be always performed on closed plans.  Even after everything has been cleaned up.

sseago: So regarding checking the closed condition. Yeah, right now this keeps checking (and presumably not finding) things to delete. We should probably change the `if plan.Spec.Closed` Condition to `if plan.Spec.Closed && !plan.Status.HasCondition(Closed)` -- that would only run this if the plan has been flagged for closure via `Spec.Closed` but for which closure has not been completed.

sseago: As for validation on closed plans, I just looked back at the validate function. I was initially concerned that a plan that failed validation shouldn't attempt to close, in case something was missing that validation checked, but at least for registry validation, I don't think that's the case. We're just iterating over all Clusters -- so validate isn't even looking at clusters that have things we need to clean up but aren't connected to the plan anymore. There are 2 ways to skip validation on closed plans:
1) Move this block of code up above validation -- then we will need to add the SetReady(false) and return after ensureClosed -- and we'll have to revisit the complexity around multiple places for errors to pop up -- since we have to check for error on ensureClosed *and* try to call setReady.
2) keep the code as-is, but wrap the validate call in an `if !plan.Spec.Closed` conditional
It's not obvious to me which approach is cleaner.

jortel: When I recommended putting the _close_ logic in close.go I had envisioned _something_ like this (meant to be pseudo-code):

_reconcile loop_
```

// Closed
closed, err := r.handleClosed(plan)
if err != nil {
    if errors.IsConflict(err) {
    } else {
        return reconcile.Result{}, err
    }
} 
if closed {
}

// Validations.
err = r.validate(plan)
...
```

_close.go_
```
func (r ReconcileMigPlan) handleClosed(plan *migapi.MigPlan) (bool, error) {
    closed := plan.Spec.Closed
    if !closed || plan.Status.hasCondition(Closed) {
        return closed, nil
    }
  
    plan.Status.SetReady(false)
    err = r.Update(context.TODO(), plan)
    if err != nil {
        return closed, err
    }

    err := r.ensureClosed(plan)
    return closed, err
}
```

Complete delegation to another method provides for simpler logi and also leaves all of the code in `Reconcile()` after `handleClosed()` concerned with an _open_ plan.  Continuing into the `validate()` will clear the `Closed` condition in _re-open_ scenarios.

sseago: "Continuing into the validate() will clear the Closed condition in re-open scenarios."
Is that because validate always clears all conditions and starts over? There's nothing in validate() that touches the Closed condition explicitly. In any case, this will be easy enough to verify. If the above approach handles reopening, then setting Spec.Closed to false will trigger re-opening and recreating the registry.

sseago:
Updated again with the last round of suggested modifications made.

pkg/controller/migplan/migplan_controller.go
jortel: Yes.  

The condition _staging_ which includes BeginStagingConditions() and BeginStagingConditions() used in validation will result in the delete any conditions that have not been explicitly: SetCondition().  This approach was indented to simplify managing conditions in validation.  Since then, other conditions have been added that are not managed as part of validation.  The side effect is that non-validation conditions get cleared too.  

I will likely open a PR today to add support for (optionally) limiting _staging_ to specific condition categories with the default being those categories managed by validation.  Namely: Critical, Error, Warn.  This will make the behavior explicit, as well as, configurable.  The `Closed` condition is currently category=Critical which we _could_ reconsider changing but fine the way it is.

pkg/controller/migplan/close.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// Ensure that resources managed by the MigPlan have been cleaned up
+func (r ReconcileMigPlan) ensureClosed(plan *migapi.MigPlan) error {
+	// Migration Registry
+	err := r.ensureMigRegistriesDelete(plan)
+	if err != nil {
+		return err
+	}
+
```
jortel: Let's clear the `RegistriesEnsured` (and eventually the `StorageEnsured`) condition here as well.

djwhatle:
@sseago PR looks great, thanks for working with us :+1: ready to merge when you guys are

pkg/controller/migplan/close.go
```diff
@@ -0,0 +1,63 @@
...
+	plan.Status.DeleteCondition(RegistriesEnsured)
```
jortel: :+1:

PR #132, opened 2019-05-31
Allow PV discovery if enough of MigPlan is present

djwhatle:
UI flow has PV actions being filled out before MigStorage is created, this PR aims to allow for PV discovery if the user has supplied enough info in MigPlan, rather than bailing out of PV discovery if the MigPlan is missing _any_ info.


djwhatle:
Tested this, worked for a stateful NFS 'move' migration :+1: @eriknelson

PR #136, opened 2019-06-04
Report migration errors.

jortel:
Initial pass to report available velero failure information.  In future iterations - more detail can be provided by mining the velero pod log or information velero stashes in the cloud.

Add `Errors` array to describe errors reported by velero.  For validation errors, the velero `ValidationErrors` will be propagated.  Otherwise, an error will be added that either the backup or restore failed including the namespaced name.

Rename: `TaskPhase` to `Phase`.  We were _on the fence_ about this but I think it provides useful troubleshooting information and thinking this _tips the scales_.

The UI or CLI user can determine failure by:
- Errors is not nil (or empty).
- The phase will be `BackupFailed` or `RestoreFailed`.
- The phase is not `Complete` and `MigrationCompleted` is _true_.

For now, the user will need to troubleshoot by reviewing the pod logs.

### DIFF
--- a/config/crds/migration_v1alpha1_migmigration.yaml
+++ b/config/crds/migration_v1alpha1_migmigration.yaml
@@ -30,6 +30,8 @@ spec:
           properties:
             migPlanRef:
               type: object
+            quiescePods:
+              type: boolean
             stage:
               type: boolean
           required:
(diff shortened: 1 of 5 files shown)


config/crds/migration_v1alpha1_migmigration.yaml
```diff
@@ -30,6 +30,8 @@ spec:
           properties:
             migPlanRef:
               type: object
+            quiescePods:
```
jortel: I don't know why `make manifest` changed this.

djwhatle: It must have been missed like you say.

PR #137, opened 2019-06-04
Grab secret resource name after Create call to avoid extra Get

sseago:
This avoids a possible "not found" error return from reconcile
which results in having to retry and potentially causing
duplicate resource creation.

### DIFF
--- a/pkg/controller/migplan/registry.go
+++ b/pkg/controller/migplan/registry.go
@@ -2,9 +2,9 @@ package migplan
 
 import (
 	"context"
-	"errors"
 
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
+	kapi "k8s.io/api/core/v1"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -35,25 +35,25 @@ func (r ReconcileMigPlan) ensureMigRegistries(plan *migapi.MigPlan) error {
 		}
 
 		// Migration Registry Secret
-		err := r.ensureRegistrySecret(client, plan, storage)
+		secret, err := r.ensureRegistrySecret(client, plan, storage)
 		if err != nil {
 			return err
 		}
 
 		// Migration Registry ImageStream
-		err = r.ensureRegistryImageStream(client, plan)
+		err = r.ensureRegistryImageStream(client, plan, secret)
 		if err != nil {
 			return err
 		}
 
 		// Migration Registry DeploymentConfig
-		err = r.ensureRegistryDC(client, plan, storage)
+		err = r.ensureRegistryDC(client, plan, storage, secret)
 		if err != nil {
 			return err
 		}
 
 		// Migration Registry Service
-		err = r.ensureRegistryService(client, plan)
+		err = r.ensureRegistryService(client, plan, secret)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


pkg/controller/migplan/registry.go
```diff
@@ -83,44 +82,37 @@ func (r ReconcileMigPlan) ensureMigRegistries(plan *migapi.MigPlan) error {
 }
 
 // Ensure the credentials secret for the migration registry on the specified cluster has been created
-func (r ReconcileMigPlan) ensureRegistrySecret(client k8sclient.Client, plan *migapi.MigPlan, storage *migapi.MigStorage) error {
+func (r ReconcileMigPlan) ensureRegistrySecret(client k8sclient.Client, plan *migapi.MigPlan, storage *migapi.MigStorage) (string, error) {
```
djwhatle: Do you think there's any value in returning the secret found/created here rather than just a string to it? Wondering if we would need to access it from anywhere else in the reconcile.

jortel: > Was thinking that returning an actual `secret` type here might make it clearer what you get back from this method.

Thinking this too.

pkg/controller/migplan/registry.go
```diff
@@ -186,15 +170,7 @@ func (r ReconcileMigPlan) ensureRegistryDC(client k8sclient.Client, plan *migapi
...
-func (r ReconcileMigPlan) ensureRegistryService(client k8sclient.Client, plan *migapi.MigPlan) error {
-	registrySecret, err := plan.GetRegistrySecret(client)
-	if err != nil {
-		return err
-	}
-	if registrySecret == nil {
-		return errors.New("migration registry credentials not found")
-	}
-	name := registrySecret.GetName()
+func (r ReconcileMigPlan) ensureRegistryService(client k8sclient.Client, plan *migapi.MigPlan, name string) error {
```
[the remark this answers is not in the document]
sseago: So this is no change from current behavior, but essentially that is correct. These resources will all have the same name. This seems to be the general kube convention for this sort of thing. For example, when you expose a route for "my-service", the route is also called "my-service". When adding a service to "my-dc", the service will also be called "my-dc". While the names don't have to be the same, if the names are all different, it's going to be somewhat more error-prone since a number of these resources reference others by name (i.e. "app" labels, etc.). I don't know if it's any riskier than other cases where service names match the names of their corresponding resources.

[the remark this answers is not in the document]
sseago: I can't really think of any situation where the service names could clash, since the only time we ever use the uniquely-determined secret name as a service is right after creating (or pulling) the associated secret.

pkg/controller/migplan/registry.go
sseago: Sure. I can return the whole thing and then just call GetName() as needed down the line.

pkg/controller/migplan/registry.go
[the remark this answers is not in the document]
djwhatle: Ok, yeah I agree it seems like low possibilty of collision. I think there are ~`60,466,176` possible names here. We can address later if it's an issue.

pkg/controller/migplan/registry.go
dymurray: +1 this makes sense to me.

PR #138, opened 2019-06-04
Use fusor-dev for plugin img tag

eriknelson:
@dymurray ~~should the main velero image also use `fusor-dev`? I haven't tested with that, but I notice it has a different sha than latest in quay.~~

Realized I screwed this up; there isn't a fusor-dev tag for the plugin, but we should be using the fusor-dev tag for the main velero img?


djwhatle:
@dymurray it sounds like we still need to investigate whether this is a desired change. Let us know which tags we should be using for Velero when you know the answer.

PR #139, opened 2019-06-05
Update restic ds to use privileged scc

dymurray:


djwhatle:
Doing this for two reasons:

1. selinux type assignment for `/var/lib/kubelet/pods` (which is used by Restic to copy data out of on PV backup) is different on OpenShift 4. We need privileged scc for Restic to be able to read data out of this dir. The default scc (perhaps anyuid) worked on OpenShift 3. We may need to work with others on the OpenShift project to fix the selinux policy that resulted in `/var/lib/kubelet/pods`, but as a workaround for now we are giving Restic greater permissions to overcome this.

2. On OpenShift 4, by default our restic pod had scc `hostmount-anyuid` and wasn't able to run the restic binary to perform the restore.

PR #141, opened 2019-06-05
Atomic update of conditions to prevent oscillation.

jortel:
Issue: 
- https://github.com/fusor/mig-controller/issues/120
- https://github.com/fusor/mig-controller/issues/153

The oscillation is caused by the number of `Update()` used to update conditions.  In each controller, the `validate()` effectively clear and re-sets all conditions and updates the resource.  For controllers that manage all conditions in `validate()` the change is _atomic_.  However, the plan controller is a bit more complicated.  It manages the _error_ conditions in `validate()` but manages the _required_ (category) conditions (including `Ready`) elsewhere.  As a result, all of the _required_ conditions (including `Ready`) are removed with the 1st `Update()` and then potentially re-set in the 2nd `Update()`.  Since the plan reconciles a lot, this causes quite a bit of oscillation of the `Ready` condition.  Especially for the Plan.

The fix proposed here is to do condition staging at the reconcile level.  The pattern looks like this:

In `Reconcile()`:
1. Begin condition staging.
2. Validate.
3. Reconcile _things_.
4. Set ready condition.
5. End condition staging.
6. Update.

This pattern would be applied to all controllers.

As a result of the single update, the `Ready` condition will not oscillate unless the resource actually goes in and out of ready on each reconcile.

A potential concern: Currently, the `Ready` condition is cleared quickly at be beginning of each reconcile which means that if the resource is actually no longer ready, resources watching will detect this faster.  As a result of the proposed _single-update_,  the `Ready` condition (removal) can potentially take longer to be updated and noticed by watching resources.  This could be exacerbated by the reconcile being interrupted by an _error_ (or the scheduler).  Since there is always are race condition between resources (and the UI), I believe the potential delay is worth it.  Also, in kube, there is never a guarantee that the _main_ controller for a resource will reconcile first when multiple controllers are watching a resource.  If things are changing that affect readiness while a migration is running, chances are, things are going to get hosed anyway.

This PR includes a change to coerce the controller reconcile ordering to be deterministic for resources being watch by multiple controllers.  This will reduce _some_ of unnecessary reconciles.  But the goal is to ensure that _secondary_ controllers will reconcile only for changes already reconciled by the _main_ controller.  This will make it more intuitive and deterministic which supports easier troubleshooting.  For example, the migration controller will only reconcile changes to the plan after the plan controller has reconciled (that ResourceVersion) first.

While profiling watches and reconciles, I eliminated a few unnecessary resource updates.  As of this PR, each resource is only updated once by their controller at the end of the `Reconcile()`.

Fixes **SEGV** when discovering PVs when the source cluster ref cannot be resolved.  Needs to use `HasAnyCondition()` because `HasCondition()` returns true when matching ALL the specified conditions.

### DIFF
--- a/pkg/apis/migration/v1alpha1/condition.go
+++ b/pkg/apis/migration/v1alpha1/condition.go
@@ -158,8 +158,13 @@ func (r *Conditions) DeleteCondition(types ...string) {
 	kept := []Condition{}
 	for i := range r.List {
 		condition := r.List[i]
-		_, found := filter[condition.Type]
-		if !found {
+		_, matched := filter[condition.Type]
+		if !matched {
+			kept = append(kept, condition)
+			continue
+		}
+		if r.staging {
+			condition.staged = false
 			kept = append(kept, condition)
 		}
 	}
@@ -182,6 +187,25 @@ func (r *Conditions) HasCondition(types ...string) bool {
 	}
 
 	return len(types) > 0
+}
+
+// The collection has Any of the specified conditions.
+func (r *Conditions) HasAnyCondition(types ...string) bool {
+	if r.List == nil {
+		return false
+	}
+	for _, cndType := range types {
+		condition := r.FindCondition(cndType)
+		if condition == nil || condition.Status != True {
+			continue
+		}
+		if r.staging && !condition.staged {
+			continue
+		}
+		return true
+	}
+
+	return false
 }
 
 // The collection contains any conditions with category.
--- a/pkg/controller/migcluster/migcluster_controller.go
+++ b/pkg/controller/migcluster/migcluster_controller.go
@@ -18,13 +18,13 @@ package migcluster
 
 import (
 	"context"
+	"k8s.io/apimachinery/pkg/types"
 
 	"k8s.io/apiserver/pkg/storage/names"
 
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
 	migref "github.com/fusor/mig-controller/pkg/reference"
 	"github.com/fusor/mig-controller/pkg/remote"
-	"github.com/google/uuid"
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -123,8 +123,8 @@ func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Re
 	log = logf.Log.WithName(names.SimpleNameGenerator.GenerateName("cluster|"))
 
 	// Fetch the MigCluster
-	migCluster := &migapi.MigCluster{}
-	err := r.Get(context.TODO(), request.NamespacedName, migCluster)
+	cluster := &migapi.MigCluster{}
+	err := r.Get(context.TODO(), request.NamespacedName, cluster)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			return reconcile.Result{}, nil // don't requeue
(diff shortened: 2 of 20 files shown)


pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -123,64 +124,103 @@ func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Re
...
-		var restCfg *rest.Config
-		if migCluster.Spec.IsHostCluster {
-			restCfg, err = config.GetConfig()
-			if err != nil {
-				return reconcile.Result{}, err
-			}
-		} else {
-			restCfg, err = migCluster.BuildRestConfig(r.Client)
-			if err != nil {
-				return reconcile.Result{}, nil // don't requeue
```
jortel: Did we mean to swallow this error?

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -158,8 +158,13 @@ func (r *Conditions) DeleteCondition(types ...string) {
 	kept := []Condition{}
 	for i := range r.List {
 		condition := r.List[i]
-		_, found := filter[condition.Type]
-		if !found {
+		_, matched := filter[condition.Type]
```
djwhatle: @jortel did you find a bug in the `DeleteCondition` method, or just needed some functionality change for this PR?

pkg/controller/migmigration/migmigration_controller.go
```diff
@@ -105,6 +105,9 @@ func (r *ReconcileMigMigration) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, err // requeue
 	}
 
+	// Begin staging conditions.
```
djwhatle: My summary of what's going on here:

- At the beginning of reconcile, we start staging conditions and then run validations, which may stage some Conditions
- Additional Conditions will be staged while running the migrate() function
- Before we exit the reconcile, we apply all staged conditions to the resource

pkg/controller/migplan/migplan_controller.go
```diff
@@ -157,37 +160,25 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
 		}
 	}
 
-	if plan.Status.HasCriticalCondition() {
-		plan.Status.SetReady(false, ReadyMessage)
```
djwhatle: I guess the Ready condition isn't meant to be set directly, is that right? I thought it was supposed to be set based on the set of other existing Conditions.

pkg/controller/migplan/registry.go
```diff
@@ -70,10 +70,7 @@ func (r ReconcileMigPlan) ensureMigRegistries(plan *migapi.MigPlan) error {
 			Category: migapi.Required,
 			Message:  RegistriesEnsuredMessage,
 		})
-	} else {
-		plan.Status.DeleteCondition(RegistriesEnsured)
```
djwhatle: @jortel is this not necessary for some reason?

pkg/controller/migstorage/validation.go
```diff
@@ -73,18 +71,6 @@ func (r ReconcileMigStorage) validate(storage *migapi.MigStorage) error {
 		return err
 	}
 
-	// Ready
-	storage.Status.SetReady(
-		!storage.Status.HasBlockerCondition(),
-		ReadyMessage)
-
-	// Apply changes.
-	storage.Status.EndStagingConditions()
```
djwhatle: So validate (which runs near the beginning of reconcile) used to begin and end staging conditions, now it only begins staging. Makes sense.

pkg/apis/migration/v1alpha1/condition.go
jortel: Found a bug.  The delete should not actually remove the condition while staging.

pkg/controller/migmigration/migmigration_controller.go
jortel: :+1:

pkg/controller/migplan/migplan_controller.go
jortel: Not sure I understand what you're asking.  The `SetReady()` is just for convenience and standardization.  This -/+ is related to setting the ready condition in one place at the end.

pkg/controller/migplan/registry.go
jortel: It's not necessary because of condition _staging_.

djwhatle:
This is working great now!

PR #148, opened 2019-06-07
Additional migration validations and proper sequencing.

jortel:
Related Issues:
- https://github.com/fusor/mig-controller/issues/146
- https://github.com/fusor/mig-controller/issues/86

Changes:
- Prevent migrations from running when the plan is _closed_.  Mainly covers race condition between user updating the plan and the plan controller clearing the `Ready`.
- Prevent migrations (associated with a plan) from running concurrently.
- Prevent any migration from running after a _final_ migration has successfully completed.
- Ensures that migrations are run in the order created with the _final_ migration last.  In practice, multiple migrations should not exist but nothing to prevent the user from doing this.  The goal is deal with it as intuitively and well mannered as possible.
- Ignore reconcile on completed migrations.
- Final migration no longer closes the plan.

### DIFF
--- a/pkg/apis/migration/v1alpha1/migmigration_types.go
+++ b/pkg/apis/migration/v1alpha1/migmigration_types.go
@@ -22,10 +22,12 @@ import (
 	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
-	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
-var log = logf.Log.WithName("controller")
+// Cache Indexes.
+const (
+	PlanIndexField = "planRef"
+)
 
 // MigMigrationSpec defines the desired state of MigMigration
 type MigMigrationSpec struct {
@@ -99,9 +101,19 @@ func (r *MigMigration) MarkAsCompleted() bool {
 	return true
 }
 
+// Get whether the migration is running.
+func (r *MigMigration) IsRunning() bool {
+	return r.Status.MigrationRunning
+}
+
 // Get whether the migration has completed.
 func (r *MigMigration) IsCompleted() bool {
 	return r.Status.MigrationCompleted
+}
+
+// Get whether the migration has failed.
+func (r *MigMigration) HasFailed() bool {
+	return r.IsCompleted() && len(r.Status.Errors) > 0
 }
 
 // Add (de-duplicated) errors.
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -19,6 +19,7 @@ package v1alpha1
 import (
 	"context"
 	"errors"
+	"fmt"
 	appsv1 "github.com/openshift/api/apps/v1"
 	imagev1 "github.com/openshift/api/image/v1"
 	kapi "k8s.io/api/core/v1"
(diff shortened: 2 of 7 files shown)


pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -468,6 +468,57 @@ func (r *MigPlan) GetRegistryService(client k8sclient.Client) (*kapi.Service, er
...
+// Sorted by created timestamp with final migrations grouped last.
+func (r *MigPlan) ListMigrations(client k8sclient.Client) ([]*MigMigration, error) {
+	stage := []*MigMigration{}
+	final := []*MigMigration{}
+	qlist := MigMigrationList{}
+
+	// Search
+	err := client.List(
+		context.TODO(),
+		k8sclient.InNamespace(r.Namespace),
```
sseago: This only finds MigMigrations in the same namespace as the plan. The controller doesn't require that migrations live in a particular namespace. I just created a migration in my test env in a different namespace from the plan and it ran. Also, this pulls all migrations, even those associated with other plans. I wonder if it might make sense to add a migplan label to the migration -- perhaps actually using the existing correlation label infrastructure to add a migplan: label, or if it makes sense to use a different name here, that would be fine too. Since I don't think we can use the MigPlanRef in the selection criteria, correlation labels (or something equivalent) would allow us to find all migrations associated with this migplan regardless of namespace.

pkg/controller/migmigration/migmigration_controller.go
```diff
@@ -149,11 +163,53 @@ func (r *ReconcileMigMigration) Reconcile(request reconcile.Request) (reconcile.
...
+		return 0, nil
+	}
+	// Postpone
+	for position, uid := range pending {
+		migration.Status.SetCondition(migapi.Condition{
+			Type:     Postponed,
+			Status:   True,
+			Category: Critical,
+			Message:  PostponedMessage,
+		})
```
sseago: I'm not quite following the workflow here. It looks like we're setting the Postponed condition for this migration each time through the loop to the same thing, and then when we find the position of this migration, we use it to set the requeue time. Should this statement be outside the loop? Should it always run or only if there's at least one pending migration?

pkg/apis/migration/v1alpha1/migplan_types.go
jortel: Fixed namespace issue.
Indexed the `planRef` field and using field selector.

pkg/controller/migmigration/migmigration_controller.go
jortel: Yes, meant to put it outside the loop.  Fixed.

pkg/apis/migration/v1alpha1/migmigration_types.go
```diff
@@ -22,10 +22,11 @@ import (
 	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
-	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
-var log = logf.Log.WithName("controller")
```
jortel: Side issue. This ^ log should not have been defined here.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -468,6 +469,52 @@ func (r *MigPlan) GetRegistryService(client k8sclient.Client) (*kapi.Service, er
...
+// Sorted by created timestamp with final migrations grouped last.
+func (r *MigPlan) ListMigrations(client k8sclient.Client) ([]*MigMigration, error) {
+	stage := []*MigMigration{}
+	final := []*MigMigration{}
+	qlist := MigMigrationList{}
+
+	// Search
+	err := client.List(
+		context.TODO(),
+		k8sclient.MatchingField(
```
djwhatle: Hadn't seen this query type before. Seems useful.

jortel:
I pushed (post approval) updates to validation to refine the `HasFinalMigration` criteria.  This is based on the understanding that running a _stage_ migration after running even a failed _final_ migration could produce unpredictable and undesirable results.  Is this correct?

Another change is to delete the `HasFinalMigration` condition on non-complete migrations when a migration is deleted.  This is intended to force a reconcile so the controller can re-evaluate the world as it exists and reset the condition to match it.

djwhatle:
> I pushed (post approval) updates to validation to refine the `HasFinalMigration` criteria. This is based on the understanding that running a _stage_ migration after running even a failed _final_ migration could produce unpredictable and undesirable results. Is this correct?

I think this is correct behavior 👍 

> Another change is to delete the `HasFinalMigration` condition on non-complete migrations when a migration is deleted. This is intended to force a reconcile so the controller can re-evaluate the world as it exists and reset the condition to match it.

Also sounds like a good addition!

PR #151, opened 2019-06-10
Update migration workflow to implement 2 stage approach for copy

dymurray:
This PR significantly reworks the migration workflow. The main change is that instead of having just 1 backup and 1 restore, we have 2 backups and 2 restores. The workflow goes like this:

* Backup 1: Captures all resources in the targeted namespaces. This is called `InitialBackup`. This backup explicitly sets `includePersistentVolumes: false` so it only captures resource state, no PV data at all. This backup is *not* run on stage migrations.

* Create Stage Pods: This creates staging "dummy" pods for all pods that have restic annotations. These will be used to copy PV data. The stage pods are rhel7 pods that mount the same volumes.

* Quiesce source application: This does *not* run on stage

* Backup 2: Captures only PVs, PVCs, and Stage Pods. For `copy`, this will perform restic backup/restore on the staging pods. For `move` this will simply swing the PV over

* Restore 1: Runs a restore of Backup 2. This copies all PVCs and PV data over and brings the staging pods to the destination cluster.

* Delete stage pods: This deletes all staging rhel7 pods on both source and destination clusters

* Restore 2: Runs a restore of Backup 1. This brings all application resources over including the original application pods.

**Future Enhancements**

* We could run a check on all PVs to see if there are zero `copy` PVs specified. If there are none, we could simple do the old approach with 1 backup and 1 restore. The backup will set `includePersistentVolumes` to `true`

### DIFF
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -70,7 +70,7 @@ func (t *Task) ensureInitialBackup() error {
 
 // Ensure the second backup on the source cluster has been created and has the
 // proper settings.
-func (t *Task) ensureCopyBackup() error {
+func (t *Task) ensureStageBackup() error {
 	newBackup, err := t.buildBackup(nil)
 	if err != nil {
 		return err
@@ -91,7 +91,7 @@ func (t *Task) ensureCopyBackup() error {
 		return err
 	}
 	if foundBackup == nil {
-		t.CopyBackup = newBackup
+		t.StageBackup = newBackup
 		client, err := t.getSourceClient()
 		if err != nil {
 			return err
@@ -102,7 +102,7 @@ func (t *Task) ensureCopyBackup() error {
 		}
 		return nil
 	}
-	t.CopyBackup = foundBackup
+	t.StageBackup = foundBackup
 	if !t.equalsBackup(newBackup, foundBackup) {
 		client, err := t.getSourceClient()
 		if err != nil {
@@ -483,7 +483,7 @@ func (t *Task) areStagePodsCreated(resticAnnotationCount int) (bool, error) {
 		}
 	}
 	// Check if all pods are ready
-	if readyCount == podCount {
+	if readyCount == podCount && podCount != 0 {
 		return true, nil
 	}
 	return false, nil
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -515,14 +516,35 @@ func (t *Task) createStagePods() error {
 		if strings.HasSuffix(pod.Name, stagePodSuffix) {
 			continue
 		}
+		resticVolumeNames := strings.Split(pod.Annotations[resticPvBackupAnnotationKey], ",")
 		stagePod := corev1.Pod{}
-		stagePod.Spec.Containers = pod.Spec.Containers
+		stagePod.Spec.Containers = []corev1.Container{}
+		stagePod.Spec.Volumes = []corev1.Volume{}
+		for _, volume := range resticVolumeNames {
+			for _, podVolume := range pod.Spec.Volumes {
+				if volume == podVolume.Name {
+					stagePod.Spec.Volumes = append(stagePod.Spec.Volumes, podVolume)
+				}
+			}
+		}
 		stagePod.Spec.Volumes = pod.Spec.Volumes
 		// Swap image ref, command, and args and create a new stage pod to be backed up
-		for i, _ := range stagePod.Spec.Containers {
-			stagePod.Spec.Containers[i].Image = rhel7ImageRef
-			stagePod.Spec.Containers[i].Command = []string{"sleep"}
-			stagePod.Spec.Containers[i].Args = []string{"infinity"}
+		for i, c := range pod.Spec.Containers {
+			newContainer := corev1.Container{
+				Name:    fmt.Sprintf("sleep-%s", strconv.Itoa(i)),
+				Image:   rhel7ImageRef,
+				Command: []string{"sleep"},
+				Args:    []string{"infinity"},
+			}
+			newContainer.VolumeMounts = []corev1.VolumeMount{}
+			for _, vm := range c.VolumeMounts {
+				for _, resticVolume := range resticVolumeNames {
+					if resticVolume == vm.Name {
+						newContainer.VolumeMounts = append(newContainer.VolumeMounts, vm)
+					}
+				}
+			}
+			stagePod.Spec.Containers = append(stagePod.Spec.Containers, newContainer)
 		}
 		stagePod.Name = fmt.Sprintf("%s-%s", pod.Name, stagePodSuffix)
 		stagePod.Namespace = pod.Namespace
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -144,6 +144,10 @@ func (t Task) getBackup(copyBackup bool) (*velero.Backup, error) {
 	// Find proper backup to return
 	if len(list.Items) > 0 {
 		for i, backup := range list.Items {
+			// Avoid nil annotation lookup
+			if backup.Annotations == nil {
+				backup.Annotations = make(map[string]string)
+			}
 			if backup.Annotations[copyBackupRestoreAnnotationKey] != "" && copyBackup {
 				return &list.Items[i], nil
 			}
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -11,28 +11,32 @@ import (
 	velero "github.com/heptio/velero/pkg/apis/velero/v1"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var resticPodPrefix = "restic-"
+var rhel7ImageRef = "registry.access.redhat.com/rhel7"
 
-// Ensure the backup on the source cluster has been created
-// and has the proper settings.
-func (t *Task) ensureBackup() error {
-	newBackup, err := t.buildBackup()
+// Ensure the initial backup on the source cluster has been created and has the
+// proper settings.
+func (t *Task) ensureInitialBackup() error {
+	includeClusterResources := false
+	newBackup, err := t.buildBackup(&includeClusterResources)
 	if err != nil {
 		log.Trace(err)
 		return err
 	}
-	foundBackup, err := t.getBackup()
+	delete(newBackup.Annotations, MigQuiesceAnnotationKey)
+	foundBackup, err := t.getBackup(false)
 	if err != nil {
 		log.Trace(err)
 		return err
 	}
 	if foundBackup == nil {
-		t.Backup = newBackup
+		t.InitialBackup = newBackup
 		client, err := t.getSourceClient()
 		if err != nil {
 			log.Trace(err)
@@ -45,7 +49,7 @@ func (t *Task) ensureBackup() error {
 		}
 		return nil
 	}
-	t.Backup = foundBackup
+	t.InitialBackup = foundBackup
 	if !t.equalsBackup(newBackup, foundBackup) {
 		client, err := t.getSourceClient()
 		if err != nil {
@@ -56,6 +60,55 @@ func (t *Task) ensureBackup() error {
 		err = client.Update(context.TODO(), foundBackup)
 		if err != nil {
 			log.Trace(err)
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Ensure the second backup on the source cluster has been created and has the
+// proper settings.
+func (t *Task) ensureCopyBackup() error {
+	newBackup, err := t.buildBackup(nil)
+	if err != nil {
+		return err
+	}
+	uniqueBackupLabelKey := fmt.Sprintf("%s-%s", pvBackupLabelKey, t.PlanResources.MigPlan.UID)
+	labelSelector := metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			uniqueBackupLabelKey:                         "true",
+			fmt.Sprintf("%s-copy", uniqueBackupLabelKey): "true",
+		},
+	}
+
+	// Set included resources to stage resources
+	newBackup.Spec.IncludedResources = stagingResources
+	newBackup.Spec.LabelSelector = &labelSelector
+	foundBackup, err := t.getBackup(true)
+	if err != nil {
+		return err
+	}
+	if foundBackup == nil {
+		t.CopyBackup = newBackup
+		client, err := t.getSourceClient()
+		if err != nil {
+			return err
+		}
+		err = client.Create(context.TODO(), newBackup)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+	t.CopyBackup = foundBackup
+	if !t.equalsBackup(newBackup, foundBackup) {
+		client, err := t.getSourceClient()
+		if err != nil {
+			return err
+		}
+		err = client.Update(context.TODO(), foundBackup)
+		if err != nil {
 			return err
 		}
 	}
@@ -74,7 +127,7 @@ func (t *Task) equalsBackup(a, b *velero.Backup) bool {
 }
 
 // Get an existing Backup on the source cluster.
-func (t Task) getBackup() (*velero.Backup, error) {
+func (t Task) getBackup(copyBackup bool) (*velero.Backup, error) {
 	client, err := t.getSourceClient()
 	if err != nil {
 		return nil, err
@@ -88,8 +141,16 @@ func (t Task) getBackup() (*velero.Backup, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Find proper backup to return
 	if len(list.Items) > 0 {
-		return &list.Items[0], nil
+		for i, backup := range list.Items {
+			if backup.Annotations[copyBackupRestoreAnnotationKey] != "" && copyBackup {
+				return &list.Items[i], nil
+			}
+			if backup.Annotations[copyBackupRestoreAnnotationKey] == "" && !copyBackup {
+				return &list.Items[i], nil
+			}
+		}
 	}
 
 	return nil, nil
@@ -132,7 +193,7 @@ func (t *Task) getVSL() (*velero.VolumeSnapshotLocation, error) {
 }
 
 // Build a Backups as desired for the source cluster.
-func (t *Task) buildBackup() (*velero.Backup, error) {
+func (t *Task) buildBackup(includeClusterResources *bool) (*velero.Backup, error) {
 	// Get client of source cluster
 	client, err := t.getSourceClient()
 	if err != nil {
@@ -143,12 +204,20 @@ func (t *Task) buildBackup() (*velero.Backup, error) {
 		log.Trace(err)
 		return nil, err
 	}
+	// If includeClusterResources isn't set, this means it is the second backup,
+	// first restore to satisfy moving the persistent storage over
+	if includeClusterResources == nil {
+		annotations[copyBackupRestoreAnnotationKey] = "true"
+	}
 	backup := &velero.Backup{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:       t.Owner.GetCorrelationLabels(),
 			GenerateName: t.Owner.GetName() + "-",
 			Namespace:    migapi.VeleroNamespace,
 			Annotations:  annotations,
+		},
+		Spec: velero.BackupSpec{
+			IncludeClusterResources: includeClusterResources,
 		},
 	}
 	err = t.updateBackup(backup)
@@ -171,14 +240,15 @@ func (t *Task) updateBackup(backup *velero.Backup) error {
 	backup.Spec = velero.BackupSpec{
 		StorageLocation:         backupLocation.Name,
 		VolumeSnapshotLocations: []string{snapshotLocation.Name},
-		TTL:                     metav1.Duration{Duration: 720 * time.Hour},
-		IncludedNamespaces:      namespaces,
-		ExcludedNamespaces:      []string{},
-		IncludedResources:       t.BackupResources,
-		ExcludedResources:       []string{},
+		TTL:                metav1.Duration{Duration: 720 * time.Hour},
+		IncludedNamespaces: namespaces,
+		ExcludedNamespaces: []string{},
+		IncludedResources:  t.BackupResources,
+		ExcludedResources:  []string{},
 		Hooks: velero.BackupHooks{
 			Resources: []velero.BackupResourceHookSpec{},
 		},
+		IncludeClusterResources: backup.Spec.IncludeClusterResources,
 	}
 
 	return nil
@@ -199,7 +269,12 @@ func (t *Task) getReplicatedBackup() (*velero.Backup, error) {
 	if err != nil {
 		return nil, err
 	}
-	if len(list.Items) > 0 {
+	// If this is stage we only need to wait for 1 backup to be replicated
+	if len(list.Items) == 1 && t.stage() {
+		return &list.Items[0], nil
+	}
+	// If this is not a stage, we need to find 2 backups before continuing
+	if len(list.Items) > 1 {
 		return &list.Items[0], nil
 	}
 
@@ -271,12 +346,14 @@ func (t *Task) bounceResticPod() error {
 }
 
 // Annotate all resources with PV action data
-func (t *Task) annotateStorageResources() error {
+// Return the number of copy pods so that we know if we need to create stage
+// pods
+func (t *Task) annotateStorageResources() (error, int) {
 	// Get client of source cluster
 	client, err := t.getSourceClient()
 	if err != nil {
 		log.Trace(err)
-		return err
+		return err, 0
 	}
 	uniqueBackupLabelKey := fmt.Sprintf("%s-%s", pvBackupLabelKey, t.PlanResources.MigPlan.UID)
 	namespaces := t.PlanResources.MigPlan.Spec.Namespaces
@@ -292,7 +369,7 @@ func (t *Task) annotateStorageResources() error {
 			&resource)
 		if err != nil {
 			log.Trace(err)
-			return err
+			return err, 0
 		}
 		if resource.Annotations == nil {
 			resource.Annotations = make(map[string]string)
@@ -305,6 +382,7 @@ func (t *Task) annotateStorageResources() error {
 		client.Update(context.TODO(), &resource)
 	}
 
+	resticPodCount := 0
 	for _, ns := range namespaces {
 		// Find all pods in our target namespaces
 		list := corev1.PodList{}
@@ -312,10 +390,15 @@ func (t *Task) annotateStorageResources() error {
 		err = client.List(context.TODO(), options, &list)
 		if err != nil {
 			log.Trace(err)
-			return err
+			return err, 0
 		}
 		// Loop through all pods to find all volume claims
 		for _, pod := range list.Items {
+			// Don't need to do this for staging pods. Adds unnecessary work
+			if pod.Labels[fmt.Sprintf("%s-copy", uniqueBackupLabelKey)] != "" {
+				continue
+			}
+
 			resticVolumes := []string{}
 			for _, volume := range pod.Spec.Volumes {
 				claim := volume.VolumeSource.PersistentVolumeClaim
@@ -339,12 +422,14 @@ func (t *Task) annotateStorageResources() error {
 					pvc.Labels = make(map[string]string)
 				}
 				pvc.Labels[uniqueBackupLabelKey] = pvBackupLabelValue
+				pvc.Labels[fmt.Sprintf("%s-copy", uniqueBackupLabelKey)] = pvBackupLabelValue
 				err = client.Update(context.TODO(), &pvc)
 				if action == migapi.PvCopyAction {
 					resticVolumes = append(resticVolumes, volume.Name)
 				}
 			}
 			if len(resticVolumes) > 0 {
+				resticPodCount += 1
 				if pod.Annotations == nil {
 					pod.Annotations = make(map[string]string)
 				}
@@ -358,6 +443,93 @@ func (t *Task) annotateStorageResources() error {
 		}
 	}
 
+	return nil, resticPodCount
+}
+
+func (t *Task) areStagePodsCreated() (bool, error) {
+	client, err := t.getSourceClient()
+	if err != nil {
+		return false, err
+	}
+	// check if we have already created dummy pods
+	// if so, return true
+	uniqueBackupLabelKey := fmt.Sprintf("%s-%s", pvBackupLabelKey, t.PlanResources.MigPlan.UID)
+	labelSelector := map[string]string{
+		fmt.Sprintf("%s-copy", uniqueBackupLabelKey): "true",
+	}
+	options := k8sclient.MatchingLabels(labelSelector)
+	podList := corev1.PodList{}
+	err = client.List(context.TODO(), options, &podList)
+	if err != nil {
+		return false, err
+	}
+	podCount := len(podList.Items)
+	if podCount == 0 {
+		return false, nil
+	}
+	readyCount := 0
+	for _, pod := range podList.Items {
+		if pod.Status.Phase == corev1.PodRunning {
+			readyCount += 1
+		}
+	}
+	// Check if all pods are ready
+	if readyCount == podCount {
+		return true, nil
+	}
+	return false, nil
+}
+
+func (t *Task) createStagePods() error {
+	client, err := t.getSourceClient()
+	if err != nil {
+		return err
+	}
+
+	// Dummy pods haven't been created yet, lets create them
+	uniqueBackupLabelKey := fmt.Sprintf("%s-%s", pvBackupLabelKey, t.PlanResources.MigPlan.UID)
+	labelSelector := map[string]string{
+		uniqueBackupLabelKey: pvBackupLabelValue,
+	}
+	options := k8sclient.MatchingLabels(labelSelector)
+	podList := corev1.PodList{}
+	err = client.List(context.TODO(), options, &podList)
+	if err != nil {
+		return err
+	}
+	for _, pod := range podList.Items {
+		if pod.Annotations[resticPvBackupAnnotationKey] == "" {
+			continue
+		}
+		dummyPod := corev1.Pod{}
+		dummyPod.Spec.Containers = pod.Spec.Containers
+		dummyPod.Spec.Volumes = pod.Spec.Volumes
+		// Swap image ref, command, and args and create a new dummy pod to be backed up
+		for i, _ := range dummyPod.Spec.Containers {
+			dummyPod.Spec.Containers[i].Image = rhel7ImageRef
+			dummyPod.Spec.Containers[i].Command = []string{"sleep"}
+			dummyPod.Spec.Containers[i].Args = []string{"infinity"}
+		}
+		dummyPod.Name = fmt.Sprintf("%s-dummy-pod", pod.Name)
+		dummyPod.Namespace = pod.Namespace
+		if dummyPod.Labels == nil {
+			dummyPod.Labels = make(map[string]string)
+		}
+		dummyPod.Labels[uniqueBackupLabelKey] = "true"
+		dummyPod.Labels[fmt.Sprintf("%s-copy", uniqueBackupLabelKey)] = "true"
+		if dummyPod.Annotations == nil {
+			dummyPod.Annotations = make(map[string]string)
+		}
+		dummyPod.Annotations[resticPvBackupAnnotationKey] = pod.Annotations[resticPvBackupAnnotationKey]
+		err = client.Create(context.TODO(), &dummyPod)
+		// This shouldn't happen with check at beginning of function, but we don't
+		// want to return an error if we somehow get here
+		if k8serrors.IsAlreadyExists(err) {
+			return nil
+		} else if err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -258,11 +258,12 @@ func (t *Task) updateBackup(backup *velero.Backup) error {
 	return nil
 }
 
-// Get a Backup that has been replicated by velero on the destination cluster.
-func (t *Task) getReplicatedBackup() (*velero.Backup, error) {
+// Determine whether backups are replicated by velero on the destination
+// cluster.
+func (t *Task) areBackupsReplicated() (bool, error) {
 	client, err := t.getDestinationClient()
 	if err != nil {
-		return nil, err
+		return false, err
 	}
 	list := velero.BackupList{}
 	labels := t.Owner.GetCorrelationLabels()
@@ -271,18 +272,18 @@ func (t *Task) getReplicatedBackup() (*velero.Backup, error) {
 		k8sclient.MatchingLabels(labels),
 		&list)
 	if err != nil {
-		return nil, err
+		return false, err
 	}
 	// If this is stage we only need to wait for 1 backup to be replicated
 	if len(list.Items) == 1 && t.stage() {
-		return &list.Items[0], nil
+		return true, nil
 	}
 	// If this is not a stage, we need to find 2 backups before continuing
 	if len(list.Items) > 1 {
-		return &list.Items[0], nil
+		return true, nil
 	}
 
-	return nil, nil
+	return false, nil
 }
 
 // Delete the running restic pod in velero namespace
@@ -386,7 +387,7 @@ func (t *Task) annotateStorageResources() (error, int) {
 		client.Update(context.TODO(), &resource)
 	}
 
-	resticPodCount := 0
+	resticAnnotationCount := 0
 	for _, ns := range namespaces {
 		// Find all pods in our target namespaces
 		list := corev1.PodList{}
@@ -398,6 +399,9 @@ func (t *Task) annotateStorageResources() (error, int) {
 		}
 		// Loop through all pods to find all volume claims
 		for _, pod := range list.Items {
+			if pod.Labels == nil {
+				pod.Labels = make(map[string]string)
+			}
 			// Don't need to do this for staging pods. Adds unnecessary work
 			if pod.Labels[fmt.Sprintf("%s-copy", uniqueBackupLabelKey)] != "" {
 				continue
@@ -433,7 +437,7 @@ func (t *Task) annotateStorageResources() (error, int) {
 				}
 			}
 			if len(resticVolumes) > 0 {
-				resticPodCount += 1
+				resticAnnotationCount += 1
 				if pod.Annotations == nil {
 					pod.Annotations = make(map[string]string)
 				}
@@ -447,15 +451,15 @@ func (t *Task) annotateStorageResources() (error, int) {
 		}
 	}
 
-	return nil, resticPodCount
+	return nil, resticAnnotationCount
 }
 
-func (t *Task) areStagePodsCreated() (bool, error) {
+func (t *Task) areStagePodsCreated(resticAnnotationCount int) (bool, error) {
 	client, err := t.getSourceClient()
 	if err != nil {
 		return false, err
 	}
-	// check if we have already created dummy pods
+	// check if we have already created stage pods
 	// if so, return true
 	uniqueBackupLabelKey := fmt.Sprintf("%s-%s", pvBackupLabelKey, t.PlanResources.MigPlan.UID)
 	labelSelector := map[string]string{
@@ -468,9 +472,6 @@ func (t *Task) areStagePodsCreated() (bool, error) {
 		return false, err
 	}
 	podCount := len(podList.Items)
-	if podCount == 0 {
-		return false, nil
-	}
 	readyCount := 0
 	for _, pod := range podList.Items {
 		if pod.Status.Phase == corev1.PodRunning {
@@ -478,7 +479,7 @@ func (t *Task) areStagePodsCreated() (bool, error) {
 		}
 	}
 	// Check if all pods are ready
-	if readyCount == podCount {
+	if readyCount == podCount && podCount == resticAnnotationCount {
 		return true, nil
 	}
 	return false, nil
@@ -491,7 +492,7 @@ func (t *Task) createStagePods() error {
 		return err
 	}
 
-	// Dummy pods haven't been created yet, lets create them
+	// Stage pods haven't been created yet, lets create them
 	uniqueBackupLabelKey := fmt.Sprintf("%s-%s", pvBackupLabelKey, t.PlanResources.MigPlan.UID)
 	labelSelector := map[string]string{
 		uniqueBackupLabelKey: pvBackupLabelValue,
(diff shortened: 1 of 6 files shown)


djwhatle:
@dymurray how far does this go in implementing copy? Is this a complete implementation?

pkg/controller/migmigration/migrate.go
```diff
@@ -90,11 +90,20 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (int, er
 		return 0, err
 	}
 
+	// TODO: SYNC WITH JEFF TO GET OPINION ON THIS HACK
```
dymurray: This is the part I think we might want to rework. Basically we are waiting until we know the copy has been completed and we set an annotation that is used to know if we can skip creating stage pods the next go round. I think that setting a temp var at the beginning of `task.Run()` will give us the same ability. Curious what you think @jortel.

pkg/controller/migmigration/quiesce.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"context"
+
+	appsv1 "github.com/openshift/api/apps/v1"
+	coreappsv1 "k8s.io/api/apps/v1"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// All quiesce functionality should be put here
```
dymurray: TODO: We should add ability to quiesce other things like DaemonSets, StatefulSets, and ReplicaSets. To do this we will need to do a hacky workaround like selecting a bunk nodeSelector.

pkg/controller/migmigration/backup.go
```diff
@@ -199,7 +269,12 @@ func (t *Task) getReplicatedBackup() (*velero.Backup, error) {
 	if err != nil {
 		return nil, err
 	}
-	if len(list.Items) > 0 {
+	// If this is stage we only need to wait for 1 backup to be replicated
+	if len(list.Items) == 1 && t.stage() {
```
djwhatle: If I'm understanding this correctly, `t.stage()` being set to true will avoid running the `initialBackup`, which is why we only need 1 Backup associated with our Migration (`copyBackup`) to proceed.

pkg/controller/migmigration/backup.go
```diff
@@ -199,7 +269,12 @@ func (t *Task) getReplicatedBackup() (*velero.Backup, error) {
...
+		return &list.Items[0], nil
+	}
+	// If this is not a stage, we need to find 2 backups before continuing
+	if len(list.Items) > 1 {
 		return &list.Items[0], nil
```
djwhatle: Is it intentional that when the list contains multiple items, we only return the first item in the list here?

pkg/controller/migmigration/backup.go
```diff
@@ -271,12 +346,14 @@ func (t *Task) bounceResticPod() error {
 }
 
 // Annotate all resources with PV action data
-func (t *Task) annotateStorageResources() error {
+// Return the number of copy pods so that we know if we need to create stage
```
djwhatle: What are copy pods and what determines how many of them we'll end up with?

P.S. I know stage pods are also called dummy pods, it would be good to unify the language across this PR if it hasn't happened already.

pkg/controller/migmigration/backup.go
```diff
@@ -305,17 +382,23 @@ func (t *Task) annotateStorageResources() error {
 		client.Update(context.TODO(), &resource)
 	}
 
+	resticPodCount := 0
```
djwhatle: What does `resticPodCount` represent? Is this the number of pods in the restic daemonset? e.g. == the number of nodes in the cluster?

pkg/controller/migmigration/backup.go
```diff
@@ -305,17 +382,23 @@ func (t *Task) annotateStorageResources() error {
...
 		err = client.List(context.TODO(), options, &list)
 		if err != nil {
 			log.Trace(err)
-			return err
+			return err, 0
 		}
 		// Loop through all pods to find all volume claims
 		for _, pod := range list.Items {
+			// Don't need to do this for staging pods. Adds unnecessary work
+			if pod.Labels[fmt.Sprintf("%s-copy", uniqueBackupLabelKey)] != "" {
```
djwhatle: Do we need to check that `pod.Labels` isn't nil before checking a value in its map?

pkg/controller/migmigration/backup.go
djwhatle: I'm seeing that a "restic pod" == "copy pod" now, but still don't know for sure what either of these are. I can assume that restic pods are the number of restic pods in the daemonset but want to verify that with you.

pkg/controller/migmigration/backup.go
```diff
@@ -358,6 +443,93 @@ func (t *Task) annotateStorageResources() error {
 		}
 	}
 
+	return nil, resticPodCount
+}
+
+func (t *Task) areStagePodsCreated() (bool, error) {
```
djwhatle: Just realized, don't we create these "stage pods" even when we're running a "final migration"? I suppose it makes sense if you think of a "stage" as the part of the process that something that always moves (stages) data over.

pkg/controller/migmigration/backup.go
```diff
@@ -358,6 +443,93 @@ func (t *Task) annotateStorageResources() error {
...
+	client, err := t.getSourceClient()
+	if err != nil {
+		return false, err
+	}
+	// check if we have already created dummy pods
```
djwhatle: Standardize on language

pkg/controller/migmigration/backup.go
```diff
@@ -358,6 +443,93 @@ func (t *Task) annotateStorageResources() error {
...
+		return false, nil
+	}
+	readyCount := 0
+	for _, pod := range podList.Items {
+		if pod.Status.Phase == corev1.PodRunning {
+			readyCount += 1
+		}
+	}
+	// Check if all pods are ready
+	if readyCount == podCount {
```
djwhatle: This function doesn't have a **specific number of stage pods** its trying to verify the existence of, it's just querying for the fact that all existing Pods are Ready. 

@dymurray do you see any opportunity to failure here if we were only partially successful in creating Pods? Guessing we wouldn't get past the previous step of creating the stage pods in that case.

pkg/controller/migmigration/backup.go
```diff
@@ -358,6 +443,93 @@ func (t *Task) annotateStorageResources() error {
...
+	return false, nil
+}
+
+func (t *Task) createStagePods() error {
+	client, err := t.getSourceClient()
+	if err != nil {
+		return err
+	}
+
+	// Dummy pods haven't been created yet, lets create them
```
djwhatle: Standardize language.

pkg/controller/migmigration/backup.go
```diff
@@ -358,6 +443,93 @@ func (t *Task) annotateStorageResources() error {
...
+	}
+	for _, pod := range podList.Items {
+		if pod.Annotations[resticPvBackupAnnotationKey] == "" {
+			continue
+		}
+		dummyPod := corev1.Pod{}
+		dummyPod.Spec.Containers = pod.Spec.Containers
+		dummyPod.Spec.Volumes = pod.Spec.Volumes
+		// Swap image ref, command, and args and create a new dummy pod to be backed up
+		for i, _ := range dummyPod.Spec.Containers {
```
djwhatle: Has this been fixed up to the point where it no longer copies over unneeded (breaking) info from other the original Pod definitions?

pkg/controller/migmigration/backup.go
```diff
@@ -358,6 +443,93 @@ func (t *Task) annotateStorageResources() error {
...
+	podList := corev1.PodList{}
+	err = client.List(context.TODO(), options, &podList)
+	if err != nil {
+		return err
+	}
+	for _, pod := range podList.Items {
+		if pod.Annotations[resticPvBackupAnnotationKey] == "" {
+			continue
+		}
+		dummyPod := corev1.Pod{}
```
djwhatle: Standardize language

pkg/controller/migmigration/backup.go
```diff
@@ -358,6 +443,93 @@ func (t *Task) annotateStorageResources() error {
...
+		dummyPod.Spec.Containers = pod.Spec.Containers
+		dummyPod.Spec.Volumes = pod.Spec.Volumes
+		// Swap image ref, command, and args and create a new dummy pod to be backed up
+		for i, _ := range dummyPod.Spec.Containers {
+			dummyPod.Spec.Containers[i].Image = rhel7ImageRef
+			dummyPod.Spec.Containers[i].Command = []string{"sleep"}
+			dummyPod.Spec.Containers[i].Args = []string{"infinity"}
+		}
+		dummyPod.Name = fmt.Sprintf("%s-dummy-pod", pod.Name)
```
djwhatle: Standardize language

pkg/controller/migmigration/backup.go
```diff
@@ -358,6 +443,93 @@ func (t *Task) annotateStorageResources() error {
...
+			dummyPod.Labels = make(map[string]string)
+		}
+		dummyPod.Labels[uniqueBackupLabelKey] = "true"
+		dummyPod.Labels[fmt.Sprintf("%s-copy", uniqueBackupLabelKey)] = "true"
+		if dummyPod.Annotations == nil {
+			dummyPod.Annotations = make(map[string]string)
+		}
+		dummyPod.Annotations[resticPvBackupAnnotationKey] = pod.Annotations[resticPvBackupAnnotationKey]
+		err = client.Create(context.TODO(), &dummyPod)
+		// This shouldn't happen with check at beginning of function, but we don't
```
djwhatle: @dymurray could you elaborate on what check happened at the beginning of the function that will prevent an error from occurring here? And why don't we want to return an error?

pkg/controller/migmigration/quiesce.go
djwhatle: This workaround worked well when I tried it, the strange part about it is likely that folks won't expect to have a new, invalid nodeselector set on their resources on the source cluster. It's not something you'd ever do in normal operation (unlike scaling down a deployment to 0, for example)

pkg/controller/migmigration/migrate.go
```diff
@@ -90,11 +90,20 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (int, er
...
+	// Setting this annotation to not create stage pods after copy restore has
+	// run to completion
+	if task.Phase == DeletingStagePods {
+		if migration.Annotations == nil {
+			migration.Annotations = make(map[string]string)
+		}
+		migration.Annotations["openshift.io/stage-completed"] = "true"
```
djwhatle: This is the annotation currently used to track state for whether we've already created and deleted "stage pods" or "dummy pods". Will likely be rolled into a phase check soon.

pkg/controller/migmigration/quiesce.go
```diff
@@ -0,0 +1,78 @@
...
+		options := k8sclient.InNamespace(ns)
+		err = client.List(
+			context.TODO(),
+			options,
+			&list)
+		if err != nil {
+			return err
+		}
+		for _, dc := range list.Items {
+			dc.Spec.Replicas = 0
```
djwhatle: We should only send an update call if replica count if it isn't already set to 0.

pkg/controller/migmigration/quiesce.go
```diff
@@ -0,0 +1,78 @@
...
+	return nil
+}
+
+// Scales down all Deployments
+func (t *Task) scaleDownDeployments() error {
+	client, err := t.getSourceClient()
+	if err != nil {
+		return err
+	}
+	zeroReplica := int32(0)
```
djwhatle: Do Deployments use a different internal datatype for replicas compared to DCs? Curious why you needed the `int32(0)` here.

pkg/controller/migmigration/quiesce.go
```diff
@@ -0,0 +1,78 @@
...
+		options := k8sclient.InNamespace(ns)
+		err = client.List(
+			context.TODO(),
+			options,
+			&list)
+		if err != nil {
+			return err
+		}
+		for _, dep := range list.Items {
+			dep.Spec.Replicas = &zeroReplica
```
djwhatle: Again, think we can skip the `Update` step if replicas is already 0.

pkg/controller/migmigration/quiesce.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	for _, ns := range t.PlanResources.MigPlan.Spec.Namespaces {
+		list := appsv1.DeploymentConfigList{}
+		options := k8sclient.InNamespace(ns)
+		err = client.List(
+			context.TODO(),
+			options,
+			&list)
+		if err != nil {
+			return err
```
djwhatle: We should run `log.Trace()` on these before returning the err.

pkg/controller/migmigration/restore.go
```diff
@@ -56,6 +56,46 @@ func (t *Task) ensureRestore() error {
 	return nil
 }
 
+// Ensure the first restore on the destination cluster has been
+// created  and has the proper settings.
+func (t *Task) ensureCopyRestore() error {
+	newRestore, err := t.buildRestore(nil)
```
djwhatle: Again, wondering what the behavior of passing in `nil` rather than true/false to buildRestore() is. Feels weird.

pkg/controller/migmigration/restore.go
```diff
@@ -56,6 +56,46 @@ func (t *Task) ensureRestore() error {
...
+	if err != nil {
+		return err
```
djwhatle: Think we should run `log.Trace()` before returning err in these places.

pkg/controller/migmigration/restore.go
```diff
@@ -78,15 +118,23 @@ func (t Task) getRestore() (*velero.Restore, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Find proper restore to return
 	if len(list.Items) > 0 {
-		return &list.Items[0], nil
+		for i, restore := range list.Items {
+			if restore.Annotations[copyBackupRestoreAnnotationKey] != "" && copyRestore {
```
djwhatle: Will `restore.Annotations[copyBackupRestoreAnnotationKey]` return nil in any case?

pkg/controller/migmigration/task.go
```diff
@@ -87,54 +104,123 @@ func (t *Task) Run() error {
...
+		}
+	}
+
+	t.Phase = InitialBackupCompleted
+
+	// Annotate persistent storage resources with PV actions
+	// This will also return the number of pods we have annotated to be backed up
+	// by restic. This is useful for knowing whether or not we need to create
+	// staging pods
+	err, resticPodCount := t.annotateStorageResources()
```
djwhatle: Ah, finally found what `resticPodCount` means. Thoughts on `resticAnnotationCount` instead? Now that I know what this is I don't think it will confuse me, but initially thought this was referring to the number of actual `restic` pods existing on the cluster.

pkg/controller/migmigration/task.go
```diff
@@ -87,54 +104,123 @@ func (t *Task) Run() error {
...
+	if err != nil {
+		return err
+	}
+
+	// Check if stage pods are created and running
+	created, err := t.areStagePodsCreated()
```
djwhatle: Would it make sense to pass `resticPodCount` into `areStagePodsCreated()` to make sure that the expected count matches the actual?

pkg/controller/migmigration/task.go
```diff
@@ -275,3 +406,85 @@ func (t *Task) addErrors(errors []string) {
 		t.Errors = append(t.Errors, error)
 	}
 }
+
+// Remove stage pods on source and destination cluster
+func (t *Task) removeStagePods() error {
```
djwhatle: I wonder if any rate limiting on this type of action will be needed. I could see us sending a bunch of deletes on pods that are in the process of deleting.

pkg/controller/migmigration/backup.go
```diff
@@ -11,28 +11,32 @@ import (
...
+// proper settings.
+func (t *Task) ensureInitialBackup() error {
+	includeClusterResources := false
+	newBackup, err := t.buildBackup(&includeClusterResources)
 	if err != nil {
 		log.Trace(err)
 		return err
 	}
-	foundBackup, err := t.getBackup()
+	delete(newBackup.Annotations, MigQuiesceAnnotationKey)
```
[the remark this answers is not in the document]
dymurray: Yeah so this is definitely temporary as I discussed with you and Scott. Ideally the plugin doesn't act on quiesce annotations.

But I need to delete this currently because if we were to quiesce on the first backup (and by that I mean set the annotation with the current plugin), then the pod will go down and we won't be able to annotate it and create a stage pod in before the next backup.

pkg/controller/migmigration/backup.go
```diff
@@ -11,28 +11,32 @@ import (
...
+	foundBackup, err := t.getBackup(false)
 	if err != nil {
 		log.Trace(err)
 		return err
 	}
 	if foundBackup == nil {
-		t.Backup = newBackup
+		t.InitialBackup = newBackup
```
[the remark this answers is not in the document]
dymurray: Correct. A future enhancement can be to not always do this workflow if we detect that there are no PVs set to `copy`. I can add that to the description as well.

pkg/controller/migmigration/backup.go
```diff
@@ -88,8 +141,16 @@ func (t Task) getBackup() (*velero.Backup, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Find proper backup to return
 	if len(list.Items) > 0 {
-		return &list.Items[0], nil
+		for i, backup := range list.Items {
+			if backup.Annotations[copyBackupRestoreAnnotationKey] != "" && copyBackup {
```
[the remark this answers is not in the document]
dymurray: It shouldn't because when we build a backup we always set annotations, but it certainly cannot hurt I will add it.

pkg/controller/migmigration/backup.go
```diff
@@ -11,28 +11,32 @@ import (
...
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var resticPodPrefix = "restic-"
+var rhel7ImageRef = "registry.access.redhat.com/rhel7"
```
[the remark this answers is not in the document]
dymurray: Yup publicly accessible.

pkg/controller/migmigration/task.go
dymurray: Ah that is a good idea, I will do that. Then we can just check the bool `created`

pkg/controller/migmigration/backup.go
```diff
@@ -63,6 +67,55 @@ func (t *Task) ensureBackup() error {
...
+			fmt.Sprintf("%s-copy", uniqueBackupLabelKey): "true",
+		},
+	}
+
+	// Set included resources to stage resources
+	newBackup.Spec.IncludedResources = stagingResources
+	newBackup.Spec.LabelSelector = &labelSelector
+	foundBackup, err := t.getBackup(true)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
dymurray: Yes I checked with him in slack. Done.

pkg/controller/migmigration/backup.go
```diff
@@ -11,28 +11,32 @@ import (
 	velero "github.com/heptio/velero/pkg/apis/velero/v1"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
```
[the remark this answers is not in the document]
dymurray: Well, I have to alias here actually. Right above we include `github.com/pkg/errors`

pkg/controller/migmigration/backup.go
```diff
@@ -63,6 +67,55 @@ func (t *Task) ensureBackup() error {
 	return nil
 }
 
+// Ensure the second backup on the source cluster has been created and has the
+// proper settings.
+func (t *Task) ensureCopyBackup() error {
+	newBackup, err := t.buildBackup(nil)
```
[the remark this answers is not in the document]
jortel: The `nil` seems to be a *bool which I find confusing as well.

pkg/controller/migmigration/backup.go
```diff
@@ -63,6 +67,55 @@ func (t *Task) ensureBackup() error {
 	return nil
 }
 
+// Ensure the second backup on the source cluster has been created and has the
+// proper settings.
+func (t *Task) ensureCopyBackup() error {
```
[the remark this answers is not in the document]
sseago: I agree with the basic point here -- "Copy" in `ensureCopyBackup` looks like a verb rather than an adjective here. If it's important to keep the word "Copy" in there, then `BackupForCopy` would be better. I think either of the suggested options "with/withoutPVs" or "1/2" work just fine. I'm not sure that I have an opinion on which is better.

pkg/controller/migmigration/backup.go
```diff
@@ -63,6 +67,55 @@ func (t *Task) ensureBackup() error {
...
+
+	// Set included resources to stage resources
+	newBackup.Spec.IncludedResources = stagingResources
+	newBackup.Spec.LabelSelector = &labelSelector
+	foundBackup, err := t.getBackup(true)
+	if err != nil {
+		return err
+	}
+	if foundBackup == nil {
+		t.CopyBackup = newBackup
```
sseago: This is related to the nomenclature question above around `ensureCopyBackup` -- whatever gets decided there would presumably show up here, too. `t.CopyBackup` looks too much like an action rather than an attribute here.

pkg/controller/migmigration/backup.go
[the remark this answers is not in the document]
dymurray: This is what `includeClusterScopedResources` field accepts: https://github.com/heptio/velero/blob/master/pkg/apis/velero/v1/backup.go#L57.

If it was a simple `bool` there would be no way to know if it is being unspecified or explicitly set to false. Setting it to `nil` tells Velero to automatically determine if it should include cluster scoped resources.

pkg/controller/migmigration/backup.go
dymurray: It's intentional but I'm not sure if it's really the clearest way to do this... basically I just wanted to return *something* to tell the task runner that the backups were both replicated. The task runner doesn't do anything with it AFAIK. I think that I will change this function to simply return a `bool` to tell us if the backup or backups were properly replicated

pkg/controller/migmigration/backup.go
dymurray: Changing this to `resticAnnotationCount`. This is the number of pods that have a restic annotation

pkg/controller/migmigration/backup.go
dymurray: I have updated the function to also verify it has created the same number of stage pods as restic annotations we have found. And it is possible that we could fail here if the pods aren't successful in coming up, but if that's the case then the migration cannot continue if the PVs are set to copy so I think failing is the right thing to do

pkg/controller/migmigration/backup.go
dymurray: Not yet, doing that today.

pkg/controller/migmigration/backup.go
dymurray: Sorry, referenced old code. With the `areStagePodsCreated` being implemented I removed the check at the beginning of the function to prevent duplication. Changed the comment.

pkg/controller/migmigration/quiesce.go
dymurray: Yeah they do unfortunately. First learned of this in the plugin. Not sure why they do but they do so we have to do this. Hate how ugly it looks.

pkg/controller/migmigration/restore.go
dymurray: See above. `includePersistentVolumes` field on the backup spec expects a `*bool`

pkg/controller/migmigration/task.go
dymurray: Good point. I suppose we could add a check to see if the pod is in a `Terminating` state and don't send the delete?

pkg/controller/migmigration/task.go
```diff
@@ -1,29 +1,44 @@
...
 	velero "github.com/heptio/velero/pkg/apis/velero/v1"
+	corev1 "k8s.io/api/core/v1"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // Annotation Keys
-const MigQuiesceAnnotationKey = "openshift.io/migrate-quiesce-pods"
+const (
+	MigQuiesceAnnotationKey        = "openshift.io/migrate-quiesce-pods"
+	copyBackupRestoreAnnotationKey = "openshift.io/copy-backup-restore"
```
sseago: Is there a reason one annotation name is exported but the other is not?

pkg/controller/migmigration/backup.go
```diff
@@ -63,6 +68,55 @@ func (t *Task) ensureBackup() error {
...
 }
 
+// Ensure the second backup on the source cluster has been created and has the
+// proper settings.
+func (t *Task) ensureCopyBackup() error {
+	newBackup, err := t.buildBackup(nil)
+	if err != nil {
+		return err
+	}
+	uniqueBackupLabelKey := fmt.Sprintf("%s-%s", pvBackupLabelKey, t.PlanResources.MigPlan.UID)
```
jortel: Why does the label reference the plan UID instead of the migration UID?

pkg/controller/migmigration/backup.go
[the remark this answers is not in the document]
dymurray: Certainly not opposed to renaming it. If others agree on this I will happily change it.

pkg/controller/migmigration/backup.go
```diff
@@ -88,8 +142,20 @@ func (t Task) getBackup() (*velero.Backup, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Find proper backup to return
 	if len(list.Items) > 0 {
```
jortel: Why test the length of the list.Items before iterating?  The _range_ on an empty collection will have the same result.

pkg/controller/migmigration/backup.go
```diff
@@ -171,24 +245,26 @@ func (t *Task) updateBackup(backup *velero.Backup) error {
...
-		ExcludedResources:       []string{},
+		TTL:                metav1.Duration{Duration: 720 * time.Hour},
+		IncludedNamespaces: namespaces,
+		ExcludedNamespaces: []string{},
+		IncludedResources:  t.BackupResources,
+		ExcludedResources:  []string{},
 		Hooks: velero.BackupHooks{
 			Resources: []velero.BackupResourceHookSpec{},
 		},
+		IncludeClusterResources: backup.Spec.IncludeClusterResources,
```
jortel: Battling Go _fmt_.  Weird.

pkg/controller/migmigration/migrate.go
jortel: Seems like the consensus is to use the phase assuming the following changes:
- Rewind would be prevented.  It will always reflect the _highest_ phase completed.
- The phase would support order comparisons like phase _After()_ or _>_, _<_.

pkg/controller/migmigration/quiesce.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+// All quiesce functionality should be put here
+func (t *Task) quiesceApplications() error {
+	// If this is stage don't quiesce anything
+	if t.stage() {
+		return nil
+	}
+	err := t.scaleDownDCs()
+	if err != nil {
```
jortel: log.Trace(err)

pkg/controller/migmigration/quiesce.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+	err = t.scaleDownDeployments()
```
jortel: `+ log.Trace(err)`

pkg/controller/migmigration/quiesce.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+	for _, ns := range t.PlanResources.MigPlan.Spec.Namespaces {
+		list := appsv1.DeploymentConfigList{}
+		options := k8sclient.InNamespace(ns)
+		err = client.List(
+			context.TODO(),
+			options,
+			&list)
+		if err != nil {
```
jortel: `+ log.Trace(err)`

pkg/controller/migmigration/quiesce.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+	err = t.scaleDownDeployments()
+	return err
+}
+
+// Scales down Deployment Configs on source cluster
+func (t *Task) scaleDownDCs() error {
+	client, err := t.getSourceClient()
+	if err != nil {
```
jortel: `+ log.Trace(err)`

pkg/controller/migmigration/quiesce.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			return err
+		}
+		for _, dc := range list.Items {
+			if dc.Spec.Replicas == 0 {
+				continue
+			}
+			dc.Spec.Replicas = 0
+			err = client.Update(context.TODO(), &dc)
+			if err != nil {
```
jortel: `+ log.Trace(err)`

pkg/controller/migmigration/restore.go
```diff
@@ -78,15 +124,27 @@ func (t Task) getRestore() (*velero.Restore, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Find proper restore to return
 	if len(list.Items) > 0 {
-		return &list.Items[0], nil
+		for i, restore := range list.Items {
```
jortel: I don't think the `if len(list.Items)` ^above^ adds value.

pkg/controller/migmigration/restore.go
```diff
@@ -78,15 +124,27 @@ func (t Task) getRestore() (*velero.Restore, error) {
...
+			// Avoid nil dereference
+			if restore.Annotations == nil {
+				restore.Annotations = make(map[string]string)
+			}
+			if restore.Annotations[copyBackupRestoreAnnotationKey] != "" && copyRestore {
```
jortel: What is the circumstance that the annotation would be in the map with a value of `""` ?  How/when/where does it get set that way.  

Or:

Is this a _Go_ thing?
The same as `_, found := restore.Annotations[copyBackupRestoreAnnotationKey]` and checking `if !found`?  If so, is it safe?

pkg/controller/migmigration/task.go
```diff
@@ -1,29 +1,44 @@
...
-	BackupStarted           = "BackupStarted"
-	BackupCompleted         = "BackupCompleted"
-	BackupFailed            = "BackupFailed"
+	InitialBackupStarted    = "InitialBackupStarted"
+	InitialBackupCompleted  = "InitialBackupCompleted"
+	InitialBackupFailed     = "InitialBackupFailed"
+	CopyBackupStarted       = "CopyBackupStarted"
+	CopyBackupCompleted     = "CopyBackupCompleted"
+	CopyBackupFailed        = "CopyBackupFailed"
+	CreatingStagePods       = "CreatingStagePods"
```
jortel: `CreateStagePods` would be more consistent with other phases.

sseago: Or possibly `CreateStagePodsStarted` and `CreateStagePodsCompleted` for this one and the one following.

pkg/controller/migmigration/backup.go
dymurray: Yeah good point. Can definitely change it to whatever we agree on.

pkg/controller/migmigration/task.go
dymurray: I don't remember why we exported the quiesce annotation. I think we should not export the quiesce annotation, it's only used in this package

pkg/controller/migmigration/backup.go
dymurray: No idea. Changed to reference the migration UID

pkg/controller/migmigration/restore.go
dymurray: `copyBackupRestoreAnnotationKey` is set to `""` if this is the final restore. I.e. restore #2 of backup #1. 

Yes your snippet is identical, I can change it if that is easier to read. (Probably is)

pkg/controller/migmigration/backup.go
dymurray: We agreed on the phone to instead call this `StageBackup`

jortel:
LGTM, @dymurray!
Deferring to @djwhatle for approval.

djwhatle:
I'm still investigating the failure, but so far on a basic "[stateless" scenario](https://github.com/fusor/mig-controller/tree/master/docs/scenarios/stateless) I'm getting failed migrations on phase `StageRestoreFailed`. Inspecting the respective restore shows that it is `PartiallyFailed`.

```
Errors:
  Velero:   backup does not contain top level resources directory
```

Going to proceed to other scenarios and see if I get different results. It is desired that this PR will not break existing known working migration scenarios even though it is dramatically changing the flow of migrations to support copy.

djwhatle:
@jortel and I are currently hitting an issue attempting to migrate using `copy` from NFS to EBS while running the StageBackup. 

```
{"level":"info","ts":1561568366.717578,"logger":"migration|rncwx","msg":"Migration: waiting.","name":"migmigration-sample","phase":"WaitOnResticRestart","backup":"","restore":""}
{"level":"info","ts":1561568371.3125923,"logger":"migration|qf2tv","msg":"Migration: resumed.","name":"migmigration-sample","phase":"WaitOnResticRestart"}
{"level":"info","ts":1561568371.3126671,"logger":"migration|qf2tv","msg":"Waiting for restic pod to be ready"}
{"level":"info","ts":1561568371.3128345,"logger":"migration|qf2tv","msg":"Migration: waiting.","name":"migmigration-sample","phase":"WaitOnResticRestart","backup":"","restore":""}
{"level":"info","ts":1561568376.723304,"logger":"migration|vpcjf","msg":"Migration: resumed.","name":"migmigration-sample","phase":"WaitOnResticRestart"}
{"level":"info","ts":1561568376.7233653,"logger":"migration|vpcjf","msg":"Waiting for restic pod to be ready"}
{"level":"info","ts":1561568376.7235482,"logger":"migration|vpcjf","msg":"Restic pod successfully restarted"}
{"level":"info","ts":1561568376.7410655,"logger":"migration|vpcjf","msg":"Migration: waiting.","name":"migmigration-sample","phase":"InitialBackupStarted","backup":"migmigration-sample-n54fd","restore":""}
{"level":"info","ts":1561568380.1979518,"logger":"migration|wjff7","msg":"Migration: resumed.","name":"migmigration-sample","phase":"InitialBackupStarted"}
{"level":"info","ts":1561568380.1982908,"logger":"migration|wjff7","msg":"Migration: waiting.","name":"migmigration-sample","phase":"InitialBackupStarted","backup":"migmigration-sample-n54fd","restore":""}
{"level":"info","ts":1561568389.862849,"logger":"migration|v7xz2","msg":"Migration: resumed.","name":"migmigration-sample","phase":"InitialBackupStarted"}
{"level":"info","ts":1561568389.8975227,"logger":"migration|v7xz2","msg":"Migration: waiting.","name":"migmigration-sample","phase":"CreateStagePodsStarted","backup":"migmigration-sample-n54fd","restore":""}
{"level":"info","ts":1561568398.6692595,"logger":"migration|fvlmv","msg":"Migration: resumed.","name":"migmigration-sample","phase":"CreateStagePodsStarted"}
{"level":"info","ts":1561568398.695923,"logger":"migration|fvlmv","msg":"Migration: waiting.","name":"migmigration-sample","phase":"CreateStagePodsStarted","backup":"migmigration-sample-n54fd","restore":""}
{"level":"info","ts":1561568408.0310137,"logger":"migration|j29gp","msg":"Migration: resumed.","name":"migmigration-sample","phase":"CreateStagePodsStarted"}
{"level":"info","ts":1561568408.0593767,"logger":"migration|j29gp","msg":"Migration: waiting.","name":"migmigration-sample","phase":"CreateStagePodsStarted","backup":"migmigration-sample-n54fd","restore":""}
{"level":"info","ts":1561568416.996085,"logger":"migration|dcts9","msg":"Migration: resumed.","name":"migmigration-sample","phase":"CreateStagePodsStarted"}
{"level":"info","ts":1561568417.0266328,"logger":"migration|dcts9","msg":"Migration: waiting.","name":"migmigration-sample","phase":"StageBackupStarted","backup":"migmigration-sample-n54fd","restore":""}
{"level":"info","ts":1561568426.545885,"logger":"migration|57b5b","msg":"Migration: resumed.","name":"migmigration-sample","phase":"StageBackupStarted"}
{"level":"info","ts":1561568426.5637126,"logger":"migration|57b5b","msg":"Migration: waiting.","name":"migmigration-sample","phase":"StageBackupFailed","backup":"migmigration-sample-n54fd","restore":""}
```

```
time="2019-06-26T17:00:17Z" level=info msg="[common-backup] Entering common backup plugin" backup=velero/migmigration-sample-th4tp cmd=/plugins/velero-plugins logSource="/go/src/github.com/fusor/ocp-velero-plugin/velero-plugins/common/backup.go:25" pluginName=velero-plugins
time="2019-06-26T17:00:17Z" level=info msg="Executing custom action" backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/item_backupper.go:307" name=pv8 namespace=mysql-persistent resource=pods
time="2019-06-26T17:00:17Z" level=info msg="[pv-backup] Entering Persistent Volume backup plugin" backup=velero/migmigration-sample-th4tp cmd=/plugins/velero-plugins logSource="/go/src/github.com/fusor/ocp-velero-plugin/velero-plugins/pv/backup.go:30" pluginName=velero-plugins
time="2019-06-26T17:00:17Z" level=info msg="Executing takePVSnapshot" backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/item_backupper.go:372" name=pv8 namespace=mysql-persistent resource=pods
time="2019-06-26T17:00:17Z" level=info msg="Skipping persistent volume snapshot because volume has already been backed up with restic." backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/item_backupper.go:390" name=pv8 namespace=mysql-persistent persistentVolume
=pv8 resource=pods
time="2019-06-26T17:00:17Z" level=info msg="Executing custom action" backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/item_backupper.go:307" name=mysql namespace=mysql-persistent resource=pods
time="2019-06-26T17:00:17Z" level=info msg="[common-backup] Entering common backup plugin" backup=velero/migmigration-sample-th4tp cmd=/plugins/velero-plugins logSource="/go/src/github.com/fusor/ocp-velero-plugin/velero-plugins/common/backup.go:25" pluginName=velero-plugins
time="2019-06-26T17:00:17Z" level=info msg="Executing custom action" backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/item_backupper.go:307" name=mysql-1-ff8xn-stage-pod namespace=mysql-persistent resource=pods
time="2019-06-26T17:00:17Z" level=info msg="[common-backup] Entering common backup plugin" backup=velero/migmigration-sample-th4tp cmd=/plugins/velero-plugins logSource="/go/src/github.com/fusor/ocp-velero-plugin/velero-plugins/common/backup.go:25" pluginName=velero-plugins
time="2019-06-26T17:00:21Z" level=info msg="1 errors encountered backup up item" backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/resource_backupper.go:265" name=mysql-1-ff8xn-stage-pod namespace=mysql-persistent resource=pods
time="2019-06-26T17:00:21Z" level=error msg="Error backing up item" backup=velero/migmigration-sample-th4tp error="pod volume backup failed: error getting volume path on host: expected one matching path, got 0" error.file="/go/src/github.com/heptio/velero/pkg/restic/backupper.go:173" error.function="github.com/heptio/velero/pkg/restic.(*backupper).BackupPodVolumes" group=v1 logSource="pkg/backup/resource_backupper.go:269" name=mysql-1-ff8xn-stage-pod namespace=mysql-persistent resource=pods
```

djwhatle:
I also had trouble running an NFS `move` with this PR currently, the app on the destination side failed to start up with error despite quiescing on the source side:
```
2019-06-26T15:25:59.189006Z 0 [ERROR] InnoDB: The innodb_system data file 'ibdata1' must be writable
```

I was able to reproduce this multiple times, and noted that it does not occur on current master. We will need to resolve this and the above issues before merging.

pranavgaikwad:
> @jortel and I are currently hitting an issue attempting to migrate using `copy` from NFS to EBS while running the StageBackup.
> 
> ```
> {"level":"info","ts":1561568366.717578,"logger":"migration|rncwx","msg":"Migration: waiting.","name":"migmigration-sample","phase":"WaitOnResticRestart","backup":"","restore":""}
> {"level":"info","ts":1561568371.3125923,"logger":"migration|qf2tv","msg":"Migration: resumed.","name":"migmigration-sample","phase":"WaitOnResticRestart"}
> {"level":"info","ts":1561568371.3126671,"logger":"migration|qf2tv","msg":"Waiting for restic pod to be ready"}
> {"level":"info","ts":1561568371.3128345,"logger":"migration|qf2tv","msg":"Migration: waiting.","name":"migmigration-sample","phase":"WaitOnResticRestart","backup":"","restore":""}
> {"level":"info","ts":1561568376.723304,"logger":"migration|vpcjf","msg":"Migration: resumed.","name":"migmigration-sample","phase":"WaitOnResticRestart"}
> {"level":"info","ts":1561568376.7233653,"logger":"migration|vpcjf","msg":"Waiting for restic pod to be ready"}
> {"level":"info","ts":1561568376.7235482,"logger":"migration|vpcjf","msg":"Restic pod successfully restarted"}
> {"level":"info","ts":1561568376.7410655,"logger":"migration|vpcjf","msg":"Migration: waiting.","name":"migmigration-sample","phase":"InitialBackupStarted","backup":"migmigration-sample-n54fd","restore":""}
> {"level":"info","ts":1561568380.1979518,"logger":"migration|wjff7","msg":"Migration: resumed.","name":"migmigration-sample","phase":"InitialBackupStarted"}
> {"level":"info","ts":1561568380.1982908,"logger":"migration|wjff7","msg":"Migration: waiting.","name":"migmigration-sample","phase":"InitialBackupStarted","backup":"migmigration-sample-n54fd","restore":""}
> {"level":"info","ts":1561568389.862849,"logger":"migration|v7xz2","msg":"Migration: resumed.","name":"migmigration-sample","phase":"InitialBackupStarted"}
> {"level":"info","ts":1561568389.8975227,"logger":"migration|v7xz2","msg":"Migration: waiting.","name":"migmigration-sample","phase":"CreateStagePodsStarted","backup":"migmigration-sample-n54fd","restore":""}
> {"level":"info","ts":1561568398.6692595,"logger":"migration|fvlmv","msg":"Migration: resumed.","name":"migmigration-sample","phase":"CreateStagePodsStarted"}
> {"level":"info","ts":1561568398.695923,"logger":"migration|fvlmv","msg":"Migration: waiting.","name":"migmigration-sample","phase":"CreateStagePodsStarted","backup":"migmigration-sample-n54fd","restore":""}
> {"level":"info","ts":1561568408.0310137,"logger":"migration|j29gp","msg":"Migration: resumed.","name":"migmigration-sample","phase":"CreateStagePodsStarted"}
> {"level":"info","ts":1561568408.0593767,"logger":"migration|j29gp","msg":"Migration: waiting.","name":"migmigration-sample","phase":"CreateStagePodsStarted","backup":"migmigration-sample-n54fd","restore":""}
> {"level":"info","ts":1561568416.996085,"logger":"migration|dcts9","msg":"Migration: resumed.","name":"migmigration-sample","phase":"CreateStagePodsStarted"}
> {"level":"info","ts":1561568417.0266328,"logger":"migration|dcts9","msg":"Migration: waiting.","name":"migmigration-sample","phase":"StageBackupStarted","backup":"migmigration-sample-n54fd","restore":""}
> {"level":"info","ts":1561568426.545885,"logger":"migration|57b5b","msg":"Migration: resumed.","name":"migmigration-sample","phase":"StageBackupStarted"}
> {"level":"info","ts":1561568426.5637126,"logger":"migration|57b5b","msg":"Migration: waiting.","name":"migmigration-sample","phase":"StageBackupFailed","backup":"migmigration-sample-n54fd","restore":""}
> ```
> 
> ```
> time="2019-06-26T17:00:17Z" level=info msg="[common-backup] Entering common backup plugin" backup=velero/migmigration-sample-th4tp cmd=/plugins/velero-plugins logSource="/go/src/github.com/fusor/ocp-velero-plugin/velero-plugins/common/backup.go:25" pluginName=velero-plugins
> time="2019-06-26T17:00:17Z" level=info msg="Executing custom action" backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/item_backupper.go:307" name=pv8 namespace=mysql-persistent resource=pods
> time="2019-06-26T17:00:17Z" level=info msg="[pv-backup] Entering Persistent Volume backup plugin" backup=velero/migmigration-sample-th4tp cmd=/plugins/velero-plugins logSource="/go/src/github.com/fusor/ocp-velero-plugin/velero-plugins/pv/backup.go:30" pluginName=velero-plugins
> time="2019-06-26T17:00:17Z" level=info msg="Executing takePVSnapshot" backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/item_backupper.go:372" name=pv8 namespace=mysql-persistent resource=pods
> time="2019-06-26T17:00:17Z" level=info msg="Skipping persistent volume snapshot because volume has already been backed up with restic." backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/item_backupper.go:390" name=pv8 namespace=mysql-persistent persistentVolume
> =pv8 resource=pods
> time="2019-06-26T17:00:17Z" level=info msg="Executing custom action" backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/item_backupper.go:307" name=mysql namespace=mysql-persistent resource=pods
> time="2019-06-26T17:00:17Z" level=info msg="[common-backup] Entering common backup plugin" backup=velero/migmigration-sample-th4tp cmd=/plugins/velero-plugins logSource="/go/src/github.com/fusor/ocp-velero-plugin/velero-plugins/common/backup.go:25" pluginName=velero-plugins
> time="2019-06-26T17:00:17Z" level=info msg="Executing custom action" backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/item_backupper.go:307" name=mysql-1-ff8xn-stage-pod namespace=mysql-persistent resource=pods
> time="2019-06-26T17:00:17Z" level=info msg="[common-backup] Entering common backup plugin" backup=velero/migmigration-sample-th4tp cmd=/plugins/velero-plugins logSource="/go/src/github.com/fusor/ocp-velero-plugin/velero-plugins/common/backup.go:25" pluginName=velero-plugins
> time="2019-06-26T17:00:21Z" level=info msg="1 errors encountered backup up item" backup=velero/migmigration-sample-th4tp group=v1 logSource="pkg/backup/resource_backupper.go:265" name=mysql-1-ff8xn-stage-pod namespace=mysql-persistent resource=pods
> time="2019-06-26T17:00:21Z" level=error msg="Error backing up item" backup=velero/migmigration-sample-th4tp error="pod volume backup failed: error getting volume path on host: expected one matching path, got 0" error.file="/go/src/github.com/heptio/velero/pkg/restic/backupper.go:173" error.function="github.com/heptio/velero/pkg/restic.(*backupper).BackupPodVolumes" group=v1 logSource="pkg/backup/resource_backupper.go:269" name=mysql-1-ff8xn-stage-pod namespace=mysql-persistent resource=pods
> ```

This is because `hostPath` in restic daemonset does not match the one `kubelet` uses in your OCP environment.

djwhatle:
Realized yesterday that it may be intended for the current iteration of this PR to work with pre-provsioned NFS on the destination cluster, so going to try that out today.

djwhatle:
Wanted to leave this branch as-is rather than making edits directly to Dylan's branch for additional commits that need to be added before merge to master, so @jortel has created a new "feature branch" on fusor that we will submit further PRs to, and that we will eventually merge to master.

https://github.com/fusor/mig-controller/tree/migration-with-pv-copy

Closing this PR.

PR #159, opened 2019-06-17
Error handling and logging.

jortel:
Problem:
---
The error handling philosophy in the controller is to _bubble up_ all errors to `Reconcile()` which in-turn returns the _error_.  As a result, the _caller_, in controller-runtime framework, logs the _error_ with a stack trace which ends at the Reconcile() and does not the controller code where the _error_ was encountered.

Solution Summary:
---
- Errors continue to be _bubble up_ (be returned) to the Reconcile().
- Errors are logged where they are detected.  `If err != nil`
- The Reconcile() will return `reconcole.Result{Requeue: true}, nil` instead of `reconcole.Result{}, err` 

Details:
---
 In order for stack traces to be useful, An _error_ need to be logged using `Logger.Error()` where the error is encountered.  Because the policy in the controller is to return errors up the call stack, it's not clear as to whether an error is logged or not.  So, each:
```
if err != nil {
  return nil
}
``` 
should be something like:
```
if err != nil {
  log.Error(err, "...")
  return nil
}
``` 
In some sense, this feels excessive.  However, it ensures that all errors are logged with a useful stack trace. and can be easily verified through code inspection/review.

However, doing this with a _stock_ logger will likely result in the same error being logged many times.  Also, there are errors we don't want to log.  Such as `IsConflict(err)` and `IsNotFound(err)`.  Adding the `if error.IsConflict(err)` in so many places adds clutter and seems avoidable.

To address all of these issues, the PR includes a custom logger that provides:

- `WithName()` factory method to create a logger named as (example): `migplan|x9f4j`.  Currently this is done in every controller.
- Ensures each error logged during a Reconcile() is only logged once.
- Omit _Conflict_ and _NotFound_ errors.
- Provides a `Logger.Trace()` convenience method. In most cases, the stack trace contains all of the information needed and an additional _message_ is likely to just be a rewording of the name of _top_ function in the stack.  

This PR includes the custom `Logger` and changes:
```
if err != nil {
  return nil
}
``` 
To:
```
if err != nil {
  log.Trace(err)
  return nil
}
``` 
Everywhere.

The `Trace()` can be replace with `Error()` where a _message_ will provide value.  Or, key-value pairs can be added to the `Trace()` as needed.

### DIFF
--- a/pkg/logging/logger.go
+++ b/pkg/logging/logger.go
@@ -1,0 +1,88 @@
+package logging
+
+import (
+	"fmt"
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apiserver/pkg/storage/names"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+// Logger
+// Delegates functionality to the wrapped `Real` logger.
+// Provides:
+//   - Prevents duplicate logging of the same error.
+//   - Provides a `Trace()` method for convenience and brevity.
+//   - Prevent spamming the log with `Conflict` errors.
+type Logger struct {
+	Real    logr.Logger
+	history map[error]bool
+	name    string
+}
+
+// Get a named logger.
+func WithName(name string) Logger {
+	logger := Logger{
+		Real: logf.Log.WithName(name),
+		name: name,
+	}
+	logger.Reset()
+	return logger
+}
+
+// Reset the logger.
+// Updates the generated correlation suffix in the name and
+// clears the reported error history.
+func (l *Logger) Reset() {
+	name := fmt.Sprintf("%s|", l.name)
+	name = names.SimpleNameGenerator.GenerateName(name)
+	l.Real = logf.Log.WithName(name)
+	l.history = make(map[error]bool)
+}
+
+// Logs at info.
+func (l Logger) Info(message string, kvpair ...interface{}) {
+	l.Real.Info(message, kvpair...)
+}
+
+// Logs an error.
+// Previously logged errors are ignored.
+// `Conflict` errors are not logged.
+func (l Logger) Error(err error, message string, kvpair ...interface{}) {
+	if err == nil {
+		return
+	}
+	_, found := l.history[err]
+	if found || errors.IsConflict(err) {
+		return
+	}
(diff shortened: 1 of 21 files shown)


pkg/logging/logger.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Logs an error.
+// Previously logged errors are ignored.
+func (l Logger) Error(err error, message string, kvpair ...interface{}) {
+	if err == nil {
+		return
+	}
+	_, found := l.history[err]
+	if found || errors.IsConflict(err) || errors.IsNotFound(err) {
```
djwhatle: Just to point this out to anyone looking at this PR: this new logger will skip printing the error for any log.Error() calls for:

- Previously printed identical (same pointer) error messages from the same reconcile

- `Conflict` (resourceVersion not as expected during update attempt, resolved by reconcile retry)

- `NotFound` (resource which was requested not found, which is expected in many places. @jortel I guess this would also prevent us from printing error messages when something is legitimately not found though? E.g. a situation where something we were expecting to exist got deleted somewhere along the way. What did you envision for dealing with a situation like I'm describing?

jortel: > @jortel I guess this would also prevent us from printing error messages when something is legitimately not found though? E.g. a situation where something we were expecting to exist got deleted somewhere along the way. What did you envision for dealing with a situation like I'm describing?

Sort of.  All of our _ensure_ logic does a  `Get()` on resources that may/may-not already exists.  The not-found is an anticipated, non-error result.

pkg/logging/logger.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"fmt"
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apiserver/pkg/storage/names"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+// Logger
+// Delegates functionality to the wrapped `Real` logger.
```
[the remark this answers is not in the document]
jortel: Probably a good idea.

jortel:
> As I type this, we are still discussing whether the IsNotFound errors should be dropped by the error logger, will leave that up to your discretion.

On second thought, I think we should log NotFound should it ever be returned.  The normal action taken when NotFound is detected is to create the resource.

PR #162, opened 2019-06-18
Fail migrations for velero PartiallyFailed output

sseago:
As of Velero 1.0, backups and restores can end with a "PartiallyFailed"
    state in addition to just a "Failed" state.
    
    While it's possible that some PartiallyFailed backups or restored are still
    usable, the cases I've seen so far have been essentially failed for
    migration purposes, so this commit handles PartiallyFailed in the same
    way as Failed, although the error message does distinguish the two cases.


sseago:
@jortel Those were pulled in as a result of updating the vendored velero dep from 0.11 to 1.0. I updated the velero dep in Gopkg.toml and re-ran `dep ensure`. I've separated the dep update commit from the controller code to make the distinction clearer.

jortel:
Please add:
```
fixes #154 
```
To the squash commit message so the related issue is closed on merge.

djwhatle:
@sseago few questions:

1. What's the motivation for moving to Velero 1.0 at this time? It seems like generally a good thing to bump from a pre-1.0 to a 1.0 release, but curious if there was a particular feature you were after in making the update.

2. Have we done testing to ensure that this doesn't break existing migration flows?

sseago:
@djwhatle 

1) As of this week we're actually running on a rebased-to-1.0 Velero install, and our plugin has already been rebased to 1.0 as well.
2) The bug that this PR fixes -- PartiallyFailed backups or restores cause the migration to get stuck in a backup/restore running state -- requires 1.0 since the PartiallyFailed phase didn't exist in 0.11. Essentially our running Velero install and the controller were out-of-sync.
3) I've tested a straight migration with this. Someone with an env set up for PV migration should probably confirm that nothing breaks there, too. In theory this is a low risk change -- the higher risk changes were the upgrades to Velero itself -- but testing is always a good thing.

sseago:
Also, even before our recent update to velero 1.0 final, we've been running on a 1.0 alpha velero codebase for at least a month now. We were already out-of-sync with 0.11 in the controller.

PR #164, opened 2019-06-19
Ordered phases.

jortel:
The _int_ phase adds (a little) complexity to the CR and but has the advantage of supporting _int_ comparisons.  However since the ordering is tied to the phase (number) phases cannot be re-ordered without changing their value.  Also, storing both a phase _name_ and _number_ in the Status is effectively two representations of the same data that could conflict (although unlikely).

Another approach is to define ordering for the existing _string_ phases and introduce a _struct_ to isolate the complexity.  This supports changing the order without impact to existing migrations and the `MigMigration.Status.Phase` is unchanged.  It also supports various methods to inspecting the phase.

### DIFF
--- a/pkg/controller/migmigration/task.go
+++ b/pkg/controller/migmigration/task.go
@@ -13,6 +13,7 @@ const MigQuiesceAnnotationKey = "openshift.io/migrate-quiesce-pods"
 
 // Phases
 const (
+	Created                 = ""
 	Started                 = "Started"
 	WaitOnResticRestart     = "WaitOnResticRestart"
 	ResticRestartCompleted  = "ResticRestartCompleted"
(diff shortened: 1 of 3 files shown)


pkg/controller/migmigration/task.go
```diff
@@ -27,6 +29,95 @@ const (
...
+	return n > 500
+}
+
+// Phase is final.
+func (p Phase) Failed() bool {
+	n, found := PhaseOrder[p.Name]
+	if !found {
+		n = 0
+	}
+	return n >= 500 && n < 1000
```
sseago: In one place the check is `>500` and in another it's `>=500` Probably should be one or the other -- is 500 considered a failed condition or do failed conditions start at 501?

pkg/controller/migmigration/task.go
```diff
@@ -27,6 +29,95 @@ const (
...
+
+// Phase is after `other`.
+func (p Phase) After(other string) bool {
+	nA, found := PhaseOrder[p.Name]
+	if !found {
+		nA = 0
+	}
+	nB, found := PhaseOrder[other]
+	if !found {
+		nB = 0
```
sseago: Handling of `!found` values seems a little weird here, since they're compared as if they're valued at zero. InvalidPhase.after(validPhase) will always be false, and InvalidPhase.before(validPhase) will always be true. Reverse the results for validPhase.after(invalidPhase), and both invalid will produce false for both before and after. Maybe this doesn't matter so much since we're already validating phase at the beginning of the task, though.

pkg/controller/migmigration/task.go
jortel: Good catch.  500 is the first error code and the comparison should be `>= 500`.

pkg/controller/migmigration/task.go
jortel: Dealing with `!found` is difficult to deal with.  However, I don't think this is worse than how it currently works (or would work with _int_ phases).  If either the task.Phase or the value used in the comparison isn't one of the _const_ values, the result will be unpredictable.  Would be nice if Go supported enums so we should have the compiler catch it.

jortel:
Merge **after** #151.  It will be much easier to resolve merge conflicts here.

jortel:
Rebased and tested.

PR #165, opened 2019-06-20
Remove references to mig-ui in README

eriknelson:
These statements were never correct, and the old deployment template used deploys a very old and mocked out UI. They should be removed because it's leading people to believe they're going to get a working UI when following these instructions.


djwhatle:
Closed by https://github.com/fusor/mig-controller/pull/166 since I somehow missed that @eriknelson had spun a PR for this already :1st_place_medal:

PR #175, opened 2019-06-25
Update sa-secret-remote.yaml sample to include b64 encoding sample

djwhatle:


djwhatle:
@sseago agree, it is already in the README, but I was going back through filling these samples out and couldn't remember the command off the top of my head so decided it might be helpful to others as well.

PR #179, opened 2019-07-01
Stateless migration with new migration workflow

pranavgaikwad:
Fixes problems with stateless migrations. 

Current workflow creates empty backups when there are no staging pods created, the empty backups cause restores to fail on the destination. This PR skips the stage steps when there's no need (when there are no PVs to be migrated).


pranavgaikwad:
closing this since we are not gonna use this solution anymore

PR #180, opened 2019-07-01
use new plugin repo

sseago:
The github repo for the velero plugin has been split into two repositories:
1) a general-purpose backup/restore velero plugin for openshift workloads
2) a migration-specific plugin, used by the controller which includes 1)

The old plugin image location ( quay.io/ocpmigrate/velero-plugin ) will
be the general-purpose plugin, and  quay.io/ocpmigrate/migration-plugin
will be the migration plugin used by the controller.

### DIFF
--- a/hack/deploy/manifests/20-velero-deployment.yaml
+++ b/hack/deploy/manifests/20-velero-deployment.yaml
@@ -63,7 +63,7 @@ spec:
         - name: scratch
           emptyDir: {}
       initContainers:
-        - image: quay.io/ocpmigrate/velero-plugin:latest
+        - image: quay.io/ocpmigrate/migration-plugin:latest
           imagePullPolicy: Always
           name: velero-plugin
           resources: {}


sseago:
@djwhatle  "assuming builds have been figured out for the new plugin location" -- I have pushed a build there to the 'master' branch tag and added the 'latest' tag to that, so we're at the same place with this build as we are with the old one. I'm going to have to refactor the Makefile logic into a Dockerfile before we can add it to autobuild. I hope to get to that once I've made some progress with the StorageClass work.

sseago:
It would be good to have this tested in whatever use cases others are using this for before merging. I've tested with the "local images" scenario and with the jenkins pipeline bug reproducer case, but it would be useful if you or someone else could test this image with some of the PV-related scenarios before we merge the change.

djwhatle:
@sseago ACK, will try this out today on a few scenarios.

djwhatle:
@sseago What is the difference between :latest and :master for the migration plugin repo? Which should the controller point at?

PR #181, opened 2019-07-02
Fixing stateless migration empty backup issue

pranavgaikwad:
This PR addresses the issue with stateless backups in new backup workflow.

By adding a temporary label to one namespace, we force Velero to create a backup which can be restored in stage phase even when there are no stage pods. This prevents restore step from failing.

### DIFF
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


pkg/controller/migmigration/backup.go
```diff
@@ -490,6 +490,52 @@ func (t *Task) areStagePodsCreated(resticAnnotationCount int) (bool, error) {
...
+// addLabelsToNamespace forces velero to include namespaces
+// in the second backup to avoid creating an empty backup
+func (t *Task) addLabelsToNamespace() error {
+	client, err := t.getSourceClient()
+	if err != nil {
+		log.Trace(err)
+		return err
+	}
+	uniqueBackupLabelKey := fmt.Sprintf("%s-%s", pvBackupLabelKey, t.Owner.UID)
+	namespace := t.PlanResources.MigPlan.Spec.Namespaces[0]
```
sseago: Do we need to check len(Namespaces) here before assuming Namespaces[0] exists?

sseago: I just checked and it is possible to have a migplan with an empty Namespace list, although that results in a critical condition rendering the migplan not ready. Still, it's probably safer here to ensure that there is at least one namespace before attempting to add a label to it.

sseago:
Oh -- one other thing. Do we need to add a label to every namespace, or is adding one to the first sufficient? For a stateless migration with two namespaces, will adding this to only one prevent the velero error from appearing?

pranavgaikwad:
@sseago adding label to just one namespace would make the error disappear.

sseago:
Ahh, good. So yeah, the only thing needed would be the nil Namespaces list check, then.

PR #185, opened 2019-07-05
Use velero deployed by mig-operator in the mig namespace

jmontleon:
Our velero migration plugin makes velero unsuitable for use with generic velero backup/restore operations. Since this is the case we shouldn't expect there to be no current velero instance or no desire for a future velero instance to handle generic backup/restore operations in the velero namespace.

### DIFF
--- a/hack/deploy/deploy_velero.sh
+++ b/hack/deploy/deploy_velero.sh
@@ -21,7 +21,7 @@ echo
 echo "===================================================="
 echo "Adding privileged scc to velero service account"
 echo "===================================================="
-oc adm policy add-scc-to-user privileged system:serviceaccount:velero:velero
+oc adm policy add-scc-to-user privileged system:serviceaccount:mig:velero
 
 echo
 echo "=========="
--- a/pkg/apis/migration/v1alpha1/migstorage_types.go
+++ b/pkg/apis/migration/v1alpha1/migstorage_types.go
@@ -129,7 +129,7 @@ func (r *MigStorage) BuildBSL() *velero.BackupStorageLocation {
 	location := &velero.BackupStorageLocation{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:       r.GetCorrelationLabels(),
-			Namespace:    "velero",
+			Namespace:    VeleroNamespace,
 			GenerateName: r.Name + "-",
 		},
 		Spec: velero.BackupStorageLocationSpec{
@@ -204,7 +204,7 @@ func (r *MigStorage) BuildVSL() *velero.VolumeSnapshotLocation {
 	location := &velero.VolumeSnapshotLocation{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:       r.GetCorrelationLabels(),
-			Namespace:    "velero",
+			Namespace:    VeleroNamespace,
 			GenerateName: r.Name + "-",
 		},
 		Spec: velero.VolumeSnapshotLocationSpec{
@@ -250,7 +250,7 @@ func (r *MigStorage) BuildCloudSecret(client k8sclient.Client) (*kapi.Secret, er
 	secret := &kapi.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:    r.GetCorrelationLabels(),
-			Namespace: "velero",
+			Namespace: VeleroNamespace,
 			Name:      "cloud-credentials",
 		},
 	}
--- a/hack/deploy/manifests/10-velero-crds.yaml
+++ b/hack/deploy/manifests/10-velero-crds.yaml
@@ -187,7 +187,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: velero
-  namespace: velero
+  namespace: mig
   labels:
     component: velero
 
@@ -200,7 +200,7 @@ metadata:
     component: velero
 subjects:
   - kind: ServiceAccount
-    namespace: velero
+    namespace: mig
     name: velero
 roleRef:
   kind: ClusterRole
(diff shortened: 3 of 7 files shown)


hack/deploy/deploy_velero.sh
```diff
@@ -21,7 +21,7 @@ echo
 echo "===================================================="
 echo "Adding privileged scc to velero service account"
 echo "===================================================="
-oc adm policy add-scc-to-user privileged system:serviceaccount:velero:velero
+oc adm policy add-scc-to-user privileged system:serviceaccount:mig:velero
```
sseago: I'd prefer using "mig-velero" as the namespace. They've always run in separate namespaces before, so keeping them separate reduces one variable in terms of what we're changing. It also makes it easier to reinstall everything velero (or everything controller) later without affecting the other, which is often useful in developemnt.

jmontleon:
This PR is submitted to deal with https://github.com/fusor/mig-controller/issues/182

pkg/apis/migration/v1alpha1/migstorage_types.go
```diff
@@ -129,7 +129,7 @@ func (r *MigStorage) BuildBSL() *velero.BackupStorageLocation {
 	location := &velero.BackupStorageLocation{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:       r.GetCorrelationLabels(),
-			Namespace:    "velero",
+			Namespace:    VeleroNamespace,
```
sseago: It's good that these will now use the const, regardless of what namespace velero is in.

hack/deploy/deploy_velero.sh
jmontleon: I honestly don't prefer installing what is effectively one application and seeing multiple namespaces created.  If velero were capable of being used independently I'd feel differently about it, but this instance holds no value outside of migration.

jmontleon:
I was able to successfully test it, so barring the question of same or different namespace I think it's ready to go.

hack/deploy/deploy_velero.sh
jmontleon: @jwmatthews @djwhatle @jortel thoughts?

jmontleon: Two points I'll add:
- There are no conflicting resources between velero and our work that I know of.
- it may simplify tasks related to permissions if we're confined to one namespace, if we ever try to get rid of cluster-admin

sseago: The value for me is convenience in development -- if nobody else sees value there, I can just change the const value in my own mig-controller checkout, since I"m not currently using an operator-deployed velero anyway. Once we go to the next step and make the velero location configurable on a per-cluster basis, this issue goes away anyway.

sseago: The other possible value is that the separation makes it slightly less likely that we trigger a bug where velero clobbers something currently managed in the mig namespace and vice versa. I have no idea how likely that is, so it's unclear whether that alone is sufficient motivation.

sseago: Oh -- on the permissions point, I don't think that really helps us long-term, since @jwmatthews has indicated that we should move in the direction of making velero location configurable in the future (which would allow using a shared velero instance once we make the necessary changes in the plugin to support that).

jmontleon: And if we make it configurable and the user puts it in the same namespace? If we or velero is clobbering something the other is managing we should fix that.

sseago: I don't disagree with "we should fix that". Actually, I just realized that there's possibly a more substantial reason that we shouldn't put them in the same namespace (by default) -- they logically belong to different clusters. The controller resources defined in the mig namespace belong to the cluster running the controller. The velero-related resource in whatever namespace we're assigning them to belong to the various source and destination clusters used by the MigPlan resources. Coincidentally, one of these velero-containing source and destination clusters might be the same cluster running the controller, but not necessarily. A different default namespace makes it less confusing when trying to sort out which cluster resources are integral to the controller and which ones are associated with a migration source or target. When looking at secrets, for example, the split makes it more obvious which ones belong to the src/dest cluster (velero credentials, migration registry credentials, etc.) and which ones are associated with the controller cluster (MigStorage credentials, etc.).

jwmatthews: I suggest we deploy Velero and our controller in same namespace for now, with thought as we proceed with more integration and testing over next few weeks we will get a better feel if separating to it's own namespace helps.

sseago: @jwmatthews OK -- in that case we need to be really clear in any docs that these are components which belong to logically separate clusters (controller cluster vs. src/dest cluster), and at some point we should probably produce a doc which identifies which resources will exist in the cluster and which they belong to (i.e. X, Y, and Z are resources that belong to the velero "mig" namespace which exist on every src/dest cluster, and A, B and C are resource that belong to the controller "mig" namespace and will exist on the controller cluster. Also, just to be clear -- they're only sharing a namespace in cases where a cluster is used both to host a controller instance *and* as a migration source or destination. If a cluster is only a src or destination, it will only have velero/registry-related resources. If it's only a controller, it will only have controller-related resources.

jwmatthews: @sseago let's revisit the separate namespace for Velero next week after we've hit the milestone for 7/12 demo.   I need some help from you to understand more of your concerns, probably best we get some time after a scrum to chat.

jwmatthews:
@djwhatle @jortel @sseago any objections to merging this?

djwhatle:
LGTM!

sseago:
LGTM as well.

PR #187, opened 2019-07-08
Controller work for supporting storage class changes

sseago:
includes the API changes for supporting storage
class changes, logic for choosing the default value of the dest
storage class, validation, and annotations for the velero plugin

In addition, this adds AccessModes to API, PVC annotation, and migplan validation


pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -597,30 +597,44 @@ const (
...
-	Capacity         resource.Quantity `json:"capacity,omitempty"`
-	StorageClass     string            `json:"storageClass,omitempty"`
-	SupportedActions []string          `json:"supportedActions"`
-	Action           string            `json:"action,omitempty"`
-	PVC              PVC               `json:"pvc,omitempty"`
-	staged           bool
+	Name         string            `json:"name,omitempty"`
+	Capacity     resource.Quantity `json:"capacity,omitempty"`
+	StorageClass string            `json:"storageClass,omitempty"`
+	Supported    Supported         `json:"supported"`
```
jortel: You can probably just include the `Supported` and `Selection` inline since you are not renaming them.

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -34,6 +35,7 @@ type MigClusterSpec struct {
 	IsHostCluster           bool                  `json:"isHostCluster"`
 	ClusterRef              *kapi.ObjectReference `json:"clusterRef,omitempty"`
 	ServiceAccountSecretRef *kapi.ObjectReference `json:"serviceAccountSecretRef,omitempty"`
+	StorageClasses          []StorageClass        `json:"storageClasses,omitempty"`
```
jortel: :+1: Listing the available `StorageClasses` on the Cluster.  Nice!

pkg/controller/migcluster/storageclass.go
```diff
@@ -0,0 +1,36 @@
...
+	for _, clusterStorageClass := range clusterStorageClasses {
+		storageClass := migapi.StorageClass{
+			Name:        clusterStorageClass.Name,
+			Provisioner: clusterStorageClass.Provisioner,
+		}
+		if clusterStorageClass.Annotations != nil {
+			storageClass.Default, _ = strconv.ParseBool(clusterStorageClass.Annotations["storageclass.kubernetes.io/is-default-class"])
+		}
+		storageClasses = append(storageClasses, storageClass)
+		cluster.Spec.StorageClasses = storageClasses
```
jortel: I think you want this outside the `for` loop.  Right?
This will only update the `Spec` when `clusterStorageClasses` isn't empty.  If there are no storage classes on the cluster, shouldn't the `Spec.StorageClasses` reflect that?

sseago: Good catch. Yes, this line needs to be moved outside the loop.

pkg/controller/migcluster/storageclass.go
```diff
@@ -0,0 +1,74 @@
...
+		if clusterStorageClass.Annotations != nil {
+			storageClass.Default, _ = strconv.ParseBool(clusterStorageClass.Annotations["storageclass.kubernetes.io/is-default-class"])
+		}
+		storageClasses = append(storageClasses, storageClass)
+	}
+	cluster.Spec.StorageClasses = storageClasses
+	return nil
+}
+
+var accessModeList = map[string][]kapi.PersistentVolumeAccessMode{
```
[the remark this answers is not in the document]
sseago: I think we need to add it because we're only including those provisioners we know about. If a user has other third-party provisioners installed that we don't know about, we assume RWO only, so if they have, for example, NFS provisioners that support RWX, we'll report warnings. Now we're intentionally only warning and not causing an error condition, so they could just document for their users to ignore the warning, but it would be better if we could allow them to modify this with a configmap. If we add that, though, it should be a follow-on PR rather than holding up the base functionality here.

pkg/controller/migcluster/storageclass.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	var storageClasses []migapi.StorageClass
+	for _, clusterStorageClass := range clusterStorageClasses {
+		storageClass := migapi.StorageClass{
+			Name:        clusterStorageClass.Name,
+			Provisioner: clusterStorageClass.Provisioner,
+			AccessModes: r.accessModesForProvisioner(clusterStorageClass.Provisioner),
+		}
+		if clusterStorageClass.Annotations != nil {
+			storageClass.Default, _ = strconv.ParseBool(clusterStorageClass.Annotations["storageclass.kubernetes.io/is-default-class"])
```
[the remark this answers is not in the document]
sseago: @dymurray I can return the error here. It just felt like failing reconcile over it was overkill, since if we can't parse the "is default?" boolean, it was safe to assume that the storageclass *isn't* the default. Is there a downside to this, though?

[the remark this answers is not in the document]
dymurray: Fair, I don't see a downside at the moment. I think it's fine to eat it then

jortel:
This all LGTM.

pkg/controller/migmigration/annotation.go
```diff
@@ -103,6 +109,14 @@ func (t *Task) annotatePVCs(client k8sclient.Client, pod corev1.Pod) ([]string,
 		// Add to Restic volume list.
 		if pvAction == migapi.PvCopyAction {
 			volumes = append(volumes, pv.Name)
+			// PV storageClass annotation needed by the velero plugin.
+			storageClass := findPVStorageClass(pvs, pvc.Spec.VolumeName)
+			pvc.Annotations[PvStorageClassAnnotation] = storageClass
```
pranavgaikwad: Is it intentional to set the annotation when `storageClass` is an empty string ? unlike `accessMode` annotation. @sseago

sseago: @pranavgaikwad It is, because we should never get here if it's empty for storageClass, since an empty storageClass results in an error condition on the MigPlan. For accessMode, empty is the default, as we only set it if the user intends to override the accessMode of the src PVC. We could still check for it being empty, but I don't know if we'll ever actually get here in that case.

pkg/controller/migcluster/storageclass.go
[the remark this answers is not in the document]
KarthikDev: Hi @sseago and @dymurray  - Our provisioner name is  csi.trident.netapp.io but in your list it is just netapp.io/trident, so when my source pvc has accessmode: RWX and destination PVC also i need it as RWX. Since the provisioner names are different in our cluster compared to the list mentioned in  line#40, we are getting warnings in the status field in migplan and the storage class of destination pvcs are getting overridden to default storageclasses with accessmode RWO. I can see you discussed somewhat similar issue to be handled via configmap. has this been implemented? if so, can you share me the details of configmap name and the data it should comprise of. so that we can create the configmap in respectivie namespace inorder to prevent 1. storageclass getting overridden and 2. accessmode is picking the right one instead of default RWO.

Issue #172, opened 2019-06-24
Migrations on overlapping plans can run concurrently.


jortel:
Overview
---

Foreach migration plan, the controller runs migrations serially.  The goal is to prevent migrations from doing conflicting things on the clusters involved or otherwise allowing the migrations to interfere with each other.  However, I don't think this approach alone is sufficient.  Currently, multiple plans can be created for the same source and destination clusters, and, with overlapping namespaces.  This defeats the migration (run) orchestration.  Running 2 migrations concurrently on different plans with the same source and destination clusters is effectively the same as running 2 migrations concurrently on the same plan.

It's my understanding that the use case for multiple plans with the same source & destination cluster is to support migrating workloads by groups of namespaces in sequential passes/efforts.

The conflict
---

Let's start with a complete understanding of the _conflict_. 
- The source cluster overlap will result in a Restic pod restart to be triggered by one migration in the middle of another migration's backup.  With overlapping namespaces: the pod quiesce will conflict; duplicate stage pods will be created/deleted; The PV copy/move will likely conflict.
- The destination cluster overlap with overlapping namespaces: the stage pods will conflict; The PV copy/move will conflict (or just overlap?).

Proposed Solution
---

There are 2 options for preventing this (that I can think of).

`Option 1` seems like the best option assuming it does not block any known use cases.

**Option 1:**  Add a validation that will block running migrations on (open) plans that have the same source or destination cluster.  The validation would set a _Conflict_ condition that will prevent the _Ready_ condition which will block migrations on overlapping plans.  Would this block any known user cases?

Pros:
- Simple.
- Alerts the user to the overlap.
- Seems more intuitive for the user.
- Keeps complexity low in the migration (run) orchestration algorithm.
- Promotes work flow of closing completed plans.

Cons:
- _May_ give the user impression that overlapping plans are unhealthy when they're not.
- _May_ block a use case I'm not aware of.

**Option 2:**  Enhance the migration (run) orchestration to run migrations serially across plans that have the same source or destination cluster.

Pros:
- Migrations will seem to _sort them selves out_ and just eventually run.
- Overlapping plans will remain Ready which give the user the feeling that things are healthy.

Cons:
- Migration (run) orchestration across plans adds a good bit of complexity.
- User may get confused about why the migrations on a plan are being postponed prompting question s like _"When will my migration run?"_
- The changes (edits) on one plan, will block/interrupt migrations running on another plan.
- User not alerted to the overlap.
[...]

PR #197, opened 2019-07-15
Add validation for conflicting plans.

jortel:
Fixes issue #172

This is a first pass at protecting against conflicting (_overlapping_) plans.  Plans are in conflict when they are open, both have the same _source_ or _destination_ cluster and overlapping namespaces.  When detected both plans with get an _Error_ condition that will prevent them from being ready thus preventing migrations.  To resolve the conflict, one of the plans needs to be: edited, closed or deleted.

This does not address the potential issue of one plan's migration restarting restic while another plan (not in conflict) is running a backup as part of its migration.

The goal is to take an incremental step forward.  Should this prove to be insufficient or be too restrictive to the user, we can revisit.

### DIFF
--- a/pkg/apis/migration/v1alpha1/condition_test.go
+++ b/pkg/apis/migration/v1alpha1/condition_test.go
@@ -383,7 +383,6 @@ func TestConditions_HasConditionCategoryStaging(t *testing.T) {
 	g.Expect(conditions.HasConditionCategory(Critical)).To(gomega.BeFalse())
 }
 
-
 func TestConditions_AppendingItems(t *testing.T) {
 	g := gomega.NewGomegaWithT(t)
 
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	migref "github.com/fusor/mig-controller/pkg/reference"
 	velero "github.com/heptio/velero/pkg/apis/velero/v1"
 	appsv1 "github.com/openshift/api/apps/v1"
 	imagev1 "github.com/openshift/api/image/v1"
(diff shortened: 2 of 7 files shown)


pkg/apis/migration/v1alpha1/condition_test.go
```diff
@@ -383,7 +383,6 @@ func TestConditions_HasConditionCategoryStaging(t *testing.T) {
 	g.Expect(conditions.HasConditionCategory(Critical)).To(gomega.BeFalse())
 }
 
-
```
jortel: Unrelated go _fmt_ changes by _make_.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -581,6 +587,27 @@ func (r *MigPlan) GetCloudSecret(client k8sclient.Client) (*kapi.Secret, error)
...
 }
 
+// Get whether the plan conflicts with another.
+func (r *MigPlan) HasConflict(plan *MigPlan) bool {
+	if !migref.RefEquals(r.Spec.SrcMigClusterRef, plan.Spec.SrcMigClusterRef) {
+		return false
+	}
+	if !migref.RefEquals(r.Spec.DestMigClusterRef, plan.Spec.DestMigClusterRef) {
+		return false
+	}
```
sseago: I wonder if we should also compare r.Spec.SrcMigClusterRef with plan.Spec.DestMigClusterRef (and vice versa) -- in other words if [r.Spec.SrcMigClusterRef, r.Spec.DestMigClusterRef] and [plan.Spec.SrcMigClusterRef, plan.Spec.DestMigClusterRef] intersect at all.

sseago: Actually, looking at this again, we return "no conflict" in all cases where src and dest are not identical between the two. In other words, for there to be a conflict, both src and dest must be the same. In other words, there's only a conflict when both src and dest clusters are identical *and* they overlap in at least one namespace.

sseago: Looking back at the PR comment, "both have the same source or destination cluster and overlapping namespaces" -- do you mean "or" or "and" here -- the logic here seems to take that to mean "and".

jortel: Yes, I meant `or` and the logic needs to be corrected.

jortel: @sseago Fixes pushed.

PR #200, opened 2019-07-17
Fix delete PV annotations.

### LINKED ISSUE #199: deletePVAnnotations deletes PVC annotations instead of PV annotations
deletePVAnnotations needs to remove annotations on PVs rather than on PVCs. We already have deletePVCAnnotations to deal with PVC deletion.

https://github.com/fusor/mig-controller/blob/29391193fe710575ced68dae4747a3f1a8728347/pkg/controller/migmigration/annotation.go#L353

jortel:
fixes #199

Also corrected comments and renaming for consistency.

### DIFF
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -60,7 +60,7 @@ func (t *Task) annotateStageResources() error {
 		return err
 	}
 	// PV & PVCs
-	err = t.annotatePvs(client)
+	err = t.annotatePVs(client)
 	if err != nil {
 		log.Trace(err)
 		return err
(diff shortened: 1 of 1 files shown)


pkg/controller/migmigration/annotation.go
```diff
@@ -60,7 +60,7 @@ func (t *Task) annotateStageResources() error {
 		return err
 	}
 	// PV & PVCs
-	err = t.annotatePvs(client)
+	err = t.annotatePVs(client)
```
jortel: unrelated: renamed for consistency.

Danil-Grigorev:
run tests

PR #201, opened 2019-07-17
Ensure quiesced pods have terminated.

### LINKED ISSUE #194: Quiesce does not wait for pods to finish terminating
Starting a final migration in the UI after completing a stage migration caused the pod to start terminating with Quiesce Pods: True, however the backup began and actually completed before the pod finished terminating.

On the destination pod the mariadb instance would not start, complaining at a minimum about lock files that exist. 

This was the migmigration that was created:
```
Name:         02939920-a3e0-11e9-a78c-f30556ee7e15
Namespace:    mig
Labels:       <none>
Annotations:  touch=e0b875f2-7cf9-4df6-b311-dd3ea74fad4d
API Version:  migration.openshift.io/v1alpha1
Kind:         MigMigration
Metadata:
  Creation Timestamp:  2019-07-11T13:30:12Z
  Generation:          22
  Resource Version:    5473255
  Self Link:           /apis/migration.openshift.io/v1alpha1/namespaces/mig/migmigrations/02939920-a3e0-11e9-a78c-f30556ee7e15
  UID:                 02b45491-a3e0-11e9-a219-0ed41539d98e
Spec:
  Mig Plan Ref:
    Name:        stage
    Namespace:   mig
  Quiesce Pods:  true
  Stage:         false
Status:
  Conditions:
    Category:              Advisory
    Durable:               true
    Last Transition Time:  2019-07-11T13:32:15Z
    Message:               The migration has completed successfully.
    Reason:                Completed
    Status:                True
    Type:                  Succeeded
  Phase:                   Completed
  Start Timestamp:         2019-07-11T13:30:12Z
Events:                    <none>
```

jortel:
fixes #194 

The approach is to wait for Pods (in the namespace) to terminate based on `OwnerReference` _Kind_.  The thinking is that if we scale down all: Deployment, DeploymentConfig, StatefulSet in the namespace, we can safely wait on any `Pod` with OwnerReference _Kind_ one of: ReplicationController, ReplicaSet, StatefulSet to terminate.

### DIFF
--- a/pkg/controller/migmigration/quiesce.go
+++ b/pkg/controller/migmigration/quiesce.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	appsv1 "github.com/openshift/api/apps/v1"
 	coreappsv1 "k8s.io/api/apps/v1"
+	"k8s.io/api/core/v1"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -14,7 +15,7 @@ func (t *Task) quiesceApplications() error {
 		log.Trace(err)
 		return err
 	}
-	err = t.scaleDownDCs(client)
+	err = t.scaleDownDeploymentConfigs(client)
 	if err != nil {
 		log.Trace(err)
 		return err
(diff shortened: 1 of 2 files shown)


pkg/controller/migmigration/quiesce.go
```diff
@@ -23,12 +24,12 @@ func (t *Task) quiesceApplications() error {
 	if err != nil {
 		log.Trace(err)
 	}
-
+
```
sseago: stray whitespace addition?

pkg/controller/migmigration/quiesce.go
```diff
@@ -85,3 +86,67 @@ func (t *Task) scaleDownDeployments(client k8sclient.Client) error {
...
+		options := k8sclient.InNamespace(ns)
+		err := client.List(
+			context.TODO(),
+			options,
+			&list)
+		if err != nil {
+			log.Trace(err)
+			return err
+		}
+		for _, deployment := range list.Items {
```
sseago: this variable name should probably be statefulSet to avoid confusion.

pkg/controller/migmigration/quiesce.go
```diff
@@ -85,3 +86,67 @@ func (t *Task) scaleDownDeployments(client k8sclient.Client) error {
...
+				log.Trace(err)
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Ensure scaled down pods have terminated.
+func (t *Task) ensureQuiescedPodsTerminated() (bool, error) {
```
dymurray: A comment describing the `bool` as essentially `terminated` or not would be great

jortel: Done.

pkg/controller/migmigration/quiesce.go
jortel: Seems like it.  Fixed.

Issue #202, opened 2019-07-17
Improved progress reporting.


jortel:
The follow proposal has been discussed.

The progress information will be provided in the _Running_ condition.  The `Reason` will be set to the current _phase_ (as is currently).  The `Message` will include "Step: _n_ `/`_total_.

For example:

```
Type:      Running
Category:  Advisory
Status:    True
Reason:    QuiesceApplications
Message:   Step: 12/26
Last Transition Time:  2019-07-17T16:04:32Z
``` 

We can revisit grouping phases into steps as needed.

Support: https://github.com/fusor/mig-ui/issues/331

PR #203, opened 2019-07-17
Add Progress Reporting.

jortel:
fixes: #202

### DIFF
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -174,6 +174,12 @@ func (t Task) hasBackupCompleted(backup *velero.Backup) (bool, []string) {
 				backup.Name))
 	case velero.BackupPhaseFailedValidation:
 		reasons = backup.Status.ValidationErrors
+		reasons = append(
+			reasons,
+			fmt.Sprintf(
+				"Backup: %s/%s validation failed.",
+				backup.Namespace,
+				backup.Name))
 		completed = true
 	}
 
--- a/pkg/controller/migmigration/restore.go
+++ b/pkg/controller/migmigration/restore.go
@@ -184,6 +184,12 @@ func (t Task) hasRestoreCompleted(restore *velero.Restore) (bool, []string) {
 				restore.Name))
 	case velero.RestorePhaseFailedValidation:
 		reasons = restore.Status.ValidationErrors
+		reasons = append(
+			reasons,
+			fmt.Sprintf(
+				"Restore: %s/%s validation failed.",
+				restore.Namespace,
+				restore.Name))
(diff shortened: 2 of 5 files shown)


pkg/controller/migmigration/backup.go
```diff
@@ -174,6 +174,12 @@ func (t Task) hasBackupCompleted(backup *velero.Backup) (bool, []string) {
 				backup.Name))
 	case velero.BackupPhaseFailedValidation:
 		reasons = backup.Status.ValidationErrors
+		reasons = append(
+			reasons,
+			fmt.Sprintf(
+				"Backup: %s/%s validation failed.",
+				backup.Namespace,
+				backup.Name))
```
jortel: While troubleshooting, I noticed the backup information was not being added to `Errors` for validation failures.  This really helps with investigation.

pkg/controller/migmigration/restore.go
```diff
@@ -184,6 +184,12 @@ func (t Task) hasRestoreCompleted(restore *velero.Restore) (bool, []string) {
 				restore.Name))
 	case velero.RestorePhaseFailedValidation:
 		reasons = restore.Status.ValidationErrors
+		reasons = append(
+			reasons,
+			fmt.Sprintf(
+				"Restore: %s/%s validation failed.",
+				restore.Namespace,
+				restore.Name))
```
jortel: While troubleshooting, I noticed the restore information was not being added to `Errors` for validation failures.  This really helps with investigation.

pkg/controller/migmigration/task.go
```diff
@@ -486,3 +516,17 @@ func (t *Task) addErrors(errors []string) {
 		t.Errors = append(t.Errors, error)
 	}
 }
+
+// Get the current step.
+// Returns: name, n, total.
+func (t *Task) getStep() (string, int, int) {
+	n := -1
+	total := len(Step)
+	for i, step := range Step {
```
sseago: Would making Step a map instead of a slice make this easier -- then instead of looping over the full slice to find the element number you could just use the map to find it directly on each getStep call. Not a huge deal, as this should work fine, too.

jortel: The advantage of the list is it provides order, enumeration and total number of steps.
I considered building a Map but thinking that iterating over a slice containing ~20 items vs. the overhead of hashing the string for the map lookup is pretty equal.

sseago: It's not a big deal either way. It probably looks more cumbersome than it is in practice, but I suspect there are golang style issues at work here, too, since go doesn't provide an `index` method on arrays/slices like other languages do.

PR #206, opened 2019-07-19
Add finalizers on plan and cluster.

jortel:
Add finalizers to the plan and cluster.

The plan finalizer will ensure that resources associated with plan on all _known_ clusters are deleted when the plan is deleted.

The cluster finalizer is needed to ensure that all resources created by the _migration_ application are deleted before the cluster (or the entire namespace) is deleted.

It's important to note that the finalizers are _best effort_.  On failure, they will only retry 3 times.

Added a new `part-of` _correlation_  label. Just as the _Mig_ resource specific labels provide correlation to a specific `MigResource`, the `part-of` label provides correlation to the _Mig_ application.  It may be used to identify _any_ resource created by the controller. The label _name_ was selected from kubernetes [common labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/).  The `part-of` label obsoletes the `KnownLabels` map used by the remote watch predicate.

The storage and repository _clean up_ logic was removed from `storage.go` and `repository.go` and consolidated in `MigCluster.DeleteResources()`.

After adding the finalizer support methods in the plan controller `.go` file, it seemed better to move the (2) methods in `close.go` there as well (I know, I pushed for the close.go file).


pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -177,10 +214,154 @@ func buildRestConfig(clusterURL string, bearerToken string) *rest.Config {
...
+	}
+	for _, r := range sList.Items {
+		err = client.Delete(context.TODO(), &r)
+		if err != nil && !k8serror.IsNotFound(err) {
+			return err
+		}
+	}
+
+	// Pod
+	pList := kapi.PodList{}
```
sseago: I wonder whether we should delete pods later. Deleting them first might result in deployments and dcs starting to respawn them while we're trying to delete the dc.

jortel: This is intended to deal with _stage_ Pods, not Pods related to the _registry_.  The migration does good job keeping them cleaned up but thinking there could be cases where they're _leaked_.

jortel: I changed the order so that Deployment and DeploymentConfig get deleted before Pods to minimize thrashing.

Issue #213, opened 2019-07-25
Unable to delete mig namespace/other objects (authorization issue)


fbladilo:
After performing migrations, I'm unable to delete either the mig namespace, migplans or migclusters via oc commands, they block and never return: 
```
[root@jenkins-me:~/ansible/dev/mig-ci/mig-e2e] (master %=)# oc -n mig delete migplans local-images-migplan-1564086899
migplan.migration.openshift.io "local-images-migplan-1564086899" deleted
```
Unlike when trying to delete the namespace, with an attempted migplan delete we can still see the controller logs errors : 
```
E0725 22:43:38.095514       1 reflector.go:134] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:196: Failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:mig:default" cannot list resource "deployments" in API group "apps" at the cluster scope
E0725 22:43:39.096983       1 reflector.go:134] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:196: Failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:mig:default" cannot list resource "deployments" in API group "apps" at the cluster scope
E0725 22:43:40.098480       1 reflector.go:134] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:196: Failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:mig:default" cannot list resource "deployments" in API group "apps" at the cluster scope
E0725 22:43:41.104789       1 reflector.go:134] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:196: Failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:mig:default" cannot list resource "deployments" in API group "apps" at the cluster scope
E0725 22:43:42.106185       1 reflector.go:134] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:196: Failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:mig:default" cannot list resource "deployments" in API group "apps" at the cluster scope
```
I checked the mig:default SA is actually is present in the role-binding to manager-role : 
```
oc describe clusterrolebindings manager-rolebinding
Name:         manager-rolebinding
Labels:       <none>
[...]

PR #216, opened 2019-07-26
Temporarily disable finalizers.

jortel:
Mitigate #213 and unblock folks.

There is likely an unrelated permission issue exposed by the _cleanup_ which should affect _closing_ as well.  Otherwise the plan and cluster finalizers work just fine.  The blocking problem is when users delete the entire namespace.

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -83,6 +83,9 @@ func init() {
 
 // Ensure finalizer.
 func (r *MigPlan) EnsureFinalizer() bool {
+	if !FinalizerEnabled {
+		return false
+	}
 	if r.Finalizers == nil {
 		r.Finalizers = []string{}
 	}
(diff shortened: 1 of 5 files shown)


pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -666,9 +669,8 @@ func (r *MigPlan) HasConflict(plan *MigPlan) bool {
 
 // PV Actions.
 const (
-	PvMoveAction    = "move"
-	PvCopyAction    = "copy"
-	VeleroNamespace = "mig"
```
jortel: Relocated _VeleroNamespace_ as it is not associated with PV.

pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -155,6 +145,15 @@ func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Re
 			retry = r.retryFinalizer(cluster)
 		}
 		return reconcile.Result{Requeue: retry}, nil
+	} else {
```
jortel: According to the [The Kubebuilder Book](https://book.kubebuilder.io/reference/using-finalizers.html) it's a better practice to add the finalizer only when the `DeletionTimestamp` not set.  Makes sense.

PR #235, opened 2019-08-02
Add pod affinity selector to stage pod definition

sseago:
Stage pods need to run on the same node as the related app pod, if possible.
Some storage types (glusterblock, for example) will cause the stage
pods to fail to start if they're on a different node as the app
pod. This PR defines a soft affinity, so it will go on the same
node if it is able to schedule it there, but if not, it will schedule
it elsewhere, limiting the failure scenario to only failing if the
desired node is unavailable *and* it's one of the affected storage
types.

### DIFF
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -29,6 +29,9 @@ const (
 	// Resources included in the stage backup.
 	// Referenced by the Backup.LabelSelector. The value is the Task.UID().
 	IncludedInStageBackupLabel = "migration-included-stage-backup"
+	// For pods included in the stage backup, this label contains a pod-unique
+	// value (namespace "/"+name)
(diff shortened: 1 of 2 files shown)


sseago:
One thing worth noting here, but I think it's probably not a huge issue for migration -- from the kubernetes docs: 
"Note: Inter-pod affinity and anti-affinity require substantial amount of processing which can slow down scheduling in large clusters significantly. We do not recommend using them in clusters larger than several hundred nodes."

My current impression is that this is probably more relevant if we were using this for normal everyday scheduling of all workloads. Since we're only using it for scheduling restic pods during migration, I suspect this isn't a major concern, but it's worth keeping in mind if we hit performance issues. Still, I"m not really aware of any obvious more-efficient alternatives that will predictably give us the desired behavior here.

PR #248, opened 2019-08-06
Use a new label for the application pods instead of stage label

### LINKED ISSUE #244: Stage migration is including the application pod
My stage restores are finishing with `PartiallyFailed` status. When you do `velero describe restore <>` I see:
```
Errors:
  Velero:     <none>
  Cluster:    <none>
  Namespaces:
    mssql-persistent:  error restoring pods/mssql-persistent/mssql-deployment-5-7m98j: pods "mssql-deployment-5-7m98j" is forbidden: error looking up service account mssql-persistent/mssql-persistent-sa: serviceaccount "mssql-persistent-sa" not found 
```

Forgetting the question of whether or not the service account should be used with the corresponding stage pod, the application pod should not be included during a stage. The stage migration is using a label selector, and it appears the application pod is getting caught up in it since it's not being quiesced on stage.

dymurray:
Fixes #244 

While before we were deleting this UID label after creating the stage pod to try and keep it from being included in the backup, it causes a race condition because the resource may not be updated by the time the stage backup runs. This is far safer and the label is still properly removed since it uses the same key.

Test this PR at `docker.io/dymurray/mig-controller:stageFix`

### DIFF
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -217,7 +217,7 @@ func (t *Task) annotatePods(client k8sclient.Client) error {
 			if pod.Labels == nil {
 				pod.Labels = make(map[string]string)
 			}
-			pod.Labels[IncludedInStageBackupLabel] = t.UID()
+			pod.Labels[IncludedInStageBackupLabel] = t.StagePodSearchLabel()
 			pod.Labels[StagePodAffinityLabel] = pod.Name
 			// Update
 			err = client.Update(context.TODO(), &pod)
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -30,6 +30,9 @@ const (
 	// Resources included in the stage backup.
 	// Referenced by the Backup.LabelSelector. The value is the Task.UID().
 	IncludedInStageBackupLabel = "migration-included-stage-backup"
+	// Application pods requiring restic/stage backups.
+	// Used to create stage pods. The value is the Task.UID()
+	ApplicationPodLabel = "migration-application-pod"
 	// For pods included in the stage backup, this label contains a pod-unique
 	// value (namespace "/"+name)
 	StagePodAffinityLabel = "migration-stage-pod-affinity"
@@ -217,7 +220,7 @@ func (t *Task) annotatePods(client k8sclient.Client) error {
 			if pod.Labels == nil {
 				pod.Labels = make(map[string]string)
 			}
-			pod.Labels[IncludedInStageBackupLabel] = t.StagePodSearchLabel()
+			pod.Labels[ApplicationPodLabel] = t.UID()
 			pod.Labels[StagePodAffinityLabel] = pod.Name
 			// Update
 			err = client.Update(context.TODO(), &pod)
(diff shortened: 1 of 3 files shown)


jortel:
I don't understand how this label value is different or safer.

pkg/controller/migmigration/task.go
```diff
@@ -467,6 +468,11 @@ func (t *Task) UID() string {
 	return string(t.Owner.UID)
 }
 
+// Migration Stage Pod Search Label
+func (t *Task) StagePodSearchLabel() string {
+	return fmt.Sprintf("%s-search", t.UID())
```
jortel: I don't understand how this label value is different or safer.

pkg/controller/migmigration/annotation.go
```diff
@@ -217,7 +217,7 @@ func (t *Task) annotatePods(client k8sclient.Client) error {
 			if pod.Labels == nil {
 				pod.Labels = make(map[string]string)
 			}
-			pod.Labels[IncludedInStageBackupLabel] = t.UID()
+			pod.Labels[IncludedInStageBackupLabel] = t.StagePodSearchLabel()
```
jortel: I get the method to create the label value is for consistency but thinking just `t.UID() + "-pod"` clearer.

pkg/controller/migmigration/task.go
jortel: Related to the other comment, I'm not sure a (1) line function that is used in (2) places is worthwhile.

pkg/controller/migmigration/pod.go
```diff
@@ -168,7 +168,7 @@ func (t *Task) ensureStagePodsCreated() (int, error) {
 	}
 
 	labelSelector := map[string]string{
-		IncludedInStageBackupLabel: t.UID(),
+		IncludedInStageBackupLabel: t.StagePodSearchLabel(),
```
jortel: I get the method to create the label value is for consistency but thinking just t.UID() + "-pod" is sufficient and perhaps clearer.

pkg/controller/migmigration/task.go
jortel: Suggest adding something like: `The -excluded suffix not matched by the label selector.` to the label comment.

https://github.com/fusor/mig-controller/blob/b1e56b34baac15a269b2b2b4511605a6f11db34b/pkg/controller/migmigration/annotation.go#L31

jortel:
Perhaps a better solution would be to add a new label like:
```
// Application pods. The value is the Task.UID().
ApplicationPodLabel = migration-application-pod
```
And add it to application pods instead of the `IncludedInStageBackupLabel`.  Search on it when creating stage pods.  Then, just add a line to `deletePodAnnotations()` to remove it as well.

This may be clearer.

pkg/controller/migmigration/pod.go
dymurray: Cool, if you like that better I'm fine with that. That's actually how I initially had it and only moved it to remove importing `fmt` in two places. I can update this.

dymurray:
@jortel Thanks Jeff great idea. I don't know why I was so insistent on using the same label.

PR #250, opened 2019-08-07
Changed default velero namespace.

Danil-Grigorev:


pkg/apis/migration/v1alpha1/resource.go
```diff
@@ -8,7 +8,7 @@ const (
 	TouchAnnotation  = "touch"
 	FinalizerEnabled = false
```
Danil-Grigorev: Re-enabling finalizers caused the same issues at this point, not allowing to remove the migPlan.  Will leave it as is for now.

jwmatthews:
I’d like to hold off on merging this until we have verified the newer work with Velero 1.1 beta is stable for gluster to ceph.  

Then I’d like to tag some images for end of sprint. 

Once we have that in place good to merge this in.

PR #251, opened 2019-08-07
Suspend PV discovery while migrations running or final migration succeeded.

jortel:
https://github.com/fusor/mig-controller/issues/253

Introduces the concept of a plan `suspended` condition while migrations are running or the _final_ migration has completed successfully.  This is primarily intended to keep the PV list from changing while migrations are running (constantly reduced as pods are migrated) and after the _final_ migration has succeeded.

The remote watch events are propagated through the plan which causes a lot of reconciles even when the MigPlan has not changed.  To reduce load on all clusters, reconcile latency and event propagation to the migration controller the follow is also not performed while `suspended`:
- Validation:
  - velero namespace exist.
  - namespaces listed in the plan exist.
- Ensure registry resources.
- Ensure storage resources.

The plan's _suspended_ status is managed and communicated to the user using a `Suspended` condition.

The plan controller adds a watch on migrations. This is new and does introduce a _circular_ watch since the migration CR has a reference to the plan.  However, this is safe because the plan watch is filtered by a predicate:
- _Create_ events are ignored.
- _Update_ events are forwarded to the referenced plan **only** when the _Running_ condition is added/removed.
- _Create_ events are ignored.  _Delete_ events are always forwarded.

The decision to have the plan controller determine the `suspended` condition instead of the migration controller (or the user) setting the plan in/out of the suspended condition is based on  _separation-of-concerns_.  It seemed more appropriate for the plan to controller to _self_ determine when the condition exists, as well as, how the condition impacts the reconcile.

### DIFF
--- a/pkg/apis/migration/v1alpha1/condition.go
+++ b/pkg/apis/migration/v1alpha1/condition.go
@@ -186,6 +186,23 @@ func (r *Conditions) SetCondition(condition Condition) {
 	}
 }
 
+// Stage an existing condition by type.
+func (r *Conditions) StageCondition(types ...string) {
+	if r.List == nil {
+		return
+	}
+	filter := make(map[string]bool)
+	for _, t := range types {
+		filter[t] = true
+	}
+	for i := range r.List {
+		condition := &r.List[i]
+		if _, found := filter[condition.Type]; found {
+			condition.staged = true
+		}
+	}
+}
+
 // Delete conditions by type.
 func (r *Conditions) DeleteCondition(types ...string) {
 	if r.List == nil {
--- a/pkg/apis/migration/v1alpha1/condition_test.go
+++ b/pkg/apis/migration/v1alpha1/condition_test.go
@@ -197,6 +197,32 @@ func TestConditions_SetCondition(t *testing.T) {
 		}))
 }
 
+func TestConditions_StageCondition(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
+
+	// Setup
+	conditions := Conditions{
+		List: []Condition{
+			{Type: "A"},
+			{Type: "B"},
+			{Type: "C"},
+			{Type: "D"},
+		},
+	}
+
+	// Test
+	conditions.StageCondition("A", "C", "X")
+
+	// Validation
+	g.Expect(conditions.staging).To(gomega.BeFalse())
+	g.Expect(conditions.List).To(gomega.Equal([]Condition{
+		{Type: "A", staged: true},
+		{Type: "B", staged: false},
+		{Type: "C", staged: true},
+		{Type: "D", staged: false},
+	}))
+}
+
 func TestConditions_DeleteCondition(t *testing.T) {
 	g := gomega.NewGomegaWithT(t)
 
--- a/pkg/controller/migplan/migplan_controller.go
+++ b/pkg/controller/migplan/migplan_controller.go
@@ -19,6 +19,7 @@ package migplan
 import (
 	"context"
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
+	migctl "github.com/fusor/mig-controller/pkg/controller/migmigration"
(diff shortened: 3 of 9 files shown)


pkg/controller/migplan/migplan_controller.go
```diff
@@ -344,3 +350,37 @@ func (r ReconcileMigPlan) ensureClosed(plan *migapi.MigPlan) error {
...
+// completed successfully. While suspended, reconcile is limited to basic validation
+// and PV discovery and ensuring resources is not performed.
+func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
+	limited := false
+	migrations, err := plan.ListMigrations(r)
+	if err != nil {
+		log.Trace(err)
+		return err
+	}
+	for _, m := range migrations {
```
djwhatle: Limits migrations when:
1. a migration is running
2. a final migration has succeeded

pkg/controller/migplan/migplan_controller.go
```diff
@@ -344,3 +350,37 @@ func (r ReconcileMigPlan) ensureClosed(plan *migapi.MigPlan) error {
 
 	return nil
 }
+
+// Get whether the plan is `suspended`.
+// A plan is considered`suspended` when a migration is running or the final migration has
+// completed successfully. While suspended, reconcile is limited to basic validation
+// and PV discovery and ensuring resources is not performed.
+func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
```
djwhatle: This function really does more than 'get' whether a plan is suspended, it also sets the suspended condition depending on migrations associated with the plan. The wording in the comment confused me initially.

pkg/apis/migration/v1alpha1/condition_test.go
```diff
@@ -197,6 +197,32 @@ func TestConditions_SetCondition(t *testing.T) {
 		}))
 }
 
+func TestConditions_StageCondition(t *testing.T) {
```
djwhatle: :pray: thank you for writing tests :+1:

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -186,6 +186,23 @@ func (r *Conditions) SetCondition(condition Condition) {
 	}
 }
 
+// Stage an existing condition by type.
+func (r *Conditions) StageCondition(types ...string) {
```
djwhatle: What does it mean to stage a condition by type? Can there be only one condition of a type attached to a migration this way?

jortel: The `SetCondition()` will add or update an existing condition.  The`StageCondition()` is not additive.  It's like _restoring_ it or just keeping the condition if it already exists.

pkg/controller/migplan/storage.go
```diff
@@ -13,6 +13,10 @@ func (r ReconcileMigPlan) ensureStorage(plan *migapi.MigPlan) error {
 	var client k8sclient.Client
 	nEnsured := 0
 
+	if plan.Status.HasCriticalCondition() || plan.Status.HasAnyCondition(Suspended) {
+		plan.Status.StageCondition(StorageEnsured)
```
djwhatle: Does setting the storageEnsured condition here prevent 'storageEnsure' from running?

pkg/controller/migplan/migplan_controller.go
jortel: It originally determined whether the plan was suspended and returned a boolean.  Then, updated to set the condition.  I s/Get/Determine/.  Does that make it clear enough?

djwhatle: I think 'determine' makes it clearer than 'get' for sure :+1:

pkg/controller/migplan/storage.go
jortel: No, it just preserves the `StageEnsured` condition if it already exists.  The `Suspended` condition prevents the _insure storage logic_ from running.

PR #252, opened 2019-08-07
[WIP] Update Restic Restore to work with Velero 1.1 Beta

dymurray:
With the move to the 1.1 Beta (https://github.com/fusor/velero/pull/27), ~~Restic now looks for a separate snapshot annotation for restic restores (https://heptio.github.io/velero/master/restic.html)~~  Velero uses PVBs as the source of truth when running restic restores. This is causing problems outlined below. The 1.1 Beta image is needed to get CSI working with Restic

### DIFF
--- a/pkg/controller/migmigration/pod.go
+++ b/pkg/controller/migmigration/pod.go
@@ -2,6 +2,7 @@ package migmigration
 
 import (
 	"context"
+	"fmt"
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


dymurray:
Moving this to WIP as we learned from Velero folks that this method of creating restic restores is deprecated and will be removed in 2.0. To get this working with the latest beta images we will set these annotations but we also need to be updated to work with their latest handling of PodVolumeBackups.

The 2 PRs that changed this behavior are here:
https://github.com/heptio/velero/pull/1577
https://github.com/heptio/velero/pull/1723

To be totally honest we were lucking out in the past with ResticRestores running properly. It will be nice to get this completely updated to work with the beta images in the proper manner.

sseago:
It may be that we don't need any of this and what we're seeing is the result of a PVB backup bug. This PR isn't actually helping the situation in my cluster, since I'm only getting one PVB out of five into the backup json file (although all five are created on my cluster).

dymurray:
Agree with Scott. I will say currently with my testing using the 1.1 beta images, without this PR I see:

1. PVB is replicated to destination side
1. PVR is created
1. PVR sits there and is never acted upon by restic daemonsets

With this PR, I'm still seeing an issue with multiple volumes:

1. 2 volumes on source side are included in backup
1. 2 PVBS are created on source side
1. 1 PVB is replicated on source side
1. 2 PVRs are created on destination side from this snapshot annotation

There must be a bug somewhere in the new restic restore logic.

dymurray:
Closing as this is no longer needed.

PR #261, opened 2019-08-14
Update aws cloudprovider to work with minio/noobaa

dymurray:
To test this you can deploy minio using `mig-agnosticd` and enter the route as the `AwsS3URL`

EDIT: Got Noobaa installed testing Noobaa out now

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -388,6 +388,10 @@ func (r *MigPlan) BuildRegistryDC(storage *MigStorage, name, dirName string) (*a
 
 // Update a Registry DeploymentConfig as desired for the specified cluster.
 func (r *MigPlan) UpdateRegistryDC(storage *MigStorage, deploymentconfig *appsv1.DeploymentConfig, name, dirName string) error {
+	region := storage.Spec.BackupStorageConfig.AwsRegion
+	if region == "" {
+		region = "us-east-1"
+	}
 	deploymentconfig.Spec = appsv1.DeploymentConfigSpec{
 		Replicas: 1,
 		Selector: map[string]string{
@@ -426,7 +430,11 @@ func (r *MigPlan) UpdateRegistryDC(storage *MigStorage, deploymentconfig *appsv1
 							},
 							kapi.EnvVar{
 								Name:  "REGISTRY_STORAGE_S3_REGION",
-								Value: storage.Spec.BackupStorageConfig.AwsRegion,
+								Value: region,
+							},
+							kapi.EnvVar{
+								Name:  "REGISTRY_STORAGE_S3_REGIONENDPOINT",
+								Value: storage.Spec.BackupStorageConfig.AwsS3URL,
 							},
 							kapi.EnvVar{
 								Name:  "REGISTRY_STORAGE_S3_ROOTDIRECTORY",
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/aws.go
```diff
@@ -142,31 +175,37 @@ func (p *AWSProvider) Test(secret *kapi.Secret) error {
...
+			region:         p.GetRegion(),
+			disableSSL:     p.GetDisableSSL(),
+			forcePathStyle: p.GetForcePathStyle(),
+			bucket:         p.Bucket,
+			secret:         secret,
 		}
 		err = test.Run()
 	case VolumeSnapshot:
-		test := Ec2Test{
+		// Disable volume snapshot test until
```
dymurray: Removing VSL verification since it's never being used in the migration process and is broken until we make UX changes.

pkg/cloudprovider/aws.go
```diff
@@ -20,6 +21,7 @@ import (
 const (
 	AwsAccessKeyId     = "aws-access-key-id"
 	AwsSecretAccessKey = "aws-secret-access-key"
+	AwsS3DefaultRegion = "us-east-1"
```
jortel: This _const_ block is for the credentials secret data.  Please add in a new _const_ block.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -426,7 +430,11 @@ func (r *MigPlan) UpdateRegistryDC(storage *MigStorage, deploymentconfig *appsv1
 							},
 							kapi.EnvVar{
 								Name:  "REGISTRY_STORAGE_S3_REGION",
-								Value: storage.Spec.BackupStorageConfig.AwsRegion,
+								Value: region,
+							},
+							kapi.EnvVar{
+								Name:  "REGISTRY_STORAGE_S3_REGIONENDPOINT",
+								Value: storage.Spec.BackupStorageConfig.AwsS3URL,
```
sseago: Have you confirmed that this still works when REGIONENDPOINT is empty? I'm wondering whether we should only set this var when we have a value for it.

dymurray: Yes, when `AwsS3URL` is an empty string the value is not set inside of the DC. I have tested using minio (specifying an endpoint) and using AWS S3 (no endpoint) and saw no trouble

dymurray: Example if the endpoint URL is not specified

```
      spec:                    
        containers:      
        - env:         
          - name: REGISTRY_STORAGE                    
            value: s3                   
          - name: REGISTRY_STORAGE_S3_ACCESSKEY
            valueFrom:                  
              secretKeyRef:                    
                key: access_key
                name: registry-fgdgs-fkf2p
          - name: REGISTRY_STORAGE_S3_BUCKET
            value: velero-5f3e46ba0dc70309b91132c3db936347
          - name: REGISTRY_STORAGE_S3_REGION
            value: us-east-2
          - name: REGISTRY_STORAGE_S3_REGIONENDPOINT
          - name: REGISTRY_STORAGE_S3_ROOTDIRECTORY
            value: /fgdgs-registry-78e01032-bec9-11e9-842d-0a60c6315958
          - name: REGISTRY_STORAGE_S3_SECRETKEY
```

dymurray:
Fixes #261

PR #267, opened 2019-08-15
Discontinue using registry cluster.

### LINKED ISSUE #173: Remove use of registry-cluster.
As a result of a discussion with Paul Morie, turns out Federation had an API review and it was suggested that their current usage of cluster-registry exposed them to a potential security escalation, hence they've removed usage of cluster-registry for time being and are using the cluster details in their own resource.   Recommendation is for us to do similar and REMOVE cluster-registry usage, move the cluster details into MigCluster and just use MigCluster with secretRefs for now.

jortel:
fixes #173 

Although, this is not a [beta-blocker], I have some concern that finalizers will fail (occasionally) if the _registry_ `Cluster` is deleted before the `MigCluster` when deleting the namespace.  This will have broad impact so just a _draft_ PR for now but would like to have it ready so we can pull the trigger quickly when the time comes. 

The main change for users is to set the `MigCluster.URL` instead of the `ServiceAddress` in the registry `Cluster` endpoint.

Requires: https://github.com/fusor/mig-operator/pull/49 and UI update before merge.


config/crds/migration_v1alpha1_migcluster.yaml
```diff
@@ -28,8 +28,8 @@ spec:
           type: object
         spec:
           properties:
-            clusterRef:
-              type: object
+            certificateAuthority:
```
djwhatle: This may be better suited for a ConfigMap or Secret. I also wonder if `caBundle` is a more idiomatic var name for this type of thing inside of k8s, feel like I've seen that places before but I could be off base.

djwhatle: I appreciate the simplicity of keeping this directly on the spec but don't know if it would make the MigCluster unwieldy or if there are size limitations. Simpler would be preferred if it won't be a burden (e.g. keeping it on the MigCluster).

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -30,18 +29,21 @@ import (
...
+// SA secret keys.
+const (
+	SaToken = "saToken"
+)
+
 // MigClusterSpec defines the desired state of MigCluster
 type MigClusterSpec struct {
 	IsHostCluster           bool                  `json:"isHostCluster"`
-	ClusterRef              *kapi.ObjectReference `json:"clusterRef,omitempty"`
+	URL                     string                `json:"url"`
```
djwhatle: :+1: I think this is the only thing we need from the Cluster CR.

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -123,148 +125,66 @@ func (r *MigCluster) DeleteFinalizer() bool {
...
 		}
 	*/
-	// If can't reuse a client from anywhere, build one from scratch without a cache
 	restConfig, err := m.BuildRestConfig(c)
 	if err != nil {
 		return nil, err
 	}
-
-	client, err := m.buildClient(restConfig)
+	client, err := k8sclient.New(restConfig, k8sclient.Options{Scheme: scheme.Scheme})
```
djwhatle: :+1: good simplification

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -123,148 +125,66 @@ func (r *MigCluster) DeleteFinalizer() bool {
...
 	}
 
-	return true, nil
+	return nil
 }
 
-// BuildRestConfig creates a remote cluster RestConfig from a MigCluster and a local client
+// Build a REST configuration.
 func (m *MigCluster) BuildRestConfig(c k8sclient.Client) (*rest.Config, error) {
-	// Get first K8s endpoint from ClusterRef
```
djwhatle: Glad to be ripping all of this out.

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -123,148 +125,66 @@ func (r *MigCluster) DeleteFinalizer() bool {
...
-
-	saTokenKey := "saToken"
-	saTokenData, ok := saSecret.Data[saTokenKey]
-	if !ok {
-		return nil, fmt.Errorf("MigCluster [%s/%s] references SA token secret [%s/%s] with empty 'saToken' field",
-			m.Namespace, m.Name, saSecretRef.Namespace, saSecretRef.Name)
+	restConfig := &rest.Config{
+		Host:        m.Spec.URL,
+		BearerToken: string(secret.Data[SaToken]),
+		TLSClientConfig: rest.TLSClientConfig{
```
djwhatle: I see you figured out how to get `Insecure: true` set on the initial object. I was struggling with that.

config/crds/migration_v1alpha1_migcluster.yaml
jortel: Yeah, looking through k8s code it seems `caBundle` is commonly used and is a `[]byte`.  Changing to match.

jortel:
Rebased and tested.

PR #277, opened 2019-08-19
Quiesce pods in stage itinerary based on QuiescePods only.

### LINKED ISSUE #268: Quiesce pods regardless of PV count.
The task phase _itinerary_ includes quiesce phases when `QuiescePods` is `true` and there is at least (1) PV being migrated.  In this case, ALL pods (including stateless pods) in the namespace are scaled down. It has been determined that the correct behavior is quiesce pods solely based on the `QuiescePods` setting.  The task _itinerary_ needs to be adjusted.

jortel:
fixes #268

Update the task _itinerary_ so that Pods are quiesced when `quiescePods: true` regardless of whether any PVs are being migrated.

### DIFF
--- a/pkg/controller/migmigration/task.go
+++ b/pkg/controller/migmigration/task.go
@@ -147,7 +147,11 @@ func (t *Task) Run() error {
 			if t.hasPVs() {
 				t.Phase = AnnotateResources
 			} else {
-				t.Phase = Completed
+				if t.quiesce() {
+					t.Phase = QuiesceApplications
+				} else {
+					t.Phase = Completed
+				}
 			}
 		} else {
 			t.Phase = EnsureInitialBackup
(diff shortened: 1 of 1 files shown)


pkg/controller/migmigration/task.go
```diff
@@ -147,7 +147,7 @@ func (t *Task) Run() error {
 			if t.hasPVs() {
 				t.Phase = AnnotateResources
 			} else {
-				t.Phase = Completed
+				t.Phase = QuiesceApplications
```
sseago: The `QuiesceApplications` phase should only be entered if `t.quiesce()` is true.
I think this section of code should probably read:
```
			if t.hasPVs() {
				t.Phase = AnnotateResources
			} else if t.quiesce() {
				t.Phase = QuiesceApplications
			} else {
				t.Phase = Completed
			}
```

jortel: Yes!  Good catch.  Corrected.

PR #278, opened 2019-08-20
Switch mig-controller to use ubi8-minimal as base image

djwhatle:
https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image


jmontleon:
When possible we should use ubi. mig-ui does, this will get controller on board, the updated Dockerfile for migration-plugin is switched, and when there are a few minutes I'll rename/update https://github.com/fusor/velero/blob/fusor-dev/Dockerfile.el7#L12

jmontleon:
Note: there is a ubi-thin as well that we may wish to investigate and make images even smaller.

PR #279, opened 2019-08-20
Update restic restart to handle multiple pods

dymurray:


pkg/controller/migmigration/pod.go
```diff
@@ -50,14 +50,14 @@ func (t *Task) restartResticPod() error {
 			log.Trace(err)
 			return err
 		}
-		return nil
 	}
+	time.Sleep(time.Second * 3)
```
dymurray: Would love some input on whether or not this is bad to put here. We debated this in the last iteration of this PR, but with the recent phase work done I think this is reasonable to put in. If we immediately return here, it's very likely that the delete requests have not been acted upon by Kubernetes the next time reconcile runs (will address that case in a second). Sleeping for 3 seconds guarantees that at least 1 of the restic pods has moved to a `Terminating` state.

IF the reconcile runs again before Kubernetes moves the pods to a `Terminating` state, I actually think it's not the worst thing in the world. I've tested running restic backups where the pod has not started when the backup runs, and restic is operating on `PodVolumeBackup` objects, so as soon as it starts up it finds the PVB and acts upon it. However, I think we should be doing everything we can to prevent this from happening as it could lead to unexpected behavior.

jortel: I doubt that a sleep will be deterministic.
What if we add an annotation to the pod before deleting.  Then, `haveResticPodsStarted()` could wait for at least (1) pod to be _Running_ without the annotation.

djwhatle: > Sleeping for 3 seconds guarantees that at least 1 of the restic pods has moved to a Terminating state.

Don't think this will always be true. Idea pitched by @jortel (annotate pods before sending delete event, then wait for pods with annotation to be gone) sounds promising and could be a simple + deterministic way to get this done.

Expanding on what onto Jeff has written above, I think we would really want to wait for _all_ of the restic pods (3 nodes = 3 restic pods) with our added annotation to be gone before proceeding to the next phase.

PR #298, opened 2019-08-30
Phase itinerary

jortel:
After the initial rewrite of the `Task.Run()` into a _stage-machine_ like _switch_ statement, I had some concerns about how hard it would be to visualize and maintain the flow through the phases.  This has proven to be difficult. As a result, I've had this work on the back-burner for a while.  

The main idea is to evolve the `Step` list into an `Itinerary` that provides both the order and conditional path(s) through the phases.  Rather than branching the phase paths using conditional logic in each phase (case), the _next_ phase is defined by the itinerary. 

Each phase (case) will:
- Implements the phase logic.
- Determines whether to advance to the next phase (or not).
- When a phase fails, jump directly to the appropriate _fail_ phase for clean up.
- Adjust the requeueAfter value.  

Each itinerary `Step` contains the phase _name_ and flags to define optional criteria used to qualify the phase such as: has-PVs or quiescePods (true). Additional flags can be added as needed.


pkg/controller/migmigration/task.go
```diff
@@ -39,52 +39,82 @@ const (
...
+	PVS     = 0x02 // Only when PVs migrated.
+)
+
+type Itinerary []Step
+
+var StageItinerary = Itinerary{
+	{phase: Created},
+	{phase: Started},
+	{phase: Prepare},
+	{phase: AnnotateResources},
```
sseago: Prior logic would have this with `flags: PVS` as well.

pkg/controller/migmigration/task.go
```diff
@@ -39,52 +39,82 @@ const (
...
+	{phase: EnsureStagePods, flags: PVS},
+	{phase: StagePodsCreated, flags: PVS},
+	{phase: RestartRestic, flags: PVS},
+	{phase: ResticRestarted, flags: PVS},
+	{phase: QuiesceApplications, flags: QUIESCE},
+	{phase: EnsureQuiesced, flags: QUIESCE},
+	{phase: EnsureStageBackup},
```
sseago: Also, to follow prior logic every step from EnsureStageBackup to Ensure AnnotationsDeleted would have `flags: PVS`

pkg/controller/migmigration/task.go
```diff
@@ -39,52 +39,82 @@ const (
...
-	FinalRestoreFailed:  true,
-	StageRestoreFailed:  true,
-	Completed:           true,
+var FinalItinerary = Itinerary{
+	{phase: Created},
+	{phase: Started},
+	{phase: Prepare},
+	{phase: EnsureInitialBackup},
+	{phase: InitialBackupCreated},
+	{phase: AnnotateResources},
```
sseago: `flags: PVS` here too?

pkg/controller/migmigration/task.go
```diff
@@ -39,52 +39,82 @@ const (
...
+	{phase: EnsureStagePods, flags: PVS},
+	{phase: StagePodsCreated, flags: PVS},
+	{phase: RestartRestic, flags: PVS},
+	{phase: ResticRestarted, flags: PVS},
+	{phase: QuiesceApplications, flags: QUIESCE},
+	{phase: EnsureQuiesced, flags: QUIESCE},
+	{phase: EnsureStageBackup},
```
sseago: `flags: PVS` for EnsureStageBackup through EnsureAnnotationsDeleted

pkg/controller/migmigration/task.go
```diff
@@ -202,31 +220,22 @@ func (t *Task) Run() error {
...
-		t.Phase = EnsureStagePods
+		t.next()
 	case EnsureStagePods:
-		count, err := t.ensureStagePodsCreated()
+		_, err := t.ensureStagePodsCreated()
 		if err != nil {
 			log.Trace(err)
 			return err
 		}
-		if count > 0 {
```
sseago: The other change is we're no iterating over StagePodsCreated, RestartRestic, and ResticRestarted even if there are no stage pods. As long as those operations are still successful when there are no stage pods, this should be fine. The prior logic skipped these steps if there were no stage pods.

jortel: As you know, the current logic generates the PVs list based on application pods that are bound to PVs.  And for each of those (app) pods, we create a stage pod.  So, if there are PVs, there will be stage pods.  Going forward, if we find a way to include PVs not assocated with pods, we can change the logic to count the stage pods.

For now, I think it's safe to use the PVS flag to qualify the phases as you indicated.

pkg/controller/migmigration/task.go
jortel: Ah, right.

pkg/controller/migmigration/task.go
sseago: So if every PV is being copied by snapshot rather than by restic we may not have any stage pods.

pkg/controller/migmigration/task.go
jortel: Fixed.

pkg/controller/migmigration/task.go
jortel: Fixed.

pkg/controller/migmigration/task.go
jortel: Ah.  Okay.

pkg/controller/migmigration/task.go
jortel: Fixed.

pkg/controller/migmigration/task.go
jortel: Fixed.

pkg/controller/migmigration/task.go
jortel: Fixed.

jortel:
This PR also resolves an issue raised whereby stage pods are being created even when application pods are not being quiesced.  Apparently this was noticed and reported because it aggravated an issue related to node affinity.

jortel:
> This PR also resolves an issue raised whereby stage pods are being created even when application pods are not being quiesced. Apparently this was noticed and reported because it aggravated an issue related to node affinity.

Reverted this change.

PR #305, opened 2019-09-04
Change namespace to migration

jmontleon:


jmontleon:
Should probably get an ACK from @jwmatthews

jmontleon:
The goal is to change the namespace name to something we can universally use a ProjectRequest api call to create. We want to use a ProjectRequest for OpenShift 3.7 and 3.9, but we can't on Openshift 3.11 and 4 for a namespace starting with `openshift-`

PR #306, opened 2019-09-06
Added migration identifier to all migration related logs

Danil-Grigorev:
```

### CODE REVIEW

File: pkg/controller/migmigration/task.go [hunk 1]
```diff
@@ -118,8 +119,6 @@ type Task struct {
 func (t *Task) Run() error {
 	t.Log.Info(
```
jortel: This can probably go on (1) line as well.

jortel:
@Danil-Grigorev I have a branch with some logger fixes: https://github.com/fusor/mig-controller/pull/308 and added `SetValues()` so it's not necessary to modify the `Logger` internals (Real).  Please consider reviewing and basing your PR on this instead.

jortel:
Looking at:
```
"Migration [RUN]","migration":"foo"
```
The word _Migration_ seems kind of redundant with the _migration_ keyword.  Perhaps the logging in the Task should just be:
```
"RUN","migration":"foo"
```
I still like the all CAPS because it makes those entries easier to pick out.

Thoughts?

Danil-Grigorev:
I'll include all the changes on top of the #308. Agree with all suggestions.

sseago:
We need to be careful assuming all migmigrations are in the same namespace. If it's just for logging purposes, it's probably OK, but nothing prevents CLI users from putting migrations in whatever namespaces they want to use. I've tested this and migrations work fine, even if created in a different namespace.

PR #313, opened 2019-09-09
Fix examples for cluster-registry removed.

jortel:
Looks like the examples did not get updated in https://github.com/fusor/mig-controller/pull/267.


config/releases/latest/manifest.yaml
```diff
@@ -52,26 +52,6 @@ rules:
...
-  - patch
-  - delete
-- apiGroups:
-  - clusterregistry.k8s.io
-  resources:
-  - clusters/status
-  verbs:
-  - get
-  - update
-  - patch
```
jortel: I'm not sure how this file got/gets generated.  Doing a `make manifests` does not seem to update it.  I also don't know what it's used for but seems like it's better for the clusterregistry to be removed.

djwhatle: @jortel this was created to give us an easy way to deploy the controller outside of `make run` in the early days. I think this is irrelevant now that we have mig-operator, should be safe to delete the `releases` dir.

jortel: Okay.  `config/releases` deleted.

PR #316, opened 2019-09-12
Add top-level "migration-controller-generated" annotation

sseago:
To support a future general-purpose velero plugin image, this
adds a "migration-controller-generated" annotation to backup and
restore CRs so that the plugin can conditionally exclude migration-
specific content within the plugin without having to guess based
on the presence of more specific annotations (stage, final, etc.).

### DIFF
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -13,11 +13,12 @@ import (
 
 // Velero Plugin Annotations
 const (
-	StageOrFinalAnnotation   = "openshift.io/migrate-copy-phase"   // (stage|final)
-	PvActionAnnotation       = "openshift.io/migrate-type"         // (move|copy)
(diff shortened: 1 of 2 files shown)


jortel:
Added 2 months ago as part of the finalizer work.  It's used by the `MigCluster.DeleteResources()`.  I'm not particularly concerned about another app named creating velero resources with the same label but perhaps changing the value to `openshift-migration` would be more appropriate and possibly reduce the chance of collisions.  If we do change it, we should cherry-pick to release-1.0 so that an upgraded installation will not leak resources.

sseago:
OK, so it sounds like using that label is probably a better plan than adding a new annotation. I think that was also done before we decided to call things "openshift-migration" -- I think changing the label value to that would make more sense than just "migration".

sseago:
Closing this PR since we'll use the part-of annotation instead.

PR #318, opened 2019-09-13
Ensure current velero cloud-secret is mounted in the pod.

jortel:
Fixes: https://github.com/fusor/mig-controller/issues/84

Workaround to deal with (~2 min) delay in the updated cloud-credentials secret resulting the secret mounted the _velero_ pod(s).  During the migration, a `cat /credentials/cloud` command is executed on the velero pod(s) and the output is compared to the content of the _cloud-credentials_ secret.  The phase effectively waits until they match.

This is considered very desirable for release-1.0.  Inclusion depends on getting good _mileage_ on it by QE and/or the development communities.  We need to be really sure it does not introduce any regressions.

Uses `NewSPDYExecutor` which requires `github.com/docker/spdystream`.
```
vendor/k8s.io/apimachinery/pkg/util/httpstream/spdy/connection.go:25:2: cannot find package "github.com/docker/spdystream"
```

This will need RBAC addition:
```
- apiGroups:
  - ""
  resources:
  - pods/exec
  verbs:
  - create
```

**REQUIRES**: https://github.com/fusor/mig-operator/pull/92


pkg/controller/migmigration/pod.go
```diff
@@ -312,3 +313,71 @@ func (t *Task) ensureStagePodsDeleted() error {
...
+// Ensure the velero cloud-credentials secret content has been
+// mounted into all velero pods on both source and destination clusters.
+func (t *Task) veleroPodCredSecretPropagated(cluster *migapi.MigCluster) (bool, error) {
+	list, err := t.findVeleroPods(cluster)
+	if err != nil {
+		log.Trace(err)
+		return false, err
+	}
+	if len(list) == 0 {
+		log.Info("Not velero pods found.")
```
sseago: Typo: "No velero pods found."

jortel: Fixed.  Thanks for the review @sseago.

PR #321, opened 2019-09-17
Added validation for pods via checks on attached replica sets

Danil-Grigorev:
Adding health checks for the app as a part of migplan.
New status field:
```
 kind: MigMigration
...
  status:
    conditions:
    - category: Warn
      lastTransitionTime: 2019-10-14T10:36:59Z
      message: The destination cluster has unhealthy migrated namespaces [nginx-example].
        See status.namespaces for details.
      status: "True"
      type: UnhealthyDestinationNamespaces
    - category: Advisory
      durable: true
      lastTransitionTime: 2019-10-14T10:36:59Z
      message: The migration has completed successfully.
      reason: Completed
      status: "True"
      type: Succeeded
    namespaces:
    - name: nginx-example
      workloads:
      - name: Deployment/nginx-deployment
        resources:
        - Pod/nginx-deployment-9b96654cd-l27kh
    phase: Completed
    startTimestamp: 2019-10-14T10:35:28Z
```


pkg/controller/migmigration/postValidate.go
```diff
@@ -0,0 +1,298 @@
...
+		podNames = append(podNames, pod.GetName())
+	}
+
+	correlatedEvents := collectRelatedEvents(podNames, eventList.Items)
+	eventsJSON, err := json.MarshalIndent(correlatedEvents, "", " ")
+	if err != nil {
+		return false, err
+	}
+
+	if failedPodCounter > 0 {
```
jortel: Looks like only the _category_ is different between the `if` and `else`.  Perhaps it would be good to reduce the duplication by:
```
category := Advisory
if failedPodCounter > 0 {
    category = Critical
}
```

Then:  `Category: category`.

pkg/controller/migmigration/postValidate.go
```diff
@@ -0,0 +1,298 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// EventDisplay is a structure for displaying occured events
+type EventDisplay struct {
```
jortel: Why not just use the `core.Event`.  Also, the `Display` in the name describes a specific usage of the data instead of being descriptive of the data itself.  Suggest dropping the _display_ terminology.

pkg/controller/migmigration/postValidate.go
```diff
@@ -0,0 +1,298 @@
...
+	Pod           string    `json:"pod"`
+	LastTimestamp time.Time `json:"lastTimestamp,omitempty"`
+	Message       string    `json:"message,omitempty"`
+	Reason        string    `json:"reason,omitempty"`
+}
+
+const (
+	// ValidationFailed is a failed validation state
+	ValidationFailed = iota
```
jortel: The term `Validation` has a pretty specific meaning in our controller/CRs.  I pertains to assessing the stage of the world prior to running a migration.  Suggest something like `Verification` for the _post_ checks would be clearer.

That said, I think you can replace these return values by having the task code chec the `Reason = (Running|Succeeded|Failed)` in the `PostVerification` (or some other name) condition.  Will describe condition changes elsewhere.

pkg/controller/migmigration/postValidate.go
```diff
@@ -0,0 +1,298 @@
...
+// collectRelatedEvents will extract all warning events
+// which occured in connection to the resource
+func collectRelatedEvents(
+	objectNames []string,
+	eventList []corev1.Event) []EventDisplay {
+
+	filteredEventMessages := []EventDisplay{}
+
+	for _, event := range eventList {
+		if !involvedObjectMatch(event, objectNames) || event.Type != corev1.EventTypeWarning {
```
jortel: Put the `event.Type != corev1.EventTypeWarning` first in the `||` for efficiency.

pkg/controller/migmigration/postValidate.go
Danil-Grigorev: Thanks for pointing it out. Going to replace the naming, sorry for confusion.

pkg/controller/migmigration/postValidate.go
```diff
@@ -0,0 +1,298 @@
...
+			Type:     Validating,
+			Status:   True,
+			Reason:   t.Phase,
+			Category: Advisory,
+			Message:  string(eventsJSON),
+		})
+	}
+
+	t.Owner.Status.SetCondition(migapi.Condition{
+		Type:     ValidationDetails,
```
jortel: A separate `details` condition is a bit odd.  Let's consider a single condition.  Our conditions have an `Items []string` field that is handy for this.

Thinking something like this:

```
t.Owner.Status.SetCondition(migapi.Condition{
    Type:     Verification,
    Status:   True,
    Reason:   Running,
    Category: Advisory,
    Message:  "Verification running."
    Items: []string{}
})
```
Then when a Pod fails verification.
```
condition := t.Owner.Status.FindCondition(Verification)
for _, event := range events {
    condition.Items = append(condition.Items, string(event-json))
}
```

At the end,
```
condition := t.Owner.Status.FindCondition(Verification)
if len(condition.Items) > 0 {
    condition.Category = Critical
    condition.Reason = Failed
    condition.Message = "Verification failed: []"  // The [] is replaced Items by when staging ends.
} else {
    condition.Reason = Succeeded
    condition.Message = "Verification succeeded." 
}
```

pkg/controller/migmigration/task.go
```diff
@@ -41,6 +41,9 @@ const (
 	EnsureFinalRestore            = "EnsureFinalRestore"
 	FinalRestoreCreated           = "FinalRestoreCreated"
 	FinalRestoreFailed            = "FinalRestoreFailed"
+	ValidateResourcesMigrated     = "ValidateResourcesMigrated"
+	ValidationInProgress          = "ValidationInProgress"
```
jortel: s/Validation/Verification/g

A separate `InProgress` phase deviates from other phases that can also be long running.  Also, I believe that the `ValidateResourcesMigrated` phase checks the static resources like DCs.  Suggest something like:

VerifyPodOwners
VerifyPods

.. or something :)

pkg/controller/migmigration/validation.go
```diff
@@ -15,6 +15,8 @@ const (
 	Running           = "Running"
 	Succeeded         = "Succeeded"
 	Failed            = "Failed"
+	ValidationDetails = "ValidationDetails"
+	Validating        = "Validating"
```
jortel: Suggest (1) phase named: `Verification`.

pkg/controller/migmigration/task.go
```diff
@@ -465,8 +470,32 @@ func (t *Task) Run() error {
 		}
 		t.Requeue = NoReQ
 		t.next()
-	case InitialBackupFailed, FinalRestoreFailed:
+	case InitialBackupFailed, FinalRestoreFailed, ResourceValidationFailed:
 		t.next()
+	case ValidateResourcesMigrated:
+		migratedSuccessfully, err := t.validateStaticResourcesPresence()
```
jortel: I'm not convinced this adds value.  It is essentially testing that velero backed up and restored all of the resources expected. Should this fail for one migration, our controller is broken and it will fail for all of them.  Seems this should be caught in e2e tests.

pkg/controller/migmigration/postValidate.go
jortel: Thinking this should not be a `Critical` category but rather see `Reason: Failed` below.

pkg/controller/migmigration/postValidate.go
jortel: As we discussed: Rather than listing events, let's provide the user with a list of resources that failed (Deployment|ReplicaSet, ..)
```
condition := t.Owner.Status.FindCondition(Verification)
for _, r := range resources {
    ref := fmt.Sprintf("%s/%s:%s", r.Kind, r.Namespace, r.Name)
    condition.Items = append(condition.Items, ref)
}
```

pkg/controller/migmigration/task.go
Danil-Grigorev: Removed after short discussion. This check will make migration fail every time, and it should be a part of pre-migration health checks. @alexander-demichev We probably need to check if our migration tooling tag all the resources in the namespace, so that velero could pick it up. For example `ReplicaSet` created by an older version of the `Deployment` is left in the namespace, but is not picked by our migration. @jwmatthews Is this expected behaviour?

pkg/controller/migmigration/validation.go
Danil-Grigorev: Updated.

pkg/controller/migmigration/task.go
jortel: You'll need to consider the owner references.  For example: for a Deployment <- ReplicaSet <- Pod  only the `Deployment` should be backed/Restored by velero since the Deployment will _spin up_ the ReplicaSet which will, in turn, _spin up_ the Pod.

pkg/controller/migmigration/verify.go
```diff
@@ -0,0 +1,329 @@
...
+			// Skip downscaled owner resources
+			if expectedReplicas == 0 {
+				continue
+			}
+
+			readyReplicas, found, err := unstructured.NestedFieldCopy(podManager.Object, "status", "readyReplicas")
+			if err != nil {
+				return VerificationFailed, err
+			}
+			// Skip resource with uprepared status
```
dymurray: NIT: `unprepared`

pkg/controller/migmigration/verify.go
```diff
@@ -0,0 +1,329 @@
...
+			if !found {
+				_, found, err := unstructured.NestedFieldCopy(podManager.Object, "status", "unavailableReplicas")
```
dymurray: This block feels weird... So if the `readyReplicas` field isn't found, we then check for `unavailableReplicas`. If `unavailableReplicas` aren't found, we return an error that `replicas field was not found in kind`.

Do we need two conditionals here? Won't `err` be set if `NestedFieldCopy` fails when `unavailableReplicas` doesn't exist? If that's the case we can remove the `if !found` block on L140. If the error message is nonsensical then I can understand having the two conditions here.

If `unavailableReplicas` are found I'm assuming we are continuing to get the failing pods health?

pkg/controller/migmigration/verify.go
```diff
@@ -0,0 +1,329 @@
...
+		if pod.Status.Phase == corev1.PodFailed {
+			failingPods = append(failingPods, fmt.Sprintf("Pod:%s/%s", pod.Namespace, pod.Name))
+			resourceNames = append(resourceNames, pod.Name)
+			continue
+		}
+
+		// Consider suspicious containers as an error state
+		podContainersStatuses := append(pod.Status.InitContainerStatuses, pod.Status.ContainerStatuses...)
+		for _, containerStatus := range podContainersStatuses {
+			if !containerStatus.Ready && containerStatus.RestartCount >= CrashLoopBackOffMargin {
```
dymurray: I'm okay with this. We are saying ONLY if the pod is not ready will we do a check to see if there are more than 3 restarts.

pkg/controller/migmigration/verify.go
Danil-Grigorev: After some investigation, I removed this part. Just looking into the `Pods` is enough. The error suggests only that there was some non string accessor in the structure, so it does not give anything. And another thing - the `unavailableReplicas` is also for pods in `Pending` state :(

dymurray:
In my opinion it sounds like the preferred method here would be #2. I think closing a plan should prevent subsequent migrations from occurring, but validating the state of that migration plan should occur after the plan is closed.

My only concern with that is that we would be running validation indefinitely on the migration plan until it is deleted. I'm not quite sure that is desired behavior considering that the controller will be allocating resources to a plan that could be days old and the state of the application wouldn't have changed. Perhaps once the validation succeeds completely then we set another field such as `appVerified` which prevents the verification loop from running.

The more I write I realize I'm basically talking about option 1, but I think it's important that the terminology remains the same. I don't think app verification should affect the existing workflow with how `closed` is being used today from the UI. So that is my hard requirement with a solution here. Curious what @jortel thinks.

pkg/apis/migration/v1alpha1/verify.go
```diff
@@ -0,0 +1,208 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// UnhealthyResources is a store for unhealthy namespaces across clusters
+type UnhealthyResources struct {
+	SourceUnhealthy      []UnhealthyNamespace `json:"sourceUnhealthy,omitempty"`
```
jortel: I think that simply `Souce` and `Destination` field names would be clearer.

pkg/apis/migration/v1alpha1/verify.go
```diff
@@ -0,0 +1,208 @@
...
+		*cluster = append(*cluster, unhealthyNamespace)
+	} else {
+		resources.Workloads = append(resources.Workloads, workload)
+	}
+}
+
+// AddUnhealthyResources is adding unhealthy workloads and resolves owner references
+func (u *UnhealthyResources) AddUnhealthyResources(
+	client k8sclient.Client,
+	targetCluster *[]UnhealthyNamespace,
```
dymurray: Consider a better name for this. Even `unhealthyNamespaces` would suffice. `targetCluster` makes me think the object is a client or something.

jortel:
I don't think there should be any further processing on a closed plan.  I also don't think we want the verification to run on every plan reconcile.  Instead, it should only run after a successful _final_ migration.  The question is - how should the verification get triggered?  Let's discuss.

Danil-Grigorev:
> Per the discussion on the email thread, I believe this PR is still missing what we agreed upon for resolving this with plan closure. The `verify` boolean on the plan spec is supposed to be used by the migration controller to determine whether or not we will run application verification on the destination cluster at the end of migration.

Changed.
> Currently this PR is validating the application state of the source and destination everytime the plan controller reconciles. You can leave the source validation code on the plan controller, but the destination cluster verification needs to be done on the migration controller as a phase at the end of migration.

I removed the part of the code responsible for source health checks. The health should be shown on the migPlan status. Should be added later by @alexander-demichev, me or anyone else.

pkg/apis/migration/v1alpha1/resource.go
```diff
@@ -6,6 +6,7 @@ import (
 
 const (
 	TouchAnnotation   = "touch"
+	VerifyAnnotation  = "verify"
```
jortel: The annotation is no longer needed.

pkg/apis/migration/v1alpha1/resource.go
```diff
@@ -70,6 +71,26 @@ func (r *MigPlan) Touch() {
 	r.Annotations[TouchAnnotation] = uuid.New().String()
 }
 
+func (r *MigPlan) GetVerify() (verify bool) {
```
jortel: The annotation is no longer needed.  Just a `bool`.

pkg/apis/migration/v1alpha1/verify.go
```diff
@@ -0,0 +1,248 @@
...
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Types
```
jortel: To keep condition `const` in (1) place, we have been defining them in `validation.go` even though this is verification.  Please put them there.

pkg/apis/migration/v1alpha1/verify.go
```diff
@@ -0,0 +1,248 @@
...
+const (
+	UnhealthySourceNamespaces      = "UnhealthySourceNamespaces"
```
jortel: The migration controller verification only deals with the destination.  No need for Source/Destination differentiation or for both.  Just `UnhealthyNamespaces`.

pkg/apis/migration/v1alpha1/verify.go
```diff
@@ -0,0 +1,248 @@
...
+)
+
+// Reasons
+const (
+	ErrorsDetected = "ErrorsDetected"
+)
+
+// Messages
+const (
+	UnhealthyDestination = "The destination cluster has unhealthy migrated namespaces []. See status.namespaces for details."
```
jortel: Since the `UnhealthyNamespaces` stucture contains the full list/description, I think the message should contain the list `[]` as well.

pkg/apis/migration/v1alpha1/verify.go
```diff
@@ -0,0 +1,248 @@
...
+	Workloads []Workload `json:"workloads"`
+}
+
+// IsHealthy will report if there are any unhealthy namespaces were present
+func (m *MigMigration) IsHealthy() bool {
+	return m.Status.UnhealthyResources.Namespaces == nil || len(m.Status.UnhealthyResources.Namespaces) == 0
+}
+
+// VerifyPods would collect, display and notify about the pod health in a specific namespace
+func (u *UnhealthyResources) VerifyPods(client k8sclient.Client, options *k8sclient.ListOptions) (*[]unstructured.Unstructured, error) {
```
jortel: `VerifyPods()` seems like a method on the Reconciler not the `UnhealthyResources` object.

Danil-Grigorev: So, this mean that until the `spec.verify` is set on the migPlan, the migrations will verify target namespaces on every reconcile?

pkg/apis/migration/v1alpha1/verify.go
```diff
@@ -0,0 +1,248 @@
...
+		default:
+			continue
+		}
+
+		return true
+	}
+
+	return false
+}
+
```
jortel: I would expect these _find_ methods to be public and used by the controller, reconciler.

Danil-Grigorev: Sure, could be fixed.

pkg/apis/migration/v1alpha1/verify.go
```diff
@@ -0,0 +1,248 @@
...
+			if subWorkload == name {
+				return &workload
+			}
+		}
+	}
+	return nil
+}
+
+// AddUnhealthyNamespace is adding a workload to selected list of unhealthy namespaces
+func (u *UnhealthyResources) AddUnhealthyNamespace(cluster *[]UnhealthyNamespace, namespace, name string, resourceNames ...string) {
```
jortel: This method is scoped to the `UnhealthyResources` object.  Anything more than `AddNamespace()` is redundant.

pkg/apis/migration/v1alpha1/verify.go
```diff
@@ -0,0 +1,248 @@
...
+			Workloads: []Workload{workload},
+		}
+		*cluster = append(*cluster, unhealthyNamespace)
+	} else {
+		resources.Workloads = append(resources.Workloads, workload)
+	}
+}
+
+// AddUnhealthyResources is adding unhealthy workloads and resolves owner references
+func (u *UnhealthyResources) AddUnhealthyResources(
```
jortel: This method is scoped to the `UnhealthyResources` object.  Anything more than `AddResources()` is redundant.

pkg/apis/migration/v1alpha1/verify.go
Danil-Grigorev: From what I see in the message, there is a list `[]` already. Could you please clarify?

pkg/apis/migration/v1alpha1/verify.go
Danil-Grigorev: I suppose this part could be used at any moment, and not only by migration/plan reconsile.

pkg/apis/migration/v1alpha1/verify.go
jortel: Meant should **not** include the `[] ` list.

pkg/apis/migration/v1alpha1/verify.go
jortel: No, the migration controller will only do the verification when verify=true and the migration is a _final_ migration that has completed successfully.

pkg/apis/migration/v1alpha1/resource.go
Danil-Grigorev: Sorry, this part was reintroduced after rebase, will clean it up.

pkg/controller/migmigration/task.go
```diff
@@ -41,6 +41,7 @@ const (
 	EnsureFinalRestore            = "EnsureFinalRestore"
 	FinalRestoreCreated           = "FinalRestoreCreated"
 	FinalRestoreFailed            = "FinalRestoreFailed"
+	Verify                        = "Verify"
```
jortel: The name `Verification` seems more like phase name.  Eg.  _"The migration is in the verification phase."_

Danil-Grigorev: Updated as well.

Danil-Grigorev:
Some structure improvements are needed, as the feature complexity is higher then expected.

pkg/controller/migmigration/validation.go
```diff
@@ -46,6 +48,8 @@ const (
 	RunningMessage           = "Step: %d/%d"
 	FailedMessage            = "The migration has failed.  See: Errors."
 	SucceededMessage         = "The migration has completed successfully."
+	UnhealthyDestination     = "The destination cluster has unhealthy migrated namespaces. See status.namespaces for details."
+	UnhealthySource          = "The source cluster has unhealthy namespaces. See status.namespaces for details."
```
jortel: The `UnhealthySource` would be used by the plan controller and should be defined there.  Also, by convention, the _message_ const should end with `Message`.  Recommend: `UnhealthyNamespacesMessage` to match `UnhealthyNamespaces` condition type.

PR #324, opened 2019-09-18
Add pod health check during validation

alexander-demichev:
https://github.com/fusor/mig-controller/issues/286

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -42,6 +42,7 @@ const (
 
 // MigPlanSpec defines the desired state of MigPlan
 type MigPlanSpec struct {
+	UnhealthyResources
 	PersistentVolumes `json:",inline"`
 	Namespaces        []string              `json:"namespaces,omitempty"`
 	SrcMigClusterRef  *kapi.ObjectReference `json:"srcMigClusterRef,omitempty"`
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -7,10 +7,12 @@ import (
 	"strings"
 
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
+	"github.com/fusor/mig-controller/pkg/health"
 	migref "github.com/fusor/mig-controller/pkg/reference"
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/fields"
+	runtime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -771,17 +773,20 @@ func (r ReconcileMigPlan) validatePods(plan *migapi.MigPlan) error {
 	}
 
 	for _, ns := range plan.Spec.Namespaces {
-		options := k8sclient.InNamespace(ns)
-		podList := kapi.PodList{}
-		err := client.List(context.TODO(), options, &podList)
+		unhealthyPods, err := health.PodsUnhealthy(client, &k8sclient.ListOptions{
+			Namespace: ns,
+		})
 		if err != nil {
 			log.Trace(err)
 			return err
 		}
 
-		for _, pod := range podList.Items {
-			if pod.Status.Phase == kapi.PodRunning || pod.Status.Phase == kapi.PodSucceeded {
-				continue
+		for _, unstrucredPod := range *unhealthyPods {
+			pod := &kapi.Pod{}
+			err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstrucredPod.UnstructuredContent(), pod)
+			if err != nil {
+				log.Trace(err)
+				return err
 			}
 
 			condition := plan.Status.FindCondition(SourcePodsNotHealthy)
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -109,6 +109,7 @@ const (
 	RegistriesEnsuredMessage              = "The migration registry resources have been created."
 	PvsDiscoveredMessage                  = "The `persistentVolumes` list has been updated with discovered PVs."
 	ClosedMessage                         = "The migration plan is closed."
+	SourcePodsNotHealthyMessage           = "Source namespace(s) contain unhealthy pods. See: `unhealthyNamespaces` for details."
 )
 
 // Valid AccessMode values
@@ -772,6 +773,7 @@ func (r ReconcileMigPlan) validatePods(plan *migapi.MigPlan) error {
 		return err
 	}
 
+	unhealthyResources := migapi.UnhealthyResources{}
 	for _, ns := range plan.Spec.Namespaces {
 		unhealthyPods, err := health.PodsUnhealthy(client, &k8sclient.ListOptions{
 			Namespace: ns,
@@ -781,6 +783,9 @@ func (r ReconcileMigPlan) validatePods(plan *migapi.MigPlan) error {
 			return err
 		}
 
+		workload := migapi.Workload{
+			Name: "Pods",
+		}
 		for _, unstrucredPod := range *unhealthyPods {
 			pod := &kapi.Pod{}
 			err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstrucredPod.UnstructuredContent(), pod)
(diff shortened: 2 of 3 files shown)


pkg/controller/migplan/validation.go
```diff
@@ -654,3 +665,79 @@ func containsAccessMode(modeList []kapi.PersistentVolumeAccessMode, accessMode k
...
+
+	if len(failedPodSlice) > 0 {
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     FailedPodStatus,
+			Status:   True,
+			Reason:   NotDone,
+			Category: Warn,
+			Message:  string(podsJSON),
+		})
+	}
```
sseago: Looking at the associated issue, I see that @jwmatthews  had suggested that we include information on all pods, not just error ones:
"Thinking for first phase we add this for pods in every namespace, we look for their condition and update CR. Was thinking we update with all pod found so that the UI has information so it can display on the validation screen for a MigPlan that ~20 pods are in a good state and one is unhealthy or similar."

This is only warning on unhealthy pods. I'm actually not sure we want to list all pods as suggested, though. Currently the infrastructure is more set up for warning about just the unexpected things. If there are no pods in an error/failed state, then perhaps it's best to just not warn at all, like is done here.

pkg/controller/migplan/validation.go
```diff
@@ -38,6 +41,7 @@ const (
 	RegistriesEnsured              = "RegistriesEnsured"
 	PvsDiscovered                  = "PvsDiscovered"
 	Closed                         = "Closed"
+	FailedPodStatus                = "FailedPodStatus"
```
jortel: Perhaps `SourcePodsNotHealthy` would be more descriptive condition.

alexander-demichev: +1

pkg/controller/migplan/validation.go
```diff
@@ -654,3 +665,79 @@ func containsAccessMode(modeList []kapi.PersistentVolumeAccessMode, accessMode k
 	}
 	return false
 }
+
+type NamespacePods struct {
+	Namespace string         `json:"namespace"`
+	Pods      []PodWithPhase `json:"pods"`
```
jortel: I'm not convinced the `phase` is useful given the short list of possible failure phases.  Given this, suggest `types.NamespacedName` is sufficient instead of a custom type.

pkg/controller/migplan/validation.go
```diff
@@ -654,3 +665,79 @@ func containsAccessMode(modeList []kapi.PersistentVolumeAccessMode, accessMode k
...
+
+// Validate the pods, all should be healthy befor migration
+func (r ReconcileMigPlan) validatePods(plan *migapi.MigPlan) error {
+	cluster, err := plan.GetSourceCluster(r)
+	if err != nil {
+		log.Trace(err)
+		return err
+	}
+
+	if cluster == nil || !cluster.Status.IsReady() {
```
jortel: :+1:

pkg/controller/migplan/validation.go
```diff
@@ -654,3 +665,79 @@ func containsAccessMode(modeList []kapi.PersistentVolumeAccessMode, accessMode k
...
+	Pods      []PodWithPhase `json:"pods"`
+}
+
+type PodWithPhase struct {
+	Name   string        `json:"name"`
+	Status kapi.PodPhase `json:"status"`
+}
+
+// Validate the pods, all should be healthy befor migration
+func (r ReconcileMigPlan) validatePods(plan *migapi.MigPlan) error {
```
jortel: This migration should be skipped while the plan is _suspended_ (migrations are running or after the _final_ migration has succeeded).
Add:
```
if plan.Status.HasAnyCondition(Suspended) {
    plan.Status.StageCondition(FailedPodStatus)
    return nil
}
```

pkg/controller/migplan/validation.go
```diff
@@ -654,3 +665,79 @@ func containsAccessMode(modeList []kapi.PersistentVolumeAccessMode, accessMode k
...
+	for _, ns := range plan.Spec.Namespaces {
+		options := k8sclient.InNamespace(ns)
+		podList := kapi.PodList{}
+		err := client.List(context.TODO(), options, &podList)
+		if err != nil {
+			log.Trace(err)
+			return err
+		}
+
+		if len(podList.Items) == 0 {
```
jortel: The `continue` here does not add value since iterating an empty list below will have the same affect.

pkg/controller/migplan/validation.go
```diff
@@ -654,3 +665,79 @@ func containsAccessMode(modeList []kapi.PersistentVolumeAccessMode, accessMode k
...
+			continue
+		}
+
+		podPhaseSlice := make([]PodWithPhase, 0)
+
+		for _, pod := range podList.Items {
+			if pod.Status.Phase == kapi.PodRunning || pod.Status.Phase == kapi.PodSucceeded {
+				continue
+			}
+			podPhaseSlice = append(podPhaseSlice, PodWithPhase{Name: pod.Name, Status: pod.Status.Phase})
```
jortel: The `Condition` supports and `Items` field that is designed to make this easy.
Suggest:
```
for _, pod := range podList.Items {
    if pod.Status.Phase == kapi.PodRunning || pod.Status.Phase == kapi.PodSucceeded {
        continue
    }
    condition := plan.Status.FindCondition(SourcePodsNotHealthy)
    if condition == nil {
        condition = migapi.Condition{
            Type:     SourcePodsNotHealthy,
            Status:   True,
            Reason:   NotHealthy,
            Category: Warn,
            Message:  "Pods not healthy on the source cluster [].",
            Items: []string{}
        }
        plan.Status.SetCondition(condition)
    }
    ref := fmt.Sprintf("%s:%s", pod.Namespace, pod.Name)
    condition.Items = append(condition.Items, ref)
}
```

pkg/controller/migplan/validation.go
alexander-demichev: @sseago @jwmatthews If total amount of pods in selected namespaces reaches 1000+, then displaying info about all pods is not a good idea.

pkg/controller/migplan/validation.go
alexander-demichev: @jortel Your suggestion was my initial solution, but I removed it later. I want the message to be readable for UI. An array of JSON objects - `{"namespace": ..., "pods": []}`  is much more easy to parse in JS than grouping namespaces and pods from an array of strings [namespace1:pod1, namespace1:pod2, namespace2:pod2].

jortel: While appreciate your intention - according to kubernetes API conventions, The `Message` should be a _"human-readable message indicating details about last transition"_ not machine parsable data.  I don't think a big hunk of json is particularly readable.  Also, the API is intended for general use and not just to support the UI.  This approach deviates from all other `MIG` conditions.

Has the UI expressed a requirement to parse the list of pods?

[1] https://github.com/eBay/Kubernetes/blob/master/docs/devel/api-conventions.md#typical-status-properties

alexander-demichev: @jwmatthews @eriknelson Hi, can you take a look at this ^^ ? What's more desired approach from UI point of view?

dymurray: I commented in the google doc about this as well, I'm inclined to agree with Jeff here. The message field inside of a condition should be human readable, I liked the approach you mentioned in the google doc around having a separate field on the status (`verify` or something) where this JSON can be stored and read from. Curious what others think about this

alexander-demichev: @jortel @dymurray I've just replaced the JSON with suggested error message:)

pkg/controller/migplan/validation.go
```diff
@@ -654,3 +666,60 @@ func containsAccessMode(modeList []kapi.PersistentVolumeAccessMode, accessMode k
 	}
 	return false
 }
+
+// Validate the pods, all should be healthy befor migration
```
dymurray: NIT: `before`

pkg/controller/migplan/validation.go
```diff
@@ -654,3 +666,60 @@ func containsAccessMode(modeList []kapi.PersistentVolumeAccessMode, accessMode k
...
+			}
+
+			condition := plan.Status.FindCondition(SourcePodsNotHealthy)
+			if condition == nil {
+				condition = &migapi.Condition{
+					Type:     SourcePodsNotHealthy,
+					Status:   True,
+					Reason:   NotHealthy,
+					Category: Warn,
+					Message:  "Pods not healthy on the source cluster [].",
```
dymurray: Is the message supposed to have `[]` at the end? Or was the intention to put the cluster name?

Danil-Grigorev: This part of the condition will be filled with content of the `Items` before being displayed.

dymurray: Ah thank you I didn't realize that's how it worked.

jortel: Hey Alex, I'd asked Danil to fill you in on the _high level design_ intended to be shared between the (pre) migration validation and the (post) migration verification.  The 1st aspect was for both the migration and plan controllers to leverage the _health_ package which I see you have incorporated.  The 2nd, was to add the `UnhealthyResources` in the MigPlan spec to list the unhealthy namespace instead of listing them in the condition.  The thinking here is that unlike other conditions the list could be very long and it would have symmetry with how similar information is included in the migration.  Please use the `UnhealthyResources` etc in `verify.go`.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -48,6 +48,7 @@ type MigPlanSpec struct {
 	DestMigClusterRef *kapi.ObjectReference `json:"destMigClusterRef,omitempty"`
 	MigStorageRef     *kapi.ObjectReference `json:"migStorageRef,omitempty"`
 	Closed            bool                  `json:"closed,omitempty"`
+	UnhealthyResources
```
jortel: Nit: For consistency, let's group _inline_ structs at the beginning of the struct definition.

pkg/controller/migplan/validation.go
```diff
@@ -737,3 +748,63 @@ func containsAccessMode(modeList []kapi.PersistentVolumeAccessMode, accessMode k
...
+		}
+
+		if len(workload.Resources) != 0 {
+			unhealthyNamespace := migapi.UnhealthyNamespace{
+				Name:      ns,
+				Workloads: []migapi.Workload{workload},
+			}
+			unhealthyResources.Namespaces = append(unhealthyResources.Namespaces, unhealthyNamespace)
+		}
+	}
```
jortel: Please set the `SourcePodsNotHealthy` condition when UnhealthyNamespaces is not empty.

pkg/apis/migration/v1alpha1/migplan_types.go
alexander-demichev: done

pkg/controller/migplan/validation.go
alexander-demichev: done

pkg/controller/migplan/validation.go
```diff
@@ -105,6 +109,7 @@ const (
 	RegistriesEnsuredMessage              = "The migration registry resources have been created."
 	PvsDiscoveredMessage                  = "The `persistentVolumes` list has been updated with discovered PVs."
 	ClosedMessage                         = "The migration plan is closed."
+	SourcePodsNotHealthyMessage           = "Not healthy pods in some namespaces."
```
jortel: Suggest: _Source namespaces contain unhealthy pods.  See: \`unhealthyNamespaces\` for details._

alexander-demichev: ready, thanks

PR #325, opened 2019-09-18
Change application label to "openshift-migration"

sseago:


sseago:
Note that when this gets merged, anyone updating to this with currently-open migplans will need to close these plans first (and let the controller clean up) before upgrading to the a build with this PR. Since VSLs must be created with predictable names, currently they're generated using the migplan UID, so with the new app label, the controller will attempt to create using the same name that already exists, resulting in a reconcile error.

PR #326, opened 2019-09-23
Pull s3-specific registry creation bits into provider

sseago:
Creates two new cloud provider methods:
- UpdateRegistrySecret
- UpdateRegistryDC

These methods pull the aws/s3-specific bits of the migration
registry creation into the aws provider implementation,
in preparation for supporting GCP and AWS cloud storage providers.


sseago:
Update: Closing and re-opening plans with the refactored code creates the registry Secret and DC with the appropriate fields set. Migrations run successfully with the created registry.

PR #327, opened 2019-09-24
Replace PodAffinity with explicit NodeName selection for stage pods

sseago:
Replaces PodAffinity with explicit NodeName selection for stage pods,
although we only set the NodeName for pods with at least one RWO
volume. If all volumes are mounted RWX, then we don't constrain
node selection.

At the moment we don't handle the case where the selected NodeName
doesn't exist in the cluster. If it doesn't exist, then the pod will
disappear, which might confuse the controller into thinking it hasn't been
created yet. However, since the pod name is taken from the (currently
running) application pods, it's not clear whether this is actually possible,
except as a race condition where a migration is running while a node name
change is being processed, or if it's possible for a running pod to
*not* be updated of a previous NodeName change.

### DIFF
--- a/pkg/controller/migmigration/pod.go
+++ b/pkg/controller/migmigration/pod.go
@@ -133,26 +133,10 @@ func (t *Task) buildStagePod(pod *corev1.Pod) *corev1.Pod {
 			SecurityContext:              pod.Spec.SecurityContext,
 			ServiceAccountName:           pod.Spec.ServiceAccountName,
 			AutomountServiceAccountToken: pod.Spec.AutomountServiceAccountToken,
-			Affinity: &corev1.Affinity{
-				PodAffinity: &corev1.PodAffinity{
-					PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
-						{
-							Weight: 100,
-							PodAffinityTerm: corev1.PodAffinityTerm{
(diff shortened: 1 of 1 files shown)


pkg/controller/migmigration/annotation.go
```diff
@@ -11,6 +11,11 @@ import (
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// Controller Annotations
+const (
+	PodRWOPvAnnotation = "openshift.io/pod-rwo-pvs"
```
jortel: Please add more documentation.  Something like:
```
    // The application pod has at lease (1) RWO PV.
    PodRWOPvAnnotation = "openshift.io/pod-rwo-pvs"  // "true"
```

sseago:
Looks like I'm going to have to modify the part where we don't specify NodeName for RWX volumes. Apparently for GCE RWX volumes, the volume can be mounted writable (RWX) on at most one node at a time -- any other nodes sharing access must be read-only (RWO). This won't work for our stage pod scheme as currently implemented.
https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/readonlymany-disks

sseago:
It looks like we will either need to take storage plugin/storageclass into account or we'll have to tighten the restriction and *always* specify NodeName.

sseago:
This has been updated to simplify it to always set the NodeName for stage pods.

PR #328, opened 2019-09-25
Initial GCP cloud-provider support.

jortel:
Implement the GCP _cloudprovider_ for:
- Validation
- Bucket read/write/delete Test.
- Update the cloud-credentials secret.
- BSL configuration.
- Registry DC and secret.

New dependency:
- google.golang.org/api

Add fields to `MigStorage` CRD:
- GcpBucket

Example `MigStorage` _snip_:
```
  backupStorageConfig:
    credsSecretRef:
      name: migration-credentials
    gcpBucket: jortel-migration
  backupStorageProvider: gcp

```

Add fields to the _migration-credentials_ secret:
- gcp-credentials

Value from (assumes you read velero doc [1]):
```
$ gcloud iam service-accounts keys create credentials-velero \
    --iam-account $SERVICE_ACCOUNT_EMAIL
```
```
$ base64 -w0 credentials-velero
```

[1] https://github.com/heptio/velero/blob/master/site/docs/master/gcp-config.md

---

In the development environment, you'll need to:
- `$ make manifests`
- `$ oc apply -f config/crds/migration_v1alpha1_migstorage.yaml`

Also, the velero deployment needs to be edited on both all clusters.
```
$ oc edit deployment velero
```
Add `GOOGLE_APPLICATION_CREDENTIALS`.
```
        env:
        - name: AWS_SHARED_CREDENTIALS_FILE
          value: /credentials/cloud
        - name: GOOGLE_APPLICATION_CREDENTIALS  <---- add this
          value: /credentials/cloud
```


pkg/cloudprovider/gcp.go
```diff
@@ -1,47 +1,205 @@
...
+
+	return client, nil
+}
+
+func (r *GcsTest) upload(client *storage.Client) error {
+	bucket := client.Bucket(r.bucket)
+	object := bucket.Object(r.key)
+	writer := object.NewWriter(context.Background())
+	_, err := writer.Write([]byte{0})
+	writer.Close()
```
sseago: Need error checking on `Close` here. Connection checking is currently failing with 403 errors in my cluster, but the error is only being reported on the attempted read, even though the attempted write should be failing too. In looking at the docs for the vendor code here, the following comes from the `Write` function:
```
// Write appends to w. It implements the io.Writer interface.
//
// Since writes happen asynchronously, Write may return a nil
// error even though the write failed (or will fail). Always
// use the error returned from Writer.Close to determine if
// the upload was successful.
```
I think the right thing to do here is to store the return error for Close in a different local var and if either err or <whatever you call the Close return val> is non-nil, then return the non-nil error -- this way we still at least attempt to close the writer even if Write returned an error.

pkg/cloudprovider/gcp.go
```diff
@@ -1,47 +1,209 @@
...
+
+func (r *GcsTest) download(client *storage.Client) error {
+	bucket := client.Bucket(r.bucket)
+	object := bucket.Object(r.key)
+	reader, err := object.NewReader(context.Background())
+	if err != nil {
+		return err
+	}
+	_, err = reader.Read(make([]byte, 1))
+	reader.Close()
```
sseago: I don't know how likely it is for reader.Close() to return an error, but we should probably check the return value here like we're doing with writer.Close() since it's possible for reader.Close() to fail.

pkg/cloudprovider/gcp.go
jortel: Fixed.

pkg/cloudprovider/gcp.go
jortel: Done.

jortel:
Made the requested changes; retested and squashed into 2 commits:
1. Feature.
2. Dependency

djwhatle:
Docs on this moved to https://github.com/vmware-tanzu/velero/blob/master/site/docs/v1.1.0/gcp-config.md

PR #337, opened 2019-10-04
Add support for namespace change

alexander-demichev:
This PR adds ability to pass namespace mapping for velero restore using `Namespaces` field of migplan, example: 
```
 namespaces:
    - foo1:bar1
    - bar2
```

### DIFF
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -160,7 +160,7 @@ func (t *Task) annotatePVCs(client k8sclient.Client, pod corev1.Pod) ([]string, 
 
 // Add label to namespaces
 func (t *Task) labelNamespaces(client k8sclient.Client) error {
-	for _, ns := range t.srcNamespaces() {
+	for _, ns := range t.sourceNamespaces() {
 		namespace := corev1.Namespace{}
 		err := client.Get(
 			context.TODO(),
@@ -196,7 +196,7 @@ func (t *Task) labelNamespaces(client k8sclient.Client) error {
 // Returns a set of referenced service accounts.
 func (t *Task) annotatePods(client k8sclient.Client) (ServiceAccounts, error) {
 	serviceAccounts := ServiceAccounts{}
-	for _, ns := range t.srcNamespaces() {
+	for _, ns := range t.sourceNamespaces() {
 		list := corev1.PodList{}
 		options := k8sclient.InNamespace(ns)
 		err := client.List(context.TODO(), options, &list)
@@ -309,7 +309,7 @@ func (t *Task) annotatePVs(client k8sclient.Client) error {
 
 // Add label to service accounts.
 func (t *Task) labelServiceAccounts(client k8sclient.Client, serviceAccounts ServiceAccounts) error {
-	for _, ns := range t.srcNamespaces() {
+	for _, ns := range t.sourceNamespaces() {
 		names, found := serviceAccounts[ns]
 		if !found {
 			continue
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -348,14 +348,14 @@ func (t *Task) labelServiceAccounts(client k8sclient.Client, serviceAccounts Ser
 
 // Delete temporary annotations and labels added.
 func (t *Task) deleteAnnotations() error {
-	clients, err := t.getBothClients()
+	clients, namespaceList, err := t.getBothClientsWithNamespaces()
 	if err != nil {
 		log.Trace(err)
 		return err
 	}
 
-	for _, client := range clients {
-		err = t.deletePVCAnnotations(client)
+	for i, client := range clients {
+		err = t.deletePVCAnnotations(client, namespaceList[i])
 		if err != nil {
 			log.Trace(err)
 			return err
@@ -365,12 +365,12 @@ func (t *Task) deleteAnnotations() error {
 			log.Trace(err)
 			return err
 		}
-		err = t.deletePodAnnotations(client)
+		err = t.deletePodAnnotations(client, namespaceList[i])
 		if err != nil {
 			log.Trace(err)
 			return err
 		}
-		err = t.deleteNamespaceLabels(client)
+		err = t.deleteNamespaceLabels(client, namespaceList[i])
 		if err != nil {
 			log.Trace(err)
 			return err
@@ -386,8 +386,8 @@ func (t *Task) deleteAnnotations() error {
 }
 
 // Delete Pod stage annotations and labels.
-func (t *Task) deletePodAnnotations(client k8sclient.Client) error {
-	for _, ns := range t.srcNamespaces() {
+func (t *Task) deletePodAnnotations(client k8sclient.Client, namespaceList []string) error {
+	for _, ns := range namespaceList {
 		options := k8sclient.InNamespace(ns)
 		podList := corev1.PodList{}
 		err := client.List(context.TODO(), options, &podList)
@@ -437,8 +437,8 @@ func (t *Task) deletePodAnnotations(client k8sclient.Client) error {
 }
 
 // Delete stage label from namespaces
-func (t *Task) deleteNamespaceLabels(client k8sclient.Client) error {
-	for _, ns := range t.srcNamespaces() {
+func (t *Task) deleteNamespaceLabels(client k8sclient.Client, namespaceList []string) error {
+	for _, ns := range namespaceList {
 		namespace := corev1.Namespace{}
 		err := client.Get(
 			context.TODO(),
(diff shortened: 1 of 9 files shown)


pkg/controller/migplan/validation.go
```diff
@@ -361,6 +363,8 @@ func (r ReconcileMigPlan) validateSourceNamespaces(plan *migapi.MigPlan) error {
 	ns := kapi.Namespace{}
 	notFound := make([]string, 0)
 	for _, name := range namespaces {
+		namespaceMapping := strings.Split(name, ":")
+		name = namespaceMapping[0]
```
alexander-demichev: If namespace name that comes from spec has format of "foo:bar", we should only take "foo" for validation

pkg/controller/migmigration/restore.go
```diff
@@ -222,8 +224,25 @@ func (t *Task) buildRestore(backupName string) (*velero.Restore, error) {
...
+		}
+		if len(mapping) == 2 {
+			if mapping[1] == "" {
+				namespaceMapping[mapping[0]] = mapping[0]
+			} else {
+				namespaceMapping[mapping[0]] = mapping[1]
+			}
+		}
+	}
+
```
alexander-demichev: Not sure about this behaviour, but above snippet always passes mapping to restore. 
For example: 
 ```
namespaces:
    - foo1:bar1
    - bar2
``` 
becomes
`{"foo1":"bar1", "bar2":"bar2"}`

File: pkg/controller/migmigration/restore.go [hunk 3]
```diff
@@ -222,8 +224,25 @@ func (t *Task) buildRestore(backupName string) (*velero.Restore, error) {
...
+	namespaceMapping := make(map[string]string)
+	for _, namespace := range t.namespaces() {
+		mapping := strings.Split(namespace, ":")
+		if len(mapping) == 1 {
+			namespaceMapping[namespace] = namespace
+		}
+		if len(mapping) == 2 {
+			if mapping[1] == "" {
+				namespaceMapping[mapping[0]] = mapping[0]
```
dymurray: I get why this check is here, but it is weird that a user could validly enter `test:` and have it be valid.

dymurray: I get why this check is here, but it feels weird a user could have `test:` as a valid namespace mapping.

sseago:
@dymurray regarding upstream changes, we definitely need to investigate that, but hopefully any changes we need can be pulled in with the 1.2 beta whenever that drops later this month. With the velero org change, once we update to velero master branch (or 1.2-related tags) we're going to have to change all of our velero imports, and possibly the way we pull in the storage plugin (since aws/gcp/etc are moving out-of-tree). I don't know if the plugin changes have landed yet or not. We could possibly cherry-pick the specific namespace-related fixes if they were merged before the refactoring, but I'd rather avoid that if possible, since it will make subsequent fork updates more difficult. I guess the determining factor is whether the code in 1.1 is missing fixes that will block development and testing of our plugin changes or if it's just fixes for things that we need before release but don't necessarily need immediately.

pkg/controller/migmigration/restore.go
sseago: So based on this code, "test", "test:", and "test:test" are all equivalent. As long as velero interprets the three this way, this seems like the right way to handle it.

sseago:
If this is the issue in question: https://github.com/vmware-tanzu/velero/issues/1888
I don't *think* this affects us, since we're not specifying include-namespaces on restore, but I may be misunderstanding the upstream PR and it may still hit us since we are specifying namespaces on backup.

pkg/controller/migmigration/restore.go
sseago: Actually, looking at the upstream code, I think it would be cleaner to omit namespaces from the NamespaceMapping if they're not being mapped to something else. Looking at this code, an unmapped namespace will just keep its original value:
```
                        // get target namespace to restore into, if different
                        // from source namespace
                        targetNamespace := namespace
                        if target, ok := ctx.restore.Spec.NamespaceMapping[namespace]; ok {
                                targetNamespace = target
                        }
```

pkg/controller/migmigration/restore.go
sseago: See the above comment, but I think the appropriate value is actually `{"foo1":"bar1"}`

pkg/controller/migmigration/restore.go
```diff
@@ -222,8 +224,25 @@ func (t *Task) buildRestore(backupName string) (*velero.Restore, error) {
...
+			}
+		}
+	}
+
 	restore.Spec = velero.RestoreSpec{
-		BackupName: backupName,
-		RestorePVs: &restorePVs,
+		BackupName:       backupName,
+		RestorePVs:       &restorePVs,
+		NamespaceMapping: namespaceMapping,
```
sseago: I'm not quite sure whether we should explicitly list an empty mapping like this or whether we should just omit it in the case where we're not changing any namespaces (which we'll know because the mapping will be empty).

alexander-demichev:
@sseago @jortel @dymurray Thank you all for reviews! I added `GetSourceNamespaces()` function to MigPlan.

pkg/controller/migmigration/annotation.go
```diff
@@ -387,7 +387,7 @@ func (t *Task) deleteAnnotations() error {
 
 // Delete Pod stage annotations and labels.
 func (t *Task) deletePodAnnotations(client k8sclient.Client) error {
-	for _, ns := range t.namespaces() {
+	for _, ns := range t.srcNamespaces() {
```
sseago: This is run on both src and destination, so we really need srcNamespaces when we're calling it on src and destNamespaces() when calling it on dest. This might require refactoring deleteAnnotations() to explicitly grab src and dest client rather than calling getBothClients() -- and also a separate destNamespaces() function.

pkg/controller/migmigration/annotation.go
```diff
@@ -438,7 +438,7 @@ func (t *Task) deletePodAnnotations(client k8sclient.Client) error {
 
 // Delete stage label from namespaces
 func (t *Task) deleteNamespaceLabels(client k8sclient.Client) error {
-	for _, ns := range t.namespaces() {
+	for _, ns := range t.srcNamespaces() {
```
sseago: Same issue as above with src vs dest namespaces.

pkg/controller/migmigration/annotation.go
```diff
@@ -467,7 +467,7 @@ func (t *Task) deleteNamespaceLabels(client k8sclient.Client) error {
 
 // Delete PVC stage annotations and labels.
 func (t *Task) deletePVCAnnotations(client k8sclient.Client) error {
-	for _, ns := range t.namespaces() {
+	for _, ns := range t.srcNamespaces() {
```
sseago: Same issue as above with src vs dest namespaces

pkg/controller/migmigration/annotation.go
sseago: Actually, thinking about it a bit more, we still want to be able to iterate over both clients instead of duplicating code. Maybe add a new fun `getBothClientsAndNamespaceList` which returns, instead of `{srcClient, destCLient}` two "client,namespaceList" pairs -- i.e. `{ {srcClient, srcNamespaces}, {destClient, destNamespaces}` -- then you can pass in the associated namespace list with the right client.

pkg/controller/migmigration/pod.go
```diff
@@ -264,7 +264,7 @@ func (t *Task) ensureStagePodsDeleted() error {
 	cLabel, _ := t.Owner.GetCorrelationLabel()
 	for _, client := range clients {
 		podList := corev1.PodList{}
-		for _, ns := range t.namespaces() {
+		for _, ns := range t.srcNamespaces() {
```
sseago: same as above with the bothClients refactor

pkg/controller/migmigration/task.go
```diff
@@ -590,16 +600,25 @@ func (t *Task) getBothClusters() []*migapi.MigCluster {
...
-func (t *Task) getBothClients() ([]k8sclient.Client, error) {
-	list := []k8sclient.Client{}
-	for _, cluster := range t.getBothClusters() {
-		client, err := cluster.GetClient(t.Client)
-		if err != nil {
-			log.Trace(err)
-			return nil, err
-		}
-		list = append(list, client)
+func (t *Task) getBothClientsWithNamespaces() ([]k8sclient.Client, [][]string, error) {
```
jortel: This seems like a worthwhile addition.  However, I'm not convinced it should replace `getBothClients()`.  Although, there are no more usages it may be better _functional decomposition_ and forward thinking to layer them.  That is, add `getBothClientsWithNamespaces()` but have the implementation _use_ `getBothClients()`.

PR #339, opened 2019-10-15
Add cloud provider watch; routinely run provider tests.

### LINKED ISSUE #292: Bug 1738564 - MigStorage object not updated when AWS bucket is deleted from AWS
https://bugzilla.redhat.com/show_bug.cgi?id=1738564

Add ability to periodically check the connection details of MigStorage and detect when external entities have changed, in this case AWS bucket has been deleted from AWS S3.

jortel:
Add custom _watch_ Source used to monitor the health of cloud provider.  Routinely runs provider tests and triggers reconcile when that status changes.

Interval= 30 seconds.  Seem right?

fixes #292

### DIFF
--- a/pkg/controller/migstorage/migstorage_controller.go
+++ b/pkg/controller/migstorage/migstorage_controller.go
@@ -19,6 +19,7 @@ package migstorage
 import (
 	"context"
 	"github.com/fusor/mig-controller/pkg/logging"
+	"time"
 
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
 	migref "github.com/fusor/mig-controller/pkg/reference"
@@ -79,6 +80,17 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 					return migref.GetRequests(a, migapi.MigStorage{})
 				}),
 		})
+	if err != nil {
+		log.Trace(err)
+		return err
+	}
+
+	// Watch for changes to cloud providers.
+	err = c.Watch(
+		&ProviderSource{
+			Client:   mgr.GetClient(),
+			Interval: time.Second * 30},
+		&handler.EnqueueRequestForObject{})
 	if err != nil {
 		log.Trace(err)
 		return err
--- a/pkg/controller/migstorage/provider.go
+++ b/pkg/controller/migstorage/provider.go
@@ -1,0 +1,100 @@
+package migstorage
+
+import (
+	"github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	event2 "sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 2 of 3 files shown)


pkg/controller/migstorage/provider.go
```diff
@@ -0,0 +1,100 @@
+package migstorage
+
+import (
+	"github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	event2 "sigs.k8s.io/controller-runtime/pkg/event"
```
djwhatle: Why `event2`? May be missing something but don't see `event` taken by an existing import.

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -84,6 +85,17 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to cloud providers.
+	err = c.Watch(
```
djwhatle: Looks like watches are being used to add regularly occurring events onto the queue. Cool use-case.

pkg/controller/migstorage/provider.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Enqueue a reconcile request.
+func (p *ProviderSource) enqueue(storage v1alpha1.MigStorage) {
+	event := event2.GenericEvent{
+		Meta:   &storage.ObjectMeta,
+		Object: &storage,
+	}
+	for _, predicate := range p.predicates {
+		if !predicate.Generic(event) {
```
djwhatle: What is this checking?

pkg/controller/migstorage/provider.go
```diff
@@ -0,0 +1,100 @@
...
+			return
+		}
+	}
+
+	p.handler.Generic(event, p.queue)
```
djwhatle: What does this do? Enqueue an event?

pkg/controller/migstorage/provider.go
jortel: As you recall, _predicates_ are used to filter events.  The code is checking with all of the predicates to see if the event should be processed by the _EventHandler_.

pkg/controller/migstorage/provider.go
jortel: Yes, enqueueing the event is delegated to the _EventHandler_.

pkg/controller/migstorage/provider.go
jortel: There must have already been a local variable named `event` and my IDE did that to resolve the shadowing.  Will fix.

pkg/controller/migstorage/provider.go
djwhatle: Oh interesting. I hadn't seen predicates invoked in this way before. Thanks for the explanation.

pkg/controller/migstorage/provider.go
djwhatle: Got it. 👍

PR #341, opened 2019-10-16
Azure BSL support

djwhatle:
Adds support for Azure as a BackupStorageLocation / Migration Registry storage provider.

## Test Instructions

First, ensure that you're deploying other mig components with recent version of mig-operator. This is necessary to ensure that the AZURE_CREDENTIALS_FILE env var is set. https://github.com/fusor/mig-operator/blob/master/roles/migrationcontroller/templates/velero.yml.j2#L60-L61

```
# Create Azure storage Resource Group

AZURE_RESOURCE_GROUP=Velero_Backups
az group create -n $AZURE_RESOURCE_GROUP --location CentralUS
```

```
# Create Azure Storage Account

az storage account create \
     --name $AZURE_STORAGE_ACCOUNT_ID \
     --resource-group $AZURE_RESOURCE_GROUP \
     --sku Standard_GRS \
     --encryption-services blob \
     --https-only true \
     --kind BlobStorage \
     --access-tier Hot
```

```
# Create Azure Blob Storage Container

BLOB_CONTAINER=velero
az storage container create -n $BLOB_CONTAINER --public-access off --account-name $AZURE_STORAGE_ACCOUNT_ID
```

```
# Create Service Principal for Velero to act as

AZURE_SUBSCRIPTION_ID=`az account list --query '[?isDefault].id' -o tsv`
AZURE_TENANT_ID=`az account list --query '[?isDefault].tenantId' -o tsv`
AZURE_CLIENT_SECRET=`az ad sp create-for-rbac --name "velero" --role "Contributor" --query 'password' -o tsv`
AZURE_CLIENT_ID=`az ad sp list --display-name "velero" --query '[0].appId' -o tsv`
```

```
# Dump secret credentials to file that we'll feed into creds secret

cat << EOF  > ./credentials-velero
AZURE_SUBSCRIPTION_ID=${AZURE_SUBSCRIPTION_ID}
AZURE_TENANT_ID=${AZURE_TENANT_ID}
AZURE_CLIENT_ID=${AZURE_CLIENT_ID}
AZURE_CLIENT_SECRET=${AZURE_CLIENT_SECRET}
AZURE_RESOURCE_GROUP=${AZURE_RESOURCE_GROUP}
AZURE_CLOUD_NAME=AzurePublicCloud
EOF
```

```
# Load b64 encoded contents of credentials into mig-storage-creds

cat << EOF  > ./mig-storage-creds-azure.yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: migstorage-creds
type: Opaque
data:
  azure-credentials: $(base64 credentials-velero -w 0)

EOF


oc create -f mig-storage-creds-azure.yaml
```

```
# Create migstorage configured against Azure Blob Storage

cat << EOF  > ./mig-storage-azure.yaml
---
apiVersion: migration.openshift.io/v1alpha1
kind: MigStorage
metadata:
  labels:
  name: migstorage-sample
spec:
  backupStorageProvider: azure
  volumeSnapshotProvider: azure

  backupStorageConfig:
    azureStorageAccount: ${AZURE_STORAGE_ACCOUNT_ID}
    azureStorageContainer: ${BLOB_CONTAINER}
    azureResourceGroup: ${AZURE_RESOURCE_GROUP}

    credsSecretRef:
      name: migstorage-creds

  volumeSnapshotConfig:
    azureResourceGroup: ${AZURE_RESOURCE_GROUP}
    azureApiTimeout: 30s
    credsSecretRef:
      name: migstorage-creds

EOF

oc create -f mig-storage-azure.yaml
```


djwhatle:
@jortel @dymurray ready for review

djwhatle:
@jortel squashed.

config/samples/mig-storage-creds.yaml
```diff
@@ -17,8 +17,15 @@ data:
...
+  #     AZURE_SUBSCRIPTION_ID=${AZURE_SUBSCRIPTION_ID}
+  #     AZURE_TENANT_ID=${AZURE_TENANT_ID}
+  #     AZURE_CLIENT_ID=${AZURE_CLIENT_ID}
+  #     AZURE_CLIENT_SECRET=${AZURE_CLIENT_SECRET}
+  #     AZURE_RESOURCE_GROUP=${AZURE_RESOURCE_GROUP}
+  #     AZURE_CLOUD_NAME=AzurePublicCloud
+  azure-credentials: aGVsbG8K
+
+  # [!] If using GCP, change `gcp-credentials` below to contain base64 encoded credentials
+  gcp-credentials: aGVsbG8K
```
jortel: Thanks for adding the GCP example.

pkg/cloudprovider/azure.go
```diff
@@ -1,33 +1,121 @@
...
+				SecretKeyRef: &kapi.SecretKeySelector{
+					LocalObjectReference: kapi.LocalObjectReference{Name: name},
+					Key:                  storageAccountKeyConfigKey,
+				},
+			},
+		},
+	}
 }
 
 func (p *AzureProvider) Validate(secret *kapi.Secret) []string {
```
jortel: Let's validate the credentials secret here:
1. Better aligned with how invalid configuration is reported during _validation_.  See AWS.
2. Simplify code.  Don't need to check everywhere the secret is used.

https://github.com/fusor/mig-controller/blob/master/pkg/cloudprovider/aws.go#L162

pkg/cloudprovider/azure.go
```diff
@@ -52,9 +143,218 @@ func (p *AzureProvider) Validate(secret *kapi.Secret) []string {
...
+			container: p.StorageContainer,
+			client:    blobClient,
+		}
+		err = test.Run()
+	}
+
+	return err
+}
+
+func (p *AzureProvider) getBlobClient(cloudCreds map[string]string) (*azstorage.BlobStorageClient, error) {
```
jortel: This is only used by the `AzureBlobTest`.  Seems simpler for the `AzureBlobTest` to have a field for the _secret_ and `getBlobClient()` be a method.

pkg/cloudprovider/azure.go
```diff
@@ -52,9 +143,218 @@ func (p *AzureProvider) Validate(secret *kapi.Secret) []string {
...
+	if err != nil {
+		return err
+	}
+
+	blob.CreateBlockBlobFromReader(bytes.NewReader([]byte{0}), nil)
+
+	return err
+}
+
+func (r *AzureBlobTest) objectExists() (bool, error) {
```
jortel: The other provider tests support _download()_ instead of just checking if it exists.  Why is azure different?

pkg/cloudprovider/azure.go
```diff
@@ -52,9 +143,218 @@ func (p *AzureProvider) Validate(secret *kapi.Secret) []string {
...
+	blob, err := getBlob(*r.client, r.container, r.key)
+	if err != nil {
+		return err
+	}
+	err = blob.Delete(nil)
+
+	return err
+}
+
+func getBlob(client azstorage.BlobStorageClient, bucket, key string) (*azstorage.Blob, error) {
```
jortel: Seems like `getBlob()` should be a method on `AzureBlobTest`.

pkg/cloudprovider/azure.go
```diff
@@ -52,9 +143,218 @@ func (p *AzureProvider) Validate(secret *kapi.Secret) []string {
...
+	if blob == nil {
+		return nil, errors.Errorf("unable to get blob reference for key %v", key)
+	}
+
+	return blob, nil
+}
+
+func (p *AzureProvider) getStorageAccountKey(cloudCreds map[string]string) (string, *azure.Environment, error) {
+	// 1. we need AZURE_TENANT_ID, AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, AZURE_SUBSCRIPTION_ID
+	azureCreds, err := getRequiredValues(mapLookup(cloudCreds), tenantIDEnvVar, clientIDEnvVar, clientSecretEnvVar, subscriptionIDEnvVar)
```
jortel: Line length <= 100 for easier PR review.

pkg/cloudprovider/azure.go
```diff
@@ -1,33 +1,121 @@
...
 
+// Credentials secret.
+const (
+	AzureCredentials = "azure-credentials"
+
+	tenantIDEnvVar       = "AZURE_TENANT_ID"
+	subscriptionIDEnvVar = "AZURE_SUBSCRIPTION_ID"
+	clientIDEnvVar       = "AZURE_CLIENT_ID"
+	clientSecretEnvVar   = "AZURE_CLIENT_SECRET"
+	cloudNameEnvVar      = "AZURE_CLOUD_NAME"
```
jortel: Are these really environment variables? ^^

pkg/cloudprovider/azure.go
```diff
@@ -52,9 +143,218 @@ func (p *AzureProvider) Validate(secret *kapi.Secret) []string {
...
+	}
+
+	if storageKey == "" {
+		return "", env, errors.New("No storage key with Full permissions found")
+	}
+
+	return storageKey, env, nil
+}
+
+func getRequiredValues(getValue func(string) string, keys ...string) (map[string]string, error) {
```
jortel: I'm a bit lost as to what `getRequiredValues()` and `maplookup()` provide.

pkg/cloudprovider/azure.go
```diff
@@ -52,9 +143,218 @@ func (p *AzureProvider) Validate(secret *kapi.Secret) []string {
...
+
+	blob := container.GetBlobReference(key)
+	if blob == nil {
+		return nil, errors.Errorf("unable to get blob reference for key %v", key)
+	}
+
+	return blob, nil
+}
+
+func (p *AzureProvider) getStorageAccountKey(cloudCreds map[string]string) (string, *azure.Environment, error) {
```
jortel: Seems like `getStorageAccountKey()` should be a method on the provider.

pkg/cloudprovider/azure.go
```diff
@@ -52,9 +143,218 @@ func (p *AzureProvider) Validate(secret *kapi.Secret) []string {
...
+
+func mapLookup(data map[string]string) func(string) string {
+	return func(key string) string {
+		return data[key]
+	}
+}
+
+// ParseAzureEnvironment returns an azure.Environment for the given cloud
+// name, or azure.PublicCloud if cloudName is empty.
+func parseAzureEnvironment(cloudName string) (*azure.Environment, error) {
```
jortel: Seem like `parseAzureEnvironment()` could be a method on the provider.  And, perhaps `get` instead of `parse` would be clearer.

pkg/cloudprovider/azure.go
djwhatle: The secret data containing these vars is in the `.env` file format, which looks like this:
```
AZURE_TENANT_ID=foo
AZURE_SUBSCRIPTION_ID=foo
AZURE_CLIENT_ID=foo
```

This is the format in which Velero accepts credentials. I'm open to suggestions for naming these items something else, they're the keys within an environment variable `.env` file.

pkg/cloudprovider/azure.go
djwhatle: I'm looking at our AWS implementation, it seems like what you're asking for is:
 - Read the data from the secret
 - Make sure all required values within secret are filled (this is all we can do for Azure as far as I can tell)

I'm not sure what you mean with `Simplify code. Don't need to check everywhere the secret is used.`, can you elaborate?

pkg/cloudprovider/azure.go
djwhatle: This seems like a good change, will implement.

pkg/cloudprovider/azure.go
djwhatle: My thought was that downloading the content is pointless since it isn't used, but I can see an argument in favor of downloading to ensure proper permissions are in-place (could be a different permission is needed for downloading vs. just viewing).

pkg/cloudprovider/azure.go
djwhatle: Works for me, will change this.

pkg/cloudprovider/azure.go
djwhatle: I think it already is. Typo?

pkg/cloudprovider/azure.go
djwhatle: ACK

pkg/cloudprovider/azure.go
djwhatle: `mapLookup` is a function that pulls values out of golang maps for `getRequiredValues`. Brought some of this code for transforming Azure Creds from to required vals for inclusion in registry config over from Velero, we don't really need the indirection between the two functions as we only apply `getRequiredValues` to a map.

pkg/cloudprovider/azure.go
djwhatle: sounds good to me

pkg/cloudprovider/azure.go
djwhatle: Changed.

pkg/cloudprovider/azure.go
djwhatle: Changed.

pkg/cloudprovider/azure.go
djwhatle: Changed.

pkg/cloudprovider/azure.go
djwhatle: Changed.

djwhatle:
@jortel updated in response to feedback.

pkg/cloudprovider/azure.go
jortel: Still appears to be a standalone function.

pkg/cloudprovider/azure.go
jortel: IMHO, just dropping the `EnvVar` suffix would be better since the keys are used in both the secret and the azure .env file.  I found this confusing (at first) because it suggested (to me) actual  process environment variables.

pkg/cloudprovider/azure.go
jortel: It may have been and I just misremembered it.

pkg/cloudprovider/azure.go
djwhatle: Oops, will fix

pkg/cloudprovider/azure.go
djwhatle: Removed mapLookup and getRequiredValues and replaced with validation function.

pkg/cloudprovider/azure.go
djwhatle: Replaced `EnvVar` with `Key`

djwhatle:
Rebased on master to resolve conflict.

pkg/cloudprovider/aws.go
```diff
@@ -102,11 +104,12 @@ func (p *AWSProvider) UpdateCloudSecret(secret, cloudSecret *kapi.Secret) {
 	}
 }
 
-func (p *AWSProvider) UpdateRegistrySecret(secret, registrySecret *kapi.Secret) {
+func (p *AWSProvider) UpdateRegistrySecret(secret, registrySecret *kapi.Secret) error {
```
dymurray: Maybe I'm missing something obvious here, but why add an `error` return if we never actually return an error?

pkg/cloudprovider/azure.go
```diff
@@ -52,9 +175,202 @@ func (p *AzureProvider) Validate(secret *kapi.Secret) []string {
...
+	return err
+}
+
+func (r *AzureBlobTest) download() error {
+	blob, err := r.getBlob()
+	if err != nil {
+		return err
+	}
+
+	_, err = blob.Get(nil)
```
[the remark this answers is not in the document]
djwhatle: @dymurray `blob` in this case already holds the reference to the key from the storageContainer we want (key == filename), so we can run a `blob.Get(nil)` where nil means no Get options are specified and it will retrieve the preset key name from the preset containerName.

See `r.container`, `r.key`
```
func (r *AzureBlobTest) getBlob() (*azstorage.Blob, error) {
	container := r.client.GetContainerReference(r.container)
	if container == nil {
		return nil, errors.Errorf("unable to get container reference for bucket %v", r.container)
	}

 	blob := container.GetBlobReference(r.key)
	if blob == nil {
		return nil, errors.Errorf("unable to get blob reference for key %v", r.key)
	}

 	return blob, nil
}
```

pkg/cloudprovider/aws.go
djwhatle: @dymurray needed to satisfy the `Provider` interface which is shared between AWS, GCP, Azure. For Azure, it's possible to encounter errors while updating the registry secret since we have to derive a storageAccountKey from the provided info.

```
type Provider interface {
	SetRole(role string)
	UpdateBSL(location *velero.BackupStorageLocation)
	UpdateVSL(location *velero.VolumeSnapshotLocation)
	UpdateCloudSecret(secret, cloudSecret *kapi.Secret)
	UpdateRegistrySecret(secret, registrySecret *kapi.Secret) error
	UpdateRegistryDC(dc *appsv1.DeploymentConfig, name, dirName string)
	Validate(secret *kapi.Secret) []string
	Test(secret *kapi.Secret) error
}
```

Deriving storageAccountKey from info available on secret (needed for mig-registry config)

https://github.com/fusor/mig-controller/blob/6b6701a854f9229a434a496686534da7cd4a901d/pkg/cloudprovider/azure.go#L82-L86

PR #342, opened 2019-10-21
Stop the remote watch when a cluster has connection problems

mansam:
This branch stops the running RemoteWatch for a migcluster if
that cluster has any blocker conditions. This allows it to
be restarted once the migcluster is ready again.

This branch also adds a custom watch source to watch the
connectivity of the remote clusters on a regular interval.

@jortel This issue seemed pretty similar to the storage issue, in that we need to know about the condition of a system outside of Kubernetes, so I took a similar approach. Curious to know what your thoughts are on this approach.

### DIFF
--- a/pkg/controller/migcluster/migcluster_controller.go
+++ b/pkg/controller/migcluster/migcluster_controller.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
+	"time"
 
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
@@ -68,6 +69,17 @@ func add(mgr manager.Manager, r *ReconcileMigCluster) error {
 		&source.Kind{Type: &migapi.MigCluster{}},
 		&handler.EnqueueRequestForObject{},
 		&ClusterPredicate{})
+	if err != nil {
+		log.Trace(err)
+		return err
+	}
+
+	// Watch remote clusters for connection problems
+	err = c.Watch(
+		&RemoteClusterSource{
+			Client:   mgr.GetClient(),
+			Interval: time.Second * 60},
+		&handler.EnqueueRequestForObject{})
 	if err != nil {
 		log.Trace(err)
 		return err
@@ -155,6 +167,8 @@ func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Re
 			log.Trace(err)
 			return reconcile.Result{Requeue: true}, nil
 		}
+	} else {
+		r.shutdownRemoteWatch(cluster)
 	}
 
 	// Ready
@@ -217,4 +231,15 @@ func (r *ReconcileMigCluster) setupRemoteWatch(cluster *migapi.MigCluster) error
 	log.Info("Remote watch started.", "cluster", cluster.Name)
 
 	return nil
+}
+
+func (r *ReconcileMigCluster) shutdownRemoteWatch(cluster *migapi.MigCluster) {
+	log.Info("Stopping remote watch.", "cluster", cluster.Name)
+	nsName := types.NamespacedName{
+		Namespace: cluster.Namespace,
+		Name:      cluster.Name,
+	}
+
+	StopRemoteWatch(nsName)
+	log.Info("Stopped remote watch.", "cluster", cluster.Name)
 }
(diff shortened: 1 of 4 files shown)


pkg/controller/migcluster/remote_watch.go
```diff
@@ -79,19 +80,29 @@ func StartRemoteWatch(r *ReconcileMigCluster, config remote.ManagerConfig) error
...
 	// MigClusters have a 1:1 association with a RemoteWatchCluster, so we will store the mapping
 	// to avoid creating duplicate remote managers in the future.
 	remoteWatchMap.Set(config.ParentNsName, remoteWatchCluster)
 
 	return nil
 }
+
+// StopRemoteWatch will close a remote watch's stop channel
+// and delete it from the remote watch map.
+func StopRemoteWatch(nsName types.NamespacedName) {
```
djwhatle: @mansam thanks for implementing this! Has been on my todo list for many months.

pkg/controller/migcluster/remotecluster.go
```diff
@@ -0,0 +1,75 @@
...
+	for {
+		time.Sleep(r.Interval)
+		list, err := v1alpha1.ListClusters(r.Client)
+		if err != nil {
+			log.Trace(err)
+			return
+		}
+
+		for _, cluster := range list {
+
```
jortel: Blank line ^ seem unnecessary.

pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -73,6 +74,17 @@ func add(mgr manager.Manager, r *ReconcileMigCluster) error {
 		return err
 	}
 
+	// Watch remote clusters for connection problems
+	err = c.Watch(
```
djwhatle: This is a cool application of watches, not one I would have expected.

pkg/controller/migcluster/remotecluster.go
```diff
@@ -0,0 +1,75 @@
...
+			// Enqueue if our view of the cluster disagrees with the result of the connectivity test.
+			timeout := time.Duration(time.Second * 5)
+			err := cluster.TestConnection(r.Client, timeout)
+			if (cluster.Status.HasBlockerCondition() && err == nil) || (!cluster.Status.HasBlockerCondition() && err != nil) {
```
jortel: The test needs to be skipped when the cluster is broken:
```
if cluster.Status.HasAnyCondition(
    InvalidURL,
    InvalidSaSecretRef,
    InvalidSaToken) {
    continue
}
```

Please check line length <= 100.  This is a guideline on the controller mainly so that code is visible during review without scrolling.

There are other blocker conditions besides the failed test.  We need to be specific here.
Replace: `cluster.Status.HasBlockerCondition()` 

with 

`cluster.Status.HasCondition(TestConnectFailed)`.

pkg/controller/migcluster/remotecluster.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Enqueue a reconcile request.
+func (r *RemoteClusterSource) enqueue(cluster v1alpha1.MigCluster) {
+	clusterEvent := event.GenericEvent{
+		Meta:   &cluster.ObjectMeta,
+		Object: &cluster,
+	}
+	for _, p := range r.predicates {
+		if !p.Generic(clusterEvent) {
```
djwhatle: @mansam what is this checking?

pkg/controller/migcluster/remotecluster.go
```diff
@@ -0,0 +1,75 @@
...
+			return
+		}
+	}
+
+	r.handler.Generic(clusterEvent, r.queue)
```
djwhatle: Is this calling an handler directly? I haven't seen this type of call before.

pkg/controller/migcluster/remotecluster.go
mansam: This would run any predicate tests included in the call to Watch. In this use case, there aren't any.

pkg/controller/migcluster/remotecluster.go
mansam: This is calling Generic on the handler to add the GenericEvent created above to the queue. I borrowed this from @jortel's custom storage watch source so he'd probably be better to explain it than me.

pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -73,6 +74,17 @@ func add(mgr manager.Manager, r *ReconcileMigCluster) error {
...
+		&RemoteClusterSource{
+			Client:   mgr.GetClient(),
+			Interval: time.Second * 60},
```
jortel: 60 seconds seems reasonable for this.

jortel:
One final note.  Let's keep an eye out for memory leaks just in case the garbage collected _client_ needs to be _closed_ first (or something).

PR #343, opened 2019-10-22
Avoid creating stage pod using controller

alexander-demichev:
All functionality is described here - https://github.com/fusor/mig-controller/issues/330

### DIFF
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -235,7 +235,14 @@ func (t *Task) annotatePods(client k8sclient.Client) (ServiceAccounts, error) {
 			if pod.Labels == nil {
 				pod.Labels = make(map[string]string)
 			}
+
+			labels := t.Owner.GetCorrelationLabels()
+			for label, value := range labels {
+				pod.Labels[label] = value
+			}
+
 			pod.Labels[ApplicationPodLabel] = t.UID()
+			pod.Labels[IncludedInStageBackupLabel] = t.UID()
 			// Update
 			err = client.Update(context.TODO(), &pod)
 			if err != nil {
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -30,9 +30,9 @@ const (
 	// Resources included in the stage backup.
 	// Referenced by the Backup.LabelSelector. The value is the Task.UID().
 	IncludedInStageBackupLabel = "migration-included-stage-backup"
-	// Application pods requiring restic/stage backups.
+	// Stage pods requiring restic/stage backups.
 	// Used to create stage pods. The value is the Task.UID()
-	ApplicationPodLabel = "migration-application-pod"
+	StagePodLabel = "migration-stage-pod"
 	// Designated as an `initial` Backup.
 	// The value is the Task.UID().
 	InitialBackupLabel = "migration-initial-backup"
@@ -208,11 +208,6 @@ func (t *Task) annotatePods(client k8sclient.Client) (ServiceAccounts, error) {
 			if pod.Labels == nil {
 				pod.Labels = make(map[string]string)
 			}
-			// Skip stage pods.
-			cLabel, _ := t.Owner.GetCorrelationLabel()
-			if _, found := pod.Labels[cLabel]; found {
-				continue
-			}
 			// Skip stateless pods.
 			if len(pod.Spec.Volumes) == 0 {
 				continue
@@ -236,12 +231,7 @@ func (t *Task) annotatePods(client k8sclient.Client) (ServiceAccounts, error) {
 				pod.Labels = make(map[string]string)
 			}
 
-			labels := t.Owner.GetCorrelationLabels()
-			for label, value := range labels {
-				pod.Labels[label] = value
-			}
-
-			pod.Labels[ApplicationPodLabel] = t.UID()
+			pod.Labels[StagePodLabel] = t.UID()
 			pod.Labels[IncludedInStageBackupLabel] = t.UID()
 			// Update
 			err = client.Update(context.TODO(), &pod)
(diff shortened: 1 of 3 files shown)


pkg/controller/migmigration/annotation.go
```diff
@@ -235,7 +235,14 @@ func (t *Task) annotatePods(client k8sclient.Client) (ServiceAccounts, error) {
 			if pod.Labels == nil {
 				pod.Labels = make(map[string]string)
 			}
+
+			labels := t.Owner.GetCorrelationLabels()
```
jortel: :+1:  Great use of correlation labels instead of a new label.

Not sure why this code is singled out as a separate block (surrounded by blank lines).

jortel:
When annotating pods, we no longer need to skip stage pods since they no longer exist on the source cluster.  https://github.com/fusor/mig-controller/blob/3b51d5e4bf59c0d22a7ed7b38c3cc73b24cfe354/pkg/controller/migmigration/annotation.go#L211

pkg/controller/migmigration/task.go
```diff
@@ -525,9 +515,6 @@ func (t *Task) next() {
 		if next.flags&HasPVs != 0 && !t.hasPVs() {
 			continue
 		}
-		if next.flags&HasStagePods != 0 && !t.Owner.Status.HasCondition(StagePodsCreated) {
```
jortel: The `HasStagePods` flag is still being used.  Two choices here:
1. Set the condition (perhaps renamed to `HasStagePods`) when at least 1 application pod is annotated. Continues to provide useful information to the user and supports skipping the stage pod delete/terminate phases.
2. Get rid of the flag.  Simpler.

jortel:
Please ensure that the `Task` unit tests have been updated.

pkg/controller/migmigration/pod.go
```diff
@@ -108,195 +105,45 @@ func (t *Task) haveResticPodsStarted() (bool, error) {
...
-					log.Trace(err)
-					return err
-				}
-				log.Info(
-					"Stage pod deleted.",
-					"ns",
-					pod.Namespace,
-					"name",
-					pod.Name)
+			if _, found := pod.Labels[cLabel]; !found {
```
jortel: The application pods on the destination will not have a _correlation label_ so I think this check can be removed. Instead update the `options` to filter on the StagePodLabel in addition to namespace.

pkg/controller/migmigration/pod.go
```diff
@@ -108,195 +105,45 @@ func (t *Task) haveResticPodsStarted() (bool, error) {
...
-	for i, client := range clients {
-		podList := corev1.PodList{}
-		for _, ns := range namespaceList[i] {
-			options := k8sclient.InNamespace(ns)
-			err := client.List(context.TODO(), options, &podList)
-			if err != nil {
-				return err
+	podList := corev1.PodList{}
+	for _, ns := range namespaceList {
+		options := k8sclient.InNamespace(ns)
```
jortel: The _options_ should filter on both namespace and `StagePodLabel`.

pkg/controller/migmigration/pod.go
```diff
@@ -108,195 +105,45 @@ func (t *Task) haveResticPodsStarted() (bool, error) {
...
+		err := client.List(context.TODO(), options, &podList)
+		if err != nil {
+			return err
+		}
+		for _, pod := range podList.Items {
+			// Owned by ANY migration.
```
dymurray: Not sure I understand this comment. If labels is `nil`, aren't you saying that this *can't* be a stage pod?

pkg/controller/migmigration/pod.go
```diff
@@ -377,9 +224,44 @@ func (t *Task) veleroPodCredSecretPropagated(cluster *migapi.MigCluster) (bool,
...
+	cLabel, _ := t.Owner.GetCorrelationLabel()
+	podList := corev1.PodList{}
+	for _, ns := range namespaceList {
+		options := k8sclient.InNamespace(ns)
+		err := client.List(context.TODO(), options, &podList)
+		if err != nil {
+			return false, err
+		}
+		for _, pod := range podList.Items {
+			// Owned by ANY migration.
```
dymurray: Same question as above

pkg/controller/migmigration/pod.go
jortel: It's no longer needed now that the selector includes the `Labels` which are is no longer inspected in the loop for each pod.  The original purpose was to prevent dereferencing a `nil` map.

pkg/controller/migmigration/pod.go
jortel: Ah, missed this before, Alex.
The same namespace & label selector used for _delete_ should be used here.

alexander-demichev: Done

pkg/controller/migmigration/pod.go
```diff
@@ -377,9 +224,48 @@ func (t *Task) veleroPodCredSecretPropagated(cluster *migapi.MigCluster) (bool,
...
+
+// Ensure the stage pods have been deleted.
+func (t *Task) ensureStagePodsTerminated() (bool, error) {
+	client, err := t.getDestinationClient()
+	namespaceList := t.destinationNamespaces()
+	if err != nil {
+		log.Trace(err)
+		return false, err
+	}
+	cLabel, _ := t.Owner.GetCorrelationLabel()
```
jortel: Don't need the `cLabel`.

pkg/controller/migmigration/pod.go
```diff
@@ -377,9 +224,48 @@ func (t *Task) veleroPodCredSecretPropagated(cluster *migapi.MigCluster) (bool,
...
+			&k8sclient.ListOptions{
+				Namespace:     ns,
+				LabelSelector: selector,
+			},
+			&podList)
+		if err != nil {
+			return false, err
+		}
+		for _, pod := range podList.Items {
+			if _, found := pod.Labels[cLabel]; !found {
```
jortel: Stage pods do not have _correlation labels_.

dymurray:
Dependent on: https://github.com/fusor/openshift-migration-plugin/pull/37

pkg/controller/migmigration/pod.go
alexander-demichev: Done

pkg/controller/migmigration/pod.go
alexander-demichev: Done. Thanks for finding it.

PR #345, opened 2019-10-22
 Add a S3URL/PublicURL scheme check to AWSProvider.Validate

mansam:
Adds a check to `AWSProvider.Validate` to ensure that for S3URL and PublicURL, if set, have either HTTP or HTTPS as their scheme.

Also expands MigCluster's `validateURL` to check that the URL scheme is valid and that the URL is well formed.

Fixes https://github.com/fusor/mig-controller/issues/269

@jortel the AWS checks aren't where we talked about putting them, but after looking at the code for a while this seemed to make a little more sense to me. Curious what your opinion is.

### DIFF
--- a/pkg/controller/migcluster/validation.go
+++ b/pkg/controller/migcluster/validation.go
@@ -2,6 +2,7 @@ package migcluster
 
 import (
 	"fmt"
+	"net/url"
 	"time"
 
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
@@ -26,6 +27,8 @@ const (
 	NotSet        = "NotSet"
 	NotFound      = "NotFound"
 	ConnectFailed = "ConnectFailed"
+	Malformed     = "Malformed"
+	InvalidScheme = "InvalidScheme"
 )
 
 // Statuses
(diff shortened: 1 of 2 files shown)


pkg/controller/migcluster/validation.go
```diff
@@ -37,10 +40,12 @@ const (
 // Messages
 const (
 	ReadyMessage              = "The cluster is ready."
-	InvalidURLMessage         = "The `url` is required when `isHostCluster` is false."
+	MissingURLMessage         = "The `url` is required when `isHostCluster` is false."
 	InvalidSaSecretRefMessage = "The `serviceAccountSecretRef` must reference a `secret`."
 	InvalidSaTokenMessage     = "The `saToken` not found in `serviceAccountSecretRef` secret."
 	TestConnectFailedMessage  = "Test connect failed: %s"
+	MalformedUrlMessage       = "The `url` is malformed."
+	InvalidUrlSchemeMessage   = "The `url` scheme must be 'http' or 'https'."
```
jortel: To be consistent, let's uppercase the `URL` acronym in the const name.

PR #346, opened 2019-10-24
Remove auto-generated RBAC

djwhatle:
Resolves https://github.com/fusor/mig-controller/issues/273, we are relying on mig-operator as the source of truth for RBAC now.


djwhatle:
@sseago @jortel @dymurray can we get this merged today?

PR #347, opened 2019-10-24
migcluster: support using a custom CA bundle and insecure SSL connections

### LINKED ISSUE #91: Support use of `caBundle` when connecting to remote cluster
Our _MigCluster_ CR holds a reference to a cluster-registry _Cluster_, which has a spot for the user to input a caBundle. 

We are currently ignoring any `caBundle` input provided and always establish "insecure" connections to remote clusters.

I think it would be best to switch over to "opt-in to running without a caBundle", meaning that we would require the user to turn on an `insecure: true` bool on the _MigCluster_ if they don't want to use a caBundle. 

My thinking is that if we automatically use insecure mode if a user doesn't provide a caBundle, they may not realize they are operating insecurely, which would be undesirable. We should make the user take a conscious step to operate insecurely.

https://github.com/fusor/mig-controller/blob/f47f24237cfe601803f405f8f4e7217f511fab5f/pkg/apis/migration/v1alpha1/migcluster_types.go#L172.

image

mansam:
This PR adds an `insecure` flag to the migcluster to support self-signed certificates. If `insecure` is false, then either the cluster's certificate must be trusted at a system level or provided via the `caBundle` option.

Fixes #91

### DIFF
--- a/config/crds/migration_v1alpha1_migcluster.yaml
+++ b/config/crds/migration_v1alpha1_migcluster.yaml
@@ -33,6 +33,8 @@ spec:
             caBundle:
               format: byte
               type: string
+            insecureSkipTLSVerify:
+              type: boolean
             isHostCluster:
               type: boolean
             serviceAccountSecretRef:
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -48,6 +48,7 @@ type MigClusterSpec struct {
 	CABundle                []byte                `json:"caBundle,omitempty"`
 	StorageClasses          []StorageClass        `json:"storageClasses,omitempty"`
 	AzureResourceGroup      string                `json:"azureResourceGroup,omitempty"`
+	InsecureSkipTLSVerify   bool                  `json:"insecureSkipTLSVerify"`
 }
 
 // MigClusterStatus defines the observed state of MigCluster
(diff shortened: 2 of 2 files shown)


pkg/controller/migcluster/validation.go
```diff
@@ -152,6 +158,30 @@ func (r ReconcileMigCluster) validateSaSecret(cluster *migapi.MigCluster) error
 	return nil
 }
 
+func (r ReconcileMigCluster) validateCABundle(cluster *migapi.MigCluster) {
+	if cluster.Spec.IsHostCluster {
+		return
+	}
+	if cluster.Spec.Insecure && len(cluster.Spec.CABundle) > 0 {
```
jortel: I'm not convinced that raising a `critical` condition if the user specifies the CA bundle when it won't be used is valuable. Perhaps too strict?  Thoughts, @mansam?

mansam: @jortel Yeah, agreed. The conditional in BuildRestConfig prevents the error, so the status is unnecessary. Fixing.

pkg/controller/migcluster/validation.go
```diff
@@ -65,6 +67,8 @@ func (r ReconcileMigCluster) validate(cluster *migapi.MigCluster) error {
 		return err
 	}
 
+	r.validateCABundle(cluster)
```
dymurray: :+1:

jortel:
Merge blocked by required UI support.

mansam:
I've updated the PR to make the caBundle optional, since the remote cluster might already be trusted systemwide or have a cert from a public CA.

PR #350, opened 2019-10-24
Add support for limits.

jortel:
Support general concept of _Application Settings_ loaded from environment variables mounted from _ConfigMap_ managed by the operator. 

Add support for migration Plan limits for:
- Total number of namespaces listed on the Plan.  When exceeded:
  - Set `Critical` condition.
  - [SKIP] Validate namespaces (exist) on the source cluster.
  - [SKIP] Validate pod limit.
  - [SKIP] PV discovery.
- Total number of _running_ pods each namespace. When exceeded:
  - Set `Warn` condition.
- Total number and PVs listed on the Plan (as discovered). When exceeded:
  - Set `Warn` condition.

In the future: the supported settings can be expanded to include a number of things we currently have hard coded constants for in the controller such as the _VeleroNamespace_.  Also, the limits can be expanded and/or refined.

### DIFF
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -2,6 +2,8 @@ package migplan
 
 import (
 	"context"
+	"fmt"
+	"k8s.io/apimachinery/pkg/fields"
 	"reflect"
 	"strings"
 
@@ -10,6 +12,7 @@ import (
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // Types
@@ -25,6 +28,8 @@ const (
 	InvalidDestinationCluster      = "InvalidDestinationCluster"
 	NsNotFoundOnSourceCluster      = "NamespaceNotFoundOnSourceCluster"
 	NsNotFoundOnDestinationCluster = "NamespaceNotFoundOnDestinationCluster"
+	NsLimitExceeded                = "NamespaceLimitExceeded"
+	PodLimitExceeded               = "PodLimitExceeded"
 	PlanConflict                   = "PlanConflict"
 	PvInvalidAction                = "PvInvalidAction"
 	PvNoSupportedAction            = "PvNoSupportedAction"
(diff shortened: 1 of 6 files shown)


pkg/controller/migplan/validation.go
```diff
@@ -206,6 +223,68 @@ func (r ReconcileMigPlan) validateNamespaces(plan *migapi.MigPlan) error {
...
+	limit := Settings.Plan.PodLimit
+	for _, name := range plan.GetSourceNamespaces() {
+		list := kapi.PodList{}
+		options := k8sclient.InNamespace(name)
+		err := client.List(context.TODO(), options, &list)
+		if err != nil {
+			log.Trace(err)
+			return err
+		}
+		count := 0
```
djwhatle: Does this count get set back to 0 for each namespace? My impression was that we wanted to limit the total number of pods being migrated rather than # pods per namespace.

pkg/settings/settings.go
```diff
@@ -0,0 +1,42 @@
+package settings
```
djwhatle: Having a settings package to consume values from the configMap is a nice touch.

pkg/settings/settings.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"errors"
+	"os"
+	"strconv"
+)
+
+// Global
+var Settings = _Settings{}
```
djwhatle: Is there meaning behind the `_Settings` type with a leading underscore? Haven't seen that notation in golang before.

pkg/settings/settings.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Load settings.
+func (r *_Settings) Load() error {
+	err := r.Plan.Load()
+	return err
+}
+
+// Get positive integer limit from the environment
+// using the specified variable name and default.
+func getEnvLimit(name string, def int) (int, error) {
```
djwhatle: Nice touch defaulting, wonder if spelling out `default` here would be easier to read since `def` means define in some languages

pkg/controller/migplan/validation.go
jortel: You're right.  I did not remember that correctly.  Fixing.

pkg/settings/settings.go
jortel: IIRC, I saw that in kubernetes code.  It's a way to avoid the name collision between the global var and the type without using a wonky name for either.  I could have made the type lowercase but really don't like making types lowercase.

pkg/controller/migplan/validation.go
jortel: Fixed.

pkg/settings/settings.go
dymurray: Huh, good idea. Thanks.

jortel:
> One question -- should our messages include not just the limit but the number of items (namespaces, PVs, pods) found? That way the user will know whether they're near the limit (and thus maybe it's OK to push through the warning) or way over (in which case they may want to restructure their migplan)?

Good idea.
All (3) conditions updated to include the number found:
```
Message:               PV limit: 1 exceeded, found: 2.
```

PR #354, opened 2019-10-28
Support setting a custom CA bundle for S3-compatible storage

### LINKED ISSUE #270: Need to allow usage of https:// with minio (or other providers)
See: https://github.com/fusor/mig-ui/issues/454

Need a way to specify CA 
Would also like to expose a means of setting this as 'insecure'

mansam:
This adds a `AwsCustomCABundle` field to the BackupStorageConfig definition, which accepts a base64 encoded CA bundle that will be used instead of the system-provided certs to validate HTTPS connections to the S3-compatible storage.

I might just be missing it, but it appears the aws-sdk for Go does not support SSL-without-verification, so connecting to an endpoint with a self-signed cert would require that a CA bundle be provided here.

For reference, documentation on the behavior of the aws-sdk session's CustomCABundle option can be found here: https://docs.aws.amazon.com/sdk-for-go/api/aws/session/

Fixes #270

### DIFF
--- a/pkg/apis/migration/v1alpha1/migstorage_types.go
+++ b/pkg/apis/migration/v1alpha1/migstorage_types.go
@@ -87,11 +87,11 @@ type BackupStorageConfig struct {
 	AwsPublicURL          string                `json:"awsPublicUrl,omitempty"`
 	AwsKmsKeyID           string                `json:"awsKmsKeyId,omitempty"`
 	AwsSignatureVersion   string                `json:"awsSignatureVersion,omitempty"`
+	S3CustomCABundle      []byte                `json:"s3CustomCABundle,omitempty"`
 	AzureStorageAccount   string                `json:"azureStorageAccount,omitempty"`
 	AzureStorageContainer string                `json:"azureStorageContainer,omitempty"`
 	AzureResourceGroup    string                `json:"azureResourceGroup,omitempty"`
 	GcpBucket             string                `json:"gcpBucket,omitempty"`
-	AwsCustomCABundle     []byte                `json:"awsCustomCABundle,omitempty"`
 }
 
 func init() {
@@ -273,7 +273,7 @@ func (r *BackupStorageConfig) GetProvider(name string) pvdr.Provider {
 			KMSKeyId:         r.AwsKmsKeyID,
 			SignatureVersion: r.AwsSignatureVersion,
 			S3ForcePathStyle: r.AwsS3ForcePathStyle,
-			CustomCABundle:   r.AwsCustomCABundle,
+			CustomCABundle:   r.S3CustomCABundle,
 		}
 	case Azure:
 		provider = &pvdr.AzureProvider{
--- a/pkg/apis/migration/v1alpha1/migstorage_types.go
+++ b/pkg/apis/migration/v1alpha1/migstorage_types.go
@@ -91,6 +91,7 @@ type BackupStorageConfig struct {
 	AzureStorageContainer string                `json:"azureStorageContainer,omitempty"`
 	AzureResourceGroup    string                `json:"azureResourceGroup,omitempty"`
 	GcpBucket             string                `json:"gcpBucket,omitempty"`
+	AwsCustomCABundle     []byte                `json:"awsCustomCABundle,omitempty"`
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


pkg/apis/migration/v1alpha1/migstorage_types.go
```diff
@@ -91,6 +91,7 @@ type BackupStorageConfig struct {
 	AzureStorageContainer string                `json:"azureStorageContainer,omitempty"`
 	AzureResourceGroup    string                `json:"azureResourceGroup,omitempty"`
 	GcpBucket             string                `json:"gcpBucket,omitempty"`
+	AwsCustomCABundle     []byte                `json:"awsCustomCABundle,omitempty"`
```
djwhatle: I know this name is just following the prior convention for S3 related stuff `Aws[..]`, but still feels strange to me since the only time we would fill this value out would be if our S3 store was _not provided by AWS_. 

Anyone have thoughts on `S3CustomCABundle` for this param name instead?

djwhatle: I like having the caBundle directly on the MigStorage because it simplifies the process of providing this information. 

I am wondering if this will cause a burdensome CLI editing experience for MigStorage, since caBundles can be very large and can contain newlines, and putting this directly on resource means it won't be b64 encoded and will keep those newline characters intact.

Something to keep in mind but not recommending we change the current approach.

mansam: > Also, nit, but thoughts on moving this param up to be grouped with other AWS / S3 params?

Yeah, I'll move that up with the others.

mansam: Yeah, I'll move that up with the others.

mansam: > and putting this directly on resource means it won't be b64 encoded and will keep those newline characters intact.

Sorry, I'm not following-- why does having this on the resource inherently mean it won't be b64 encoded?

djwhatle: You *can* have b64 encoded data directly on a resource, but *all* data isn't shown in b64 *by default* when you look at `oc edit migstorage foo` like it would be if you were looking at a secret with `oc edit secret foo`.

Now that I think of it, configMaps also aren't b64 encoded, and I've seen caBundles dumped in configMaps without issue before so don't think my original concern was valid. Marking this as resolved.

mansam: Gotcha, thanks for the explanation!

jortel: The _custom_ in the name seems redundant.  Isn't this a custom bundle by definition?  Could it be anything else?  Seems `S3CABundler` would have been sufficient.  Along the same lines, is there any reason to only scope it to S3?

mansam: I borrowed the naming convention from the aws-sdk, it's `CustomCABundle` there, but just `CABundle` would probably be better. If we anticipate this being used in other storage backends, then there's no reason to scope it to S3.

sseago: "there's no reason to scope it to S3" -- hmm, on this point, it seems that the convention we're following is that everything is scoped to the provider.

djwhatle:
@mansam @jortel @dymurray do you guys think it would be useful to add support for skipping verification of S3 caBundle? 

Have been chatting with @pranavgaikwad and think it would save some hassle if someone doesn't care to complete this step. I don't think the technique is well documented but found this issue on the aws-sdk for golang outlining the approach:

https://github.com/aws/aws-sdk-go/issues/2404#issuecomment-454196045

djwhatle:
In the meantime, anyone trying to figure out the correct caBundle to use with their custom S3 service running on OpenShift (e.g. NooBaa) should be able to use this script to grab the value from their NooBaa URL:

```bash
#!/bin/bash
# grab_cabundle.sh
openssl s_client -showcerts -verify 5 -connect $1 < /dev/null | awk '/BEGIN/,/END/{ if(/BEGIN/){a++}; out="cert"a".crt"; print >out}' && for cert in *.crt; do newname=$(openssl x509 -noout -subject -in $cert | sed -n 's/^.*CN=\(.*\)$/\1/; s/[ ,.*]/_/g; s/__/_/g; s/^_//g;p').pem; mv $cert $newname; done
​
cat *.pem > caBundle.pem
```

Then dump caBundle.pem into the field on MigStorage.

mansam:
Just a note that until the Velero side supports this (which I am presently working on), the MigStorage validation will pass because the S3Test will succeed, but it will fail when Velero attempts to reach the bucket.

PR #357, opened 2019-10-30
Adds support for Azure snapshots

sseago:
1) Adds Spec.AzureResourceGroup to migcluster, required to be set
   on both src and dest when using Azure snapshots
2) Refactors the migstorage/cloudprovider code to allow for
   provider-specific fields from migcluster to be added to
   the cloudprovider before generating the velero secret
3) Adds logic to the azure UpdateCloudSecret method to pull
   cluster.Spec.AzureResource group, if present, into the velero
   secret for the cluster.

### DIFF
--- a/pkg/cloudprovider/azure.go
+++ b/pkg/cloudprovider/azure.go
@@ -22,11 +22,12 @@ import (
 const (
 	AzureCredentials = "azure-credentials"
 
-	tenantIDKey       = "AZURE_TENANT_ID"
-	subscriptionIDKey = "AZURE_SUBSCRIPTION_ID"
-	clientIDKey       = "AZURE_CLIENT_ID"
-	clientSecretKey   = "AZURE_CLIENT_SECRET"
-	cloudNameKey      = "AZURE_CLOUD_NAME"
+	tenantIDKey             = "AZURE_TENANT_ID"
+	subscriptionIDKey       = "AZURE_SUBSCRIPTION_ID"
+	clientIDKey             = "AZURE_CLIENT_ID"
+	clientSecretKey         = "AZURE_CLIENT_SECRET"
+	cloudNameKey            = "AZURE_CLOUD_NAME"
+	clusterResourceGroupKey = "AZURE_RESOURCE_GROUP"
 )
 
 // Registry Credentials Secret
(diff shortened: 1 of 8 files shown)


pkg/controller/migplan/storage.go
```diff
@@ -193,24 +216,84 @@ func (r PlanStorage) ensureCloudSecret() error {
...
-	if cluster != nil {
-		list = append(list, *cluster)
+
+	return secret, nil
+}
+
+// Update backup cloud-secret.
+func (r *PlanStorage) UpdateBSLCloudSecret(cloudSecret *kapi.Secret) error {
+	secret, err := r.storage.GetBackupStorageCredSecret(r.Client)
+	if err != nil {
```
jortel: Good job adding the missing `log.Trace(err)`.  Missed one here.

pkg/controller/migplan/storage.go
```diff
@@ -193,24 +216,84 @@ func (r PlanStorage) ensureCloudSecret() error {
...
-	if cluster != nil {
-		list = append(list, *cluster)
+
+	return secret, nil
+}
+
+// Update snapshot cloud-secret.
+func (r *PlanStorage) UpdateVSLCloudSecret(cloudSecret *kapi.Secret) error {
+	secret, err := r.storage.GetVolumeSnapshotCredSecret(r.Client)
+	if err != nil {
```
jortel: Add `log.Trace(err)`.

pkg/cloudprovider/azure.go
```diff
@@ -66,10 +68,23 @@ func (p *AzureProvider) UpdateVSL(vsl *velero.VolumeSnapshotLocation) {
 	}
 }
 
-func (p *AzureProvider) UpdateCloudSecret(secret, cloudSecret *kapi.Secret) {
+func (p *AzureProvider) UpdateCloudSecret(secret, cloudSecret *kapi.Secret) error {
+	// TODO: Inject AZURE_RESOURCE_GROUP = p.ClusterResourceGroup
```
jortel: The TODO can be removed, right?

sseago:
Updated based on feedback above and to fix a bug with generating the VSL secret introduced by the code changes. The one thing left to do is to test this PR with AWS migration to ensure that there are no regressions introduced here.

sseago:
AWS snapshots regression test succeeded, so I'm merging this PR now.

PR #362, opened 2019-11-06
Strip invalid chars from migplan name when creating registry resources

sseago:
Creating a migplan with a dot in the name was causing failures with
the migration registry creation. This commit strips the problematic
characters from the migplan name when generating names for the
registry resources.

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"regexp"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -211,10 +212,14 @@ const (
 func (r *MigPlan) BuildRegistrySecret(client k8sclient.Client, storage *MigStorage) (*kapi.Secret, error) {
 	labels := r.GetCorrelationLabels()
 	labels[MigrationRegistryLabel] = string(r.UID)
+	strippedName, err1 := toDnsLabel(r.GetName())
```
dymurray: NIT: Why use `err1`? Could just use `err` and reuse the var down below for `UpdateRegistrySecret`.

sseago: This was a workaround for an unexplained error that the compiler was giving me. I hadn't even noticed the redeclaration below because the compiler was complaining about something unrelated instead of this. I've updated the PR, as changing the second declaration removed the problem.

PR #363, opened 2019-11-06
Add s3CustomCABundle to migstorage crd schema

mansam:
I missed this while working on https://github.com/fusor/mig-controller/pull/354, it should have been included there.


jortel:
The CRD will need to be updated in the operator as well.

PR #373, opened 2019-11-20
separate cred secrets for azure/gcp/aws

sseago:
This allows for aws, gcp, and azure cred secrets to coexist on the same velero instance. Only one set of credentials per provider is still a limitation, but this allows the use of multiple providers at once for a given controller instance.

This PR requires a (forthcoming) mig-operator PR for azure/gcp use, since without it velero won't look in the right location for (or mount) those provider secrets.


sseago:
Required mig-operator PR: https://github.com/fusor/mig-operator/pull/166

PR #375, opened 2019-11-21
Test NFS accessibility.

jortel:
Issue: https://github.com/fusor/mig-controller/issues/214

Test NFS server accessibility on the destination cluster by performing a connection check on port 2049.
When the controller is running on the destination, the test is performed directly by the controller.  Else, the test is performed by running (pod exec) the `nc` command.  For convenience, one of the velero pods is used.  This requires: https://github.com/fusor/velero/pull/39.  if not suitable pods can be found, an `Error` condition is set.

When an NFS server cannot be accessed, the `move` action is removed from the PV supported actions list and a `Warn` condition is set.

### DIFF
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -3,18 +3,21 @@ package migplan
 import (
 	"context"
 	"fmt"
-	"reflect"
-	"strings"
-
 	migapi "github.com/fusor/mig-controller/pkg/apis/migration/v1alpha1"
 	"github.com/fusor/mig-controller/pkg/health"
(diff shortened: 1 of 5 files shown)


pkg/controller/migplan/validation.go
```diff
@@ -819,3 +817,204 @@ func (r ReconcileMigPlan) validatePods(plan *migapi.MigPlan) error {
...
+			Category: Warn,
+			Message:  NfsNotAccessibleMessage,
+			Items:    notAccessible,
+		})
+	}
+
+	return nil
+}
+
+// Find a pod suitable for command execution.
```
dymurray: This is a neat way to check.

PR #379, opened 2019-12-05
Add Prometheus metrics to track high level aggregate progress of migrations

djwhatle:
## Purpose
Instruments mig-controller with a prometheus metrics endpoint on http://localhost:2112/metrics. 

Initial implementation is intended to provide something to prove out the rest of the metrics plumbing with, and when eventually paired with Grafana dashboards will give us graphical insight on how migrations are progressing for mig-controller.

Cardinality = 8 for the `mig_migrations` metric I've added so far, which should be in bounds for the guidelines set by OpenShift.

## Preview of metrics available
```
# These metrics were pulled from a mig-controller instance where there was 
# just a single "final" MigMigration sitting idle, not yet in ready condition.

$ curl http://localhost:2112/metrics | grep mig_migration

# HELP mig_migrations Count of MigMigrations sorted by status and type
# TYPE mig_migrations gauge
mig_migrations{status="completed",type="final"} 0
mig_migrations{status="completed",type="stage"} 0
mig_migrations{status="error",type="final"} 0
mig_migrations{status="error",type="stage"} 0
mig_migrations{status="idle",type="final"} 1
mig_migrations{status="idle",type="stage"} 0
mig_migrations{status="running",type="final"} 0
mig_migrations{status="running",type="stage"} 0

```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -478,6 +478,7 @@
   packages = [
     "prometheus",
     "prometheus/internal",
+    "prometheus/promauto",
     "prometheus/promhttp",
   ]
   pruneopts = "T"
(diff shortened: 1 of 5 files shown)


sseago:
@jortel It looks like those prometheus files are already under vendor, probably because they're needed by other things we have in vendor.

smarterclayton:
These metrics names are really really opaque.  I would suggest spending a few characters and making them a bit more expressive - like `cluster_migration_records` (more of an example, not meant as a literal name suggestion)

PR #380, opened 2019-12-13
Discovery service.

jortel:
Implements the [Discovery Service](https://docs.google.com/document/d/1BWlSlsrV_uzjLyFVkoYmjHaNHs-F__exMAh8yTVqAeg/edit#heading=h.vlywthintc4j).

Built as an additional controller in the `migration-controller` (manager) using `sqlite` and `Gin`. The new controller watches `MigCluster`.  When a cluster created, it discovers resources and stores them in the DB.  It creates watches on those resources to keep up with real-time changes.

cam-discovery

Design Concepts (abstractions):
- `Container` - The _container_, contains and manages a set of _DataSources_.  When a `MigCluster` is created, updated or deleted, a corresponding data-source is created, updated or deleted.
- `DataSource` - A _data-source_ is responsible for managing the inventory for a cluster. It has a corresponding _controller-runtime_ `Manager` that is used for remote watches.  Each data-source contains and manages a set of `Collections`.
- `Collection` - A _collection_ corresponds loosely to a k8s resource `Kind`. It performs the actual discovery and maintians the inventory of that kind in the DB.  An initial discovery (reconcile) is performed upon creation.  Then, it registers watches with its _data-source_.  Collections implement the _Predicate_ interface and apply real-time changes to the DB.
- `WebServer` - The web server provides a REST API.  Each request is delegated to a `RequestHandler`.
- `RequestHandler` - Each handler supports GET requests for s specific REST resource.
- `Model` - The _model_ is a very lightweight abstraction of the DB.

**API**

Concepts:

_Pagination_ - All GET requests support pagination by passing the offset and limit parameters.

_Availability_ - To protected against returning partial datasets, each collection must be initially/completely reconciled prior to REST API fulling a GET request on the collection. Returns a 206 (Partial) when not yet available.

_Eventual Consistency_ - After a collection is initially reconciled (available), remote watches are used to add/update/delete individual resources.  This has been tuned to be very fast but the kubernetes informer subsystem has some inherent latency. As a result, a GET on a collection will represent a snapshot in time but will eventually become consistent with the inventory on the cluster.

Standard parameters:
- `offset` - The offset of the first item to be included (default: 0).
- `limit` - The number of items returned (default: unlimited). 

Headers:

- `Authorization` - Contains a bearer token.

**Try Me:**

Edit MigrationController (migration-controller):
```
mig_controller_image: quay.io/jortel/mig-controller
mig_controller_version: discovery
```

Edit configmap (migration-controller):
```
CORS_ALLOWED_ORIGINS: “”   ← list of origin regex separated by space.
```

Create the following resources:
```
---
apiVersion: v1
kind: Service
metadata:
  labels:
  name: discovery
spec:
  ports:
  - port: 80
    protocol: TCP
    targetPort: 8080
  selector:
    control-plane: controller-manager
---
apiVersion: v1
kind: Route
metadata:
  name: discovery
spec:
  to:
    kind: Service
    name: discovery
  tls:
    termination: edge
    insecureEdgeTerminationPolicy: Redirect
```
Restart the migration controller.


pkg/controller/discovery/web/web.go
```diff
@@ -0,0 +1,243 @@
...
+type WebServer struct {
+	allowedOrigins []*regexp.Regexp
+	*container.Container
+}
+
+func (w *WebServer) Start() {
+	r := gin.Default()
+	r.Use(cors.New(cors.Config{
+		AllowMethods:     []string{"GET"},
+		AllowHeaders:     []string{"Origin"},
```
Danil-Grigorev: For now, while the authorization is still in progress, to allow [UI client](https://github.com/fusor/mig-ui/pull/640) to work, this line should say
```golang
AllowHeaders:     []string{"Origin", "Authorization"},
```
to allow passing authentication token, otherwise server response returns error, as the `Authorization` field is not allowed in `OPTIONS` header, which identifies itself as a CORS issue.

jortel: Good catch!  Fixed.

pkg/controller/discovery/web/pv.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"encoding/json"
+	"github.com/fusor/mig-controller/pkg/controller/discovery/model"
+	"github.com/gin-gonic/gin"
+	"k8s.io/api/core/v1"
+	"net/http"
+)
+
+const (
+	PvsRoot = NamespaceRoot + "/persistentvolumes"
```
Danil-Grigorev: Looks like a typo here, it should be 
```golang
PvsRoot = ClusterRoot + "/persistentvolumes"
```

pkg/controller/discovery/web/plan.go
```diff
@@ -0,0 +1,257 @@
...
+	Source      *v1.ObjectReference
+	Destination *v1.ObjectReference
+}
+
+//
+// PlanPods REST resource.
+type PlanPods struct {
+	Namespace   string
+	Name        string
+	Controller  Pod
```
Danil-Grigorev: Would be simpler to have it as a list of pods here too.

jortel: How so?  There should only be 1 controller pod.  Right?

pkg/controller/discovery/web/pv.go
jortel: Oh, right.  I was wondering why the path was not what I expected when typing it :)

Danil-Grigorev: Thanks :)

pkg/controller/discovery/web/plan.go
Danil-Grigorev: Just in terms of generic implementation of the UI client, it is simpler to treat it as a list of log sources. But you are right, for now it is 1 controller pod, unless something will change (re: https://docs.google.com/document/d/1632Iw2DCnpMe7zMhL3YOmDPI9EPuEA5474hgTafxqrc/edit?disco=AAAADjiS2bE)

jortel: Easy enough. will do.

pkg/controller/discovery/web/pv.go
jortel: Fixed.

pkg/controller/discovery/web/plan.go
jortel: Done.

Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY cmd/    cmd/
 COPY vendor/ vendor/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux go build -a -o manager github.com/fusor/mig-controller/cmd/manager
+RUN CGO_ENABLED=1 GOOS=linux go build -a -o manager github.com/fusor/mig-controller/cmd/manager
```
jortel: Needed by dependencies.

pkg/controller/discovery/container/pod.go
```diff
@@ -0,0 +1,171 @@
...
+		return false
+	}
+	pod := model.Pod{
+		Base: model.Base{
+			Cluster: r.ds.cluster.PK,
+		},
+	}
+	pod.With(object)
+	pod.SetPk()
+	tx, _ := r.ds.Container.Db.Begin()
```
dymurray: Is it safe to skip the error check here?

pkg/controller/discovery/container/pv.go
```diff
@@ -0,0 +1,172 @@
...
+		return false
+	}
+	pv := model.PV{
+		Base: model.Base{
+			Cluster: r.ds.cluster.PK,
+		},
+	}
+	pv.With(object)
+	pv.SetPk()
+	tx, _ := r.ds.Container.Db.Begin()
```
dymurray: Same as above

pkg/controller/discovery/container/pv.go
```diff
@@ -0,0 +1,172 @@
...
```
dymurray: same

pkg/controller/discovery/container/pv.go
```diff
@@ -0,0 +1,172 @@
...
```
dymurray: same

pkg/controller/discovery/container/pod.go
jortel: Fixed.
Updated model methods to accept the DB as an _interface_ and removed all but (1) explicit use of transaction.  Handling the error in the remaining place.  Also added checking for tx.Commit() errors.

PR #385, opened 2019-12-18
Adjust migration metric naming

djwhatle:
Follow best practice of including unit on metrics per https://prometheus.io/docs/practices/naming/#metric-names


djwhatle:
Requested review on metrics naming from @LiliC and @smarterclayton, arrived at `cam_app_workload_migration_total` for this metric name. 

All future metrics for this project should be prefixed with `cam_*` and have some type of informative unit postfixed per the doc linked above.

PR #398, opened 2020-01-13
Suppoort separate container for cam and discovery.

jortel:
Support running the discovery controller in a separate container.

The _migration_ manager will load controllers based on a `ROLE` environment variable. The two supported roles are:
- `cam`  - Has CAM (plan, cluster, storage, migration) controllers.
- `discovery` - Has discovery controller.

The migration controller pod will contain (2) containers, one for each role.

To support developers running (1) combined manager (controller) locally, the `ROLE` supports a comma delimited list of roles.  Eg: `ROLE=cam,discovery` which is also the default when `ROLE` is not defined.

---

This required the discovery controller to fully support Pods with multiple containers. The `Pod` REST resource changed:

from:
```
{
  name: cam
  log: /namespaces/..../log
}
```
to:
```
{
  name: cam
  containers: [
    {
      name: cam
      log:  /namespaces/..../log/?container=cam
    },
  ]
}
```

---

**REQUIRED** Operator changes:  https://github.com/fusor/mig-operator/pull/211.

---

**REQUIRED** The UI will need to iterate over the containers when processing the data. The _container_ does not need to be displayed to the user for now. Instead, the logs for each container could be aggregated for each pod: https://github.com/fusor/mig-ui/pull/651


pkg/controller/discovery/web/plan.go
```diff
@@ -280,30 +279,11 @@ func (p *PlanPods) buildPods(h *PlanHandler, ref *v1.ObjectReference) ([]Pod, er
 			return nil, err
 		}
 		for _, m := range list {
```
Danil-Grigorev: This could be more expressive in my opinion
```golang
for _, model := range podModels {
```

pkg/controller/discovery/model/pv.go
```diff
@@ -103,10 +103,24 @@ func (m *PV) With(object *v1.PersistentVolume) {
...
+// Encode definition
+func (m *PV) EncodeDefinition(pv *v1.PersistentVolume) {
+	definition, _ := json.Marshal(pv)
 	m.Definition = string(definition)
 }
 
+//
+// Encode definition
+func (m *PV) DecodeDefinition() *v1.PersistentVolume {
+	pod := &v1.PersistentVolume{}
```
Danil-Grigorev: This is a `pv` :-)

pkg/controller/controller.go
```diff
@@ -17,24 +17,63 @@ limitations under the License.
...
-		if err := f(m); err != nil {
+	load := func(functions []AddFunction) error {
+		for _, f := range functions {
+			if err := f(m); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+	if settings.Settings.HasRole(settings.CamRole) {
```
Danil-Grigorev: Appreciate this `ROLE` based approach for choosing which controllers to load in the manager

pkg/controller/discovery/web/plan.go
```diff
@@ -280,30 +279,11 @@ func (p *PlanPods) buildPods(h *PlanHandler, ref *v1.ObjectReference) ([]Pod, er
...
 			pods = append(pods, pod)
 		}
 	}
 
 	return pods, nil
 }
-
-//
-// Build the URI (path) for fetching logs.
-func LogPath(cluster *model.Cluster, pod *model.Pod) string {
```
Danil-Grigorev: :+1:

pkg/controller/discovery/model/pv.go
jortel: Fixed.

PR #399, opened 2020-01-15
Fix race condition in DataSource startup.

jortel:
Fixes: https://github.com/fusor/mig-controller/issues/400

The `DataSource` startup is designed to provide a baseline snapshot which is then updated with incremental changes based on _watch_.  The initial snapshot ensures a minimum sane result for REST requests processed immediately after startup or a `MigCluster` is added.

The steps had been:
1. Full Reconcile()
2. Start remote watches.

However, the time between 1 & 2 provided window in which resource add/update/delete would be missed because watches had not yet been setup.

This PR changes the steps to:
1. Start remote watches (events are queued).
2. Full Reconcile()
3. Begin processing queued watch events.

The overlap is safe because the Primary Key for an object stored in DB includes the k8s UID. 
An insert has no affect when the object already exists.
A delete of an object that does not exist has no affect.
Object updates will be applied in order and last update wins.

### DIFF
--- a/pkg/controller/discovery/container/collection.go
+++ b/pkg/controller/discovery/container/collection.go
@@ -28,6 +28,8 @@ type Collection interface {
 	// Associate with a DataSource.
 	// Mainly to support two phase construction.
 	Bind(ds *DataSource)
+	// Get the associated DataSource.
+	GetDs() *DataSource
 	// Add k8s watches.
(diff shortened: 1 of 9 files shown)


jortel:
Channels consolidated.

jortel:
Optimization added.

PR #403, opened 2020-01-20
Fixes to edge-case issues.

jortel:
Few important fixes needed in case more than 1 MigCluster references the `host`.

---

The logger is global and needed a mutex to protect the `history`.

The sqlite3 DB driver does not support concurrent writes.  Added mutex and used appropriately.

### DIFF
--- a/pkg/controller/discovery/container/collection.go
+++ b/pkg/controller/discovery/container/collection.go
@@ -75,7 +75,6 @@ func (r *BaseCollection) IsReady() bool {
 // Reset `hasReconciled` and association with a DataSource.
 func (r *BaseCollection) Reset() {
 	r.hasReconciled = false
-	r.ds = nil
 }
 
 //
@@ -132,6 +131,8 @@ func (r *SimpleReconciler) Reconcile(collection Collection) error {
 			dpn.discovered = m
 		}
 	}
+	model.Mutex.RLock()
+	defer model.Mutex.RUnlock()
 	tx, err := r.Db.Begin()
 	if err != nil {
 		Log.Trace(err)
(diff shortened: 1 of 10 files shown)


pkg/controller/discovery/container/collection.go
```diff
@@ -75,7 +75,6 @@ func (r *BaseCollection) IsReady() bool {
 // Reset `hasReconciled` and association with a DataSource.
 func (r *BaseCollection) Reset() {
 	r.hasReconciled = false
-	r.ds = nil
```
jortel: Predicates may still be using this until the controller (watches) are completely terminated.  Has caused SEGV.
No value in setting to `nil`.

pkg/controller/discovery/container/ds.go
```diff
@@ -209,7 +210,11 @@ func (r *DataSource) buildClient(cluster *migapi.MigCluster) error {
 		Log.Trace(err)
 		return err
 	}
-	r.Client, err = cluster.GetClient(r.Container.Client)
+	r.Client, err = client.New(
```
jortel: Each `DataSource` needs it's own _client_.  The MigCluster will return the same client when IsHostCluster=true.  I have seen this cause problems when a client is used concurrently.

pkg/controller/discovery/controller.go
```diff
@@ -36,19 +36,23 @@ import (
...
 
 // Application settings.
 var Settings = &settings.Settings
 
+func init() {
+	log = logging.WithName("discovery")
+	model.Log = &log
+	container.Log = &log
+	web.Log = &log
+}
```
jortel: More appropriate to use package initialization to ensure the `Log` variables are set before they're used.

pkg/controller/discovery/model/cluster.go
```diff
@@ -278,6 +278,7 @@ func (m *Cluster) NsList(db DB, page *Page) ([]*Namespace, error) {
 			&ns.Cluster)
 		if err != nil {
 			Log.Trace(err)
+			return nil, err
```
jortel: Should return the error here.

(same below)

pkg/controller/discovery/container/collection.go
sseago: I think I just hit this problem in my env, actually.

PR #404, opened 2020-01-21
fix typo/SQL syntax error

sseago:


pkg/controller/discovery/model/pv.go
```diff
@@ -84,7 +84,7 @@ WHERE  pk = :pk;
 var PvDeleteSQL = `
 DELETE FROM pv
 WHERE
-  pk = ;pk;
+  pk = :pk;
```
jortel: Looks like this was not caught by unit tests because the PV.Delete() is missing.  Would you mind adding before merge?

sseago: Done.

PR #405, opened 2020-01-21
Embedding requeue logic for connection testing in migCluster

Danil-Grigorev:
- Added on Generation change predicate for remote watches on pods and secrets


djwhatle:
@Danil-Grigorev would you please write an issue describing the problem that is being solved with this PR?

djwhatle:
Closing this for now. Let's re-open if there's an associated issue that we need to tackle.

PR #411, opened 2020-01-28
make migration registry image configurable

sseago:
If the MIGRATION_REGISTRY_IMAGE environment variable is set on the
controller pod, this image will be used for launching the migration
registry. If it is not set, the "registry:2" default will be used instead.

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"reflect"
 	"regexp"
 	"sort"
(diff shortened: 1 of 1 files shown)


pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -207,9 +208,20 @@ func (r *MigPlan) ListMigrations(client k8sclient.Client) ([]*MigMigration, erro
...
 const (
-	MigrationRegistryLabel = "migration-registry"
+	MigrationRegistryLabel        = "migration-registry"
+	MigrationRegistryDefaultImage = "registry:2"
+	MigrationRegistryImageEnvVar  = "MIGRATION_REGISTRY_IMAGE"
 )
 
+// Return the image reference for the migration registry (defaults to "registry:2")
+func migRegistryImageRef() string {
+	envImage := os.Getenv(MigrationRegistryImageEnvVar)
```
jortel: It would be better to be consistent with other environment variables and put this logic in the `settings` package.  That said, we can merge as-is and make this change in 1.2. or 1.1.z.

dymurray:
With Scott on PTO I think we should merge this and then submit a follow up PR ourselves to fix this by putting it in the `settings` pkg I completely agree.

PR #412, opened 2020-02-03
Replace the Touch ID with observedGeneration.

### LINKED ISSUE #401: Handle stale CR data more consistently
Currently there are a few issues that have come up in the UI where CR spec fields are updated and the UI then pulls the resource before the next reconcile loop is able to respond. This could result in stale conditions on the Status -- warnings that no longer apply, ready migplans that no longer meet ready status, and PV discovery not done after an update to plan namespaces.

We need some way for the UI (and other API clients) to be able to determine whether the state of a resource has gone through a reconcile update since the client last updated the resource.

### LINKED ISSUE #410: Add validations with EBS move to catch potential issues
@sseago please help to flesh out the validations we should consider adding:
 - Same AWS account on src/dst
 - Same Availability Zone?
 - others ....

jortel:
closes #401

Replaces the `touch` annotation with `Status.ObservedGeneration`.

The Touch ID provided:
- Ensured that the `resourceVersion` and `generation` were incremented (when updated) on OCP4 even when the resource (spec or status) did not change. This was essential for watch event propagation.
- Provided a means by which watch predicates (for referenced MIG resources) could ignore update events for un-reconciled resources.

The `observedGeneration` design pattern is used by _Kubernetes_ controllers (such as DaemonSet controller) and provides more information. The `observedGeneration` indicates the last generation reconciled by the controller. This provides a means to determine if a specific generation of interest has been reconciled. When reconciles are _caught up_ to resource updates, the `ObservedGeneration` will equal the `Generation`.

### DIFF
--- a/pkg/apis/migration/v1alpha1/resource.go
+++ b/pkg/apis/migration/v1alpha1/resource.go
@@ -1,19 +1,10 @@
 package v1alpha1
 
-import (
-	"github.com/google/uuid"
-)
-
 const (
-	TouchAnnotation = "touch"
 	VeleroNamespace = "openshift-migration"
 )
 
 // Migration application CR.
-// The `touch` UUID is set by the controller after reconcile
-// and needed to support relay of remote watch events. Also to
-// provide remote watch predicates the ability to determine
-// when an updated referenced resource has been fully reconciled.
 type MigResource interface {
 	// Get a map containing the correlation label.
 	GetCorrelationLabels() map[string]string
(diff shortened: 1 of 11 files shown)


pkg/apis/migration/v1alpha1/resource.go
```diff
@@ -50,23 +41,12 @@ func (r *MigPlan) GetName() string {
...
-		return ""
-	}
-	id, found := r.Annotations[TouchAnnotation]
-	if found {
-		return id
-	}
-
-	return ""
+func (r *MigPlan) MarkReconciled() {
+	r.Status.ObservedGeneration = r.Generation + 1
```
djwhatle: This seems to say that we've "observed" as generation past the current one, which doesn't make sense to me. I'm probably missing something, @jortel can you explain?

djwhatle: If we write the resource back to the APIserver and generation gets incremented at that point to match the ObservedVersion, then this makes sense to me.

PR #413, opened 2020-02-11
GVR compare on plan reconcile

Danil-Grigorev:
Allows to compare supported GVRs between clusters, and provides per-namespace report on resources, which will need APIGroup conversion or manual exclusion from the current migration run. Those resources will not be restored on the destination cluster without manual conversion, so the migration will not be fully successful.

Sample output:
```yaml
  observedGeneration: 16674
  unsupportedNamespaces:
  - name: phronetic
    unsupportedResources:
    - batch/v2alpha1, Resource=cronjobs
    - batch/v2alpha1, Resource=scheduledjobs
```

Output from OCP4.4 (kube 16+)
```yaml
    incompatibleNamespaces:
  - gvrs:
    - group: extensions
      resoruce: daemonsets
      version: v1beta1
    - group: extensions
      resoruce: deployments
      version: v1beta1
    - group: extensions
      resoruce: replicasets
      version: v1beta1
    - group: apps
      resoruce: controllerrevisions
      version: v1beta1
    - group: apps
      resoruce: deployments
      version: v1beta1
    name: openshift-migration
  - gvrs:
    - group: extensions
      resoruce: deployments
      version: v1beta1
    - group: extensions
      resoruce: replicasets
      version: v1beta1
    - group: apps
      resoruce: deployments
      version: v1beta1
    name: nginx-deployment
  - gvrs:
    - group: batch
      resoruce: cronjobs
      version: v2alpha1
    - group: batch
      resoruce: scheduledjobs
      version: v2alpha1
    name: cron

```

Warning message:
```yaml
  - category: Warn
    lastTransitionTime: 2020-02-17T12:55:52Z
    message: 'Some namespaces are unsupported. See: `unsupportedNamespaces` for
      details.'
    status: "True"
    type: NotSupported
```

Demo: https://drive.google.com/open?id=1J7_pA5sRSpnuR17JiGiLVaVfT8iNIPZW


pkg/gvkcompare/gvkcompare.go
```diff
@@ -0,0 +1,276 @@
...
+func (r *GVKCompare) Compare() error {
+	gvDiff, err := r.compareGroupVersions()
+	if err != nil {
+		return errors.Wrap(err, "Unable to compare GroupVersions between clusters")
+	}
+
+	unsupportedGVRs, err := r.unsupportedServerResources(gvDiff)
+	if err != nil {
+		return errors.Wrap(err, "Unable to get unsupported resources for scrCluster")
+	}
```
Danil-Grigorev: Both methods will reduce the number of resources, which will be listed in each `migPlan` namespace to the minimum amount, allowing to inspect every namespace in details by listing for unsupported resources in it and checking for their presence.

pkg/gvkcompare/gvkcompare.go
```diff
@@ -0,0 +1,276 @@
...
+
+	missingGroups := missingGroups(srcGroupList.Groups, dstGroupList.Groups)
+	matchGroups(srcGroupList, dstGroupList, missingGroups)
+	missingVersions := missingVersions(srcGroupList.Groups, dstGroupList.Groups)
+
+	return append(missingGroups, missingVersions...), nil
+}
+
+func (r *GVKCompare) collectNamespaceReport(unsupportedResources []schema.GroupVersionResource) error {
```
Danil-Grigorev: Collects report for `MigPlan` namespaces only for unsupported resources.

pkg/gvkcompare/gvkcompare.go
```diff
@@ -0,0 +1,276 @@
...
+	for _, namespace := range r.Plan.GetSourceNamespaces() {
+		unsupportedGVRs := []string{}
+		for _, gvr := range unsupportedResources {
+			options := metav1.ListOptions{}
+			resourceList, err := r.SrcClient.Resource(gvr).Namespace(namespace).List(options)
```
Danil-Grigorev: Dynamic client was the simplest option to allow generic listing of unsupported resources on the source cluster, as it does not require typecasting, and encapsulates constructing of `*Unstructured` object.

pkg/gvkcompare/gvkcompare.go
```diff
@@ -0,0 +1,276 @@
...
+	unsupportedGVRs := []schema.GroupVersionResource{}
+	for _, gr := range gvDiff {
+		for _, version := range gr.Versions {
+			r, err := r.SrcDiscovery.ServerResourcesForGroupVersion(version.GroupVersion)
+			if err != nil {
+				return nil, errors.Wrap(err, "Unable to get a list of resources for a GroupVersion on srcCluster")
+			}
+
+			r.APIResources = namespaced(r.APIResources)
+			r.APIResources = excludeSubresources(r.APIResources)
```
Danil-Grigorev: Removes unnecessary [`resource/status`, `resource/logs`...] api paths.

pkg/gvkcompare/gvkcompare.go
```diff
@@ -0,0 +1,276 @@
...
+			reducedDst = append(reducedDst, group)
+		}
+	}
+
+	sort.Slice(reducedSrc, func(i int, j int) bool {
+		return reducedSrc[i].Name < reducedSrc[j].Name
+	})
+	sort.Slice(reducedDst, func(i int, j int) bool {
+		return reducedDst[i].Name < reducedDst[j].Name
+	})
```
Danil-Grigorev: Ensure the order and number of `APIGroups` for both cluster.

config/crds/migration_v1alpha1_migcluster.yaml
```diff
@@ -58,6 +58,12 @@ spec:
           - isHostCluster
           type: object
         status:
+          properties:
+            observedGeneration:
```
djwhatle: This and other CRD changes are just a leftover from artifacts not being checked in on https://github.com/fusor/mig-controller/pull/412/files

Danil-Grigorev: Don't mind importing them with the PR, as it still requires a CRD change.

pkg/controller/migplan/gvk.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		log.Error(err, "Failed to compare GVRs on both clusters")
+		return err
+	}
+
+	if len(plan.Status.UnsupportedNamespaces) > 0 {
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     NotSupported,
+			Status:   True,
+			Category: Warn,
```
djwhatle: Note to anyone else reviewing, this condition is a `Warn` at present. 
cc @jortel

config/crds/migration_v1alpha1_migcluster.yaml
jortel: Ah, looks like I forget the `make manfests` on pull/412.  Let's fix this in a separate PR and then rebase.  The reason for not combining is that if this PR got reverted, we would not want the CRD change to be reverted.

jortel: Fixed on master: https://github.com/fusor/mig-controller/pull/421

pkg/apis/migration/v1alpha1/gvk.go
```diff
@@ -0,0 +1,29 @@
+package v1alpha1
+
+// UnsupportedNamespace - namespace, which is noticed
+// to contain resources unsupported by the migration
+type UnsupportedNamespace struct {
+	Name                 string   `json:"name"`
+	UnsupportedResources []string `json:"unsupportedResources"`
```
jortel: This should just be `Resources` within the scope of `UnsupportedNamespaces`.

pkg/apis/migration/v1alpha1/gvk.go
```diff
@@ -0,0 +1,29 @@
...
+func (p *MigPlanStatus) SelectUnsupported() []string {
+	namespaces := []string{}
+	for _, ns := range p.UnsupportedNamespaces {
+		namespaces = append(namespaces, ns.Name)
+	}
+
+	return namespaces
+}
+
+func (p *MigPlanStatus) InitUnsupported() {
```
jortel: :+1: for functional decomposition (generally) but don't think this function adds value.

pkg/apis/migration/v1alpha1/gvk.go
```diff
@@ -0,0 +1,29 @@
...
+	p.UnsupportedNamespaces = []UnsupportedNamespace{}
+}
+
+func (p *MigPlanStatus) AppendUnsupported(namespace string, resources []string) {
```
jortel: Perhaps add:
```
if p.UnsupportedNamespaces == nil {
  p.UnsupportedNamespaces = []UnsupportedNamespace{}
}
```
to avoid the need for 2 step construction.

pkg/cloudprovider/aws.go
```diff
@@ -187,16 +187,16 @@ func (p *AWSProvider) UpdateRegistryDC(dc *appsv1.DeploymentConfig, name, dirNam
 		dc.Spec.Template.Spec.Containers[0].VolumeMounts = append(
 			dc.Spec.Template.Spec.Containers[0].VolumeMounts,
 			kapi.VolumeMount{
```
jortel: Let's fix this unrelated formatting in a separate PR and rebase.

pkg/apis/migration/v1alpha1/gvk.go
jortel: Rather than appending the slice in `MigPlanStatus` you can eliminate some complexity and coupling by building a []UnsupportedNamespace{} then assigning to `MigPlanStatus` already populated.

pkg/controller/migplan/gvk.go
```diff
@@ -0,0 +1,62 @@
...
+	// No spec chage this time
+	if plan.HasReconciled() {
+		return nil
+	}
+
+	plan.Status.InitUnsupported()
+
+	gvkCompare, err := r.prepareGVKCompare(plan)
+	if err != nil {
+		log.Error(err, "Failed to prepare GVK compare")
```
jortel: The convention in the controller is to `log.Trace(err)` instead.  The trace provides all the necessary information (and more).

pkg/controller/migplan/gvk.go
```diff
@@ -0,0 +1,62 @@
...
+		return err
+	}
+
+	err = gvkCompare.Compare()
+	if err != nil {
+		log.Error(err, "Failed to compare GVRs on both clusters")
```
jortel: The convention in the controller is to `log.Trace(err)` instead.  The trace provides all the necessary information (and more).

pkg/controller/migplan/gvk.go
```diff
@@ -0,0 +1,62 @@
...
+	return nil
+}
+
+func (r ReconcileMigPlan) prepareGVKCompare(plan *migapi.MigPlan) (*gvk.GVKCompare, error) {
+	gvkCompare := &gvk.GVKCompare{
+		Plan: plan,
+	}
+
+	err := gvkCompare.PrepareSourceDiscovery(r)
+	if err != nil {
```
jortel: Please add:
```
log.Trace(err)
```

pkg/controller/migplan/gvk.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil, err
+	}
+
+	err = gvkCompare.PrepareDestinationDiscovery(r)
+	if err != nil {
```
jortel: Please add:
```
log.Trace(err)
```

pkg/controller/migplan/gvk.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil, err
+	}
+
+	err = gvkCompare.PrepareSourceClient(r)
+	if err != nil {
```
jortel: Please add:
```
log.Trace(err)
```

pkg/controller/migplan/migplan_controller.go
```diff
@@ -254,6 +254,12 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	err = r.compareGVK(plan)
+	if err != nil {
+		log.Trace(err)
+		return reconcile.Result{Requeue: true}, nil
+	}
+
```
jortel: To be consistent with other validations and to keep `Reconcile()` clean, let's move this into:  `ReconcileMigPlan.validate()`.

pkg/controller/migplan/validation.go
```diff
@@ -94,6 +95,7 @@ const (
 	StorageNotReadyMessage                = "The referenced `migStorageRef` does not have a `Ready` condition."
 	NsListEmptyMessage                    = "The `namespaces` list may not be empty."
 	InvalidDestinationClusterMessage      = "The `srcMigClusterRef` and `dstMigClusterRef` cannot be the same."
+	NsNotSupported                        = "Namespaces [] not supported. See: `unsupportedNamespaces` for details."
```
jortel: Since there is a separate list (which may be large) in the status, let's not list them on the condition as well.

pkg/gvkcompare/gvkcompare.go
```diff
@@ -0,0 +1,276 @@
...
+
+	err = r.collectNamespaceReport(unsupportedGVRs)
+	if err != nil {
+		return errors.Wrap(err, "Unable to evaluate GVR gaps for migrated resources")
+	}
+
+	return nil
+}
+
+func (r *GVKCompare) PrepareSourceDiscovery(c client.Client) error {
```
jortel: Instead of Prepare__() methods, I think the idiomatic way to accomplish complex construction would be to use a `NewComparator() (*Comparator, error)` function.

pkg/gvkcompare/gvkcompare.go
```diff
@@ -0,0 +1,276 @@
+package gvkcompare
```
jortel: Perhaps the package named `gvk` would better support a broader range of tools in the future.
Within the package, suggest a `comparator.go` file would be intuitive.

pkg/gvkcompare/gvkcompare.go
```diff
@@ -0,0 +1,276 @@
...
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+
+	"github.com/pkg/errors"
+	"k8s.io/client-go/discovery"
+)
+
+// GVKCompare reconciles a GVKCompare object
+type GVKCompare struct {
```
jortel: Within the `gvk` package, suggest the type named `Comparator` would be intuitive.  Would be used as: `gvk.Comparator{}`

pkg/gvkcompare/gvkcompare.go
jortel: Actually, another question.  If only used by the plan controller, why is this a top level package instead of a sub-package under the plan controller package?

pkg/gvkcompare/gvkcompare.go
```diff
@@ -0,0 +1,276 @@
...
+
+	missingGroups := missingGroups(srcGroupList.Groups, dstGroupList.Groups)
+	matchGroups(srcGroupList, dstGroupList, missingGroups)
+	missingVersions := missingVersions(srcGroupList.Groups, dstGroupList.Groups)
+
+	return append(missingGroups, missingVersions...), nil
+}
+
+func (r *GVKCompare) collectNamespaceReport(unsupportedResources []schema.GroupVersionResource) error {
+	for _, namespace := range r.Plan.GetSourceNamespaces() {
```
jortel: This appears to make an API call for: (each namespace on the plan) **x** (all GVK combinations identified as missing on the destination).  Right?
It may be more efficient to query the GVK across all namespaces and filter/collate on namespace.

Danil-Grigorev: This appeared to me as an option. It sounds better then per-namespace requests, will update as the initial approach is now functional.

pkg/apis/migration/v1alpha1/gvk.go
Danil-Grigorev: I consider this an initial run for GVK related tasks. This structure may contain additional fields, some specific info to `why` the resource is unsupported, and what needs to be done, which it does not have at the moment. May be the name change to `ResetUnsupported()` will be sufficient?

pkg/apis/migration/v1alpha1/gvk.go
Danil-Grigorev: This will complicate understanding for this field in the status. Either the name, or some sub item should tell the reason for it being there.

pkg/controller/migplan/gvk.go
Danil-Grigorev: Should pick it up: https://github.com/fusor/mig-controller/pull/413/files#diff-ff4cf8d4f8669d3270d95d4c88c43266R17

djwhatle:
@Danil-Grigorev are vendor updates needed for this PR? I'm unable to build it locally.

```
        imports github.com/evanphx/json-patch: cannot find package "github.com/evanphx/json-patch" in any of:
        /home/dwhatley/dev/GO/src/github.com/fusor/mig-controller/vendor/github.com/evanphx/json-patch (vendor tree)
        /usr/local/go/src/github.com/evanphx/json-patch (from $GOROOT)
        /home/dwhatley/dev/GO/src/github.com/evanphx/json-patch (from $GOPATH)
```

djwhatle:
@Danil-Grigorev I'm having trouble getting a successful connection between clusters after I switched to this build of mig-controller. Any special instructions for testing? I installed using non-olm 1.1 manifests on both a 3.7 / 4.2 combo, setting these values on the `MigrationController` CR on both ends.

```
  mig_controller_image: quay.io/dgrigore/mig-controller
  mig_controller_version: latest
```

djwhatle:
Going to take another look at this PR today, will see if I can get it to run successfully and check if I see the observedGeneration increasing too quickly.

djwhatle:
@Danil-Grigorev I'm getting an error starting up remote watches on a current build of this. I guess the "work-in-progress" label means something here :)

```
{"level":"info","ts":1582755048.009608,"logger":"kubebuilder.controller","msg":"Starting workers","controller":"migplan-controller","worker count":1}
E0226 22:10:48.111303       1 runtime.go:69] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
/go/src/github.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:76
/go/src/github.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:65
/go/src/github.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:51
/usr/local/go/src/runtime/asm_amd64.s:573
/usr/local/go/src/runtime/panic.go:502
/usr/local/go/src/runtime/panic.go:63
/usr/local/go/src/runtime/signal_unix.go:388
/go/src/github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1/migcluster_types.go:151
/go/src/github.com/konveyor/mig-controller/pkg/gvk/compare.go:111
/go/src/github.com/konveyor/mig-controller/pkg/gvk/compare.go:66
/go/src/github.com/konveyor/mig-controller/pkg/controller/migplan/gvk.go:38
/go/src/github.com/konveyor/mig-controller/pkg/controller/migplan/gvk.go:16
/go/src/github.com/konveyor/mig-controller/pkg/controller/migplan/validation.go:188
/go/src/github.com/konveyor/mig-controller/pkg/controller/migplan/migplan_controller.go:217
/go/src/github.com/konveyor/mig-controller/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:215
/go/src/github.com/konveyor/mig-controller/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:158
/go/src/github.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133
/go/src/github.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134
/go/src/github.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88
/usr/local/go/src/runtime/asm_amd64.s:2361
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x108 pc=0x149ce47]

```

djwhatle:
@Danil-Grigorev @jortel ran the latest commit of this and had it successfully detect GVK differences across my clusters.

```
    unsupportedNamespaces:
    - name: phronetic
      unsupportedResources:
      - batch/v2alpha1, Resource=cronjobs
      - batch/v2alpha1, Resource=scheduledjobs
```

@Danil-Grigorev did you ever figure out what was up with the inability to detect one of the GVK differences between the 3.x and 4.4 pair we looked at on our review call earlier?

pkg/controller/migmigration/quiesce.go
```diff
@@ -88,25 +102,65 @@ func (t *Task) scaleDownDeploymentConfigs(client k8sclient.Client) error {
...
-func (t *Task) scaleDownDeployments(client k8sclient.Client) error {
+func (t *Task) scaleDownDeployments(client k8sclient.Client, kubeVersion int) error {
+	if kubeVersion >= reference.AppsGap {
+		return t.scaleDownv1Deployments(client)
+	}
+	return t.scaleDownv1betaDeployments(client)
+}
+
+// Scales down all Deployments (kubernetes < 16)
+func (t *Task) scaleDownv1betaDeployments(client k8sclient.Client) error {
```
djwhatle: I have the same question for here. If we are going with the approach Jeff outlined that automatically handles older versions, would these version-specialized methods still be necessary?

pkg/controller/migmigration/quiesce.go
```diff
@@ -182,26 +269,66 @@ const (
...
+				log.Trace(err)
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Scales down all DaemonSets. (kubernetes < 16)
+func (t *Task) scaleDownv1betaDaemonSets(client k8sclient.Client) error {
```
djwhatle: Same question here.

pkg/gvk/compare.go
```diff
@@ -0,0 +1,280 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+
+	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
+)
+
+var crdGVR = schema.GroupVersionResource{
```
djwhatle: In case anyone else reviewing  is wondering: this is used for excluding all CRDs from the comparison list since CRDs can be moved over to a dest cluster already.

pkg/gvk/compare.go
```diff
@@ -0,0 +1,280 @@
...
+
+	dstGroupList, err := r.DstDiscovery.ServerGroups()
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable to fetch server groups for a dstCluster")
+	}
+
+	gvDiff := compareGroupVersions(srcGroupList, dstGroupList)
+	unsupportedGVRs, err := r.unsupportedResources(gvDiff)
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable to get unsupported resources for scrCluster")
```
djwhatle: typo: `scrCluster` -> `srcCluster`

pkg/gvk/compare.go
```diff
@@ -0,0 +1,280 @@
...
+					unsupportedNamespaces[namespace] = []schema.GroupVersionResource{gvr}
+				}
+			}
+		}
+	}
+
+	return unsupportedNamespaces, nil
+}
+
+func (r *Compare) occureIn(gvr schema.GroupVersionResource) ([]string, error) {
```
djwhatle: typo: `occureIn`->`occurIn`

pkg/gvk/compare.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return errors.Wrap(err, "Error while listing CRDs")
+	}
+
+	crdGroups := []string{}
+	groupPath := []string{"spec", "group"}
+	for _, crd := range crdList.Items {
+		group, found, err := unstructured.NestedString(crd.Object, groupPath...)
+		if !found {
+			return errors.Wrap(err, "Error while extracting CRD group: not exist")
```
djwhatle: "not exist" -> "doesn't exist"

pkg/gvk/compare_test.go
```diff
@@ -0,0 +1,216 @@
...
+	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic/fake"
+)
+
+var (
+	sourceGroups = []metav1.APIGroup{
```
djwhatle: @Danil-Grigorev big 👍 for adding tests!

pkg/health/pods.go
```diff
@@ -87,15 +90,34 @@ func PodManagersRecreated(client k8sclient.Client, options *k8sclient.ListOption
 }
 
 // DaemonSetsRecreated checks the number of pod replicas managed by DaemonSets
-func DaemonSetsRecreated(client k8sclient.Client, options *k8sclient.ListOptions) (bool, error) {
-	daemonSetList := v1beta1.DaemonSetList{}
-
-	err := client.List(context.TODO(), options, &daemonSetList)
+func DaemonSetsRecreated(
```
djwhatle: @Danil-Grigorev same question here: is this necessary if we use @jortel change that auto-downgrades versions when necessary?

pkg/gvk/compare_test.go
Danil-Grigorev: Removing those, as unnecessary addition. The data set for the tests is to huge.

pkg/health/pods.go
Danil-Grigorev: It is unrelated to the PR

pkg/gvk/compare.go
Danil-Grigorev: Thanks

pkg/gvk/compare.go
Danil-Grigorev: :+1:

pkg/gvk/compare.go
Danil-Grigorev: :+1:

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -174,23 +188,45 @@ func (m *MigCluster) DeleteResources(client k8sclient.Client, labels map[string]
...
 	if err != nil {
 		return err
 	}
-	for _, r := range dList.Items {
-		err = client.Delete(context.TODO(), &r)
-		if err != nil && !k8serror.IsNotFound(err) {
+	options := k8sclient.MatchingLabels(labels)
+
+	// Deployment
+	if kubeVersion >= reference.AppsGap {
```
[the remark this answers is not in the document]
Danil-Grigorev: It is unrelated to the PR

pkg/controller/migmigration/migrate.go
```diff
@@ -67,13 +68,13 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (time.Du
...
 	// Run
 	task := Task{
-		Log:             log,
-		Client:          r,
-		Owner:           migration,
-		PlanResources:   planResources,
-		Phase:           migration.Status.Phase,
-		Annotations:     r.getAnnotations(migration),
-		BackupResources: r.getBackupResources(migration),
+		ReconcileMigMigration: r,
```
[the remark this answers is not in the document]
Danil-Grigorev: Sorry for confusion @djwhatle, the client-go scheme is created as a single instance on the client-go library import, and all the managers are referencing it and not creating a new one each time. @jortel approach is correct.

pkg/controller/migmigration/migmigration_controller.go
```diff
@@ -45,12 +45,12 @@ func Add(mgr manager.Manager) error {
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+func newReconciler(mgr manager.Manager) *ReconcileMigMigration {
```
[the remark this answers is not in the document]
Danil-Grigorev: There is no reason to do that. Unrelated to the PR

pkg/controller/migmigration/migmigration_controller.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/konveyor/mig-controller/pkg/logging"
 	migref "github.com/konveyor/mig-controller/pkg/reference"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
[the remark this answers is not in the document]
Danil-Grigorev: It's IDE configuration, have no idea.

Danil-Grigorev:
@djwhatle Would have to ask for a final review. 

After yesterday's testing I found out that filtering based on GroupVersions provided form the server is not sufficient. The issue you saw was due to the fact that group extensions/v1beta1 never went away in a newer cluster versions, just it's content changed. So my algorithm was marking as an existing group and existing version for the group, therefore showing nothing in unhealthyResources at the end. Moving on to use full resource discovery instead - there is no way around.

Cleaned up imports. I have no convenient way of disabling the IDE features around structuring imports. The number of unrelated files with those restructured imports is so huge due to the fact that I had to manually switch import paths from `fusor` to `konveyor` and it caused autosaving and autoformatting. Most of those are now cleaned.

Removed autodowngrade from my previous PR this was based on.

pkg/controller/migplan/gvk.go
```diff
@@ -0,0 +1,414 @@
...
+		res.APIResources = namespaced(res.APIResources)
+		res.APIResources = excludeSubresources(res.APIResources)
+		// Some resources appear not to have permissions to list, need to exclude those.
+		res.APIResources = listAllowed(res.APIResources)
+	}
+
+	return resources, nil
+}
+
+func (r *Compare) unsupportedResources(resourceDiff []*metav1.APIResourceList) ([]schema.GroupVersionResource, error) {
```
gildub: There is no need here for a function type.  (r *Compare) is not used.

Danil-Grigorev: Fixed, thanks :)

pkg/controller/migplan/gvk.go
```diff
@@ -0,0 +1,414 @@
...
+	err = gvkCompare.NewSourceClient(r)
+	if err != nil {
+		return nil, err
+	}
+
+	return gvkCompare, nil
+}
+
+func reportGVK(plan *migapi.MigPlan, unsupportedMapping map[string][]schema.GroupVersionResource) {
+	unsupportedNamespaces := []migapi.UnsupportedNamespace{}
```
djwhatle: `incompatible` would be a more accurate term across the board in this PR. Please change all instances of `unsupported` to `incompatible`

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -57,7 +58,8 @@ type MigPlanSpec struct {
 // MigPlanStatus defines the observed state of MigPlan
 type MigPlanStatus struct {
 	Conditions
-	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	UnsupportedNamespaces []UnsupportedNamespace `json:"unsupportedNamespaces,omitempty"`
```
djwhatle: Unsupported -> Incompatible

config/crds/migration_v1alpha1_migplan.yaml
```diff
@@ -96,6 +96,22 @@ spec:
             observedGeneration:
               format: int64
               type: integer
+            unsupportedNamespaces:
```
djwhatle: Unsupported -> Incompatible

Danil-Grigorev:
@djwhatle Updated with requested changes included, with a minor note - we are dealing with GVRs but not GVKs, therefore the name change. @jortel Is it ok from your point of view?

config/crds/migration_v1alpha1_migplan.yaml
```diff
@@ -96,6 +96,22 @@ spec:
...
+                    items:
+                      type: string
+                    type: array
+                required:
+                - name
+                - unsupportedResources
+                type: object
+              type: array
+          required:
+          - observedGeneration
```
jortel: This change is reverting https://github.com/konveyor/mig-controller/pull/422.  The `observedGeneration` needs to be optional.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -57,7 +58,8 @@ type MigPlanSpec struct {
 // MigPlanStatus defines the observed state of MigPlan
 type MigPlanStatus struct {
 	Conditions
-	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	Incompatible
+	ObservedGeneration int64 `json:"observedGeneration"`
```
jortel: This change ^ is reverting https://github.com/konveyor/mig-controller/pull/422.  The `observedGeneration` needs to be `omitempty`.

Danil-Grigorev: Oh, ok, I see. This resolves the issue, thanks @jortel

pkg/apis/migration/v1alpha1/gvr.go
```diff
@@ -0,0 +1,32 @@
...
+// which are being selected in the MigPlan
+type Incompatible struct {
+	Namespaces []IncompatibleNamespace `json:"incompatibleNamespaces,omitempty"`
+}
+
+// IncompatibleNamespace - namespace, which is noticed
+// to contain resources incompatible by the migration
+type IncompatibleNamespace struct {
+	Name string            `json:"name"`
+	GVRs []IncompatibleGVR `json:"gvrs"`
```
jortel: Can you help me understand why this is GVR instead of GVK?  The field seems to contain kinds (Eg: pods).  Since resources are instances of kinds, I would expect a field named GVR to identify instances by name.

Danil-Grigorev: For the most part it is explained by the dynamic client API, on which the feature was built on. It expects a `GVR` structure to fetch the resource, and the actual Kind was lost during transformations. It is not suitable, as the resources are lowercase and in plural tense, which allows it to be directly translated to the API path. You could fetch `apis/extensions/v1beta1/replicasets`, but server does not recognize `apis/extensions/v1beta1/ReplicaSet` as a valid path.

pkg/apis/migration/v1alpha1/gvr.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// IncompatibleGVR - custom sructure for printing GVRs lowercase
+type IncompatibleGVR struct {
+	Group    string `json:"group"`
+	Version  string `json:"version"`
+	Resource string `json:"resoruce"`
```
djwhatle: `json:"resoruce"` -> `json:"resource`

I see several places in this PR where "resource" is has typos in its spelling, @Danil-Grigorev let's fix these

Danil-Grigorev: Fixed 😥

djwhatle:
Testing now.

Issue #415, opened 2020-02-11
Discovery controller using non-cached client in reconciler.


jortel:
The discovery controller supports concurrent reconciles (`MaxConcurrentReconciles: 10`).  However it's been observed that using the controller-runtime cached client concurrently results in erroneous `NotFound` errors GETing resources.
Two choices here:
1. Set `MaxConcurrentReconciles: 1` (the default).
2. Build a non-cached client for the local host for the reconciler instead of using the cached client built by the `Manager`.

The goal of concurrent reconciles in the discovery controller is to expedite data availability on controller startup or other scenarios when many MigClusters need to be inventoried by the discovery controller.  For this reason, **option 2** is the desirable solution.

 https://bugzilla.redhat.com/show_bug.cgi?id=1801912

PR #416, opened 2020-02-11
Use non-cached client in the reconciler instead.

jortel:
fixes: #415

The default cached client provided by the `Manager` has been observed to result in _false_ `NotFound` errors when used in concurrent reconcoles.

Build a non-cached client for the Reconciler.


djwhatle:
@jortel so are we not using a client with cache anywhere now?

PR #417, opened 2020-02-13
migcluster: use the CABundle unless 'insecure' is true

mansam:
This is the PR (with a minor naming change) that was reverted in https://github.com/fusor/mig-controller/pull/396 because it had been mistakenly merged in for 1.1. Needed for MigCluster CA bundle support in 1.2.

### DIFF
--- a/config/crds/migration_v1alpha1_migcluster.yaml
+++ b/config/crds/migration_v1alpha1_migcluster.yaml
@@ -33,6 +33,8 @@ spec:
             caBundle:
               format: byte
               type: string
+            insecure:
+              type: boolean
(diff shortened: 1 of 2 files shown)


config/crds/migration_v1alpha1_migcluster.yaml
```diff
@@ -56,6 +58,7 @@ spec:
               type: string
           required:
           - isHostCluster
+          - insecure
```
dymurray: Let's have a default value here

mansam: done

PR #421, opened 2020-02-17
Last observed generation - CRDs

jortel:
Forgot to update the CRDs in https://github.com/fusor/mig-controller/pull/412.

### DIFF
--- a/config/crds/migration_v1alpha1_migcluster.yaml
+++ b/config/crds/migration_v1alpha1_migcluster.yaml
@@ -60,6 +60,12 @@ spec:
           - isHostCluster
           type: object
         status:
+          properties:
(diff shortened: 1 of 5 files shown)


config/crds/migration_v1alpha1_migcluster.yaml
```diff
@@ -58,8 +58,15 @@ spec:
               type: string
           required:
           - isHostCluster
+          - insecure
```
dymurray: Believe this needs to be removed

jortel: Looks like the CRD was edited manually instead of updating the Field tag.  Fixed.

PR #433, opened 2020-02-27
Validate that the service account for the remote cluster has cluster-admin

### LINKED ISSUE #420: Validate SA token has cluster-admin.
Add a validation/condition to the `Cluster` controller to validate that the SA token in the secret has _cluster-admin_.

Likely perform a SAR or SAR check on something we know requires _cluster-admin_.

mansam:
This performs a SubjectAccessReview on the service account for the remote cluster in order to validate whether the service account has cluster-admin. It does this by requesting rights to all verbs on all resources in all groups and in all namespaces.

Fixes #420

### DIFF
--- a/pkg/controller/migcluster/validation.go
+++ b/pkg/controller/migcluster/validation.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	auth "k8s.io/api/authorization/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"net/url"
 	"time"
 
@@ -14,11 +13,11 @@ import (
 
 // Types
 const (
-	InvalidURL          = "InvalidURL"
-	InvalidSaSecretRef  = "InvalidSaSecretRef"
-	InvalidSaToken      = "InvalidSaToken"
-	UnauthorizedSaToken = "UnauthorizedSaToken"
-	TestConnectFailed   = "TestConnectFailed"
+	InvalidURL           = "InvalidURL"
+	InvalidSaSecretRef   = "InvalidSaSecretRef"
+	InvalidSaToken       = "InvalidSaToken"
+	TestConnectFailed    = "TestConnectFailed"
+	SaTokenNotPrivileged = "SaTokenNotPrivileged"
 )
 
 // Categories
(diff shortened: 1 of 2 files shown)


pkg/controller/migcluster/validation.go
```diff
@@ -39,13 +44,14 @@ const (
...
-	MalformedURLMessage       = "The `url` is malformed."
-	InvalidURLSchemeMessage   = "The `url` scheme must be 'http' or 'https'."
+	ReadyMessage               = "The cluster is ready."
+	MissingURLMessage          = "The `url` is required when `isHostCluster` is false."
+	InvalidSaSecretRefMessage  = "The `serviceAccountSecretRef` must reference a `secret`."
+	InvalidSaTokenMessage      = "The `saToken` not found in `serviceAccountSecretRef` secret."
+	TestConnectFailedMessage   = "Test connect failed: %s"
+	MalformedURLMessage        = "The `url` is malformed."
+	InvalidURLSchemeMessage    = "The `url` scheme must be 'http' or 'https'."
+	UnauthorizedSaTokenMessage = "The service account token is not authorized to perform migrations."
```
jortel: Saying that the token is _unauthorized_ may be misunderstood to mean that the token is invalid.

Suggest other naming/wording:
```
SaTokenNotPrivileged = The `saToken` has insufficient privileges.
```
To be consistent with other messages, we reference the specific field in ``.  The _"to perform migrations"_ is to leave but seems unnecessarily.

pkg/controller/migcluster/validation.go
```diff
@@ -205,3 +226,37 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
 
 	return nil
 }
+
+func (r *ReconcileMigCluster) saIsClusterAdmin(cluster *migapi.MigCluster) (bool, error) {
+	// check for access to all verbs on all resources in all namespaces
+	// in order to determine if the service account has cluster-admin
+	attributes := auth.ResourceAttributes{
```
jortel: :+1: Seems like a good determination of cluster-admin.

pkg/controller/migcluster/validation.go
```diff
@@ -205,3 +226,37 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
...
+	client, err := cluster.GetClient(r.Client)
+	if err != nil {
+		serr, ok := err.(*errors.StatusError)
+		if ok && serr.ErrStatus.Reason == Unauthorized {
+			return false, nil
+		}
+		return false, err
+	}
+	err = client.Create(context.TODO(), &sar)
+	if err != nil {
```
jortel: Add:
```
log.Trace(err)
```

pkg/controller/migcluster/validation.go
```diff
@@ -205,3 +226,37 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
...
+	attributes := auth.ResourceAttributes{
+		Group:    "*",
+		Resource: "*",
+		Verb:     "*",
+		Version:  "*",
+	}
+
+	sar := auth.SelfSubjectAccessReview{
+		Spec: auth.SelfSubjectAccessReviewSpec{
+			ResourceAttributes: &attributes,
```
jortel: Separate variable is fine but could be defined here _inline_ since it's not used anywhere else.

pkg/controller/migcluster/validation.go
```diff
@@ -29,6 +33,7 @@ const (
 	ConnectFailed = "ConnectFailed"
 	Malformed     = "Malformed"
 	InvalidScheme = "InvalidScheme"
+	Unauthorized  = "Unauthorized"
```
jortel: :+1:

pkg/controller/migcluster/validation.go
```diff
@@ -205,3 +226,37 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
 
 	return nil
 }
+
+func (r *ReconcileMigCluster) saIsClusterAdmin(cluster *migapi.MigCluster) (bool, error) {
```
jortel: This validation cannot be performed when any of the existing critical conditions exist.
```
if cluster.Status.HasCriticalCondition() {
	return nil
}
```

Also, let's be consistent with other validations.  The design pattern is to name the method: validate__() and to set the condition here (in the method) instead of returning a boolean.

mansam: Fair enough. My thinking was that `saIsClusterAdmin` was just a helper function to `validateSaSecret`, but I can certainly make it its own top level validation and call it from `validate`.

Issue #432, opened 2020-02-26
mig-controller may need changes to work with OCP 4.4


jmontleon:
@Danil-Grigorev is seeing an error when trying to use mig-controller on OCP 4.4.

In the error below we have:
```
""error":"no matches for kind \"Deployment\" in version \"apps/v1beta1\"
```

Coming from:
https://github.com/konveyor/mig-controller/blob/master/pkg/controller/migplan/migplan_controller.go#L311

Which is probably being thrown here:
https://github.com/konveyor/mig-controller/blob/93bda35aabb93cec6b3aaac9a4031f67b874f0be/pkg/apis/migration/v1alpha1/migcluster_types.go#L184

apps/v1beta1 Deployments no longer exists in OpenShift 4.4 and we must use apps/v1 Deployments.

I don't know if this approach will work for the controller and whether this is cross cluster so source and destination versions need to be taken into account, but in the operator I took the approach of checking the Kubernetes minor version and if old enough using v1beta1 else v1. It was also necessary to do this for daemonsets
https://github.com/konveyor/mig-operator/pull/232/files
https://github.com/konveyor/mig-operator/pull/233/files

Full error:
```
[...]

PR #436, opened 2020-03-02
Apps version dynamic support

Danil-Grigorev:
Selecting supported resources basing on kubernetes version
    
- Included discovery client to migCluster functionality
- Checking for kubernetes cluster version before attempts to get cluster resources

Closes https://github.com/konveyor/mig-controller/issues/432

### DIFF
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -29,7 +29,6 @@ import (
 	storageapi "k8s.io/api/storage/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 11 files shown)


djwhatle:
@Danil-Grigorev does this address an existing GitHub issue? I see "closes #436" which is a self-reference back to this PR.

Danil-Grigorev:
@djwhatle Thanks for pointing out

Danil-Grigorev:
@jortel I would be glad to hear opinions how to approach this without falling back to a raw `runtime.Object` and preserving static typing. If such a change needs to take place, it needs some code refactoring in the first place, which I'm willing to do.

Danil-Grigorev:
@jortel Following the pattern 
1) get resource
2) read and update the returned object
3) put/remove object

The only place where the logic should not chage, is the case # 2. Other places require interaction with the server, which needs a conversion to the right `apiVersion` for the resource. One way to deal with it, is to use something like conversion functions, which should be included in the `scheme`.
a) old to new: https://godoc.org/k8s.io/kubernetes/pkg/apis/apps/v1beta1#Convert_v1beta1_Deployment_To_apps_Deployment
b) and back: https://godoc.org/k8s.io/kubernetes/pkg/apis/apps/v1beta1#Convert_v1beta1_Deployment_To_apps_Deployment
So that we are picking a single API versions to deal with in the logic part. Such approach requires a wrapper on the `client.Client` from controller-runtime, which seems like a dirty path.

If there are other ideas, would be great to hear. @jortel @sseago

Danil-Grigorev:
Closing this as @jortel apparently started on client wrapper already.

Issue #438, opened 2020-03-02
[MIG-399] Use go modules


gildub:
From golang 1.14 modules are defacto standard

dymurray:
Let's setup an hour meeting to discuss what we learned moving the plugin repos to go modules and help the controller contributors get up to speed on this.

djwhatle:
Closed https://github.com/konveyor/mig-controller/pull/439 due to staleness. Still plan to switch to go modules in the future.

PR #439, opened 2020-03-02
Use go modules

gildub:
https://github.com/konveyor/mig-controller/issues/438

For now k8s.io/code-generator (v0.17.0) and sigs.k8s.io/controller-tools (v0.1.10) are added to vendor by using anonymous import in cmd/manager/main.go.


jortel:
Thanks, @gildub 
This seems like something we want to do but I need to read up on `modules`.
What is the value/risk in doing it now vs later?
Let's get 3-4 ACKs before merge.

sseago:
One risk of continuing to use dep is that it's basically unsupported now. Dep's latest release doesn't support dependencies which use modules (and more and more of them are moving to modules now). Once you have a dependency on a modules-based repo, the only option with dep is to build your own binary based on an open PR. Upstream comments on the PR are basically "looks good, but we don't have the resources to test it". For the plugins, I had to do this when containers/image moved to go modules. Velero upstream has moved to modules just this week (so 1.4 will use modules). This means when we upgrade to velero 1.4, mig-controller will have to start using the same unsupported/custom build of dep if we haven't moved to modules yet.

cmd/manager/main.go
```diff
@@ -31,6 +31,21 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+
+	_ "k8s.io/code-generator/cmd/client-gen/args"
```
djwhatle: @gildub @dymurray trying to understand why this change results in new imports in our manager main.go

gildub: @djwhatle, those are development dependencies to build the project.

For instance the manifest rule of the Makefile uses `go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go`. 

By adding anonymous imports automatically includes those libraries which are used by CI too.

There might be different way of solving it.

djwhatle:
Since this has been idle for a while, I want to mention that this is still a priority.

I think the team is in agreement now that we should move over to go mod, just a matter of doing due-diligence checks vetting the new dep format for equivalence and ensuring no systems in the controller are broken when we're checking in 13 million lines.

pranavgaikwad:
\test

djwhatle:
Closing until we can take time to do a more thorough review of the effects of this switch.

PR #443, opened 2020-03-05
Provide seamless API compat.

jortel:
The approach:  The controller imports and uses only `v1` resources.  The `Cluster.GetClient()` returns a _smart_ client (wrapper) for all clusters (even the host).  The _wrapper_ will down-convert objects (as needed) before making the API call to the cluster version.  Then, up-convert the object (as needed) and return to the caller.

This way, all the controller code can be based on `v1` but be compatible with older clusters.  The logic is isolated to one place in the code.

The client (wrapper) also provides the `DiscoveryInterface` when _cast_ as follows:
```
dClient, _ := client.(DiscoveryInterface)
```

This PR only add the client (wrapper).

Future PRs will be needed to convert the code-base to use `v1` resources.
It would be good for this effort to also include changing the quiesce code using _unstructured_ for cron jobs to use structured `CronJob`.

### DIFF
--- a/pkg/compat/client.go
+++ b/pkg/compat/client.go
@@ -1,0 +1,151 @@
+package compat
+
+import (
+	"context"
+	appv1 "k8s.io/api/apps/v1"
+	appv1beta1 "k8s.io/api/apps/v1beta1"
+	extv1beta1 "k8s.io/api/extensions/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime"
+	dapi "k8s.io/client-go/discovery"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"strconv"
+	"strings"
+)
+
+//
+// A smart client.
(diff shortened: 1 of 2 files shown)


pkg/compat/client.go
```diff
@@ -0,0 +1,151 @@
...
+// Get the specified resource.
+// The resource will be converted to a compatible version as needed.
+func (c Client) Get(ctx context.Context, key k8sclient.ObjectKey, in runtime.Object) error {
+	obj := c.downConvert(in)
+	err := c.Client.Get(ctx, key, obj)
+	if err != nil {
+		return err
+	}
+	if in != obj {
+		err := scheme.Scheme.Convert(obj, in, ctx)
```
djwhatle: @jortel you said that this reference to `scheme` has the necessary context to do the conversion correct? I'm surprised we were able to skip passing scheme through, but I'll take your word for it if you've observed this working.

jortel: The `Manager.GetScheme()` returns the global `scheme.Scheme` when not constructed with a `Scheme` in the _options_.  The manager code generated by kube-builder does not construct with a Scheme.

PR #450, opened 2020-03-13
Fix for CRD exclusion

Danil-Grigorev:
Wrongly assuming a list is a pointer in the function argument.

fix (partial) #451


djwhatle:
@Danil-Grigorev I assume you've tested this already? I'm about to test now.

Issue #157, opened 2019-06-14
Recovery workflows after migration failures


sseago:
We now have basic failure handling in place to properly set Migration state if something goes wrong during backup or restore. Longer-term, we need some sort of recovery workflow. After a failure, in some cases a simple retry is all a user might need (especially if it's on backup). In other cases, we may need to clean up from the failure (remove destination namespaces, un-do quiesce actions on the source) before allowing the user to re-run another stage and then a final migration.

dymurray:
Adding GA Blocker label. We need the ability to reverse the quiesce behavior after a migration has run. We should explore as much as we can of recovering from failed migrations and rolling back successful migrations before GA.

sseago:
One component of this, reversing the quiesce action, might make sense to have exposed as a standalone function, but it also fits within the recovery workflow framework.

A couple thoughts on the implementation:
To do this, we could set an annotation (something like "openshift.io/quiesced") where the value will be whatever replicas was before we set it to 0 (for everything other than DaemonSet -- for DaemonSet the presence of the annotation is all we'll need). The "reverse quiesce" function would be to do the same iteration over resources we did for quiesce, but looking for this annotation and, when we find it, re-setting replicas (removing the custom nodeSelector key for DaemonSet) and removing the annotation. I"m not sure where we'd attach the function call for this though.
￼

dymurray:
Let's itemize this issue since this covers all functionality required by adding a "rollback" feature to the migration tool. Adding the Post 4.3 label since some of this may fall into post 4.3 buckets and some may be delivered prior to 4.3.

PR #458, opened 2020-03-19
App backscaling on demand

### LINKED ISSUE #237: Add ability to cancel a Stage/Migrate in process

Danil-Grigorev:
- Cancel process now starts immediately, once the spec field is set.
- Added `FailedItinerary` to handle failing process during migration.
- Refactored `t.next` to account current step as an argument (allows to change phase before executing current one)
- Added app reactivation on failure/cancellation
- Added better failure handling process
- Added app cleanup on destination by label `migration.openshift.io/migrated-by: <migration-uid>`

Needs a change in the operator repo: https://github.com/konveyor/mig-operator/pull/293

Visually will add annotations with resources state before quiescence takes place
```
    annotations:
      migration.openshift.io/preQuiesceReplicas: "2"
```
Those annotations will be removed only when an application will be re-activated by a `Failed` migration, or migration being `Cancelled` 

Requires to have `delete` permissions to all resources with the SA token used to communicate to the remote cluster.

[Demo](https://drive.google.com/open?id=1WbBMZXnrX-lpackfXveZ_aKKcc5iWmrZ)

Closes https://github.com/konveyor/mig-controller/issues/157

This is a necessary step needed to close #237 completely.

### DIFF
--- a/pkg/apis/migration/v1alpha1/gvk.go
+++ b/pkg/apis/migration/v1alpha1/gvk.go
@@ -51,6 +51,7 @@ func (i *Incompatible) ResourceList() (incompatible []string) {
 	return
 }
 
+// CollectResources collects all namespaced scoped apiResources from the cluster
 func CollectResources(discovery discovery.DiscoveryInterface) ([]*metav1.APIResourceList, error) {
 	resources, err := discovery.ServerResources()
 	if err != nil {
@@ -67,6 +68,7 @@ func CollectResources(discovery discovery.DiscoveryInterface) ([]*metav1.APIReso
 	return resources, nil
 }
 
+// ConvertToGVRList converts provided apiResourceList to list of GroupVersionResources from the server
 func ConvertToGVRList(resourceList []*metav1.APIResourceList) ([]schema.GroupVersionResource, error) {
 	GVRs := []schema.GroupVersionResource{}
 	for _, resourceList := range resourceList {
--- a/pkg/apis/migration/v1alpha1/gvk.go
+++ b/pkg/apis/migration/v1alpha1/gvk.go
@@ -1,11 +1,7 @@
 package v1alpha1
 
 import (
-	"strings"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/discovery"
 )
 
 // Incompatible - list of namespaces containing incompatible resources for migration
@@ -49,75 +45,4 @@ func (i *Incompatible) ResourceList() (incompatible []string) {
 		}
 	}
 	return
-}
-
-// CollectResources collects all namespaced scoped apiResources from the cluster
-func CollectResources(discovery discovery.DiscoveryInterface) ([]*metav1.APIResourceList, error) {
-	resources, err := discovery.ServerResources()
-	if err != nil {
-		return nil, err
-	}
-
-	for _, res := range resources {
-		res.APIResources = namespaced(res.APIResources)
-		res.APIResources = excludeSubresources(res.APIResources)
-		// Some resources appear not to have permissions to list, need to exclude those.
-		res.APIResources = listAllowed(res.APIResources)
-	}
-
-	return resources, nil
-}
-
-// ConvertToGVRList converts provided apiResourceList to list of GroupVersionResources from the server
-func ConvertToGVRList(resourceList []*metav1.APIResourceList) ([]schema.GroupVersionResource, error) {
-	GVRs := []schema.GroupVersionResource{}
-	for _, resourceList := range resourceList {
-		gv, err := schema.ParseGroupVersion(resourceList.GroupVersion)
-		if err != nil {
-			return nil, err
-		}
-
-		for _, resource := range resourceList.APIResources {
-			gvk := gv.WithResource(resource.Name)
-			GVRs = append(GVRs, gvk)
-		}
-	}
-
-	return GVRs, nil
-}
-
-func excludeSubresources(resources []metav1.APIResource) []metav1.APIResource {
-	filteredList := []metav1.APIResource{}
-	for _, res := range resources {
-		if !strings.Contains(res.Name, "/") {
-			filteredList = append(filteredList, res)
-		}
-	}
-
-	return filteredList
-}
-
-func namespaced(resources []metav1.APIResource) []metav1.APIResource {
-	filteredList := []metav1.APIResource{}
-	for _, res := range resources {
-		if res.Namespaced {
-			filteredList = append(filteredList, res)
-		}
-	}
-
-	return filteredList
-}
-
-func listAllowed(resources []metav1.APIResource) []metav1.APIResource {
-	filteredList := []metav1.APIResource{}
-	for _, res := range resources {
-		for _, verb := range res.Verbs {
-			if verb == "list" {
-				filteredList = append(filteredList, res)
-				break
-			}
-		}
-	}
-
-	return filteredList
 }
--- a/pkg/apis/migration/v1alpha1/gvk.go
+++ b/pkg/apis/migration/v1alpha1/gvk.go
@@ -1,6 +1,12 @@
 package v1alpha1
 
-import "k8s.io/apimachinery/pkg/runtime/schema"
+import (
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+)
 
 // Incompatible - list of namespaces containing incompatible resources for migration
 // which are being selected in the MigPlan
@@ -43,4 +49,73 @@ func (i *Incompatible) ResourceList() (incompatible []string) {
 		}
 	}
 	return
+}
+
+func CollectResources(discovery discovery.DiscoveryInterface) ([]*metav1.APIResourceList, error) {
+	resources, err := discovery.ServerResources()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, res := range resources {
+		res.APIResources = namespaced(res.APIResources)
+		res.APIResources = excludeSubresources(res.APIResources)
+		// Some resources appear not to have permissions to list, need to exclude those.
+		res.APIResources = listAllowed(res.APIResources)
+	}
+
+	return resources, nil
+}
+
+func ConvertToGVRList(resourceList []*metav1.APIResourceList) ([]schema.GroupVersionResource, error) {
+	GVRs := []schema.GroupVersionResource{}
+	for _, resourceList := range resourceList {
+		gv, err := schema.ParseGroupVersion(resourceList.GroupVersion)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, resource := range resourceList.APIResources {
+			gvk := gv.WithResource(resource.Name)
+			GVRs = append(GVRs, gvk)
+		}
+	}
+
+	return GVRs, nil
+}
+
+func excludeSubresources(resources []metav1.APIResource) []metav1.APIResource {
+	filteredList := []metav1.APIResource{}
+	for _, res := range resources {
+		if !strings.Contains(res.Name, "/") {
+			filteredList = append(filteredList, res)
+		}
+	}
+
+	return filteredList
+}
+
+func namespaced(resources []metav1.APIResource) []metav1.APIResource {
+	filteredList := []metav1.APIResource{}
+	for _, res := range resources {
+		if res.Namespaced {
+			filteredList = append(filteredList, res)
+		}
+	}
+
+	return filteredList
+}
+
+func listAllowed(resources []metav1.APIResource) []metav1.APIResource {
+	filteredList := []metav1.APIResource{}
+	for _, res := range resources {
+		for _, verb := range res.Verbs {
+			if verb == "list" {
+				filteredList = append(filteredList, res)
+				break
+			}
+		}
+	}
+
+	return filteredList
 }
--- a/pkg/apis/migration/v1alpha1/labels.go
+++ b/pkg/apis/migration/v1alpha1/labels.go
@@ -1,15 +1,17 @@
 package v1alpha1
 
 import (
+	"strings"
+
 	migref "github.com/konveyor/mig-controller/pkg/reference"
 	"k8s.io/apimachinery/pkg/types"
-	"strings"
 )
 
 // Labels
 const (
-	PartOfLabel = "app.kubernetes.io/part-of" // = Application
-	Application = "openshift-migration"
+	PartOfLabel     = "app.kubernetes.io/part-of" // = Application
+	Application     = "openshift-migration"
+	MigratedByLabel = "migration.openshift.io/migrated-by" // (migmigration UID)
 )
 
 // Build label (key, value) used to correlate CRs.
--- a/pkg/apis/migration/v1alpha1/labels.go
+++ b/pkg/apis/migration/v1alpha1/labels.go
@@ -9,9 +9,8 @@ import (
 
 // Labels
 const (
-	PartOfLabel     = "app.kubernetes.io/part-of" // = Application
-	Application     = "openshift-migration"
-	MigratedByLabel = "migration.openshift.io/migrated-by" // (migmigration UID)
+	PartOfLabel = "app.kubernetes.io/part-of" // = Application
+	Application = "openshift-migration"
 )
 
 // Build label (key, value) used to correlate CRs.
--- a/pkg/apis/migration/v1alpha1/model.go
+++ b/pkg/apis/migration/v1alpha1/model.go
@@ -2,11 +2,17 @@ package v1alpha1
 
 import (
 	"context"
+	"strconv"
+
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
-	"strconv"
 )
 
 //
@@ -160,4 +166,82 @@ func GetSecret(client k8sclient.Client, ref *kapi.ObjectReference) (*kapi.Secret
 	}
 
 	return &object, err
+}
+
+// DeleteMigrated - dispatches delete requests for migrated resources
+func DeleteMigrated(config *rest.Config, uid string) error {
+	GVRs, err := getGVRs(config)
+	if err != nil {
+		return err
+	}
+	client, err := dynamic.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+	listOptions := k8sclient.MatchingLabels(map[string]string{
+		MigratedByLabel: uid,
+	}).AsListOptions()
+	foreground := metav1.DeletePropagationForeground
+	deleteOptions := (&k8sclient.DeleteOptions{
+		PropagationPolicy: &foreground,
+	}).AsDeleteOptions()
+
+	for _, gvr := range GVRs {
+		list, err := client.Resource(gvr).List(*listOptions)
+		if err != nil {
+			return err
+		}
+		for _, r := range list.Items {
+			err = client.Resource(gvr).Namespace(r.GetNamespace()).Delete(r.GetName(), deleteOptions)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// WaitForMigratedDeletion - returns true if all of the migrated resources were deleted
+func WaitForMigratedDeletion(config *rest.Config, uid string) (bool, error) {
+	GVRs, err := getGVRs(config)
+	if err != nil {
+		return false, err
+	}
+	client, err := dynamic.NewForConfig(config)
+	if err != nil {
+		return false, err
+	}
+	listOptions := k8sclient.MatchingLabels(map[string]string{
+		MigratedByLabel: uid,
+	}).AsListOptions()
+	for _, gvr := range GVRs {
+		// Count resource occurences
+		list, err := client.Resource(gvr).List(*listOptions)
+		if err != nil {
+			return false, err
+		}
+		if len(list.Items) > 0 {
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
+
+func getGVRs(config *rest.Config) ([]schema.GroupVersionResource, error) {
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	resourceList, err := CollectResources(discoveryClient)
+	if err != nil {
+		return nil, err
+	}
+	GVRs, err := ConvertToGVRList(resourceList)
+	if err != nil {
+		return nil, err
+	}
+
+	return GVRs, nil
 }
--- a/pkg/apis/migration/v1alpha1/model.go
+++ b/pkg/apis/migration/v1alpha1/model.go
@@ -6,12 +6,7 @@ import (
 
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/discovery"
-	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/rest"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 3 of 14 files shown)


jwmatthews:
Saw the demo Danil shared and functionality wise I think this makes sense to consider for CAM 1.2.

pkg/controller/migmigration/quiesce.go
```diff
@@ -59,7 +115,7 @@ func (t *Task) quiesceApplications() error {
 }
 
 // Scales down DeploymentConfig on source cluster
-func (t *Task) scaleDownDeploymentConfigs(client k8sclient.Client) error {
+func (t *Task) quiesceDeploymentConfigs(client k8sclient.Client, suspend bool) error {
```
jortel: I was confused by `suspend` here.

I get that you want to reduce duplicate code but since the majority of the code in these methods depends on whether we are quiescing or un-quiescing, it may be cleaner to just have separate methods:

`quiesceDeploymentConfigs()`
`unQuiesceDeploymentConfigs()`

etc.

pkg/controller/migmigration/quiesce.go
```diff
@@ -12,44 +14,98 @@ import (
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// ReplicasAnnotation - annotation to set during quisce phase, which will allow to recreate the app back to it's initial state
+// in case of a migration failure or user's intent
+const (
+	SuspendAnnotation      = "migration.openshift.io/preQuiesceSuspend"
+	ReplicasAnnotation     = "migration.openshift.io/preQuiesceReplicas"
+	NodeSelectorAnnotation = "migration.openshift.io/preQuiesceNodeSelector"
+)
```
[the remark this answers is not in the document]
sseago: I don't know whether stripping them from the destination is needed. We don't quiesce until all backups have been performed, so these shouldn't be showing up on the destination cluster unless someone migrates a previously-quiesced namespace in a later migration. That being said, it wouldn't hurt to look for these and delete them anyway, just in case.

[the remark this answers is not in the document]
Danil-Grigorev: Moved annotations to the right location. With this PR such situation described by Scott would not appear that often, as the stripping is done on migration failure.

pkg/controller/migmigration/quiesce.go
```diff
@@ -12,44 +14,98 @@ import (
...
+	if err != nil {
+		log.Trace(err)
+		return err
+	}
+
+	return nil
+}
+
+// All quiesce undo functionality should be put here
+func (t *Task) reactivateApplications() error {
```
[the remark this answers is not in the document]
Danil-Grigorev: English is hard, I was actually looking for the right word to describe what is happening there :) Fixed.

pkg/controller/migmigration/quiesce.go
Danil-Grigorev: Sure.

Danil-Grigorev:
@jortel May I get a review once again?

pkg/controller/migmigration/task.go
```diff
@@ -30,6 +30,7 @@ const (
 	ResticRestarted               = "ResticRestarted"
 	QuiesceApplications           = "QuiesceApplications"
 	EnsureQuiesced                = "EnsureQuiesced"
+	ReactivateApplications        = "ReactivateApplications"
```
mansam: Maybe this should be UnQuiesceApplications for consistency?

Danil-Grigorev: Thanks! Missed this one

Danil-Grigorev:
During last sprint review @eriknelson pointed out a possible issue with up scaling the application after running the `move` migration, where the same PV template is simultaneously used from both clusters. That would corrupt data if the app at some moment will run from both cluster simultaneously. Will work on the approach to this in the same PR.

The second thing which is desirable, is to jump to the cancel workflow at once, as it appears on the resource spec, and stop the running `phase` from being executed, as opposed to the approach which is used now - wait for the step to finish. Will extend the PR with that one too.

pkg/apis/migration/v1alpha1/gvk.go
```diff
@@ -44,3 +50,72 @@ func (i *Incompatible) ResourceList() (incompatible []string) {
 	}
 	return
 }
+
+func CollectResources(discovery discovery.DiscoveryInterface) ([]*metav1.APIResourceList, error) {
```
Danil-Grigorev, line 54: `CollectResources` and `ConvertToGVRList` are reusable, so they were moved here.

pkg/apis/migration/v1alpha1/labels.go
```diff
@@ -1,15 +1,17 @@
...
-	"strings"
 )
 
 // Labels
 const (
-	PartOfLabel = "app.kubernetes.io/part-of" // = Application
-	Application = "openshift-migration"
+	PartOfLabel     = "app.kubernetes.io/part-of" // = Application
+	Application     = "openshift-migration"
+	MigratedByLabel = "migration.openshift.io/migrated-by" // (migmigration UID)
```
Danil-Grigorev, line 14: A unique label is set to select only pieces of migrated application by label selector.

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -161,3 +167,81 @@ func GetSecret(client k8sclient.Client, ref *kapi.ObjectReference) (*kapi.Secret
...
+		return err
+	}
+	client, err := dynamic.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+	listOptions := k8sclient.MatchingLabels(map[string]string{
+		MigratedByLabel: uid,
+	}).AsListOptions()
+	foreground := metav1.DeletePropagationForeground
```
Danil-Grigorev, line 184: This will ensure all of dependent resources will be removed (those which have `BlockOwnerDeletion` set on true) - https://github.com/konveyor/mig-controller/blob/e333d47d55dc6317e4a0495fafbb6cb991445204/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L420-L424

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -161,3 +167,81 @@ func GetSecret(client k8sclient.Client, ref *kapi.ObjectReference) (*kapi.Secret
...
+		if err != nil {
+			return err
+		}
+		for _, r := range list.Items {
+			err = client.Resource(gvr).Namespace(r.GetNamespace()).Delete(r.GetName(), deleteOptions)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
Danil-Grigorev, line 200: This whole block could be simplified by using `DeleteCollection` method, but I hit issues with `ConfigMaps` while doing that.

```bash
oc api-resources -o wide | grep configmap
configmaps                            cm                                                     true         ConfigMap                            [create delete deletecollection get list patch update watch]
```

`
{"level":"error","ts":1586180540.0871735,"logger":"migration|z7pc5","msg":"","migration":"eb86bc20-75a9-11ea-bada-edebfcfc413f","error":"Resource - /v1, Resource=configmaps: the server does not allow this method on the requested resource","stacktrace":"github.com/konveyor/mig-controller/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/vendor/github.com/go-logr/zapr/zapr.go:128\ngithub.com/konveyor/mig-controller/pkg/logging.Logger.Error\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/pkg/logging/logger.go:75\ngithub.com/konveyor/mig-controller/pkg/logging.Logger.Trace\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/pkg/logging/logger.go:81\ngithub.com/konveyor/mig-controller/pkg/controller/migmigration.(*Task).deleteMigrated\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/pkg/controller/migmigration/restore.go:301\ngithub.com/konveyor/mig-controller/pkg/controller/migmigration.(*Task).Run\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/pkg/controller/migmigration/task.go:534\ngithub.com/konveyor/mig-controller/pkg/controller/migmigration.(*ReconcileMigMigration).migrate\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/pkg/controller/migmigration/migrate.go:78\ngithub.com/konveyor/mig-controller/pkg/controller/migmigration.(*ReconcileMigMigration).Reconcile\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/pkg/controller/migmigration/migmigration_controller.go:195\ngithub.com/konveyor/mig-controller/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:215\ngithub.com/konveyor/mig-controller/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:158\ngithub.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133\ngithub.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134\ngithub.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/wait.Until\n\t/home/dgrigore/go/src/github.com/konveyor/mig-controller/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88"}
`

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -161,3 +167,114 @@ func GetSecret(client k8sclient.Client, ref *kapi.ObjectReference) (*kapi.Secret
 
 	return &object, err
 }
+
+// DeleteLabels - removes migration  labels from the resource after restore
+func DeleteLabels(config *rest.Config, uid string) error {
```
sseago, line 172: DeleteLabels and DeleteMIgrated follow the same structure. They're basically identical except for the extra foreground/deleteOptions block in DeleteMigrated and the actual label/delete steps within the `range list.Items` loop. It's not a huge deal, but it might be nice if these could be refactored to remove some of the duplication.

jortel, line 172: The `model.go` is for commonly used convenience functions.  both `DeleteLabels()` and `DeleteMigrated()` seem to fall completely within the concern of, and solely used by, the Migration controller.  What are these defined here?

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -161,3 +167,114 @@ func GetSecret(client k8sclient.Client, ref *kapi.ObjectReference) (*kapi.Secret
...
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// WaitForMigratedDeletion - returns true if all of the migrated resources were deleted
+func WaitForMigratedDeletion(config *rest.Config, uid string) (bool, error) {
```
jortel, line 239: Why do we need to wait?

pkg/apis/migration/v1alpha1/labels.go
jortel, line 14: Seems like this labels is solely within the concern of the migration controller.

Danil-Grigorev: The idea was to follow the pattern with having all global variables - annotations or labels, defined in a single location.

pkg/apis/migration/v1alpha1/model.go
Danil-Grigorev: The whole purpose of removing the application on the destination side will be reduced to zero, if some of the pods (or other resource interacting with the same location as the application on the source side) on the target side remain running. Also, the second time the migration will run, existing resources won't be updated so the method is needed.

pkg/controller/migmigration/restore.go
```diff
@@ -285,22 +309,130 @@ func (t *Task) deleteRestores() error {
...
 			}
-			if mapping[1] != "" {
-				namespaceMapping[mapping[0]] = mapping[1]
+		}
+	}
+
+	return nil
+}
+
+func (t *Task) waitForDeleteMigrated() (bool, error) {
```
jortel, line 346: Suggest `ensureMigratedResourcesDeleted()` would better conform to naming convention in the migration controller.

pkg/controller/migmigration/restore.go
```diff
@@ -285,22 +309,130 @@ func (t *Task) deleteRestores() error {
...
+	}).AsListOptions()
+	for _, gvr := range GVRs {
+		list, err := client.Resource(gvr).List(*listOptions)
+		if err != nil {
+			log.Trace(err)
+			return false, err
+		}
+		// Wait for resources with deletion timestamps
+		for _, r := range list.Items {
+			if r.GetDeletionTimestamp() != nil {
```
jortel, line 364: :+1: Using the deletionTimestamp is clever.

pkg/controller/migmigration/task.go
```diff
@@ -608,8 +656,8 @@ func (t *Task) next() {
 	t.Phase = Completed
 }
 
-// Phase failed.
-func (t *Task) failed(nextPhase string, reasons []string) {
+// Phase fail.
+func (t *Task) fail(nextPhase string, reasons []string) {
```
jortel, line 660: Why was this renamed?

pkg/controller/migmigration/restore.go
```diff
@@ -285,22 +309,130 @@ func (t *Task) deleteRestores() error {
...
 			}
 		}
 	}
 
-	if len(namespaceMapping) != 0 {
-		restore.Spec.NamespaceMapping = namespaceMapping
+	return true, nil
+}
+
+func (t *Task) deleteLabels() error {
```
jortel, line 373: All of the other annotation & label management methods are in `annotation.go`.
Please follow the established convention of organizing functionality in `go` files.

pkg/apis/migration/v1alpha1/gvk.go
jortel, line 54: Agreed they are generic GVK functionality.  Why in the API package?

pkg/controller/migmigration/restore.go
```diff
@@ -285,22 +309,134 @@ func (t *Task) deleteRestores() error {
...
+	if err != nil {
+		log.Trace(err)
+		return nil, nil, err
+	}
+	config, err := cluster.BuildRestConfig(t.Client)
+	if err != nil {
+		log.Trace(err)
+		return nil, nil, err
+	}
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(config)
```
jortel, line 420: The destination client _IsA_ `compat.Client` and can be type asserted as follows:
```
discoveryClient := client.(discovery.DiscoveryInterface)
```

Danil-Grigorev: Woah, didn't notice that, thanks. Will fix it now.

pkg/controller/migmigration/restore.go
Danil-Grigorev: +

pkg/controller/migmigration/task.go
Danil-Grigorev: In correlation to `t.next()` This phase is never **next**, it is just an order to do something, either `fail` or go `next`.

Also, it correlates with `t.failed() bool` check, saying if migration failed, which is used at https://github.com/konveyor/mig-controller/blob/c2752422b1e329e68e12e9e982b4579f54838d3a/pkg/controller/migmigration/task.go#L604 and there is no way around.

pkg/controller/migmigration/restore.go
Danil-Grigorev: Sounds good, fixing.

pkg/apis/migration/v1alpha1/gvk.go
Danil-Grigorev: May be it is not needed, the only problem that I hit a dependency loop at some point, trying to use an internal `gvk` implementation. If the GVK was initially dedicated to a separate package, this won't happen, but right now it is necessary evil.

pkg/apis/migration/v1alpha1/labels.go
Danil-Grigorev: Ok, now I see. Fixed.

pkg/apis/migration/v1alpha1/gvk.go
Danil-Grigorev: @jortel does it sound like a solution to you?

pkg/compat/client.go
```diff
@@ -21,6 +21,7 @@ import (
 // A smart client.
 // Provides seamless API version compatibility.
 type Client struct {
+	*rest.Config
```
Danil-Grigorev, line 24: It is beneficial to have config to be exposed in the custom client, as the client itself is short lived and non-cached, could be used in multiple occasion.

pkg/controller/migmigration/annotation.go
```diff
@@ -477,6 +487,47 @@ func (t *Task) deleteNamespaceLabels(client k8sclient.Client, namespaceList []st
...
+		log.Trace(err)
+		return err
+	}
+
+	listOptions := k8sclient.MatchingLabels(map[string]string{
+		MigratedByLabel: string(t.Owner.UID),
+	}).AsListOptions()
+
+	for _, gvr := range GVRs {
+		for _, ns := range t.destinationNamespaces() {
```
Danil-Grigorev, line 504: Having an addition listing per-namespace is explained by the issue I saw, by listing resources in cluster scope with
```go
client.Resource(gvr).List(*listOptions) 
```

Such untyped client had issues picking up CRs with label selectors across multiple namespaces, and instead returned a full list of resources across cluster.

pkg/controller/migmigration/task.go
```diff
@@ -184,11 +202,12 @@ func (t *Task) Run() error {
 	t.Log.Info("[RUN]", "stage", t.stage(), "phase", t.Phase)
 
 	t.init()
+	step := t.setStep()
```
jortel, line 205: Please explain how this change is related to and required for this feature.  Else, revert.

pkg/controller/migmigration/task.go
```diff
@@ -184,11 +202,12 @@ func (t *Task) Run() error {
...
 
 	// Run the current phase.
 	switch t.Phase {
 	case Created, Started:
-		t.next()
+		t.next(step)
```
jortel, line 210: Please explain how this change is related to and required for this feature.  Else, revert.

pkg/controller/migmigration/task.go
```diff
@@ -572,8 +614,8 @@ func (t *Task) init() {
 	}
 }
 
-// Advance the task to the next phase.
-func (t *Task) next() {
+// Set the step which will be executed.
+func (t *Task) setStep() int {
```
jortel, line 618: Please explain how this change is related to and required for this feature.  Else, revert.

pkg/controller/migmigration/task.go
```diff
@@ -583,14 +625,20 @@ func (t *Task) next() {
...
-		return
+		t.Phase = t.Itinerary[current].phase
 	}
-	for n := current + 1; n < len(t.Itinerary); n++ {
+
+	return current
+}
+
+// Advance the task to the next phase.
+func (t *Task) next(currentPhaseIndex int) {
```
jortel, line 640: Please explain how this change is related to and required for this feature.  Else, revert.

pkg/controller/migmigration/task.go
Danil-Grigorev: This is simple. That's the first bullet in the PR description - we need to cancel immediately, without allowing a stage to finish. In case it is waiting for backup, it will block for inadequate amount of time, and only then start the cancel process.

Danil-Grigorev: @jortel In other words, this is the implementation for Itinerary swaps, which happens on Failure or Cancel.

Danil-Grigorev:
@jortel I added all related code into `gvk/gvk.go` for re-usability and refactored it a little. Please review

pkg/controller/migmigration/task.go
jortel: The bullet explains the desired behavior but not the need for the infrastructure change. 
Currently:
Each phase implementation is designed to run quickly.  Nothing waits.
At the beginning of each reconcile, Task.Run(), Task.init() determines the appropriate itinerary. 
The `Spec.Cancel` will trigger a reconcile and _new_ will immediately begin processing.
This code in `next()`:
```
	if current == -1 {
		if t.canceled() {
			t.Phase = CancelItinerary[0].phase
		} else {
			t.Phase = Completed
		}
		return
	}
```
Probably needs to get moved to `init()`.  I'm thinking we may need to store the itinerary in the Status or in an annotation so that `init()` can detect when the itinerary has changed and reset the Phase to the first phase.

jortel: The itinerary will need to change as:
```
type Itinerary struct {
	Name string
	Steps []Step
}

var StageItinerary = Itinerary{
	Name: "Stage",
	Steps: []Step{
		{phase: Created},
		{phase: Started},
		{phase: Prepare},
		{phase: EnsureCloudSecretPropagated},
		{phase: AnnotateResources, all: HasPVs},
		{phase: EnsureStagePods, all: HasPVs},
		{phase: StagePodsCreated, all: HasStagePods},
		{phase: RestartRestic, all: HasStagePods},
		{phase: ResticRestarted, all: HasStagePods},
		{phase: QuiesceApplications, all: Quiesce},
		{phase: EnsureQuiesced, all: Quiesce},
		{phase: EnsureStageBackup, all: HasPVs},
		{phase: StageBackupCreated, all: HasPVs},
		{phase: EnsureStageBackupReplicated, all: HasPVs},
		{phase: EnsureStageRestore, all: HasPVs},
		{phase: StageRestoreCreated, all: HasPVs},
		{phase: EnsureStagePodsDeleted, all: HasStagePods},
		{phase: EnsureStagePodsTerminated, all: HasStagePods},
		{phase: EnsureAnnotationsDeleted, all: HasPVs},
		{phase: Completed},
	},
}
...
```
So that it can be named.

pkg/controller/migmigration/task.go
```diff
@@ -60,68 +64,97 @@ const (
...
-	{phase: EnsureStageBackup, all: HasPVs},
-	{phase: StageBackupCreated, all: HasPVs},
-	{phase: EnsureStageBackupReplicated, all: HasPVs},
-	{phase: EnsureStageRestore, all: HasPVs},
-	{phase: StageRestoreCreated, all: HasPVs},
-	{phase: EnsureStagePodsDeleted, all: HasStagePods},
-	{phase: EnsureStagePodsTerminated, all: HasStagePods},
-	{phase: EnsureAnnotationsDeleted, all: HasPVs},
-	{phase: Completed},
+	Name: "StageItinerary",
```
jortel, line 73: Please let's not include `Itinerary` as part of name.  It's redundant and results in:
```
Status:
  itinerary: "StageItinerary"
```

pkg/controller/migmigration/task.go
```diff
@@ -563,35 +617,36 @@ func (t *Task) Run() error {
...
+	} else if t.canceled() {
 		t.Itinerary = CancelItinerary
 	} else if t.stage() {
 		t.Itinerary = StageItinerary
 	} else {
 		t.Itinerary = FinalItinerary
 	}
+	if t.Owner.Status.Itenerary != t.Itinerary.Name {
+		t.Phase = t.Itinerary.Steps[0].phase
+	}
```
jortel, line 631: :+1:

pkg/controller/migmigration/migrate.go
```diff
@@ -83,6 +83,7 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (time.Du
 
 	// Result
 	migration.Status.Phase = task.Phase
+	migration.Status.Itenerary = task.Itinerary.Name
```
jortel, line 86: :+1:

pkg/controller/migmigration/restore.go
```diff
@@ -277,6 +301,7 @@ func (t *Task) deleteRestores() error {
 	for _, restore := range list.Items {
 		err = client.Delete(context.TODO(), &restore)
 		if err != nil && !k8serror.IsNotFound(err) {
+			log.Trace(err)
```
jortel, line 304: Extra `Trace()` added.  Already logged.

eriknelson:
Guys I thought we had agreed PRs to the controller that also depend on operator changes like CRD changes had to both be posted and merged in an coordinated effort. Am I clear this requires an operator PR that has not been posted yet?

CC: @jmontleon

jmontleon:
Why did this get ACKd with CRD changes without a corresponding ACKd operator PR? Are there permission changes required as well?

djwhatle:
@eriknelson @jmontleon yes, I think this needs mig-operator changes to CRDs and also perhaps some additional RBAC permissions around sending DeleteCollectionRequests. 

I think it was a mistake that this got merged before the appropriate mig-operator PRs were also posted and approved. I'll propose 3 steps for us to follow to avoid this in the future.

---

### **Proposed steps to follow for merging a mig-controller PR that depends on mig-operator changes:**
1) mig-controller PRs should not be marked as approved until the accompanying mig-operator PR is linked and approved 
2) when the appropriate mig-operator PR is approved , then the mig-controller PR can be approved
3) once all PRs are approved, the author of said PRs is responsible for merging all PRs in appropriate order

Are we in agreement on these steps?

eriknelson:
Honestly I'd really like to explore some kind of automated linter that's integrated with github, because I don't think a process that relies on people to "just remember" is going to be sufficient; I'm certainly guilty of this exact thing on a number of occasions. Maybe that's too ambitious or not even possible, but at least worth discussing. Also want to point out this is an issue across all our repos, not just the controller.

I think this is a good topic for the scrum of scrum calls or a dedicated brainstorming call.

djwhatle:
@eriknelson @jmontleon @Danil-Grigorev bumping, I looked again and noticed that Danil linked to his operator changes in the original PR description above: https://github.com/konveyor/mig-operator/pull/293

I think we are missing mig-operator changes for the newest changes introduced in this PR.

eriknelson:
Saw the link and it was merged but aren't we still missing something to do with an itinerary? We weren't aware of that until Jeff mentioned it on a scrum call this morning.

So my questions are:
- Are there still CRD changes that need to be added to the operator?
- Is an overall deployment broken without these changes? 

⁣Get BlueMail for Android ​

PR #460, opened 2020-03-23
Added typed conversion apps/v1 and batch/v1beta1

Danil-Grigorev:
This PR allows to do a safe conversion back and forth between api versions.

The conversions are driven by comments
```golang
// +k8s:conversion-gen=k8s.io/api/batch/v2alpha1 - conversion input
// +k8s:conversion-gen-external-types=k8s.io/api/batch/v1beta1 - conversion output (hub version)
```

Each time a `client` will need to add an additional down/up convert logic, the `pkg/compat/conversion/register.go ` should be updated with api version
```bash
.
├── appsv1 # new api version 
│   ├── conversion_generated.go # output from conversion-gen
│   ├── conversion_manual.go # manual conversion of missing fields, added `localSchemeBuilder` method for compatibility
│   └── doc.go # should contain // +k8s:conversion-gen* comments
├── batchv1beta
│   ├── conversion_generated.go
│   ├── conversion_manual.go
│   └── doc.go
└── register.go # registering new conversion functions in the scheme
```


Makefile
```diff
@@ -46,9 +46,13 @@ vet:
 	go vet ./pkg/... ./cmd/...
 
 # Generate code
-generate:
+generate: conversion-gen
 	go generate ./pkg/... ./cmd/...
 
+# Generate conversion functions
+conversion-gen:
```
jortel: Why the separate make _target_ for this instead of adding to the `generate` target?

pkg/compat/conversion/batchv1beta/conversion_generated.go
```diff
@@ -0,0 +1,261 @@
+// +build !ignore_autogenerated
```
jortel: A k8s conventions seems to be to prefix generated files with `zz_` should we do that here?

jortel:
I'm a little concerned about the gen binaries in `hack` but not sure what else to do.

pkg/compat/conversion/batchv1beta/conversion_generated.go
Danil-Grigorev: That could be adjusted via argument. Will fix it now

Makefile
Danil-Grigorev: In case we won't need to run the task every time, dependency could be removed from the generate target

pkg/compat/conversion/batchv1beta/conversion_generated.go
Danil-Grigorev: @jortel Added `zz_` as a prefix

PR #465, opened 2020-03-26
Allowing to delete backup and restore, once migration is cancelled.

### LINKED ISSUE #237: Add ability to cancel a Stage/Migrate in process

Danil-Grigorev:
Follow-up on `MIG-118 - Canceling an ongoing migration`, addition to #458, closes #237 

By cancelling migration, controller will remove:
1) backup content from remote storage (including snapshots and PV data captured by restic)
2) backup resources from the source cluster
3) restores from the destination cluster

### DIFF
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -9,7 +9,6 @@ import (
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	"github.com/pkg/errors"
 	velero "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
-	"github.com/vmware-tanzu/velero/pkg/label"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -289,12 +288,10 @@ func (t *Task) deleteBackups() error {
 		return err
 	}
 
-	labels := t.Owner.GetCorrelationLabels()
-	labels["migmigration"] = string(t.Owner.GetUID())
 	list := velero.BackupList{}
 	err = client.List(
 		context.TODO(),
-		k8sclient.MatchingLabels(labels),
+		k8sclient.MatchingLabels(t.Owner.GetCorrelationLabels()),
 		&list)
 	if err != nil {
 		log.Trace(err)
@@ -302,7 +299,15 @@ func (t *Task) deleteBackups() error {
 	}
 
 	for _, backup := range list.Items {
-		request := buildDeleteBackupRequest(backup.Name, backup.UID)
+		request := &velero.DeleteBackupRequest{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace:    migapi.VeleroNamespace,
+				GenerateName: backup.Name + "-",
+			},
+			Spec: velero.DeleteBackupRequestSpec{
+				BackupName: backup.Name,
+			},
+		}
 		if err := client.Create(context.TODO(), request); err != nil {
 			log.Trace(err)
 			return err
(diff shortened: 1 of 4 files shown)


pkg/controller/migmigration/restore.go
```diff
@@ -232,6 +233,41 @@ func (t *Task) updateRestore(restore *velero.Restore, backupName string) {
 	t.updateNamespaceMapping(restore)
 }
 
+func (t *Task) cleanRestores() error {
```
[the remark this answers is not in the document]
jortel: Seems like this could be simpler using the correlation tags.
Instead of deleting specific restores, how about listing all restores created by this migration and delete them (in a loop) using the correlation labels?

pkg/controller/migmigration/task.go
```diff
@@ -116,6 +117,7 @@ var CancelItinerary = Itinerary{
 	{phase: Canceling},
 	{phase: EnsureStagePodsDeleted, flags: HasStagePods},
 	{phase: EnsureAnnotationsDeleted, flags: HasPVs},
+	{phase: BackupDeletionRequested},
```
jortel: Shouldn't the restores get deleted on the destination when a failed migration is canceled?

pkg/controller/migmigration/task.go
```diff
@@ -45,6 +45,7 @@ const (
 	EnsureStagePodsDeleted        = "EnsureStagePodsDeleted"
 	EnsureStagePodsTerminated     = "EnsureStagePodsTerminated"
 	EnsureAnnotationsDeleted      = "EnsureAnnotationsDeleted"
+	BackupDeletionRequested       = "RequestBackupDeletion"
```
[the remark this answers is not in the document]
Danil-Grigorev: @jortel What about `DeleteBackupsAndRestores`? This is just a generic clean-up stage which does not even wait for anything. Getting a good name for it is hard.

Danil-Grigorev:
@jortel All requested changes are updated. Please review.

pkg/controller/migmigration/backup.go
```diff
@@ -281,6 +282,52 @@ func (t *Task) updateBackup(backup *velero.Backup) error {
...
+		context.TODO(),
+		k8sclient.MatchingLabels(labels),
+		&list)
+	if err != nil {
+		log.Trace(err)
+		return err
+	}
+
+	for _, backup := range list.Items {
+		request := buildDeleteBackupRequest(backup.Name, backup.UID)
```
jortel: The `buildDeleteBackupRequest()` function is effectively 1 line of code and only used here.  Seems simpler to just create the request here.

pkg/controller/migmigration/restore.go
```diff
@@ -232,6 +233,35 @@ func (t *Task) updateRestore(restore *velero.Restore, backupName string) {
...
 
+func (t *Task) deleteRestores() error {
+	client, err := t.getDestinationClient()
+	if err != nil {
+		log.Trace(err)
+		return err
+	}
+
+	labels := t.Owner.GetCorrelationLabels()
+	labels["migmigration"] = string(t.Owner.GetUID())
```
jortel: This should be already included in the migration correlation labels.

pkg/controller/migmigration/backup.go
```diff
@@ -281,6 +282,52 @@ func (t *Task) updateBackup(backup *velero.Backup) error {
...
 
+func (t *Task) deleteBackups() error {
+	client, err := t.getSourceClient()
+	if err != nil {
+		log.Trace(err)
+		return err
+	}
+
+	labels := t.Owner.GetCorrelationLabels()
+	labels["migmigration"] = string(t.Owner.GetUID())
```
jortel: This should already be included in the correlation labels.

pkg/controller/migmigration/backup.go
```diff
@@ -281,6 +282,52 @@ func (t *Task) updateBackup(backup *velero.Backup) error {
...
+}
+
+func buildDeleteBackupRequest(name string, uid types.UID) *velero.DeleteBackupRequest {
+	return &velero.DeleteBackupRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace:    migapi.VeleroNamespace,
+			GenerateName: name + "-",
+			Labels: map[string]string{
+				velero.BackupNameLabel: label.GetValidName(name),
+				velero.BackupUIDLabel:  string(uid),
```
jortel: What are these labels for?

Danil-Grigorev: These are recognized by velero, and it is a copy of a method from  https://github.com/vmware-tanzu/velero/blob/3e1b8e01a7d96f84ff9016445d5c4e36060f7348/pkg/backup/delete_helpers.go

Including this package directly evolves vendoring core velero methods, so assuming that is the only place where DeleteRequest will be used, I decided to include it here.

pkg/controller/migmigration/backup.go
Danil-Grigorev: Thanks, I didn't realize that :(

pkg/controller/migmigration/backup.go
Danil-Grigorev: Actually, there is no need for those, as the labels will be set during request processing: https://github.com/konveyor/mig-controller/blob/60c8bcae42e10d89938137bb58b0cf8b6ad4d811/vendor/github.com/vmware-tanzu/velero/pkg/controller/backup_deletion_controller.go#L238-L258

PR #466, opened 2020-03-26
Added Tiltfile for production like make run

Danil-Grigorev:
Depends on https://github.com/konveyor/mig-operator/pull/279

Peek 2020-03-26 21-20


dymurray:
@djwhatle @jortel and opposition to adding this tool to our Makefile?

I had never heard of tilt before but seems to work well. Only hang up for me is I never like running curl install scripts like that. Would prefer to find something packaged as an rpm.

djwhatle:
@Danil-Grigorev I was just thinking the other day that it would be great to have some automation around rolling out changes to mig-controller within the mig-operator flow. This is really cool, I'll try it out.

jortel:
This seems to automate a process having relatively few, simple steps that I only do once at the end of each feature/bug development cycle just prior to opening the PR.  It does not seem worth maintaining, IMHO.
No objection to merge if other find it useful.

djwhatle:
@jortel I'd argue that we might be able to avoid more mig-operator issues if devs are actually testing in the mig-operator context as opposed to using `make run`. This lowers the friction to testing in a "real world" scenario. 

My questions would be:
1) Could we accomplish the same thing with a few Makefile additions and skip requiring a new tool that isn't installed on anyone's machine by default
2) Does the new tool add substantial value over adding a bit more to our Makefile

jortel:
> @jortel I'd argue that we might be able to avoid more mig-operator issues if devs are actually testing in the mig-operator context as opposed to using `make run`. This lowers the friction to testing in a "real world" scenario.

I 100% agree that developers must test in the deployed _production_ environment.  I just don't see the need for automation.

dymurray:
I would agree with Jeff that for my particular development workflow I don't know that I would use this a ton but if others find it useful I find no harm in merging it in. 

I generally just configure my operator with an image I control and push it there and delete the pod.

Danil-Grigorev:
@djwhatle @dymurray This particular toll will help you guys to test new PRs in this repo, reducing the time cost and eliminating the error prone redeployment process of building an image, patching the migrationController and waiting for pods to come up. This way if you'd, say for example, have couple of PRs out you need to test in production, all you would need to do, is to `make tilt` and then checkout PR branches one after another. You won't need to verify if the image is up-to-date, if the spelling is correct, and the pods are not crash-looping. This all will be done automatically.

I don't see myself running `tilt up` during development process that much. The behaviour of a live update / image push is useful in scenario described earlier, and will help newcomers to test their change out with less effort.

To clarify from where the effort started from, I wanted to add that, running the code via `make tilt` identifies all the issues with production account permissions, which will not be otherwise seen in `make run` scenario, where the local `kubeconfig` file is used, with `cluster-admin` permissions by default.

@jortel Addressing your concerns for maintaining the feature - there should be no need for that, as the parts which are used in the automaton are not moving too fast. At least I assume that `migrationController` resource will not go away, and we won't remove `Dockerfile` used to build the image right now.

dymurray:
Thinking more about this, is it necessary to introduce a new tool to accomplish this automation? When I first looked at this I thought the value-add with Tilt is automating an action (building/deploying) on code change. Now, I would say that for Go I don't generally find this useful as I do for Javascript development. The actual automation tasks in this PR are fairly simple:
1. Build image
2. Push image
3. Modify migrationcontroller CR with image value
4. Redeploy the pods.

This isn't a ton of overhead which is why I think I would keep my current development workflow rather than introducing a new tool into my setup.

Here's another option, couldn't we write a simple Ansible playbook that does the above steps and put it into a `hack` dir? Then we could accomplish the same thing with a tool that ships with Fedora/RHEL. Thoughts?

I think I'm just not a fan of introducing a new tool if we aren't going to have all developers adopt it.

Danil-Grigorev:
> Thinking more about this, is it necessary to introduce a new tool to accomplish this automation? When I first looked at this I thought the value-add with Tilt is automating an action (building/deploying) on code change. Now, I would say that for Go I don't generally find this useful as I do for Javascript development. The actual automation tasks in this PR are fairly simple:
> 
>     1. Build image
> 
>     2. Push image
> 
>     3. Modify migrationcontroller CR with image value
> 
>     4. Redeploy the pods.

This is not so simple. We wait for pods to redeploy, and to know which pods are the new ones we expect to see, and which are the old ones, requires a check on the `image` sha sum, which should be stored somewhere during previous steps. In ansible that would not be good looking, and the Tilt supports this out of the box.

> 
> This isn't a ton of overhead which is why I think I would keep my current development workflow rather than introducing a new tool into my setup.

So none of you guys are having an issue with working on UI part as well as controller part at the same time, and apparently I'm alone who is dealing with this issue of testing controller in production with UI or during development of a small user-facing changes. That often requires making small adjustments to a usual controller workflow, in order to isolate the feature and see the response in UI before submitting a PR.

For example, @jortel 's discovery integration will not work with a locally running controller (`make run` case) out of the box. You need to adjust the roles to point on the ip address of your machine, and also ensure you have the ports opened and it is accessible form your cluster... - https://github.com/konveyor/mig-operator/blob/c8f92ec0bfb589b523a799957eaa6aabe8244ad8/roles/migrationcontroller/tasks/main.yml#L417

With up-to-date image this not that big of a deal.

> Here's another option, couldn't we write a simple Ansible playbook that does the above steps and put it into a `hack` dir? Then we could accomplish the same thing with a tool that ships with Fedora/RHEL. Thoughts?

This is not a part of the code we are going to ship, it is purely for our convenience. So I don't see this as a hard requirement.

> I think I'm just not a fan of introducing a new tool if we aren't going to have all developers adopt it.

Tilt allows some fancy stuff like intelligent waiting for resource to deploy, which otherwise is needed to be done manually via some ansible code, which tends to break very often and actually requires maintenance.

Danil-Grigorev:
@djwhatle This PR already provides the functionality of #479, does not need maintenance, it is indifferent to changes of `migration-controller` SA name or any inconsistencies with the current way of deploying `mig-controller` as it is using mig-operator repo.

djwhatle:
Trying this is still in my backlog.

Danil-Grigorev:
I was heavily using this tool recently for testing, and encourage everyone to do so, who does not have obligations to try an optional tool to simplify development process. Tilt is heavily used in kubernetes community, and it is pretty intelligent under the hood.

PR #473, opened 2020-03-31
Add support for proxy environment

dymurray:


eriknelson:
Just making note, the transport change on this PR to the S3 check is also critical to proxy functionality working. Instantiating your own Transport and only setting something like the TLS config will **zero out** the other settings that are normally on a default transport, like timeout settings and the default behavior to proxy from environment. It's important to use the default transport and its settings, and to override your selected fields to preserve default behavior, as Dylan has done here.

PR #475, opened 2020-03-31
Replaced ObservedGeneration with ObservedDigest.

jortel:
The `generation` field is not managed (incremented) in OCP3 so cannot apply the `observedGeneration` pattern.
Instead, generating a SHA256 (hex) digest for the `Spec` and storing in the status as: `observedDigest`.  Gives us most of the same benefits and compatible with older clusters.

Requires: https://github.com/konveyor/mig-operator/pull/291

### DIFF
--- a/config/crds/migration_v1alpha1_migcluster.yaml
+++ b/config/crds/migration_v1alpha1_migcluster.yaml
@@ -61,9 +61,8 @@ spec:
           type: object
         status:
           properties:
-            observedGeneration:
-              format: int64
-              type: integer
+            observedDigest:
+              type: string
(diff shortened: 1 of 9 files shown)


jortel:
> I know we were previously tweaking observedGen to trigger a reconcile but I don’t see any mention of that here, is that piece gone from the code base already?
We discussed doing that but I don't think the changes in the predicates to use `HasReconciled()` instead of `DeepCopy()` of the spec got made.  In any case, this change does not impact that support.  Instead of decrementing the observedGeneration, changing (perhaps to "") would be equivalent.

PR #476, opened 2020-04-01
Ignore cancelled final migration in validation step

Danil-Grigorev:
- Prevent from blocking non-cancelled final migration from being executed.
- Also includes some spelling fixes, required by @vconzola's UX design

Issue was discovered while testing a fix for https://github.com/konveyor/mig-ui/issues/753

### DIFF
--- a/pkg/controller/migmigration/validation.go
+++ b/pkg/controller/migmigration/validation.go
@@ -11,10 +11,8 @@ const (
 	InvalidPlanRef      = "InvalidPlanRef"
 	PlanNotReady        = "PlanNotReady"
 	PlanClosed          = "PlanClosed"
-	Cancelling          = "Cancelling"
 	HasFinalMigration   = "HasFinalMigration"
 	Postponed           = "Postponed"
-	Canceled            = "Canceled"
 	Running             = "Running"
 	Succeeded           = "Succeeded"
 	Failed              = "Failed"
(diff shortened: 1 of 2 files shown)


pkg/controller/migmigration/validation.go
```diff
@@ -152,10 +150,8 @@ func (r ReconcileMigMigration) validateFinalMigration(plan *migapi.MigPlan, migr
 	}
 	hasCondition := false
 	for _, m := range migrations {
-		if m.UID == migration.UID {
-			continue
-		}
-		if !!m.Spec.Stage {
+		// ignore canceled
```
jortel, line 153: Recommend ordering the check by importance and reflecting in the comment.
```
// Ignored: `This` migration, canceled and stage.
if m.UID == migration.UID || m.Spec.Canceled || m.Spec.Stage {
    continue
}
```

pkg/controller/migmigration/validation.go
```diff
@@ -152,10 +150,8 @@ func (r ReconcileMigMigration) validateFinalMigration(plan *migapi.MigPlan, migr
 	}
 	hasCondition := false
 	for _, m := range migrations {
-		if m.UID == migration.UID {
-			continue
-		}
-		if !!m.Spec.Stage {
```
jortel, line 158: Ha!  `!!m.Spec.Stage`.  That's nice :)

PR #478, opened 2020-04-02
Update migplan controller to take in identity secret

### LINKED ISSUE #389: [non-admin] Enable migplan controller to take in a user's identity token for src + dest
As part of the non-admin work, we need the controller and the discovery svc to be able to identify the user which initiated the migration. This means that when a `migplan` is created, there needs to be a field in the spec to allow the user to specify an identity token. This token will be used for all actions the controller makes. The discovery svc and controller are also expected to perform SAR with this token to ensure that the originating user is not creating the migplan with a set of namespaces the user does not have access to.

Migplan validation should verify:
1. Identity Secret Ref exists and is a valid secret ref
1. Identity Secret contents are valid and properly populated
1. SrcToken on Identity secret is capable of `get` access on the selected list of namespaces

dymurray:
Fixes #389 

Also adds the design document in here.

Related operator PR: https://github.com/konveyor/mig-operator/pull/304

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	migauth "github.com/konveyor/mig-controller/pkg/auth"
 	pvdr "github.com/konveyor/mig-controller/pkg/cloudprovider"
 	migref "github.com/konveyor/mig-controller/pkg/reference"
 	"github.com/konveyor/mig-controller/pkg/settings"
@@ -56,6 +57,7 @@ type MigPlanSpec struct {
 	SrcMigClusterRef  *kapi.ObjectReference `json:"srcMigClusterRef,omitempty"`
 	DestMigClusterRef *kapi.ObjectReference `json:"destMigClusterRef,omitempty"`
 	MigStorageRef     *kapi.ObjectReference `json:"migStorageRef,omitempty"`
+	IdentitySecretRef *kapi.ObjectReference `json:"identitySecretRef,omitempty"`
 	Closed            bool                  `json:"closed,omitempty"`
 }
 
@@ -98,10 +100,102 @@ func (r *MigPlan) GetSourceCluster(client k8sclient.Client) (*MigCluster, error)
 	return GetCluster(client, r.Spec.SrcMigClusterRef)
 }
 
+// GetSourceIdentity - Get the source identity object using the secret
+// Returns nil if unable to build
+func (r *MigPlan) GetSourceIdentity(client k8sclient.Client) (*migauth.Identity, error) {
+	srcToken, err := r.GetSourceIdentityToken(client)
+	if err != nil {
+		return nil, err
+	}
+	srcCluster, err := r.GetSourceCluster(client)
+	if err != nil {
+		return nil, err
+	}
+	if srcCluster == nil || !srcCluster.Status.IsReady() {
+		return nil, errors.New("source cluster is not in a ready state")
+	}
+	srcRestCfg, err := srcCluster.BuildRestConfig(client)
+	if err != nil {
+		return nil, err
+	}
+	srcIdentity := &migauth.Identity{
+		Token:   srcToken,
+		RestCfg: *srcRestCfg,
+	}
+	return srcIdentity, nil
+}
+
+// GetDestinationIdentity - Get the destination identity object using the secret
+// Returns nil if unable to build
+func (r *MigPlan) GetDestinationIdentity(client k8sclient.Client) (*migauth.Identity, error) {
+	destToken, err := r.GetDestinationIdentityToken(client)
+	if err != nil {
+		return nil, err
+	}
+	destCluster, err := r.GetDestinationCluster(client)
+	if err != nil {
+		return nil, err
+	}
+	if destCluster == nil || !destCluster.Status.IsReady() {
+		return nil, errors.New("destination cluster is not in a ready state")
+	}
+	destRestCfg, err := destCluster.BuildRestConfig(client)
+	if err != nil {
+		return nil, err
+	}
+	destIdentity := &migauth.Identity{
+		Token:   destToken,
+		RestCfg: *destRestCfg,
+	}
+	return destIdentity, nil
+}
+
 // GetDestinationCluster - Get the referenced destination cluster.
 // Returns `nil` when the reference cannot be resolved.
 func (r *MigPlan) GetDestinationCluster(client k8sclient.Client) (*MigCluster, error) {
 	return GetCluster(client, r.Spec.DestMigClusterRef)
+}
+
+// GetSourceIdentity - Get the source identity token
+// Returns "" if not found
+func (r *MigPlan) GetSourceIdentityToken(client k8sclient.Client) (string, error) {
+	secret, err := r.GetIdentitySecret(client)
+	if err != nil {
+		return "", err
+	}
+	if secret == nil {
+		return "", errors.New("identity secret not found")
+	}
+	if secret.Data == nil {
+		return "", errors.New("identity secret misconfigured")
+	}
+	if secret.Data["srcToken"] == nil {
+		return "", errors.New("identity secret misconfigured")
+	}
+	return string(secret.Data["srcToken"]), nil
+}
+
+// GetDestinationIdentity - Get the destination identity token
+// Returns "" if not found
+func (r *MigPlan) GetDestinationIdentityToken(client k8sclient.Client) (string, error) {
+	secret, err := r.GetIdentitySecret(client)
+	if err != nil {
+		return "", err
+	}
+	if secret == nil {
+		return "", errors.New("identity secret not found")
+	}
+	if secret.Data == nil {
+		return "", errors.New("identity secret misconfigured")
+	}
+	if secret.Data["destToken"] == nil {
+		return "", errors.New("identity secret misconfigured")
+	}
+	return string(secret.Data["destToken"]), nil
+}
+
+func (r *MigPlan) GetIdentitySecret(client k8sclient.Client) (*kapi.Secret, error) {
+	return GetSecret(client, r.Spec.IdentitySecretRef)
 }
 
 // GetStorage - Get the referenced storage..
@@ -146,6 +240,21 @@ func (r *MigPlan) GetRefResources(client k8sclient.Client) (*PlanResources, erro
 	}
 	if destMigCluster == nil {
 		return nil, errors.New("destination cluster not found")
+	}
+
+	// Identity
+	identitySecret, err := r.GetIdentitySecret(client)
+	if err != nil {
+		return nil, err
+	}
+	if identitySecret == nil {
+		return nil, errors.New("identity secret not found")
+	}
+	if identitySecret.Data["srcToken"] == nil {
+		return nil, errors.New("identity secret doesn't contain source token")
+	}
+	if identitySecret.Data["destToken"] == nil {
+		return nil, errors.New("identity secret doesn't contain source token")
 	}
 
 	resources := &PlanResources{
--- a/pkg/auth/identity.go
+++ b/pkg/auth/identity.go
@@ -1,0 +1,84 @@
+package auth
+
+import (
+	"context"
+
+	"k8s.io/api/authentication/v1beta1"
+	authapi "k8s.io/api/authorization/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Authorized map[string]bool
+
+type Identity struct {
+	Token   string
+	RestCfg rest.Config
+	client  k8sclient.Client
+}
+
+func (r *Identity) HasRead(namespaces []string) (Authorized, error) {
+	authorized := Authorized{}
+	err := r.buildClient()
+	if err != nil {
+		return authorized, err
+	}
+	for _, namespace := range namespaces {
+		sar := authapi.SelfSubjectAccessReview{
+			Spec: authapi.SelfSubjectAccessReviewSpec{
+				ResourceAttributes: &authapi.ResourceAttributes{
+					Resource:  "namespaces",
+					Namespace: namespace,
+					Verb:      "get",
+				},
+			},
+		}
+		err := r.client.Create(context.TODO(), &sar)
+		if err != nil {
+			return authorized, err
+		}
+		authorized[namespace] = sar.Status.Allowed
+	}
+
+	return authorized, nil
+}
+
+func (r *Identity) HasMigrate(namespaces []string) (Authorized, error) {
+	authorized := Authorized{}
+	err := r.buildClient()
+	if err != nil {
+		return authorized, err
+	}
+
+	return authorized, nil
+}
+
+func (r *Identity) Authenticates(client k8sclient.Client) (bool, error) {
+	tokenReview := v1beta1.TokenReview{
+		Spec: v1beta1.TokenReviewSpec{
+			Token: r.Token,
+		},
+	}
+	err := client.Create(context.TODO(), &tokenReview)
+	if err != nil {
+		return false, err
+	}
+	return tokenReview.Status.Authenticated, nil
+}
+
+func (r *Identity) buildClient() error {
+	if r.client != nil {
+		return nil
+	}
+	// build client using r.RestCfg and replacing the token with r.Token.
+	restCfg := r.RestCfg
+	restCfg.BearerToken = r.Token
+	client, err := k8sclient.New(&restCfg, k8sclient.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		return err
+	}
+	r.client = client
+
+	return nil
+}
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -2,6 +2,7 @@ package migplan
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"reflect"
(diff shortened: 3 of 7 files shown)


pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -100,6 +102,34 @@ func (r *MigPlan) GetDestinationCluster(client k8sclient.Client) (*MigCluster, e
...
+		Spec: v1beta1.TokenReviewSpec{
+			Token: string(secret.Data["srcToken"]),
+		},
+	}
+	err = client.Create(context.TODO(), &tokenReview)
+	if err != nil {
+		return "", err
+	}
+	if !tokenReview.Status.Authenticated {
+		return "", errors.New("token did not authenticate with source cluster")
```
jortel, line 123: Small suggestion: Given the scope of the nature of the method, seems `token not authenticated` would be more concise and sufficient.

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -161,3 +161,27 @@ func GetSecret(client k8sclient.Client, ref *kapi.ObjectReference) (*kapi.Secret
 
 	return &object, err
 }
+
+// Read identity secret and get token
+func LoadIdentitySecret(client k8sclient.Client, ref *kapi.SecretReference) (*kapi.Secret, error) {
```
jortel, line 166: Any reason we can't use: `GetSecret()` provided in `model.go` instead?

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -100,6 +102,34 @@ func (r *MigPlan) GetDestinationCluster(client k8sclient.Client) (*MigCluster, e
...
 
+// GetSourceIdentity - Get the corresponding user for a given source identity
+// token
+// Returns "" if the srcToken doesn't authenticate
+func (r *MigPlan) GetSourceIdentity(client k8sclient.Client) (string, error) {
+	secret, err := r.GetIdentitySecret(client)
+	if err != nil {
+		return "", err
+	}
+	tokenReview := v1beta1.TokenReview{
```
jortel, line 113: I'm not convinced that authenticating the token each time a controller _gets_ it is a good idea.  The cluster token is only authenticated during validation (reconcile) and reports as a condition.  Seems like code using the token will just try to use it and stack trace either way.  Not sure it's worth the additional overhead.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -52,6 +53,7 @@ type MigPlanSpec struct {
 	SrcMigClusterRef  *kapi.ObjectReference `json:"srcMigClusterRef,omitempty"`
 	DestMigClusterRef *kapi.ObjectReference `json:"destMigClusterRef,omitempty"`
 	MigStorageRef     *kapi.ObjectReference `json:"migStorageRef,omitempty"`
+	IdentitySecretRef *kapi.SecretReference `json:"identitySecretRef,omitempty"`
```
jortel, line 56: :+1:

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -100,6 +102,34 @@ func (r *MigPlan) GetDestinationCluster(client k8sclient.Client) (*MigCluster, e
...
+// GetSourceIdentity - Get the corresponding user for a given source identity
+// token
+// Returns "" if the srcToken doesn't authenticate
+func (r *MigPlan) GetSourceIdentity(client k8sclient.Client) (string, error) {
+	secret, err := r.GetIdentitySecret(client)
+	if err != nil {
+		return "", err
+	}
+	tokenReview := v1beta1.TokenReview{
+		Spec: v1beta1.TokenReviewSpec{
```
jortel, line 114: Note to add TokenReview to RBAC in operator.

pkg/controller/migmigration/backup.go
```diff
@@ -267,11 +267,11 @@ func (t *Task) updateBackup(backup *velero.Backup) error {
...
-		TTL:                     metav1.Duration{Duration: 720 * time.Hour},
-		IncludedNamespaces:      t.sourceNamespaces(),
-		ExcludedNamespaces:      []string{},
-		IncludedResources:       t.BackupResources,
-		ExcludedResources:       []string{},
+		TTL:                metav1.Duration{Duration: 720 * time.Hour},
+		IncludedNamespaces: t.sourceNamespaces(),
+		ExcludedNamespaces: []string{},
+		IncludedResources:  t.BackupResources,
+		ExcludedResources:  []string{},
```
jortel, line 274: Please run `make fmt` to avoid IDE formatting changes that I think will be contradictory.

dymurray: Yeah weirdly my version of `go fmt` has always modified this file. I accidentally checked it in this time. I will try updating my `go fmt` to see if it resolves this.

pkg/controller/migplan/validation.go
```diff
@@ -581,6 +592,83 @@ func (r ReconcileMigPlan) validateConflict(plan *migapi.MigPlan) error {
...
+			Message:  InvalidNamespaceSelectionMessage,
+			Items:    invalidNamespaces,
+		})
+		return errors.New("selected namespaces are invalid")
+	}
+
+	return nil
+}
+
+func canUserAccess(client k8sclient.Client, user, namespace string) (bool, error) {
```
jortel, line 651: I'm wondering if putting this, and what the migration controller will need to do in a new top level `auth` package.
```
package auth

type Authorized map[string]bool

func HasRead(token string, namespaces []string) (Authorized, error)

func HasMigrate(token string, namespace []string) (Authorized, error)
```
For these reasons:
- The HasRead can be shared by the discovery controller.
- In both cases, a new client needs to be constructed and can be reused to check each namespace.
- Centralizes optimizations.
- Seems helpful to have one place to maintain how auth is done.

pkg/controller/migplan/validation.go
```diff
@@ -581,6 +592,83 @@ func (r ReconcileMigPlan) validateConflict(plan *migapi.MigPlan) error {
...
+	}
+
+	identity, err := plan.GetSourceIdentity(client)
+	if err != nil {
+		log.Trace(err)
+		return err
+	}
+
+	for i := 0; i < len(plan.Spec.Namespaces); i++ {
+		access, err := canUserAccess(client, identity, plan.Spec.Namespaces[i])
```
jortel, line 626: Pretty sure this is the wrong `client` being passed here.

pkg/controller/migplan/validation.go
```diff
@@ -243,7 +254,7 @@ func (r ReconcileMigPlan) validateStorage(plan *migapi.MigPlan) error {
 }
 
 // Validate the referenced assetCollection.
-func (r ReconcileMigPlan) validateNamespaces(plan *migapi.MigPlan) error {
+func (r ReconcileMigPlan) validateNamespaceCount(plan *migapi.MigPlan) error {
```
jortel, line 257: :+1:

pkg/controller/migplan/validation.go
```diff
@@ -581,6 +592,83 @@ func (r ReconcileMigPlan) validateConflict(plan *migapi.MigPlan) error {
 	return nil
 }
 
+// Validiate Namespace Selection
+func (r ReconcileMigPlan) validateNamespaceSelection(plan *migapi.MigPlan) error {
```
jortel, line 596: Perhaps `validateNamespaceAuthroized()`

pkg/controller/migplan/validation.go
```diff
@@ -581,6 +592,83 @@ func (r ReconcileMigPlan) validateConflict(plan *migapi.MigPlan) error {
...
+		if !access {
+			invalidNamespaces = append(invalidNamespaces, plan.Spec.Namespaces[i])
+			continue
+		}
+	}
+	if len(invalidNamespaces) > 0 {
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     InvalidNamespaceSelection,
+			Status:   True,
+			Reason:   NotDone,
```
jortel, line 640: How about `Reason: NotAuthorized`?

pkg/controller/migplan/validation.go
```diff
@@ -581,6 +592,83 @@ func (r ReconcileMigPlan) validateConflict(plan *migapi.MigPlan) error {
 	return nil
 }
 
+// Validiate Namespace Selection
+func (r ReconcileMigPlan) validateNamespaceSelection(plan *migapi.MigPlan) error {
+	invalidNamespaces := make([]string, 0)
```
jortel, line 597: Seems like `unauthorized` would be clearer.  I don't think it's a question of valid/invalid.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -100,6 +102,34 @@ func (r *MigPlan) GetDestinationCluster(client k8sclient.Client) (*MigCluster, e
...
+	}
+	err = client.Create(context.TODO(), &tokenReview)
+	if err != nil {
+		return "", err
+	}
+	if !tokenReview.Status.Authenticated {
+		return "", errors.New("token did not authenticate with source cluster")
+	}
+
+	return tokenReview.Status.User.Username, nil
```
jortel, line 126: This will not be set for SA tokens.

pkg/controller/migplan/validation.go
```diff
@@ -581,6 +592,83 @@ func (r ReconcileMigPlan) validateConflict(plan *migapi.MigPlan) error {
...
+			Items:    invalidNamespaces,
+		})
+		return errors.New("selected namespaces are invalid")
+	}
+
+	return nil
+}
+
+func canUserAccess(client k8sclient.Client, user, namespace string) (bool, error) {
+	sar := authapi.SubjectAccessReview{
```
jortel, line 652: To support SA and user identity tokens, we would need to use a `SelfSubjectAccessReview` and just deal with the token.

pkg/apis/migration/v1alpha1/migplan_types.go
dymurray: The more I think about it, is tokenReview even necessary? I thought it was due to using `SubjectAccessReview`, but in my code refactor it seems that `SelfSubjectAccessReview` allows us to simply use the token in the client without needing to determine the originating user/sa.

My current implementation isn't using it, so I believe we don't need to update any operator RBAC rules unless someone sees a need to use it.

pkg/auth/identity.go
```diff
@@ -0,0 +1,70 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Authorized map[string]bool
+
+type Identity struct {
+	Token   string
+	RestCfg *rest.Config
```
jortel, line 16: Perhaps make the `RestCfg` not be a pointer since we want it to be copy that we can modify in `buildClient()`.

pkg/controller/migplan/validation.go
```diff
@@ -79,6 +84,7 @@ const (
 	NotDistinct   = "NotDistinct"
 	LimitExceeded = "LimitExceeded"
 	NotDone       = "NotDone"
+	NotAuthorized = "NotAuthorized"
```
jortel, line 87: :+1:

pkg/controller/migplan/validation.go
```diff
@@ -37,10 +39,13 @@ const (
 	InvalidDestinationCluster                  = "InvalidDestinationCluster"
 	NsNotFoundOnSourceCluster                  = "NamespaceNotFoundOnSourceCluster"
 	NsNotFoundOnDestinationCluster             = "NamespaceNotFoundOnDestinationCluster"
+	InvalidNamespaceAuthorization              = "InvalidNamespaceAuthorization"
```
jortel, line 42: The `Invalid` seem a bit odd.  The condition is that the identity has insufficient privileges to operation on 1+ of the namespaces.  This working suggests the authorization itself is invalid.  Perhaps consider: `UnauthorizedNamespaces`.

pkg/controller/migplan/validation.go
```diff
@@ -128,6 +136,7 @@ const (
 	PvsDiscoveredMessage                              = "The `persistentVolumes` list has been updated with discovered PVs."
 	ClosedMessage                                     = "The migration plan is closed."
 	SourcePodsNotHealthyMessage                       = "Source namespace(s) contain unhealthy pods. See: `unhealthyNamespaces` for details."
+	InvalidNamespaceAuthorizationMessage              = "Namespaces [] are not authorized."
```
jortel, line 139: Suggest: "Source identity not authorized for namespaces []."

pkg/auth/identity.go
```diff
@@ -0,0 +1,70 @@
...
+	client  k8sclient.Client
+}
+
+func (r *Identity) HasRead(namespaces []string) (Authorized, error) {
```
jortel, line 20: Seems like this should only raise an error when it fails.
- General client API fail.
- Authentication failed.

The caller can use the authorized list to determine if any of the namespaces are not authorized.

Same for `HasMigrate()`.

jortel: Perhaps a separate `Authenticate()` that does a `TokenReview` and report that as a separate token authentication condition.

pkg/controller/migplan/validation.go
```diff
@@ -595,6 +611,135 @@ func (r ReconcileMigPlan) validateConflict(plan *migapi.MigPlan) error {
...
+	_, err = plan.GetDestinationIdentity(r.Client)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Sets identities on ReconcileMigPlan struct
+// Returns if already set
+func (r *ReconcileMigPlan) SetIdentities(plan *migapi.MigPlan) error {
```
jortel, line 703: Seems unexpected for this to be done here (in validation).

pkg/controller/migplan/migplan_controller.go
```diff
@@ -157,9 +158,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 
+type Identities struct {
+	SrcIdentity  *migauth.Identity
+	DestIdentity *migauth.Identity
+}
+
 // ReconcileMigPlan reconciles a MigPlan object
 type ReconcileMigPlan struct {
 	client.Client
+	Identities
```
jortel, line 169: This is a interesting convenience but a little concerned.  The pattern so far has been to pass around the reconciled resource (or related information).  I wonder if we can achieve the same convenience by adding the identities to the `PlanResources`.

That said, perhaps this is a good _new_ president.  Now that a lot of code has been written for the Reconcilers stashing the plan (and related information) may be better than passing it around everywhere.  Wonder why we never did that?

In either case, I'm not convinced the `Identities` aggregation provides much value.  Just my $0.02.
```
type ReconcileMigPlan struct {
	client.Client
	scheme *runtime.Scheme
        plan *migapi.MigPlan
        srcIdentity *migauth.Identity
        dstIdentity *migauth.Identity
```

The `plan` would be `set` at the beginning of the reconcile.  The identities could be set after validate().  both in the main (though identities probably in a separate method  like you have it outside the loop).
Sure would clean up a lot of method signatures.

What do you think?

pkg/auth/identity.go
```diff
@@ -0,0 +1,70 @@
...
+	authorized := Authorized{}
+	err := r.buildClient()
+	if err != nil {
+		return authorized, err
+	}
+	for _, namespace := range namespaces {
+		sar := authapi.SelfSubjectAccessReview{
+			Spec: authapi.SelfSubjectAccessReviewSpec{
+				ResourceAttributes: &authapi.ResourceAttributes{
```
mansam, line 29: Seems like this is asking whether the user has permission to `get` namespaces in the namespace, which doesn't quite make sense to me. Is that a valid way to check for access to a given namespace? I couldn't work that out from the SAR documentation.

dymurray: Yeah so this something weird that took me a lot of trial and error to figure out. Surprisingly specifying the resource name for namespaces doesn't work as you'd expect.

Here's some example output from the oc client:
```
[dymurray@pups mig-controller]$ oc auth can-i get namespaces/dylancantaccess
Warning: resource 'namespaces' is not namespace scoped
yes
[dymurray@pups mig-controller]$ oc auth can-i get namespaces/dylan
Warning: resource 'namespaces' is not namespace scoped
yes
[dymurray@pups mig-controller]$ oc auth can-i get namespaces -n dylancantaccess
Warning: resource 'namespaces' is not namespace scoped
no - no RBAC policy matched
[dymurray@pups mig-controller]$ oc auth can-i get namespaces -n dylan
Warning: resource 'namespaces' is not namespace scoped
yes
```

I don't know if I'm doing it wrong but I think since `namespaces` is a cluster scoped resource that you need to use the `Namespace` param in order to specify which namespace this SAR check is run against. @jortel any thoughts on this?

pkg/controller/migplan/validation.go
dymurray: Good point, I will move this to `Reconcile`

pkg/controller/migplan/migplan_controller.go
dymurray: Have a couple questions about this I will sync up with you in the morning.

pkg/auth/identity.go
jortel: I have seen the same behavior.  May be worth running past David E. at some point but think we should run with it for now.

pkg/controller/migplan/migplan_controller.go
```diff
@@ -157,9 +158,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 var _ reconcile.Reconciler = &ReconcileMigPlan{}
 
+type Identities struct {
+	SrcIdentity  *migauth.Identity
+	DestIdentity *migauth.Identity
```
jortel, line 163: All of the fields of a struct are properties of (describe) the _type_ so each filed containing the name (form of) of the _type_ is redundant.  This reads like: `Identities.dstIdentity`.

Recommend.
```
type Identities struct {
    source *migauth.Identity
    destination *migauth.Identity
}
```

pkg/auth/identity.go
dymurray: @jortel I'm a little confused here, this func currently only returns an error for those two cases you described. it returns an error if:
1. building the client fails
2. Running the client SAR check fails to execute (this is when the authentication fails)

For this reason I felt that it was superfluous to even use `TokenReview` and/or to have a separate `Authenticate()` func. Thoughts?

dymurray: I was wrong, I ended up doing this.

pkg/controller/migplan/migplan_controller.go
dymurray: Done.

djwhatle:

dymurray:
As discussed this morning we are going to merge this and make any other changes before final PR is submitted against master branch.

PR #479, opened 2020-04-02
Assume migration-controller SA during 'make run'

djwhatle:
**Purpose**
Don't use currently logged in k8s user credentials during `make run`, instead use `migration-controller` SA so that permission gaps will be spotted quicker

**How it works**
Creates a temporary KUBECONFIG logged in as `migration-controller` SA that the controller will run as when using `make run`. Does _not_ modify your environments KUBECONFIG env var or its content.
 
**Request for feedback**
Will _always_ running as the migration-controller SA make development more cumbersome, should we provide an escape hatch if you want to override and run as the currently logged in user? 

```
$  make run
./hack/conversion-gen-`go env GOOS` --go-header-file ./hack/boilerplate.go.txt --output-file-base zz_conversion_generated -i ./pkg/compat/conversion/...
go generate ./pkg/... ./cmd/...
go fmt ./pkg/... ./cmd/...
go vet ./pkg/... ./cmd/...
./hack/controller-sa-login.sh
Found migration-controller SA. Performing token login...
KUBECONFIG=/Users/dwhatley/.kube/awsconfig-mig-controller go run ./cmd/manager/main.go
{"level":"info","ts":1585860935.230756,"logger":"entrypoint","msg":"setting up prometheus endpoint :2112/metrics"}
{"level":"info","ts":1585860935.231097,"logger":"entrypoint","msg":"setting up client for manager"}
{"level":"info","ts":1585860935.233898,"logger":"entrypoint","msg":"setting up manager"}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,11 @@ run: generate fmt vet
 	./hack/controller-sa-login.sh
 	KUBECONFIG=$(KUBECONFIG)-${KUBECONFIG_POSTFIX} go run ./cmd/manager/main.go
 
+
+# Run against the configured Kubernetes cluster in ~/.kube/config, skip login to mig-controller SA
+run-skip-sa-login: generate fmt vet
+	go run ./cmd/manager/main.go
+
 # Install CRDs into a cluster
 install: manifests
 	kubectl apply -f config/crds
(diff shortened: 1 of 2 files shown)


djwhatle:
@jortel @dymurray @jmontleon  I added a Makefile target `run-skip-sa-login` to allow folks to skip the new SA login step if desired, but still keeping SA login as the default so that it will have to be a conscious decision after you've hopefully already realized RBAC issues are present.

PR #483, opened 2020-04-08
Bug 1820630 - Migration plans with long names cannot create registry pods and break the migration

Danil-Grigorev:
Using UID instead of a plan name to prevent hitting name length limit

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -232,19 +232,14 @@ func migRegistryImageRef() string {
 func (r *MigPlan) BuildRegistrySecret(client k8sclient.Client, storage *MigStorage) (*kapi.Secret, error) {
 	labels := r.GetCorrelationLabels()
 	labels[MigrationRegistryLabel] = string(r.UID)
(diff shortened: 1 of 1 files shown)


pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -232,19 +232,14 @@ func migRegistryImageRef() string {
...
 	labels := r.GetCorrelationLabels()
 	labels[MigrationRegistryLabel] = string(r.UID)
-	strippedName, err := toDnsLabel(r.GetName())
-	if err != nil {
-		return nil, err
-	}
 	secret := &kapi.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:       labels,
-			GenerateName: "registry-" + strippedName + "-",
```
sseago, line 242: It's easier to see what's going on when the plan name is used. UID works, but it's slightly easier to know what's going on when I can look at the openshift-migration namespace and see which plans are using the resources just from 'oc get all'. Another way to do this would be to manually truncate strippedName to a max length of 46 (so "registry-"+strippedName+"-"+<generateName chars> + "-1" result in no more than 63 chars for the replicationcontroller name

PR #485, opened 2020-04-09
Refactored stage pods approach to work with all PVCs available

### LINKED ISSUE #204: Migrate PVCs that aren't attached to running pods
Current stage migration behavior only migrates PVCs associated with running pods. The restic implementation currently used makes use of the existing pods, redefined on restore as dummy pods. Longer-term, should we be migrating PVCs which are bound to PVs but aren't associated with currently-running pods? Will any of these have data we care about? In addition to unconnected PVCs, there are also the cases where their pods are currently shut down, either deliberately, or due to a recent pod crash.

Doing this might require spawning temporary migration-only pods. Any concerns here other than the implementation challenges? Is this desirable in the first place?

Danil-Grigorev:
Refactored stage pods approach to work with PVCs available, even when there is no running pod mounting the volume. In order to make it work with CRs specific for the customer case, introducing some management tactics for pods, there is a new `spec` field, called `templates`.

This field may contain a list of additional CRs, which were downscaled before migration starts be the administrator running the migration, so no pods will be found to create a stage pod from and mount a PVC, therefore such PVC would not be mounted by any pod and wont be otherwise migrated.

Example:
```yaml
kind: MigPlan
...
spec:
 templates:
 - resource:  "cronjob.batch",
     templatePath: ".spec.jobTemplate.spec.template",
 - resource: "deployment.apps",
     templatePath: ".spec.template",
```
The default template set includes: 
- job
- cronjob
- deployment
- deploymentconfig
- replicationcontroller
- daemonset
- statefulset
- replicaset

-------

For all completely unused PVCs there will be created a generic stage pod, which will migrated data from those if it would be desired. For such PVCs there is a new `skip` action, allowing to drop those on backup.

Example:
```yaml
      name: pvc-774cc225-f2bb-4922-a65f-201d182fb351
      pvc:
        accessModes:
        - ReadWriteOnce
        name: mysql1
        namespace: mysql-persistent-test-move-2
      selection:
        action: skip <- here
        copyMethod: filesystem
        storageClass: gp2
      storageClass: gp2
      supported:
        actions:
        - skip <- here
        - copy
        - move
        copyMethods:
        - filesystem
        - snapshot
```

Those PVC that are mounted (or will be mounted on pod startup) in at least one of the discovered templates, include `pvc.hasReference` field

Example:
```yaml
    - capacity: 10Gi
      name: pvc-bf71bc65-af71-482c-8def-78288ed156d4
      pvc:
        accessModes:
        - ReadWriteOnce
        hasReference: true <- here
        name: mysql2
        namespace: mysql-persistent-test-move-2
      selection:
        action: copy
        copyMethod: filesystem
        storageClass: gp2
      storageClass: gp2
      supported:
        actions:
        - copy
        - move
        copyMethods:
        - filesystem
        - snapshot
    srcMigClusterRef:
      name: host
```
------
Related operator PR: https://github.com/konveyor/mig-operator/pull/310
Closes #204


sseago:
One thing that's not clear to me right now is how we're getting the list of PVCs that aren't connected to a running pod. We need to create stage pods for all of the PVCs that don't already have a stage pod. It looks like here we're creating stage pods based on the PodTemplateSpec in each CronJob, but that won't catch other PVs not connected to a pod, such as scaled down Deployments, non-running DaemonSets, etc. There's even the possibility of a PVC that's still there although the deployment referencing it has been deleted. I'm not sure what the best way of incorporating the others might be, though. Would it be possible to just take the PVC list, filter out the PVCs that have already been included by the stage pods based on running pods, and then create new stage pods from a dummy template for the remaining PVCs?

Danil-Grigorev:
> One thing that's not clear to me right now is how we're getting the list of PVCs that aren't connected to a running pod. We need to create stage pods for all of the PVCs that don't already have a stage pod. It looks like here we're creating stage pods based on the PodTemplateSpec in each CronJob, but that won't catch other PVs not connected to a pod, such as scaled down Deployments, non-running DaemonSets, etc. There's even the possibility of a PVC that's still there although the deployment referencing it has been deleted. I'm not sure what the best way of incorporating the others might be, though. Would it be possible to just take the PVC list, filter out the PVCs that have already been included by the stage pods based on running pods, and then create new stage pods from a dummy template for the remaining PVCs?

I'm afraid of hitting a list of issues with namespace limitations, node affinity and security contexts, if we don't take in account those values:

sseago:
Also, we need to make sure this works with Snapshot copy too. For that we don't need new stage pods, but we probably need to make sure the PVCs are labeled for inclusion in the stage backup. I don't remember how that's done right now, though.

jwmatthews:
> > One thing that's not clear to me right now is how we're getting the list of PVCs that aren't connected to a running pod. We need to create stage pods for all of the PVCs that don't already have a stage pod. It looks like here we're creating stage pods based on the PodTemplateSpec in each CronJob, but that won't catch other PVs not connected to a pod, such as scaled down Deployments, non-running DaemonSets, etc. There's even the possibility of a PVC that's still there although the deployment referencing it has been deleted. I'm not sure what the best way of incorporating the others might be, though. Would it be possible to just take the PVC list, filter out the PVCs that have already been included by the stage pods based on running pods, and then create new stage pods from a dummy template for the remaining PVCs?
> 
> @sseago I'm afraid of hitting a list of issues with namespace limitations, node affinity and security contexts, if we don't take in account those values:
> 
> * https://github.com/konveyor/mig-controller/pull/485/files#diff-594d82bf8714da004d0670163b41f133L141
> * https://github.com/konveyor/mig-controller/pull/485/files#diff-594d82bf8714da004d0670163b41f133L135-L136
> 
> There is no way to know those values for the application, unless you know from where the pod which is mimicked by a stage pod came from. The `PVCs` does not provide such data.
> 

Are we sure there is a problem here?

Assume we need to retain the filesystem permissions and ownership, that will be retained by Restic, so we don't need to know what UID the pod was running as, Restic will be privileged and be able to copy and retain permissions.  (I am assuming, this should be confirmed).

NodeAffinity, would impact us if there was a specific Volume that is only accessible on a given Node.  Do we know of situations where this would occur?  Perhaps with Local Persistent Volumes?  

For the SecurityContext I assume we just need to ensure we run with a SecurityContext sufficient for restic to do it's job, I would assume it's likely our stage/backup pod is running with higher privileges than the original pod.

Any other concerns?

I would prefer to see us find an approach that is able to look for all PVCs in a namespace, find a way to determine which will be snapshotted versus restic copy, then create a generic Pod spec that can satisfy Restic's requirements and add the annotation.  Do we know of blockers that will prevent this?



> I've added missing deployments, statefulsets etc. to the list, as they are part of default set of templates. The idea is to have a list of resource templates, which could be extended by user with his `CustomResources`, he will have to downscale manually, as our tool may not support it.
> 
> It will consist from `GroupKind` for the resource, and a json path to the `PodTemplateSpec`, where the template will be captured from. That way the stage pods have an ability to preserve necessary values from the application

sseago:
@jwmatthews "find a way to determine which will be snapshotted versus restic copy"
This part is easy enough. The disconnected PVs will be added to the migplan's PV list via PV discovery, so that will be determined via the UI in the same way as the other PVs are determined.

djwhatle:
Related to issue https://github.com/konveyor/mig-controller/issues/204

.github/labeler.yml
```diff
@@ -0,0 +1,3 @@
+# Add 'update-operator' to any changes within 'config/crds' dir
+update-operator:
+  - config/crds/*
```
Danil-Grigorev, line 3: @djwhatle This is something to add into gitignore, I don't remember adding it to my pr.

pkg/apis/migration/v1alpha1/template.go
```diff
@@ -0,0 +1,60 @@
...
+type TemplateResource struct {
+	Resource     string `json:"resource"`
+	TemplatePath string `json:"templatePath"`
+}
+
+var DefaultTemplates = []TemplateResource{
+	{
+		Resource:     "cronjob.batch",
+		TemplatePath: ".spec.jobTemplate.spec.template",
+	},
```
sseago, line 22: I think job (job.batch) should be included here too.

.github/labeler.yml
djwhatle: This needs to be checked in to master. It's a github action that auto-labels our PRs, not local config. See https://github.com/konveyor/mig-controller/pull/508

djwhatle:
@Danil-Grigorev since this changes CRDs, please link to mig-operator update PR as a prereq for merge.

pkg/apis/migration/v1alpha1/template.go
Danil-Grigorev: Thanks! Missed this part.

djwhatle:
@Danil-Grigorev would you please add a _"Testing"_ section to your original PR comment describing how to test this PR? 

My assumption is that I would test by following these steps:

---

**Test Steps**
1. Create a PVC + an associated Pod that will write to the PVC
2. Stop the running Pod
3. Run a migration
4. Confirm existence of PVC on destination
5. Check that PVC contents were moved over successfully

---

If the above steps sound good as-is, feel free to move them up to your comment so others can see.

sseago:
@djwhatle So there are 3 different situations a PVC could be in:
1) connected to a running pod (which may or may not be associated with a higher level object (deploymentconfig, etc.)
2) associated with a PodTemplateSpec-containing resource that currently has no pods (deploymentconfig, cronjob, etc.)
3) completely unconnected to any pod-related resources

Case 1) is the only one that works in the merged code. Your test scenario would describe case 2) if the running pod you describe belonged to a deploymentconfig and was scaled down to zero replicase. Case 3) would be descibed by your test scenario if it was a standalone pod that was simply deleted. Alternatively, case 3) might be represented by a PVC created without a pod, but only if it was bound to a PV. unbound PVCs are excluded from everything.

Danil-Grigorev:
\test-with-operator 310

Danil-Grigorev:
@sseago @djwhatle Ready for review.

Danil-Grigorev:
\test-with-operator 310

pranavgaikwad:
\test-with-operator 310

jwmatthews:
@djwhatle  @eriknelson  @sseago @jortel  please be aware intent of this PR is to address https://issues.redhat.com/browse/MIG-179

Plan is to consider this for CAM 1.2.1 release to reduce risk to CAM 1.2.0

pranavgaikwad:
\test-with-operator 310

eriknelson:
> Plan is to consider this for CAM 1.2.1 release to reduce risk to CAM 1.2.0

+1, what is the proposed plan to keep this out of master? This PR is configured to come into master and will end up in 1.2.0 if this is merged.

We need to start branching to include the z-stream, i.e. our next release branch should be `release-v1.2.0` and we should expect a `release-v1.2.1`. This is necessary to repair our community-operator.

Can we put a do-not-merge on this until the above strategy is resolved so this doesn't end up in 1.2.0?

sseago:
@eriknelson I think the driver for whether we need a feature branch for this is when we want to branch for 1.2. If it will happen soon, we can just hold off merging until then. If it's not until end of sprint, we may want a feature branch for this work now that can be merged to master right after 1.2 release branch. In any case, we should probably merge this before branching for 1.2.1 (unless it's not ready for some reason, in which case we can merge later and cherry-pick to 1.2.1).

eriknelson:
@sseago I would expect a decent amount of change intended for 1.2.0 that we would want in master without this change given that we started any QE on 1.2.0 yet, and our target is for stage builds by the end of the sprint. Based on that, I would argue this should go into a feature branch that gets rebased or merged into master post-1.2.0 branch.

sseago:
@eriknelson So it sounds like we're assuming that for this sprint at least, master is still 1.2.0-only. 
@jortel Do we have any naming (or other) conventions for feature branches here? Seems like this PR (and some related follow-up work I'll be doing this sprint) will need to go into the same feature branch.

djwhatle:
@Danil-Grigorev added `do-not-merge` since we are close to cutting 1.2.0 and this PR functionality is targeted to 1.2.1.

eriknelson:
This PR needs to be updated to merge into `feature-1.2.1`, and it also needs a corresponding operator PR created into the operator `feature-1.2.1` branch and linked in this PR.

Danil-Grigorev:
\test-with-operator 310

Danil-Grigorev:
\test-with-operator 310

djwhatle:
PR is now targeted at correct branch, removed `do-not-merge`.

pranavgaikwad:
\test-with-operator 310

djwhatle:
**Notes review today:**
 - @Danil-Grigorev plans to remove 'templates' from CR spec
 - @sseago plans to provide a test-plan so that we can exercise the different PV migration paths that this PR impacts during tests.

sseago:
For each test case, consider PVs in the following scenarios:
1) volume mounted by pod running standalone (not connected to deployment, etc.)
2) volume mounted by pod running in a deploymentconfig
3) volume connected to scaled down deploymentconfig (case 2, but then setting replicas=0)
4) volume connected to CronJob
5) volume not connected to anything (previously cvonnected to deploymentconfig but the dc was deleted -- might need to configure the PVC not to be auto-deleted)

For all 5 scenarios above:
1) test PV filesystem copy, changing storageclass in some cases
2) test PV snapshot copy (src and dest gp2 volumes)
3) test PV move (src and dest NFS volumes)
4) test PV skip (should only show up as an option for scenario 5)

pkg/controller/migmigration/stage.go
```diff
@@ -0,0 +1,415 @@
...
+func (t *Task) ensureStagePodsFromOrphanedPVCs() error {
+	existingStagePods := StagePodList{}
+	stagePods := StagePodList{}
+	client, err := t.getSourceClient()
+	if err != nil {
+		log.Trace(err)
+		return err
+	}
+
+	err = existingStagePods.list(client, t.stagePodLabels())
```
sseago, line 111: So it looks like we're comparing against the stage pods from running pods but not from templates here. Shouldn't we compare against both before creating a pod for a disconnected PVC?

pkg/controller/migmigration/stage.go
```diff
@@ -0,0 +1,415 @@
...
+			Items:    []string{strconv.Itoa(created)},
+			Durable:  true,
+		})
+	}
+
+	return nil
+}
+
+// Ensure all stage pods from running pods withing the application were created
+func (t *Task) ensureStagePodsFromTemplates() error {
```
sseago, line 171: I may be missing something here, but there doesn't seem to be anything in `ensureStagePodsFromTemplates` that filters out the stage pods created from `ensureStagePodsFromRunning` -- will this result in two stage pods created here?

pkg/controller/migmigration/stage.go
Danil-Grigorev: `.list` is looking only for created stage pods - https://github.com/konveyor/mig-controller/pull/485/files#diff-656fd52107ab09d74773450b8b7d546bR90

Each time `stagePodList.merge` method is called, it is doing de-duplication with existing list within `merge` method.  And the stage pods from templates were already created in the cluster during previous phase, so we have all of them stored in a single location at that moment.

pkg/apis/migration/v1alpha1/template.go
```diff
@@ -0,0 +1,80 @@
...
+// Formatting is as follows:
+//
+// templates:
+// - resource:     "cronjob.batch",
+// 	 templatePath: ".spec.jobTemplate.spec.template",
+// - resource:     "deployment.apps",
+// 	 templatePath: ".spec.template",
+type TemplateResources struct {
+	Templates []TemplateResource `json:"templates,omitempty"`
+}
```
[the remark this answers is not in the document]
Danil-Grigorev: Just left it for a reference, may remove if it is desired.

pkg/controller/migmigration/stage.go
```diff
@@ -0,0 +1,415 @@
...
+		return err
+	}
+
+	err = existingStagePods.list(client, t.stagePodLabels())
+	if err != nil {
+		log.Trace(err)
+		return nil
+	}
+
+	pvcMounted := func(list StagePodList, claimName string) bool {
```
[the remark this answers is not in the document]
Danil-Grigorev: It will, but I think this is better option then allow to select different behaviour for 2 PVCs in different namespaces. No matter what, if the first case ignores `skip` option for the PVC, it will be selected for copy there. During the stage pod creation phase, only one stage pod will be created, due to the fact that both stage pods will have equal mount fields. So `copy` + `skip` or `copy` + `copy` will have the same outcome here.

pkg/controller/migmigration/stage.go
```diff
@@ -0,0 +1,415 @@
...
+// BuildStagePods - creates a list of stage pods from a list of pods
+func BuildStagePods(list *corev1.PodList) StagePodList {
+	stagePods := StagePodList{}
+	for _, pod := range list.Items {
+		stagePods.merge(buildStagePodFromPod(migref.ObjectKey(&pod), pod.Labels, &pod))
+	}
+	return stagePods
+}
+
+func (p StagePod) equals(pod StagePod) bool {
```
sseago, line 34: This equals method only confirms that all of the volumes from this pod's volumes are present in the passed-in volume. If p1 includes vol1 and vol2, and p2 includes vol1, vol2, and vol3, then p1.equals(p2) is true, while p2.equals(p1) is false.

It looks like this doesn't break anything, though, based on the order things are being called.

pkg/controller/migmigration/stage.go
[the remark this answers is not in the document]
sseago: But if pvcMounted() returns true for a volume that's orphaned, it means the stage pod won't be created, and the volume won't be copied. PVCs are namespaced, which means the same named pvcs in two different namespaces are completely different PVCs. They point to different PVs.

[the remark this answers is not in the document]
Danil-Grigorev: Good catch :D

pkg/controller/migmigration/stage.go
Danil-Grigorev: bYeah, it is filtering those during `stagePods.create` call by comparing existing pods for equality in the volume.VolumeSource fields on line https://github.com/konveyor/mig-controller/pull/485/files#diff-656fd52107ab09d74773450b8b7d546bR77

pkg/controller/migplan/pvlist.go
```diff
@@ -409,3 +420,15 @@ func isRWO(accessModes []core.PersistentVolumeAccessMode) bool {
 	}
 	return false
 }
+
+func pvcInPodVolumes(pvc core.PersistentVolumeClaim, pods []core.Pod) bool {
+	for _, pod := range pods {
+		for _, volume := range pod.Spec.Volumes {
+			if volume.PersistentVolumeClaim != nil && volume.PersistentVolumeClaim.ClaimName == pvc.Name {
```
sseago, line 427: We also need to compare pvc namespace and pod namespace. If pvc and pod are in different namespace, we can skip the inner loop entirely.

sseago: Also, we're now looking at pods. Do we also need to look at templates if it's not found in the pods? Wondering whether skip should be a valid option for PVCs that belong to scaled-down DeploymentConfigs. @dymurray what do you think here?

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -218,6 +221,55 @@ func (r *MigPlan) ListMigrations(client k8sclient.Client) ([]*MigMigration, erro
...
+			unstructuredTemplate, found, err := unstructured.NestedMap(resource.Object, template.Path()...)
+			if err != nil {
+				return nil, err
+			}
+			if !found {
+				return nil, fmt.Errorf("Template path %s was not found on resource: %s", template.TemplatePath, template.Resource)
+			}
+			err = runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredTemplate, &podTemplate)
+			if err != nil {
+				return nil, fmt.Errorf("Unable to convert resource filed '%s' to 'PodTemplateSpec': %w", template.TemplatePath, err)
```
sseago, line 257: `%w` is new as of go 1.13 (https://blog.golang.org/go1.13-errors ). It fails in my go1.11 local environment, and our Dockerfile uses an earlier version:
```
FROM golang:1.10.3 as builder
```

pkg/controller/migmigration/stage.go
```diff
@@ -340,9 +343,19 @@ func (t *Task) stagePodLabels() map[string]string {
 }
 
 // Build a stage pod based on existing pod.
-func buildStagePodFromPod(ref k8sclient.ObjectKey, labels map[string]string, pod *corev1.Pod) StagePod {
+func buildStagePodFromPod(ref k8sclient.ObjectKey, labels map[string]string, pod *corev1.Pod) *StagePod {
+	pvcVolumes := []corev1.Volume{}
+	for _, volume := range pod.Spec.Volumes {
+		if volume.PersistentVolumeClaim != nil {
+			pvcVolumes = append(pvcVolumes, volume)
```
sseago, line 350: To avoid creating stage pods when we're not using restic for any volumes, it might be worth filtering based on the migplan's PV list here as well. The hard part is that this is logic we're already going through in the set annotations phase, and we need to avoid duplicating too much or this may end up being too slow, but we really need similar filtering here. We only want to include volumes that:
1) are PVC volumes
2) action=copy
3) copyMethod=filesystem

Danil-Grigorev: Added. Please review.

sseago:
This PR will fix the issues I ran into with snapshot copy. It's made against your PR branch, so you should be able to merge it to pull it into this PR. This fixes two things:
1) snapshot-only stage pods skipped in the "from running pods" and "from templates" steps were being created as orphaned stage pods.
2) orphaned stage pod creation wasn't filtering on copyMethod: the fix is to filter here too.
3) when a stage pod wasn't created (for snapshots), the PVCs weren't labeled or annotated since annotatePVCs is only called from within annotatePod. The fix is to annotate PVCs along with PVs.

https://github.com/Danil-Grigorev/mig-controller/pull/1

sseago:
So I've now tested the essential use cases:
1) pod PV (standalone and DC), quiesced DC PV, and disconnected PV restic copy with and without storageclass change (3.11->4.2, 4.2->4.4)
2) pod PV (standalone and DC), quiesced DC PV, and disconnected PV snapshot copy gp2 4.2->4.4
3) disconnected PV skip action 4.2->4.4
4) pod PV, quiesced DC PV, and disconnected PV move action (NFS 3.11->4.4)
It might be worth confirming that this works with a CronJob as well, although the use case is basically equivalent to the quiesced DC. My current setup doesn't have a stateful CronJob app definition to work with, so I haven't tested that particular case.

sseago:
I also found a bug in the PR I created against this PR, so I'll need to create a new one. I'll get that done later this evening.

sseago:
@Danil-Grigorev I've replaced the above PR with https://github.com/Danil-Grigorev/mig-controller/pull/2 -- this should resolve the remaining issues I had with this. The above testing I did was with that commit on top of your PR plus the #520 fix.

sseago:
Now depends on https://github.com/konveyor/mig-operator/pull/327 rather than the prior mig-operator PR

pkg/reference/ref.go
```diff
@@ -1,11 +1,32 @@
...
 )
 
+func ObjectKey(obj metav1.Object) client.ObjectKey {
+	return client.ObjectKey{
+		Name:      obj.GetName(),
+		Namespace: obj.GetNamespace(),
+	}
+}
+
+func TruncateName(name string) string {
```
jortel, line 20: I don't understand why this is part of the `ref` package.  It's seems completely unrelated.  Also, it's only used in 1 place.

pkg/reference/ref.go
```diff
@@ -1,11 +1,32 @@
...
 	"reflect"
+	"regexp"
 	"strings"
+
+	kapi "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+func ObjectKey(obj metav1.Object) client.ObjectKey {
```
jortel, line 13: I don't understand why this is part of the `ref` package.  The usages don't seem like a valuable change.  It's effectively changing 1 line of code with another.

pkg/reference/ref.go
sseago: It looks like Danil used it in 3 places, from two different packages. Perhaps that's why he put it under ref. Where should it go?

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -218,6 +221,55 @@ func (r *MigPlan) ListMigrations(client k8sclient.Client) ([]*MigMigration, erro
 	return list, nil
 }
 
+// ListTemplatePods - get list of pod templates, associated with a plan resource
+func (r *MigPlan) ListTemplatePods(client k8sclient.Client) ([]corev1.Pod, error) {
```
jortel, line 225: I don't understand why this method is on MigPlan.  The existing `List` or `Get` convenience methods *only* pertain to resources referenced directly (or indirectly) by the Plan. Further, the functionality is solely the concern of the migration controller.

pkg/reference/ref.go
sseago: Looks like this converts a kubernetes resource (with more parameters than just a namespaced name) into an object key. It's only used in 2 places, though, so leaving it inline would be fine. Danil won't be working on this PR anymore, though, so the easiest thing to do with this sort of thing would be a follow-on PR with whatever changes are desired from this one.

pkg/controller/migmigration/task.go
```diff
@@ -749,9 +774,8 @@ func (t *Task) anyFlags(step Step) bool {
 }
 
 // Phase fail.
-func (t *Task) fail(nextPhase string, reasons []string) {
+func (t *Task) fail(nextPhase string, reasons ...string) {
```
jortel, line 777: This signature was changed to support the _exception_ use case which seems undesirable and resulted in a lot of unrelated code changes.

pkg/apis/migration/v1alpha1/migplan_types.go
sseago: I think an earlier iteration of this wasn't on migplan. I think this was because we need to call GetSourceNamespaces on the plan within this method. Another possibility would be to make it a top level function that takes either the migplan, or the migplan's list of src namespaces as an argument.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -218,6 +221,55 @@ func (r *MigPlan) ListMigrations(client k8sclient.Client) ([]*MigMigration, erro
...
+	pods := []corev1.Pod{}
+	for _, template := range DefaultTemplates {
+		// Get resources
+		list := unstructured.UnstructuredList{}
```
jortel, line 229: Listing resources with templates using the _unstructured_ api seems overly complicated and contrary to approach used for other _like_ processing.  Given this is a well known set of resources kinds, it seems more consistent and straight forward to just list them specifically.

pkg/controller/migmigration/task.go
sseago: It's not obvious to me exactly what's going on here. I'm not seeing any obvious places where reasons needs the change, and the method in question isn't doing anything different with reasons than before.

pkg/reference/ref.go
jortel: Not sure but putting it here is arbitrary and we've worked hard to keep our packages clean.
Any idea why we need to build _special_ names instead of using labels?

pkg/reference/ref.go
jortel: Do you think the undesirable changes could be enumerated in an issue?  I'm concerned they'll fall through the cracks.

pkg/apis/migration/v1alpha1/migplan_types.go
sseago: I'm guessing this is a holdover from the original implementation which allowed for arbitrary user-specified resources on the migplan (which is no longer here). In any case, we have a list of 8 different resource types, but the template path is different for each of them. So when you say "list them specifically", do you mean something along the line of instead of the current slice of:
```
type TemplateResource struct {
￼	Resource     string `json:"resource"`
￼	TemplatePath string `json:"templatePath"`
￼}
item0 = TemplateResource{
		Resource:     "deployment.apps",
		TemplatePath: ".spec.template",
	}
```
we'd have something like this:
```
item0 = TemplateResource{
		Resource:    ocappsv1.DeploymentConfig{},
		TemplatePath: ".spec.template",
}
```
(and whatever struct definition would match the explicit resource object).
Would that work with the fact that the template path is different for each? We can't just say resource.Spec.Template since in some cases it would be resource.spec.JobTemplate.Spec.Template.

pkg/reference/ref.go
sseago: Well we don't want a label. We need an object name. Specifically, an input to generateName. For example, here's one use:
```
				GenerateName: migref.TruncateName("stage-"+ref.Name) + "-",
￼```
Creating a stage pod name where ref.Name might be an existing pod name, it might be a DeploymentConfig name, or it might be a PVC name (depending on context).

pkg/apis/migration/v1alpha1/migplan_types.go
jortel: Seems like it should just go in `pvlist.go` and pass in the `plan` where it's used.

pkg/reference/ref.go
sseago: Sure. That makes sense.

pkg/reference/ref.go
sseago: That said, I don't see any reason to use TruncateName where it's used in migplan/registry.go. If we remove that instance, it's only needed in stage.go and it can be an unexported local function.

pkg/apis/migration/v1alpha1/migplan_types.go
sseago: It's also used in migmigration/stage.go. I know in the past, things that are called from multiple controllers end up pushed into the apis pkg. Is it OK for it to go in migplan/pvlist.go (but as an exported function) which would then be called from stage.go as well?

jortel: Ah, didn't realize it's used by multiple controllers.  In that case, `MigPlan` is a reasonable place for it for the reasons you mentioned.

pkg/apis/migration/v1alpha1/migplan_types.go
jortel: I was wondering why not just list Deployments etc and inspect the template part of the spec in a structured way to build the list of pods.

Issue #471, opened 2020-03-30
Update all controllers to watch resources in the sandbox namespace


dymurray:
All controllers should consume an environment variable from the operator which will tell the controller which namespace the sandbox was installed. This namespace will be where the controller is watching for creation of mig resources.

The creation of Velero resources will still occur in the protected `openshift-migration` namespace (or wherever CAM is installed). Let's ensure this namespace is also a configurable setting.

Consider adding this to the `settings` pkg.

PR #488, opened 2020-04-09
Allow the mig resource controllers to watch a specific namespace

mansam:
Following discussion with @jortel and @dymurray this splits the CAM role into individual roles for each of the mig resource controllers, and updates the watch predicates to filter based on the SANDBOX_NAMESPACE environment variable if it's present. The cluster and plan controllers also use a SINGLETON_NAMESPACE environment variable, if present, to restrict where it will look for cluster updates.


dymurray:
@mansam FYI please target this for the non-admin feature branch. I am about to rebase it with master

djwhatle:

pkg/controller/controller.go
```diff
@@ -31,19 +31,12 @@ import (
...
-	migmigration.Add,
-	migstorage.Add,
-	migplan.Add,
-}
-
-//
-// List of controller add functions for the Discovery role.
-var DiscoveryControllers = []AddFunction{
-	discovery.Add,
+var Controllers = map[string]AddFunction{
```
dymurray, line 34: :+1:

pkg/controller/migcluster/predicate.go
```diff
@@ -14,6 +15,9 @@ type ClusterPredicate struct {
 }
 
 func (r ClusterPredicate) Create(e event.CreateEvent) bool {
+	if !common.IsInSingletonNamespace(e.Meta.GetNamespace()) {
```
dymurray, line 18: Can you help explain to me what exactly this is doing? I'm not super familiar with the predicate code.

pkg/controller/common/common.go
```diff
@@ -0,0 +1,11 @@
+package common
```
jortel, line 1: Let's avoid package, .go files and functions named: "util", "common", "helper".  They lead to _buckets of functions_. I strongly believe that any functionality added should be scoped and grouped by concern.

These functions look like a good fit in the _settings_ package and likely a method on `_Settings`.

pkg/controller/common/common.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "github.com/konveyor/mig-controller/pkg/settings"
+
+func IsInSingletonNamespace(namespace string) bool {
```
jortel, line 5: Perhaps we can find a more descriptive name than "Singleton".  Perhaps "Privileged"?

pkg/controller/controller.go
jortel: :+1:

pkg/controller/common/common.go
mansam: Sounds good.

pkg/controller/common/common.go
mansam: Privileged sounds much better to me. @dymurray?

sseago: "Singleton" generally refers to an object for which only one is able to exist its type/class. There isn't generally a notion of "privileged" there. If privileges are relevent, then I'd say "PrivilegedNamespace" makes more sense.

jortel:
It seem reasonable to finalize the _admin_ vs _non-admin_ installation as either: In the sandbox or in the privileged namespaces.  As such, I think the code simply needs to answer the question: "_Are we in the sandbox or not."_

Suggest something like:

In settings:
- SandboxNamespace
- PrivilegedNamespace

```
+ func (s *_Settings) InSandbox(r MigResource) bool {
    return r.GetNamespace() == s.SandboxNamespace
}
```

Using the interface:
```
type MigResource interface {
+ InSandbox() bool
}
...
func (r *MigPlan) InSandbox() bool {
    return settings.Settings.InSandbox(r)
```
In the predicates:

```
if cluster.InSandbox() {
    return false
}
...
if !plan.InSandbox() {
    return false
}
```

pkg/controller/common/common.go
dymurray: @mansam yes `privileged` sounds good to me. Or `admin`.

mansam:
\test

mansam:
\test-with-operator 312

mansam:
\test-with-operator 312

mansam:
\test-with-operator 312

mansam:
\test-with-operator 312

mansam:
\test-with-operator 312

mansam:
\test-with-operator 312

mansam:
\test-with-operator 312

dymurray:
Fixes #471

Issue #393, opened 2020-01-07
Controller should pre-create destination namespaces


dymurray:
Since the Velero restore will still be run as a privileged user, we need to ensure that the user will have access to the namespace after the migration runs. To ensure this, the controller will use the user's identity token to pre create the namespaces on the destination cluster prior to the restore running. This will force Velero to skip the recreation of the namespaces allowing for the user to have access to the resources post restore.

PR #494, opened 2020-04-15
Pre-create the namespaces to be migrated on the destination cluster

mansam:
https://github.com/konveyor/mig-controller/issues/393

This might need to change depending on how https://github.com/konveyor/mig-controller/pull/478 and its follow ups shake out.

@dymurray @jortel It seems plausible that the destination cluster might already have namespaces with names matching those to be migrated, in which case EnsureNamespacesCreated will not create them. Do we want to have another step in the itinerary that checks whether the user has permissions in those destination namespaces?

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -122,6 +122,7 @@ func (r *MigPlan) GetSourceIdentity(client k8sclient.Client) (*migauth.Identity,
 		Token:   srcToken,
 		RestCfg: *srcRestCfg,
 	}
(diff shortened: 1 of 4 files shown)


dymurray:
@mansam yes I think that makes sense. And in the case where a user has run a stage migration and then a full migration we don't want to recreate the namespace, but rather ensure the token they supplied is still valid. I think we can reuse the `HasRead` func on the auth pkg in #478.

djwhatle:
Converting this to a draft PR, let's use that system to denote WIP.

djwhatle:

mansam:
\test

PR #496, opened 2020-04-16
Fix disappearing GVK condition

### LINKED ISSUE #495: "GVKsIncompatible" condition disappears after MigPlan reconcile
Condition shows up but is removed by subsequent reconcile.  IncompatibleNamespaces list containing resources remains, but UI needs the condition to persist as a trigger for showing appropriate warning.

djwhatle:
Resolves #495

### DIFF
--- a/pkg/controller/migplan/gvk.go
+++ b/pkg/controller/migplan/gvk.go
@@ -29,6 +29,7 @@ type Compare struct {
 func (r ReconcileMigPlan) compareGVK(plan *migapi.MigPlan) error {
 	// No spec chage this time
 	if plan.HasReconciled() || !clustersReady(plan) {
+		plan.Status.StageCondition(GVKsIncompatible)
 		return nil
 	}
 


pkg/controller/migplan/gvk.go
```diff
@@ -84,19 +88,6 @@ func reportGVK(plan *migapi.MigPlan, incompatibleMapping map[string][]schema.Gro
 		}
 		incompatibleNamespaces = append(incompatibleNamespaces, incompatibleNamespace)
```
jortel, line 89: This will create a new slice and assign to `incompatibleNamespaces`.  The Incompatible.Namespaces will never be populated.

pkg/controller/migplan/gvk.go
```diff
@@ -84,19 +88,6 @@ func reportGVK(plan *migapi.MigPlan, incompatibleMapping map[string][]schema.Gro
...
 	}
-
```
jortel, line 87: I think you'll still need to still do:
```
plan.Status.Incompatible = incompatibleNamespaces
```
here.

djwhatle:
@jortel I was having trouble getting the desired behavior using the second method of modification that we discussed, and need to get Ian an image to test with, so went with the one-liner change we originally came up with. Seems to work well

```
   - category: Warn
      lastTransitionTime: 2020-04-17T20:24:09Z
      message: 'Some namespaces contain GVKs incompatible with destination cluster.
        See: `incompatibleNamespaces` for details'
      status: "True"
```

```
    incompatibleNamespaces:
    - gvks:
      - group: batch
        kind: cronjobs
        version: v2alpha1
      - group: batch
        kind: scheduledjobs
        version: v2alpha1
      name: phronetic
```

djwhatle:
Provided @ibolton336 with a build of this and he got positive results, merged.

Issue #470, opened 2020-03-30
Add authorization phase to migration controller workflow


dymurray:
As part of the migplan validation, we check if the user has `get` access on the selected list of namespaces. However, we also need to add an authorization phase to the controller's task list to ensure that the user has the needed permissions to perform all steps necessary for a migration such as:
* scaling down deployments/replicasets/etc (quiescing applications)
* creating/updating pods (for stage pods/labels/annotations)
* listing PVCs (for storage selection)listing PVCs (for storage selection)
* create namespaces on the destination cluster
* create/put/patch/update in destination namespaces mirroring same SAR checks on source
* etc.

To ensure that CAM can create Velero resources on behalf of the user for a given cluster, we need to verify they have `use` access on the associated service account secret for the given cluster.

Consider parallelizing the authorization phase for src + dest clusters in future releases.

Get with Jeff to discuss how to properly fail the migration if the auth phase fails.

jortel:
Create (2) phases: `Authorization` and `NotAuthorized`.
```
case Authorization:
    allowed, err := t.EnsureAuthorized()
    if err != nil {
        log.Trace(err)
        return err
    }
    if !allowed {
        t.fail(NotAuthorized, "your error here")
    else {
        t.next()
    }
}
		
```

PR #511, opened 2020-04-23
Add Authorization step to task itinerary

mansam:
https://github.com/konveyor/mig-controller/issues/470


pkg/controller/migmigration/task.go
```diff
@@ -19,6 +19,8 @@ const (
 	Created                       = ""
 	Started                       = "Started"
 	Prepare                       = "Prepare"
+	Authorization                 = "Authorization"
+	NotAuthorized                 = "NotAuthorized"
```
dymurray, line 23: What about `AuthorizationFailed` to match the other phases?

mansam: `NotAuthorized` was @jortel's recommendation over `AuthorizationFailed`.

mansam:
The last commit puts the `EnsureNamespacesCreated` step before `Authorization`, because we need the destination namespaces to exist in order to do SAR checks on them. That creates a need to clean up the namespaces if authorization fails, which is a bit tricky since some of the namespaces might already exist, so we'd need to keep track of the ones which were just created. One approach might be to create the namespaces during the auth step and immediately remove them if auth fails.

Alternatively, can we make any assumptions about what kind of access a user would have in namespaces created by them? Maybe we can do the destination auth check on any namespaces that exist, and then assume that we'll have rights on the namespaces we create? Not convinced of that either.

@jortel, what do you think?

PR #513, opened 2020-04-24
Add developer guide.

jortel:
Add developer guide with basic design patterns, conventions and practices.


djwhatle:
@jortel do you think we should merge this, and we can continue adding later if needed?

PR #518, opened 2020-04-29
BZ 1829390 - Check additional failure conditions and fail task

jmontleon:
Sergio found an additional way to cause a failure on a job so this is addressing it by checking the status reason. In addition to that I'm raising the backofflimit to 6 as it should be, and failing the job properly when we catch an error.

### DIFF
--- a/pkg/apis/migration/v1alpha1/mighook_types.go
+++ b/pkg/apis/migration/v1alpha1/mighook_types.go
@@ -26,7 +26,8 @@ import (
 )
 
 const (
-	PhaseLabel           = "phase"
+	HookPhaseLabel       = "phase"
+	HookOwnerLabel       = "owner"
 	PreBackupHookPhase   = "PreBackup"
 	PostBackupHookPhase  = "PostBackup"
 	PreRestoreHookPhase  = "PreRestore"
@@ -75,10 +76,11 @@ func init() {
 }
 
 // Get an existing hook job.
-func (r *MigHook) GetPhaseJob(client k8sclient.Client, phase string) (*batchv1.Job, error) {
+func (r *MigHook) GetPhaseJob(client k8sclient.Client, phase string, owner string) (*batchv1.Job, error) {
 	list := batchv1.JobList{}
 	labels := r.GetCorrelationLabels()
-	labels[PhaseLabel] = phase
+	labels[HookPhaseLabel] = phase
+	labels[HookOwnerLabel] = owner
 	err := client.List(
 		context.TODO(),
 		k8sclient.MatchingLabels(labels),
@@ -93,10 +95,11 @@ func (r *MigHook) GetPhaseJob(client k8sclient.Client, phase string) (*batchv1.J
 }
 
 // Get an existing configMap job.
-func (r *MigHook) GetPhaseConfigMap(client k8sclient.Client, phase string) (*corev1.ConfigMap, error) {
+func (r *MigHook) GetPhaseConfigMap(client k8sclient.Client, phase string, owner string) (*corev1.ConfigMap, error) {
 	list := corev1.ConfigMapList{}
 	labels := r.GetCorrelationLabels()
-	labels[PhaseLabel] = phase
+	labels[HookPhaseLabel] = phase
+	labels[HookOwnerLabel] = owner
 	err := client.List(
 		context.TODO(),
 		k8sclient.MatchingLabels(labels),
--- a/pkg/controller/migmigration/hooks.go
+++ b/pkg/controller/migmigration/hooks.go
@@ -31,6 +31,17 @@ func (t *Task) runHooks(hookPhase string) (bool, error) {
 
 	migHook := migapi.MigHook{}
 
+	svc := corev1.ServiceAccount{}
+	ref := types.NamespacedName{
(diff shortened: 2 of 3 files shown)


jmontleon:
This will now also allow running the migration again after correcting errors without having to clean up old configmaps and jobs.

jmontleon:
This now also checks to ensure the SA exists to prevent another eternal hang. If the SA does not exist the job just waits on can't create forever and never fails so we need to make sure it exists. Conversely a non-existent ns will trigger a client error when it goes to create the resource so there is no probkem there.

djwhatle:
\test

djwhatle:
@jmontleon assuming tests pass, is this good to merge?

pkg/apis/migration/v1alpha1/mighook_types.go
```diff
@@ -26,7 +26,8 @@ import (
 )
 
 const (
-	PhaseLabel           = "phase"
+	HookPhaseLabel       = "phase"
+	HookOwnerLabel       = "owner"
```
djwhatle, line 30: I missed this on initial review, and we don't implement this across the project right now, but in the future I think we should consider following k8s guidelines for label names (e.g. including `migration.openshift.io` namespace in label name)

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

pkg/controller/migmigration/hooks.go
```diff
@@ -15,7 +15,8 @@ import (
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const HookJobFailedLimit = 5
+const HookJobFailedLimit = 6
```
djwhatle, line 18: Why is this supposed to be 6 rather than 5? Just curious.

pkg/controller/migmigration/hooks.go
```diff
@@ -79,6 +91,9 @@ func (t *Task) ensureJob(job *batchv1.Job, hook migapi.MigPlanHook, migHook miga
 	} else if runningJob.Status.Failed >= HookJobFailedLimit {
 		err := fmt.Errorf("Hook job %s failed.", runningJob.Name)
 		return false, err
+	} else if len(runningJob.Status.Conditions) > 0 && runningJob.Status.Conditions[0].Reason == BackoffLimitExceededError {
```
djwhatle, line 94: I think `HasCondition` might be a good fit for this? https://github.com/konveyor/mig-controller/blob/master/pkg/apis/migration/v1alpha1/condition.go#L249

pkg/controller/migmigration/task.go
```diff
@@ -427,6 +432,7 @@ func (t *Task) Run() error {
 		status, err := t.runHooks(migapi.PostBackupHookPhase)
 		if err != nil {
 			log.Trace(err)
+			t.fail(PostBackupHooksFailed, []string{err.Error()})
```
djwhatle, line 435: @jortel is this the right way to move a migration to failed state? I don't know the convention for this, haven't had to send a migration to failure in code I've written.

djwhatle:
CI failed, looks like it could be a flake. Re-testing.

djwhatle:
\test

pkg/controller/migmigration/hooks.go
jmontleon: The default BackOffLimit (that we use) is 6, not 5. It will generate 6 failures before giving up.

pkg/controller/migmigration/hooks.go
jmontleon: Let me test and change it.

jmontleon:
\test

pkg/controller/migmigration/hooks.go
jmontleon: I'm actually not clear on how to implement this. Do I have to create a HasConditions function specifically for jobs?

pkg/controller/migmigration/task.go
jmontleon: When I asked this is what I was pointed at. It seems to work well like this. We still get the trace and the migration fails with the condition/error that specifically caused the issue.

jmontleon:
@pranavgaikwad can you help me understand why this is failing? Is it a flake?

pranavgaikwad:
@jmontleon CI test cases are not configured to use MigHooks. The MigMigration has following status : 
```bash
Status:
  Conditions:
    Category:              Advisory
    Durable:               true
    Last Transition Time:  2020-05-05T13:47:30Z
    Message:               The migration has failed.  See: Errors.
    Reason:                PreBackupHooksFailed
    Status:                True
    Type:                  Failed
  Errors:
    ServiceAccount "" not found
```

pkg/controller/migmigration/hooks.go
djwhatle: You're right, I don't think we want to implement this for jobs. Marking as resolved.

pkg/controller/migmigration/task.go
djwhatle: Cool, that sounds reasonable.

jmontleon:
\test

PR #520, opened 2020-04-29
Create backup and restore only once for each run

Danil-Grigorev:


pkg/controller/migmigration/backup.go
```diff
@@ -19,47 +18,25 @@ import (
...
 	newBackup.Labels[InitialBackupLabel] = t.UID()
 	newBackup.Spec.ExcludedResources = excludedInitialResources
 	delete(newBackup.Annotations, QuiesceAnnotation)
-	foundBackup, err := t.getInitialBackup()
+	err = client.Create(context.TODO(), newBackup)
 	if err != nil {
 		log.Trace(err)
 		return nil, err
 	}
-	if foundBackup == nil {
```
sseago, line 35: I think we still need this block. It makes sure we don't attempt to create the CR if it already exists. We do need to remove the `!t.equalsBackup...` block, though.

pkg/controller/migmigration/backup.go
```diff
@@ -84,33 +66,11 @@ func (t *Task) ensureStageBackup() (*velero.Backup, error) {
 	newBackup.Labels[StageBackupLabel] = t.UID()
 	newBackup.Spec.IncludedResources = stagingResources
 	newBackup.Spec.LabelSelector = &labelSelector
-	foundBackup, err := t.getStageBackup()
+	err = client.Create(context.TODO(), newBackup)
 	if err != nil {
 		return nil, err
 	}
-	if foundBackup == nil {
```
sseago, line 91: I think we still need this block. It makes sure we don't attempt to create the CR if it already exists. We do need to remove the !t.equalsBackup... block, though.

pkg/controller/migmigration/restore.go
```diff
@@ -27,45 +27,22 @@ func (t *Task) ensureFinalRestore() (*velero.Restore, error) {
...
 		return nil, err
 	}
 	newRestore.Labels[FinalRestoreLabel] = t.UID()
-	foundRestore, err := t.getFinalRestore()
+	err = client.Create(context.TODO(), newRestore)
 	if err != nil {
 		log.Trace(err)
 		return nil, err
 	}
-	if foundRestore == nil {
```
sseago, line 41: Similar to the backup comments above.

pkg/controller/migmigration/backup.go
Danil-Grigorev: This block is now executed before opening client and building backup: https://github.com/konveyor/mig-controller/pull/520/files#diff-bae2a9cc4cbb76d893b45a873d56cc30R21-R28 if the backup or restore was already created, then the step will be ended

sseago: Ahh. I see that now.

djwhatle:
@Danil-Grigorev @sseago this needs an associated issue or BZ. Not enough context included on this PR right now.

Is this fixing a bug that has been in our codebase for a while, or something introduced with recent changes?

djwhatle:
I saw on slack that this PR is related to another PR https://github.com/konveyor/mig-controller/pull/485. It sounds like this is a pre-req to merge the other PR. Is this accurate?

sseago:
@djwhatle It's a longstanding bug but apparently it's not something that we often encounter. So there are actually two problems with the code that's being removed:
1) After comparing the in-cluster CR to what it should look like, the update actually sets it to something wrong -- removing the excludedResources specified on the initial backup.
2) We should never update velero CRs after creating them. Velero manages them once they've been created.

I hit a problem with this code when testing #485 -- basically the controller tried to update the velero CR but because velero also modified it at the same time, it failed to update, returning an error, which then failed the migration, triggering the cleanup dest cluster code.

sseago:
@djwhatle I'll create a github issue describing what happened in my env.

djwhatle:
\test

sseago:
Fixes #524

sseago:
@djwhatle see linked github issue #524

pkg/controller/migmigration/task.go
```diff
@@ -379,10 +379,10 @@ func (t *Task) Run() error {
 			log.Trace(err)
 			return err
 		}
-		t.Requeue = 0
+		t.Requeue = NoReQ
```
djwhatle, line 382: @Danil-Grigorev @jortel why was this ever set to 0? I am perhaps misremembering but doesn't requeue 0 immediately requeue the reconcile?

djwhatle:
\test

pkg/controller/migmigration/backup.go
```diff
@@ -245,40 +219,17 @@ func (t *Task) buildBackup(includeClusterResources *bool) (*velero.Backup, error
...
+			VolumeSnapshotLocations: []string{snapshotLocation.Name},
+			TTL:                     metav1.Duration{Duration: 720 * time.Hour},
+			IncludedNamespaces:      t.sourceNamespaces(),
+			IncludedResources:       t.BackupResources,
+			Hooks: velero.BackupHooks{
+				Resources: []velero.BackupResourceHookSpec{},
+			},
 		},
 	}
-	err = t.updateBackup(backup)
```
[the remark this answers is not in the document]
sseago: @djwhatle right. The content from the updateBackup call was moved into here since we no longer need a separate update step.

sseago:
An update: I did mange to test this fix in conjunction with my local checkout of #485 (which is where the bug reproduced for me), and this seems to fix the problem without introducing any new ones. I've run several migrations with the code and they've all succeeded.

pkg/controller/migmigration/task.go
Danil-Grigorev: Reconcile set to 0 is effectively disabling reconcile. So any further updates would be triggered by velero resource change, that way we do not poll.

pkg/controller/migmigration/backup.go
```diff
@@ -225,13 +193,19 @@ func (t *Task) getVSL() (*velero.VolumeSnapshotLocation, error) {
 }
 
 // Build a Backups as desired for the source cluster.
-func (t *Task) buildBackup(includeClusterResources *bool) (*velero.Backup, error) {
-	// Get client of source cluster
-	client, err := t.getSourceClient()
+func (t *Task) buildBackup(client k8sclient.Client) (*velero.Backup, error) {
+	var includeClusterResources *bool = nil
+	annotations, err := t.getAnnotations(client)
```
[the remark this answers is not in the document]
Danil-Grigorev: Each time the client is created we.open a new connection. Ideally this should be done once per phase, but only refactored it here.

djwhatle:
@sseago we would expect this PR to pass tests on its own, right? I'm wondering why CI is failing.

pkg/controller/migmigration/task.go
djwhatle: Agree that polling doesn't seem like the right approach here.

PR #521, opened 2020-04-29
Implement HasMigrate

mansam:
This does two things:

1) Adds a `CanI(namespace, resource, verb)` convenience function on the Identity.
2) Implements the permissions checks in HasMigrate

@dymurray and I had an initial chat about what the list of resources we need to check against might be. For now this does a SAR check on each of those resources for all (`*`) verbs to avoid making a half dozen SAR checks per resource.

Nested for loops inherently give me the creeps but I don't see a better way to approach this at the moment.

### DIFF
--- a/pkg/auth/identity.go
+++ b/pkg/auth/identity.go
@@ -18,6 +18,23 @@ type Identity struct {
 	Client  k8sclient.Client
 }
 
+func (r *Identity) CanI(namespace, resource, verb string) (bool, error) {
+	sar := authapi.SelfSubjectAccessReview{
+		Spec: authapi.SelfSubjectAccessReviewSpec{
+			ResourceAttributes: &authapi.ResourceAttributes{
+				Resource:  resource,
+				Namespace: namespace,
+				Verb:      verb,
+			},
+		},
+	}
+	err := r.Client.Create(context.TODO(), &sar)
+	if err != nil {
+		return false, err
+	}
+	return sar.Status.Allowed, nil
+}
+
 func (r *Identity) HasRead(namespaces []string) (Authorized, error) {
 	authorized := Authorized{}
 	err := r.BuildClient()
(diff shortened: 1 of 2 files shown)


pkg/auth/identity.go
```diff
@@ -18,39 +18,63 @@ type Identity struct {
 	Client  k8sclient.Client
 }
 
+func (r *Identity) CanI(namespace, resource, verb string) (bool, error) {
```
dymurray, line 21: :+1:

pkg/auth/identity.go
```diff
@@ -18,39 +18,63 @@ type Identity struct {
...
 	authorized := Authorized{}
 	err := r.BuildClient()
 	if err != nil {
 		return authorized, err
 	}
 
+	for _, namespace := range namespaces {
+		authorized[namespace] = true
+		for _, resource := range migrateResources {
+			allowed, err := r.CanI(namespace, resource, "*")
```
dymurray, line 67: This may actually not be sufficient. We may end up doing a lot of individual SAR checks here and that's okay.

See:
```
[dymurray@pups mig-controller]$ oc auth can-i '*' deployments                                                                                                               
no - no RBAC policy matched
[dymurray@pups mig-controller]$ oc auth can-i update deployments 
yes
```

I think we need to do individual verb checks on these resources:
`get`, `create`, `update`.

djwhatle:
@mansam @dymurray what does `HasMigrate` mean? Is this a check for `HasMigrationPermission`?

mansam:
@dymurray Yeah, that's a check for the permissions necessary to perform a migration.

pkg/auth/identity.go
```diff
@@ -18,37 +18,65 @@ type Identity struct {
...
-		authorized[namespace] = sar.Status.Allowed
+		authorized[namespace] = allowed
 	}
 
 	return authorized, nil
 }
 
 func (r *Identity) HasMigrate(namespaces []string) (Authorized, error) {
+	resources := []string{"pods", "deployments", "deploymentconfigs", "daemonsets", "replicasets", "statefulsets", "pvcs"}
+	verbs := []string{"get", "create", "update"}
```
dymurray, line 57: Let's add `delete` since we end up deleting stage pods as well.

mansam:
This has been integrated into https://github.com/konveyor/mig-controller/pull/526

Issue #512, opened 2020-04-24
Add `MigToken` CRD for non-admin UX improvement


dymurray:
We need a new CRD `MigIdentity` which will contain a user's identity token to be consumed by a MigPlan. Today we are simply using a secretRef, but we want the controller to reconcile on the `MigIdentity` objects to validate whether they can currently authorize against the given cluster. This will be used to determine if a user's oauth token has expired.

dymurray:
The `MigIdentity` CRs will be created in the tenant's namespace. it should contain at least the following spec fields:

```
type MigIdentitySpec struct {
  identitySecretRef kapi.SecretReference
  migclusterRef kapi.ObjectReference
  type: string // The type of identity token this is (oauth/serviceaccount)
  expired: boolean // Only used with oauth, but could also indicate an SA token has been deleted
}
```

There will be a 1-1 mapping of MigIdentity to MigCluster. The controller will be responsible for reconciling on the migIdentitys at a predetermined length of time (5-10 mins?) or on the change of the identity secret. If an oauth token has expired, set the spec field accordingly.

Prior to running a migration, all migIdentities should be validated once more to ensure the token hasn't expired and is valid.

alaypatel07:
@dymurray closing this issue since it is checked in #388. Please re-open if required. Thanks

PR #526, opened 2020-05-07
Add MigToken CRD

mansam:
* Adds `MigToken` type, controller, and controller role
* Moves `Authenticates`, `HasRead`, and `HasMigrate` from the Identity type to the `MigToken`
* Removes the `auth` package and `Identity` type
* Replaces the `MigPlan` `identitySecretRef` with `srcMigTokenRef` and `destMigTokenRef`.
* Moves Authentication validation from `MigPlan` controller to `MigToken` controller.

I was not able to determine a non-Openshift specific way to retrieve the expiry date on an identity token, so I was unable to put that information in the `MigTokenStatus`.

See https://github.com/konveyor/mig-controller/issues/512


pkg/controller/migtoken/validation.go
```diff
@@ -0,0 +1,47 @@
+package migtoken
+
+import (
+	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
+)
+
```
dymurray, line 6: NIT: Consider using comments here to distinguish why these const's are in different blocks

pkg/apis/migration/v1alpha1/migtoken_types.go
```diff
@@ -0,0 +1,145 @@
...
+	authapi "k8s.io/api/authorization/v1"
+	kapi "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// MigTokenSpec defines the desired state of MigToken
+type MigTokenSpec struct {
+	IdentitySecretRef *kapi.ObjectReference `json:"identitySecretRef"`
```
jortel, line 17: The token can be an SA token so we decided to discontinue to use the term `Identity`.  Just `SecretRef` seems sufficient.

pkg/apis/migration/v1alpha1/migtoken_types.go
```diff
@@ -0,0 +1,145 @@
...
+	authorized := Authorized{}
+	_, err := r.GetClient(client)
+	if err != nil {
+		return authorized, err
+	}
+
+	return authorized, nil
+}
+
+func (r *MigToken) Authenticates(client k8sclient.Client) (bool, error) {
```
jortel, line 91: Suggest `Authenticate()` since the method will perform (do) the authentication when called.

pkg/controller/migtoken/validation.go
jortel: +1 The other controllers label the const blocks as:
// Types
// Reasons
// Messages
etc..

PR #527, opened 2020-05-08
Check for the SA on the correct cluster

jmontleon:


jmontleon:
I inadvertently used `t.Client` which also checks on the controller cluster rather than the `client` set up for the source/destination cluster.

PR #529, opened 2020-05-08
follow-on fixes from review of disconnected PVC migration

### LINKED ISSUE #525: Follow-on fixes for disconnected PVC migration
There are a few lingering fixes needed once #485 is merged:

- Move `TruncateName` from pkg/reference/ref.go to an unexported function in 'pkg/controller/migmigration/stage.go'
- Revert use of TruncateName in `pkg/controller/migplan/registry.go`
- Remove `ObjectKey` in pkg/reference/ref.go, inline the functionality where it's used in two places
- Possibly revert the signature change to `fail` in `pkg/controller/migmigration/task.go` once we determine it doesn't break anything
- consider avoiding `unstructured.UnstructuredList{}` in `pkg/apis/migration/v1alpha1/migplan_types.go` although it's not obvious off the top of my head what the alternative would look like.

sseago:
Fixes #525

This should include all of the items from #525. I haven't run through the earlier migration use cases again with this PR yet (will do that shortly) but should be complete, other than fixing any problems that might be exposed in testing.


sseago:
I've tested this with several migrations now. restic and snapshot copy, running and non-running PVCs. 3.11->4.2, 4.2->4.4

sseago:
@djwhatle actually this should be 1.2.1, since these fixes go against the 1.2.1 stage pod work. Also, the other 1.2.1 fix I'm working on is built on top of this PR which will be submitted as a new PR once this is merged.

djwhatle:
@sseago is this ready for review?

sseago:
@djwhatle yes.

djwhatle:
\test

PR #530, opened 2020-05-08
Bug 1831711: Fix incompatible GVK detection for 4.3->4.4 migrations

alaypatel07:
This patch fixes https://bugzilla.redhat.com/show_bug.cgi?id=1831711. 

The controller now uses `ServerPreferredNamespacedResources` to
list the preferred namespaced resources on source cluster and then
compare the group version with destination cluster. It adds two unit
tests to verify. As further steps, more extensive unit test needs to 
be added to understand this better


alaypatel07:
Logging the initial PR comment for history
```

This PR is meant to help in figuring out the problem for https://bugzilla.redhat.com/show_bug.cgi?id=1831711

This unit test needs to pass. My understanding is that in 4.4 only `apps/v1` is supported for Deployment. But the controllers checks if `apps/v1beta2` exists in 4.4 as it exists in 4.2

Probably using `ServerPreferredNamespacedResources`[1] for collecting source GVK's[2][3] might fix this bug.

1. https://github.com/konveyor/mig-controller/blob/e333d47d55dc6317e4a0495fafbb6cb991445204/vendor/k8s.io/client-go/discovery/discovery_client.go#L93
2. https://github.com/konveyor/mig-controller/blob/b3398c28b68679af5b6f912e02373c517c0a2d82/pkg/gvk/gvk.go#L33
3. https://github.com/konveyor/mig-controller/blob/b3398c28b68679af5b6f912e02373c517c0a2d82/pkg/gvk/gvk.go#L59
```

djwhatle:
\test

djwhatle:
This PR narrows the GVK diff list so that migrations 4.3 -> 4.4 can run. In a followup PR we plan to only notify the user of a GVK diff if the GVK diff is present in a namespace selected for migration.

Issue #504, opened 2020-04-20
Update CAM configuration for a cluster-scoped resource blacklist


dymurray:
ClusterResourceQuotas are cluster scoped resources that namespace scoped resources (Pods) depend on their existence in order to be properly restored. A more common example are PersistentVolumes. Generally, a user cannot list all PVs in the cluster, but CAM requires this to help make intelligent decisions about migrating PVCs in a given namespace.

Our plan to address this is to expose a list of cluster-scoped resources as a configuration option to allow an admin to control which cluster-scoped resources will automatically be migrated in non-admin migration.

sseago:
ResourceQuotas are created in the namespace for which they apply. So the whitelist wouldn't be needed for them. As for the general whitelist, how will that interact with velero in the case where velero itself (or the plugins) pulls resources in? PVs connected to PVCs are pulled in by velero explicitly. Same with CRDs. For others, such as SCCs, the plugin pulls those in as required to prevent migration failure.

dymurray:
Following up from some thoughts I've had after discussing this with Scott in slack... the more I've thought about this, I think that instead of passing this list of included resources in as annotations to the plugins, we should consider this to be an `alwaysInclude` list. In fact, it makes me wonder if we should add a new `include` field into the migplan spec. This would allow us to enable the ability to add label selector usage and other items in the API.

dymurray:
@sseago the more I dig into this, I'm realizing it probably just makes more sense while I'm doing this to expose the full capabilities of the includes/excludes list in Velero. What do you think about a new field on the `MigPlanSpec`:

```
type IncludesExcludes struct {
  IncludeNamespaces []string
  IncludeResources []string
  ExcludeNamespaces []string
  ExcludeResources []string
  IncludeClusterScoped bool
  LabelSelector map[string]string
}
```

That just exposes these fields to be passed into the Velero backup? And then for now the UX doesn't have to change, we can still use `IncludeNamespaces` but also leave the door open to more customized migrations down the road.

dymurray:
After discussing this further with Jeff, I think it's a bad idea to expose these APIs on the migplan at this moment and will make the validations for non-admin even more difficult than they already are. If we exposed `LabelSelector` then we would need to validate everything with the label selector can be read by the non-admin user. Not impossible, but no need to add the extra complexity right now.

In fact, the best way to solve this is to simply expose a whitelist for the admin on the `MigrationController` CR and have the tenant controllers respect this whitelist.

[...]

PR #533, opened 2020-05-11
Add cluster-scoped blacklist to settings

dymurray:
Dependent on https://github.com/konveyor/mig-operator/pull/338/


dymurray:
We need to fix the migmigration test with the new phases it looks like. Independent of my PR

pkg/settings/settings.go
```diff
@@ -77,3 +83,11 @@ func getEnvBool(name string, def bool) bool {
 
 	return boolean
 }
+
+func getClusterScopedBlacklist(name string, def []string) []string {
```
mansam, line 87: A minor nit: since this is only used by the Migration subpackage, this should probably live in that file, or possibly just be inlined into Migration.Load()

dymurray:
Fixes #504

djwhatle:
@dymurray this PR has been around for a while, thoughts on merge or close?

Issue #531, opened 2020-05-11
GVK differ reports incorrectly with multiple groups of same kind 


alaypatel07:
When multiple groups of the same kind are present and one of
the group is getting deprecated at the destination cluster, the
GVK diff complains about deprecated but Valero uses the other
group's preferred kind group-version to migrate successfully. 

Example: let's assume the source has two group of `deployment`
as follows:

```
Group: extensions
Version: v1beta1
Kind: deployments
```
and 
```
Group: apps
Version: v1
Kind: deployments
```

Assume the destination has deprecated `extensions` group so 
the GVK at the destination is: 

```
Group: apps
Version: v1
Kind: deployments
```

### Problem

The GVK differ will correctly identify that `extensions/v1beta1` deployments
is deprecated and incompatible. But because of the way Kube reads/writes
works the resource deployments will automatically be shifted to the preferred 
group version of destination. Hence extensions workloads will be migrated as
apps workloads in the above example. 

However, the controller checks through the list of incompatible GVK calculated
by the differ and checks if an object[[1]](https://github.com/konveyor/mig-controller/blob/d6e18043d84b21b82af530e930f7af41c0e6c1f2/pkg/gvk/gvk.go#L191) of any of the incompatible gvk's exist
 in the cluster and report the GVK if the resource exists. This creates a problem,
as `extensions` workloads are identified from the source cluster.

Hence the problem is for kinds with multiple groups if *at least* one group 
exists at the destination, the kind will be able to migrate and the controller
should remove all the incompatible GVKs with this kind 

### Solution
The controller should only complain if, an incompatible GVK exists for a kind with
no other group version available to migrate.

I propose adding a method here[[2]](https://github.com/konveyor/mig-controller/blob/d6e18043d84b21b82af530e930f7af41c0e6c1f2/pkg/gvk/gvk.go#L53), to filter the list of incompatible GVKs which
 do not have a compatible group kind alternative available for Valero to migrate. 
This filtered list will be the correct reflection of GVK's which cannot be migrated, 
as required

1. https://github.com/konveyor/mig-controller/blob/d6e18043d84b21b82af530e930f7af41c0e6c1f2/pkg/gvk/gvk.go#L191
2. https://github.com/konveyor/mig-controller/blob/d6e18043d84b21b82af530e930f7af41c0e6c1f2/pkg/gvk/gvk.go#L53

jwmatthews:
[...]

PR #534, opened 2020-05-11
Bug 1831711: Fix GVK diff to work for special kinds with multiple groups

alaypatel07:
This implements the fix described [here](https://github.com/konveyor/mig-controller/issues/531#issuecomment-626953888)

### DIFF
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -1,6 +1,7 @@
 package gvk
 
 import (
+	"sort"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


djwhatle:
\test

pkg/gvk/gvk.go
```diff
@@ -239,12 +267,20 @@ func resourceExist(resource metav1.APIResource, resources []metav1.APIResource)
...
+func (r *Compare) compareResources(src []*metav1.APIResourceList, dst []*metav1.APIResourceList) []*metav1.APIResourceList {
 	missingResources := []*metav1.APIResourceList{}
+	sortResources(src)
 	for _, srcList := range src {
 		missing := []metav1.APIResource{}
 		for _, resource := range srcList.APIResources {
-			if !resourceExist(resource, fingResourceList(srcList.GroupVersion, dst)) {
+			if cohabitator, found := r.CohabitatingResources[resource.Name]; found {
+				if cohabitator.seen {
+					// TODO: need to add log entry
```
djwhatle, line 278: @alaypatel07 you might be interested in learning about the controller `logging` package https://github.com/konveyor/mig-controller/blob/cac36386f9de729b65c452a453825254a6b9fe12/docs/devguide.md#logging

pkg/gvk/gvk.go
```diff
@@ -261,7 +297,7 @@ func compareResources(src []*metav1.APIResourceList, dst []*metav1.APIResourceLi
 	return missingResources
 }
 
-func fingResourceList(groupVersion string, list []*metav1.APIResourceList) []metav1.APIResource {
+func findResourceList(groupVersion string, list []*metav1.APIResourceList) []metav1.APIResource {
```
djwhatle, line 300: 👍

pkg/gvk/gvk.go
djwhatle: In mig-controller the logging philosophy is essentially "less is more". We rely heavily on CR conditions to convey information as needed. 

mig-controller only writes logs when:

1) an error occurs
2) it enters or exits stages in an active migration
3) controller or watches start/stop

djwhatle: In this case I would argue that the resulting MigPlan condition conveying GVK diff information would be sufficient, no log message required. Open to counter-argument.

pkg/gvk/gvk.go
```diff
@@ -19,12 +20,39 @@ var crdGVR = schema.GroupVersionResource{
...
+func newCohabitatingResource(resource, group1, group2 string) *cohabitatingResource {
+	return &cohabitatingResource{
+		resource:       resource,
+		groupResource1: schema.GroupResource{Group: group1, Resource: resource},
+		groupResource2: schema.GroupResource{Group: group2, Resource: resource},
+		seen:           false,
+	}
+}
+
+func NewCohabitatingResources() map[string]*cohabitatingResource {
```
djwhatle, line 39: Does this need to be exported?

alaypatel07: Yes, I believe so because it is used in another package for instantiation[[1]](https://github.com/konveyor/mig-controller/pull/534/files#diff-ff4cf8d4f8669d3270d95d4c88c43266R67).

1. https://github.com/konveyor/mig-controller/pull/534/files#diff-ff4cf8d4f8669d3270d95d4c88c43266R67

pkg/gvk/gvk.go
alaypatel07: ack, not logging here makes sense in this case. Upstream Velero logs here, but I was not sure if we want our controller to log. I'll remove the TODO.

alaypatel07:
updated as per review, PTAL whenever you get a chance @djwhatle

pkg/gvk/gvk.go
```diff
@@ -19,12 +20,39 @@ var crdGVR = schema.GroupVersionResource{
...
+	return map[string]*cohabitatingResource{
```
djwhatle, line 40: FYI for anyone looking on, this is borrowed from Velero source.

Our goal here is to accurately simulate the logic that Velero run to choose which resource to back up when cohabitating API groups are present on a source cluster.

pkg/gvk/gvk.go
```diff
@@ -290,3 +325,29 @@ func isCRDGroup(group string, crdGroups []string) bool {
 
 	return false
 }
+
+// sortResources sources resources by moving extensions to the end of the slice. The order of all
+// the other resources is preserved.
+func sortResources(resources []*metav1.APIResourceList) {
```
djwhatle, line 331: Another piece borrowed from Velero to accurately simulate the logic used in Velero to determine which resources should be prioritized in a backup.

djwhatle:
Closes #531

PR #536, opened 2020-05-14
Set memory/CPU on stage pods to avoid errors with ResourceQuotas

sseago:
Set stage pod memory to 128Mi and CPU to 100m for stage pod containers
unless there are LimitRanges specifying higher minimums, in which case
use the higher minimums.

### DIFF
--- a/pkg/controller/migmigration/stage.go
+++ b/pkg/controller/migmigration/stage.go
@@ -12,6 +12,7 @@ import (
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	migpods "github.com/konveyor/mig-controller/pkg/pods"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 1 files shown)


djwhatle:
@sseago is there an associated issue for this PR? If not, let's create one.

sseago:
Right now I think we only have the JIRAs: 
https://issues.redhat.com/browse/MIG-202
https://issues.redhat.com/browse/MIG-204

sseago:
Basically if there's a ResourceQuota in a namespace being migrated (or a ClusterResourceQuota which matches the namespace being migrated) that specifies a quota limit for CPU or memory, then any pod that doesn't specify request and limit values for CPU and memory will fail to schedule, which means that the stage pods as we were creating them would fail to schedule. The starting point for this PR is to specify limit/request values of 100m for CPU and 128Mi for memory for all stage pods. However, that initial fix causes a different sort of failure if there are LimitRanges defined with minimum CPU or memory per-pod or per-container which are higher than these defaults, so I also needed to find the largest minimum CPU and memory values from the LimitRanges for each namespace and use the larger of this and the default value in creating stage pods.

djwhatle:
@sseago so if I'm understanding correctly:


1. By default, give stage pods `limits` and `requests` of  100m CPU and 128Mi memory

2. If `LimitRanges` are defined, don't use defaults mentioned above. Stage pods will **instead** receive resource allocation within set `LimitRanges` 

3. Pod scheduler is now happy, so Stage Pod can run.

djwhatle:
\test

sseago:
@djwhatle That's mostly correct. Basically we use the larger of those defaults and the LimitRange minimum values. So we'll never use less than 100m/128Mi, just in case it's possible to create a LimitRange with a minimum value that is otherwise too low for a container to schedule successfully.

PR #540, opened 2020-05-19
Bug 1837737 - Use data instead of binaryData so hooks work w/3.7 and 3.9

jmontleon:


jmontleon:
binaryData was not introduced until Kubernetes 1.10/OpenShift 3.10 so when the ConfigMap is created on 3.7 or 3.10 the binaryData field just gets dropped.

jmontleon:
I don't think there is a BZ for this yet, but there should be one coming. QE found the issue today.

jmontleon:
https://bugzilla.redhat.com/show_bug.cgi?id=1837737

PR #542, opened 2020-05-20
Fix mutex usage of RLock() when should be Lock().

### LINKED ISSUE #541: Discovery controller has occasional database lock error
Few usages _mutex_ `RLock()` should be `Lock()`.

jortel:
#fixes #541

### DIFF
--- a/pkg/controller/discovery/container/collection.go
+++ b/pkg/controller/discovery/container/collection.go
@@ -131,8 +131,6 @@ func (r *SimpleReconciler) Reconcile(collection Collection) (err error) {
 			dpn.discovered = m
 		}
 	}
-	model.Mutex.RLock()
-	defer model.Mutex.RUnlock()
 	tx, err := r.Db.Begin()
 	if err != nil {
 		Log.Trace(err)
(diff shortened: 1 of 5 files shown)


pkg/controller/discovery/container/collection.go
```diff
@@ -131,8 +131,6 @@ func (r *SimpleReconciler) Reconcile(collection Collection) (err error) {
 			dpn.discovered = m
 		}
 	}
-	model.Mutex.RLock()
-	defer model.Mutex.RUnlock()
```
mansam, line 135: Is the mutex not necessary at all here?

jortel: The intent was to acquire the mutex for the duration of transaction.  The individual Insert(), Update() and Delete() will also call Lock() and because it is not reentrant will cause a panic.  It would be better to hold the lock for the duration of the transaction.  I'll investigate options.

djwhatle:
\test

djwhatle:
\test

djwhatle:
CI may be broken. Checking.

pranavgaikwad:
\test

pranavgaikwad:
@djwhatle a small typo in operator definition in master caused the CI to break. Fixed with this PR : 
https://github.com/konveyor/mig-operator/pull/361

djwhatle:
@jortel I don't recall the state this PR was in. Should we merge / close?

Issue #388, opened 2020-01-07
Allow for non cluster-admin users to run migrations


dymurray:
This is a tracking issue for all the work required to allow non cluster-admin users to run migrations. The design document is here: https://github.com/konveyor/enhancements/blob/master/enhancements/non-admin/cam-non-admin-design.md

Feature Branch across all repos:
* https://github.com/konveyor/mig-controller/tree/non-admin
* https://github.com/konveyor/mig-ui/tree/non-admin
* https://github.com/konveyor/mig-operator/tree/non-admin

**mig-controller**:
- [x] #389  - Enable migplan controller to take in a user's identity token for src + dest
- [ ] #391  - Discovery controller authorization should be using SAR checks with user's identity
- [x] #393  - Controller should pre-create destination namespaces
- [x] #470  - Add authorization phase to migration controller workflow
- [x] #471  - Update all controllers to watch resources in the sandbox namespace
- [x] #472  - Investigate singleton discovery controller vs 1 per controller
- [x] #504 - Update CAM configuration to allow for a blacklist of cluster-scoped resources to not allow non-admin user to migrate
- [x] #512 - Add MigToken CRD
- [x] #544 - Add helper script to create MigTenant for testing
- [x] #545 - Update MigToken CRD to include OAuth details
- [x]  allow mig-controller to watch all namespaces 
- [ ] #561 - Improve tokenreview/SAR check workflow
- [ ] #550 - Dynamically generate list of GVK's that require SAR check
- [ ] #555 - add ProjectRequest SAR check to validite MigToken
- [ ] #570 - Restricting object references to non-admin namespaces

**mig-operator**
- [x] https://github.com/konveyor/mig-operator/issues/253 - Use migration-controller SA rather than cluster-admin
- [x] https://github.com/konveyor/mig-operator/issues/285 - Add MigrationTenant CRD
- [x] https://github.com/konveyor/mig-operator/issues/286 - Update operator to watch all namespaces for MigrationTenant creation
- [x] https://github.com/konveyor/mig-operator/issues/315 - Update Ansible roles for non-admin
- [ ] https://github.com/konveyor/mig-operator/issues/287 - Ensure that 1.2+ upgrades are possible and not breaking current installations

**mig-ui**
- [ ] https://github.com/konveyor/mig-ui/issues/789 - Break out identity creation in UI

**Velero** - Optional
- [ ] https://github.com/vmware-tanzu/velero/issues/1531 - Support concurrent Restic backups
[...]

PR #544, opened 2020-05-21
Add helper script to create MigTenant, non-admin users and namespace

alaypatel07:
This adds a script for setting up test users `bob` and `alice`
on a ocp 4 cluster. Bob will have access to the host cluster,
alice will be restricted.

### DIFF
--- a/hack/setup-bob-4.4.sh
+++ b/hack/setup-bob-4.4.sh
@@ -17,6 +17,7 @@ else
 fi
 
 htpasswd -c -B -b users.htpasswd bob bob
+htpasswd -B -b users.htpasswd alice alice
 oc create secret generic htpass-secret --from-file=htpasswd=./users.htpasswd -n openshift-config
 oc patch oauth cluster --type='json' -p '[ { "op": "replace", "path": "/spec", "value": {"identityProviders": [{"htpasswd": {"fileData": {"name": "htpass-secret"}}, "mappingMethod": "claim", "name": "bob_httpassd_provider", "type": "HTPasswd"}]}}]'
 sleep 2
@@ -35,5 +36,20 @@ spec:
   olm_managed: true
   version: 1.0 (OLM)
 EOF
+
+cat <<EOF | oc apply -f -
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: openshift-migration
+  name: use-host-cluster
+rules:
+- apiGroups: ["migration.openshift.io"]
+  resources: ["migclusters"]
+  verbs: ["use"]
+  resourceNames: ["host"]
+EOF
+
+oc create rolebinding bob-use-host --role use-cluster --user bob   --namespace openshift-migration
 
 rm users.htpasswd
--- a/hack/setup-bob-4.4.sh
+++ b/hack/setup-bob-4.4.sh
@@ -1,0 +1,39 @@
+if [[ -z "$KUBECONFIG" ]]; then
+  echo "please export KUBECONFIG env var"
+  exit 1
+fi
+
+if [[ -z "$KUBEADMIN_PASSWORD" ]]; then
+  echo "please export KUBEADMIN_PASSWORD env var"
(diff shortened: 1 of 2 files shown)


djwhatle:
Related to non-admin work https://github.com/konveyor/mig-controller/issues/388

hack/setup-bob-4.4.sh
```diff
@@ -0,0 +1,39 @@
+if [[ -z "$KUBECONFIG" ]]; then
```
djwhatle, line 1: @alaypatel07 how would you feel about renaming this script to something like `setup-migtenant.sh`

hack/setup-bob-4.4.sh
```diff
@@ -0,0 +1,55 @@
...
+  echo "not a 4.x cluster, please export correct kubeconfig";
+  exit 1
+fi
+
+htpasswd -c -B -b users.htpasswd bob bob
+htpasswd -B -b users.htpasswd alice alice
+oc create secret generic htpass-secret --from-file=htpasswd=./users.htpasswd -n openshift-config
+oc patch oauth cluster --type='json' -p '[ { "op": "replace", "path": "/spec", "value": {"identityProviders": [{"htpasswd": {"fileData": {"name": "htpass-secret"}}, "mappingMethod": "claim", "name": "bob_httpassd_provider", "type": "HTPasswd"}]}}]'
+sleep 2
+oc login -u bob -p bob
```
djwhatle, line 24: We could avoid mutating the existing KUBECONFIG and requesting the KUBEADMIN password for re-login if we make a temporary copy of the active KUBECONFIG. 

https://github.com/konveyor/mig-controller/blob/master/hack/controller-sa-login.sh#L10-L12

hack/setup-bob-4.4.sh
alaypatel07: sure, good idea, may be `setup-users-4.4.sh`? the script is not actually creating a MigrationTenant resource so wanted to avoid that confusion

hack/setup-bob-4.4.sh
alaypatel07: +1, this looks exactly like what I needed to get rid of KUBEADMIN_PADDWORD, will modify

hack/setup-bob-4.4.sh
```diff
@@ -0,0 +1,55 @@
...
+oc new-project bob-destination
+oc login -u kubeadmin -p "$KUBEADMIN_PASSWORD"
+cat <<EOF | oc create -f -
+apiVersion: migration.openshift.io/v1alpha1
+kind: MigrationTenant
```
djwhatle, line 29: @alaypatel07 I'm confused, it looks to me like we _are_ creating a MigTenant. What am I missing?

alaypatel07: I am missing caffeine :) my bad

hack/setup-bob-4.4.sh
djwhatle: @alaypatel07 I think we should mention in the script name that this is related to non-admin. Also, it looks to me like we are creating a MigTenant in this script.

hack/setup-migtenant-ocp4.sh
```diff
@@ -0,0 +1,75 @@
...
+rules:
+- apiGroups: ["migration.openshift.io"]
+  resources: ["migclusters"]
+  verbs: ["use"]
+  resourceNames: ["host"]
+EOF
+
+oc create rolebinding bob-use-host --role use-cluster --user bob   --namespace openshift-migration
+
+rm users.htpasswd
```
djwhatle, line 75: I think we could add a final step here of removing the temporary KUBECONFIG that we create.

```
rm $KUBEADMIN_KUBECONFIG
```

alaypatel07: +1

dymurray:
Awesome, love that we have Bob+Alice and only Bob has `use` access. Will be very helpful.

PR #554, opened 2020-06-04
add expirasAt field and refactor SelfSAR check to SAR checks

alaypatel07:
This PR:
- add an `expiresAt` field to provide expiration time in UTC from #553 
- convert SelfSubjectAccessReview calls to SubjectAccessReview with
  username and scope from token review calls from MigToken
- refactors the code to add two getters on MigToken
- add few unexported status fields to keep the number of
     TokenReview calls low

PR #559, opened 2020-06-04
allow the controllers to watch all namespaces

alaypatel07:
This PR removes the filtering logic based on env variables
`TENANT_NAMESPACE` and `PRIVILEGED_NAMESPACE`.
The controller now watches all reconciles MigTokens 
(and other Mig*) in all namespaces 

Depends on #554


pkg/settings/settings.go
```diff
@@ -13,7 +13,6 @@ var Settings = _Settings{}
 //   Plan: Plan settings.
 type _Settings struct {
 	Discovery
-	Namespace
```
jortel, line 16: Delete the `namespace.go`?

alaypatel07: no reason, removed it. Thanks

PR #563, opened 2020-06-09
Add non-admin support in discovery controller.

jortel:
Add support for non-admin.
Changes:
- Change authorization model to perform SAR for specific resource being requested.
- Add _/wellknown_ endpoint needed by the UI to renew tokens.
- Add _/auth_ endpoint needed by the UI to determine if token (user) has (MIG) admin on a specific namespace.
- Included a fix done in earlier versions of non-admin.  Mainly, that it inventories MigPlan like any other resource.  This fixes the problem of not detecting a plan has been deleted while the controller is down.

### DIFF
--- a/pkg/controller/discovery/auth/rbac.go
+++ b/pkg/controller/discovery/auth/rbac.go
@@ -1,0 +1,342 @@
+package auth
+
+import (
+	"context"
+	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
+	"github.com/konveyor/mig-controller/pkg/logging"
+	project "github.com/openshift/api/project/v1"
+	auth "k8s.io/api/authentication/v1beta1"
+	"k8s.io/api/authorization/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/utils/pointer"
+	"net/http"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//
+// Shared logger.
+var Log *logging.Logger
+
+//
+// k8s Resources.
+const (
+	ALL       = "*"
+	Namespace = "namespaces"
+	PV        = "persistentvolumes"
+	PVC       = "persistentvolumeclaims"
+	Service   = "services"
+	Pod       = "pods"
+	PodLog    = "pods/log"
+)
+
+//
+// Verbs
+const (
+	ANY    = "*"
+	LIST   = "list"
+	GET    = "get"
+	CREATE = "create"
+	DELETE = "delete"
+	PATCH  = "patch"
+	UPDATE = "update"
+)
+
+//
+// SAR request.
+type Review = v1.ResourceAttributes
+
+//
+// RBAC authorization provider.
+type Provider interface {
+	// Authenticate the token.
+	Authenticate() (bool, error)
+	// Get whether the token was authenticated.
+	Authenticated() bool
+	// Determine of actions specified are authorized
+	// for the token.
+	Allow(*Review) (bool, error)
+	// Determine if the all of the actions
+	// specified in the matrix are authorized for
+	// the token.
+	AllowMatrix(*Matrix) (bool, error)
+	// Get the user name.
+	User() string
+}
+
+//
+// Review Matrix.
+type Matrix struct {
+	// API group.
+	Group string
+	// Namespace
+	Namespace string
+	// Resource name.
+	Name string
+	// Resources.
+	Resources []string
+	// Verbs.
+	Verbs []string
+}
+
+//
+// Provides RBAC authorization.
+type RBAC struct {
+	client.Client
+	// Cluster
+	Cluster *migapi.MigCluster
+	// The subject bearer token.
+	Token string
+	// Token review status.
+	tokenStatus auth.TokenReviewStatus
+	// Token authenticated.
+	authenticated *bool
+	// Project-based authorization.
+	pbac *PBAC
+}
+
+//
+// Allow/deny the subject review request.
+// Special (optimized) steps for namespaces:
+//   1. Populate `allowedProject`.
+//   2. Match namespace in `allowedProject`.
+//   3. Match `*` in `allowedProject`.
+//   4. SAR
+// The SAR for each namespace is last resort. For large
+// clusters this will be very slow.
+func (r *RBAC) Allow(request *Review) (bool, error) {
+	authenticated, err := r.Authenticate()
+	if err != nil {
+		Log.Trace(err)
+		return false, err
+	}
+	if !authenticated {
+		return false, nil
+	}
+	if r.pbac == nil {
+		r.pbac = &PBAC{
+			Cluster: r.Cluster,
+			Token:   r.Token,
+		}
+	}
+	switch request.Resource {
+	case Namespace:
+		switch request.Verb {
+		case LIST, GET:
+			if r.pbac.Supported() {
+				allowed := r.pbac.Allow(request.Name)
+				return allowed, nil
+			}
+		}
+		fallthrough
+	default:
+		return r.sar(request)
+	}
+}
+
+//
+// Allow all the combinations in the matrix.
+// See: Allow() for details.
+func (r *RBAC) AllowMatrix(m *Matrix) (bool, error) {
+	mAllowed := false
+	request := &Review{
+		Group:     m.Group,
+		Namespace: m.Namespace,
+		Name:      m.Name,
+	}
+	for _, resource := range m.Resources {
+		for _, verb := range m.Verbs {
+			mAllowed = true
+			request.Resource = resource
+			request.Verb = verb
+			allowed, err := r.Allow(request)
+			if err != nil {
+				Log.Trace(err)
+				return false, err
+			}
+			if !allowed {
+				return false, nil
+			}
+		}
+	}
+
+	return mAllowed, nil
+}
+
+//
+// Get whether authenticated
+func (r *RBAC) Authenticated() bool {
+	return r != nil && *r.authenticated
+}
+
+//
+// Get the user name for the token.
+func (r *RBAC) User() string {
+	if r.Authenticated() {
+		return r.tokenStatus.User.Username
+	}
+
+	return ""
+}
+
+//
+// Get the user ID for the token.
+func (r *RBAC) UID() string {
+	if r.Authenticated() {
+		return r.tokenStatus.User.UID
+	}
+
+	return ""
+}
+
+//
+// Get the user's groups.
+func (r *RBAC) Groups() []string {
+	if r.Authenticated() {
+		return r.tokenStatus.User.Groups
+	}
+
+	return []string{}
+}
+
+//
+// Get the extra fields for the token.
+func (r *RBAC) Extra() map[string]v1.ExtraValue {
+	extra := map[string]v1.ExtraValue{}
+	if r.Authenticated() {
+		for k, v := range r.tokenStatus.User.Extra {
+			extra[k] = append(
+				v1.ExtraValue{},
+				v...)
+		}
+	}
+
+	return extra
+}
+
+//
+// Do subject access review.
+func (r *RBAC) sar(request *Review) (bool, error) {
+	sar := v1.SubjectAccessReview{
+		Spec: v1.SubjectAccessReviewSpec{
+			ResourceAttributes: request,
+			Groups:             r.Groups(),
+			User:               r.User(),
+			UID:                r.UID(),
+			Extra:              r.Extra(),
+		},
+	}
+	err := r.Client.Create(context.TODO(), &sar)
+	if err != nil {
+		Log.Trace(err)
+		return false, err
+	}
+
+	return sar.Status.Allowed, nil
+}
+
+//
+// Authenticate the token.
+// Build the `subjectClient`.
+func (r *RBAC) Authenticate() (bool, error) {
+	if r.authenticated != nil {
+		return *r.authenticated, nil
+	}
+	r.authenticated = pointer.BoolPtr(false)
+	tr := &auth.TokenReview{
+		Spec: auth.TokenReviewSpec{
+			Token: r.Token,
+		},
+	}
+	err := r.Client.Create(context.TODO(), tr)
+	if err != nil {
+		Log.Trace(err)
+		return false, err
+	}
+	if !tr.Status.Authenticated {
+		return false, err
+	}
+	r.tokenStatus = tr.Status
+	r.authenticated = pointer.BoolPtr(true)
+
+	return true, nil
+}
+
+//
+// Provides project-based authorization.
+type PBAC struct {
+	// Token
+	Token string
+	// Cluster
+	Cluster *migapi.MigCluster
+	// Project API supported by the cluster.
+	supported bool
+	// List of projects.
+	projects map[string]bool
+}
+
+//
+// Project API is supported by the cluster.
+func (p *PBAC) Supported() bool {
+	p.Load()
+	return p.supported
+}
+
+//
+// The token is allowed access to the specified namespace.
+func (p *PBAC) Allow(ns string) bool {
+	p.Load()
+	_, found := p.projects[ns]
+	return found
+}
+
+//
+// Load projects (as needed).
+func (p *PBAC) Load() {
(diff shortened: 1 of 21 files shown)


pkg/controller/discovery/web/wellknown.go
```diff
@@ -0,0 +1,92 @@
...
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
+	"github.com/gin-gonic/gin"
+	"io/ioutil"
+	"k8s.io/client-go/rest"
+	"net/http"
+)
+
+// Route roots.
```
djwhatle, line 13: This is fun to say out loud.

pkg/controller/discovery/auth/sar.go
```diff
@@ -0,0 +1,331 @@
...
+	tokenStatus auth.TokenReviewStatus
+	// Token authenticated.
+	authenticated *bool
+	// Allowed project list.
+	allowedProject map[string]bool
+}
+
+//
+// Allow/deny the subject review request.
+// Special (optimized) steps for namespaces:
```
djwhatle, line 84: Summarizing the `Allow()` method in my own words for understanding. 

**For namespaces**:
 - Checks if token has `LIST` and `GET` permission on namespace using cached data
   - Grants access to project if token has both of these , returns true
 - If token doesn't have `LIST`, `GET` runs SAR check (due to `fallthrough`) which doesn't rely on cached data

**For any other kind of resource**
 - Immediately runs SAR check

---

I'm curious, are SAR checks inherently slow for namespaces moreso than other resources?

pkg/controller/discovery/container/ds.go
```diff
@@ -210,6 +218,8 @@ func (r *DataSource) buildClient(cluster *migapi.MigCluster) error {
 		Log.Trace(err)
 		return err
 	}
+	r.RestCfg.Burst = 1000
```
djwhatle, line 221: I guess these are the client-side rate controls we talked about.

pkg/controller/discovery/web/ns.go
```diff
@@ -117,29 +198,68 @@ func (h NsHandler) List(ctx *gin.Context) {
...
+		Namespace: ns,
+		Resource:  auth.ALL,
+		Verb:      auth.ALL,
+	})
+}
+
+//
+// Has basic MIG user access.
+func (h *NsHandler) hasBasic(ns string) (bool, error) {
+	allowed, err := h.hasAdmin(ns)
```
alaypatel07, line 228: uhm I am confused, why does this function call `hasAdmin`? looks like the non-admin user will always return on line https://github.com/konveyor/mig-controller/pull/563/files#diff-e12f082564b41558aa0342877bc3fd84R231 when they should be going through checks on https://github.com/konveyor/mig-controller/pull/563/files#diff-e12f082564b41558aa0342877bc3fd84R236

djwhatle: See the conversation I just "unresolved" above

pkg/controller/discovery/auth/sar.go
```diff
@@ -0,0 +1,331 @@
...
+	GET    = "get"
+	CREATE = "create"
+	DELETE = "delete"
+	PATCH  = "patch"
+	UPDATE = "update"
+)
+
+//
+// SAR request.
+type Review = v1.ResourceAttributes
```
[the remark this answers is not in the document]
jortel: Abstraction.

pkg/controller/discovery/auth/sar.go
```diff
@@ -0,0 +1,331 @@
...
+	// Verbs.
+	Verbs []string
+}
+
+//
+// Provides RBAC authorization.
+type RBAC struct {
+	client.Client
+	// Cluster
+	Cluster *model.Cluster
```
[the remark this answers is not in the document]
jortel: In discovery controller MigCluster is at the very edges.  Caller only concerned with odel.Cluster. That said, changing to MigCluster makes sense in light of plan to relocate and use more generally.

pkg/controller/discovery/auth/sar.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+}
+
+//
+// Build the subject client using the REST
+// configuration and token.
+func (r *RBAC) subjectClient() (client.Client, error) {
+	update := func(cfg *rest.Config) {
+		cfg.BearerToken = r.Token
+		cfg.Burst = 1000
```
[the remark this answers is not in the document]
jortel: Agreed.

pkg/controller/discovery/auth/sar.go
```diff
@@ -0,0 +1,331 @@
...
+		restCfg,
+		client.Options{
+			Scheme: scheme.Scheme,
+		})
+	if err != nil {
+		if stErr, cast := err.(*errors.StatusError); cast {
+			switch stErr.Status().Code {
+			case http.StatusUnauthorized,
+				http.StatusForbidden:
+				return nil, nil
```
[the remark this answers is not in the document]
jortel: No, because 401 and 403 are not errors in the failure sense but represent an invalid token which is an anticipated condition.

pkg/controller/discovery/auth/sar.go
```diff
@@ -0,0 +1,331 @@
...
+// Load project list (as needed).
+// When project API not supported, it will fall-back to
+// a SAR to determine if the subject has GET on all namespaces and
+// add the `ALL: true` entry.
+func (r *RBAC) loadProjects() {
+	if r.allowedProject != nil {
+		return
+	}
+	r.allowedProject = map[string]bool{}
+	subjectClient, err := r.subjectClient()
```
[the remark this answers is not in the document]
jortel: Typically, when error != nil, there is not guarantee about the other returned values  (in this case subjectClient) and they should not be used.  Which it's not.  In this case, we don't care about the specific error.

pkg/controller/discovery/web/ns.go
jortel: The SAR checks are very expensive in quantity.  This is an optimization for cases where the user is a cluster-admin and we're checking 2k namespaces (which can require 4k SAR).  For admins, this is intended to provided a x3 reduction.  For non-admins it adds one more check but the basic check will likely reject on the first SAR.

jortel: The discovery service is authorizing the REST API and must be low latency.  This would not be necessary if we could agree on a balance between performance and certainty.  IMHO, the goal is not to provide an absolute guarantee the user has ALL necessary permission but instead a list of namespaces based on a quick estimate.  If the user has "*" on MigPlan the will most certainly have permission for everything else.  In any case, the downside to being wrong is they get an error message in the UI when they don't have permission.

pkg/controller/discovery/auth/sar.go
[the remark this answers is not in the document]
jortel: Done.

pkg/controller/discovery/auth/sar.go
```diff
@@ -0,0 +1,331 @@
...
+	for _, p := range list.Items {
+		r.allowedProject[p.Name] = true
+	}
+}
+
+//
+// Build the subject client using the REST
+// configuration and token.
+func (r *RBAC) subjectClient() (client.Client, error) {
+	update := func(cfg *rest.Config) {
```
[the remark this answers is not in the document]
jortel: Fixed.

pkg/controller/discovery/web/ns.go
```diff
@@ -16,17 +21,84 @@ const (
...
+		if allowed {
+			content.Items = append(
+				content.Items,
+				Namespace{
+					Name: m.Name,
+				})
+		}
+	}
+
+	Log.Info("****** ROOT NS", "d", time.Since(mark))
```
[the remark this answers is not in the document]
jortel: Fixed.

pkg/controller/discovery/web/ns.go
```diff
@@ -117,29 +198,68 @@ func (h NsHandler) List(ctx *gin.Context) {
...
+		Group:     migapi.SchemeGroupVersion.Group,
+		Namespace: ns,
+		Resource:  auth.ALL,
+		Verb:      auth.ALL,
+	})
+}
+
+//
+// Has basic MIG user access.
+func (h *NsHandler) hasBasic(ns string) (bool, error) {
```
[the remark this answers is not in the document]
jortel: The goal is to short circuit the hasBasic() check when the user is an admin.  1 SAR vs 4 x each namespace.

pkg/controller/discovery/auth/sar.go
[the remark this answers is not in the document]
alaypatel07: hrm so I am talking about when error is nil and subject client is also nil (when returned from [here](https://github.com/konveyor/mig-controller/pull/563/files#r438887558)) and `subjectClient.List` is called a few lines below. Looks like controller will panic right?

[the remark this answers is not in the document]
alaypatel07: In general, I prefer functions/APIs such that when error is nil, the other values are assumed to be usable and will not create a nil pointer panic. This avoids the need to check for the validity of other return values and just check the error returned.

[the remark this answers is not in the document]
jortel: > hrm so I am talking about when error is nil and subject client is also nil (when returned from [here](https://github.com/konveyor/mig-controller/pull/563/files#r438887558)) and `subjectClient.List` is called a few lines below. Looks like controller will panic right?

Ah, Now I follow.  Having this discussion on https://github.com/konveyor/mig-controller/pull/563/files#r438887558 would have been easier to follow. 


> In general, I prefer functions/APIs such that when error is nil, the other values are assumed to be usable and will not create a nil pointer panic. This avoids the need to check for the validity of other return values and just check the error returned.

Yes. Of course.

pkg/controller/discovery/auth/sar.go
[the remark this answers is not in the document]
jortel: > wondering if we should have `return nil, err` instead?

Yes, definitely.  Fixed.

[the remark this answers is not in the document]
alaypatel07: thanks!

pkg/controller/discovery/auth/sar.go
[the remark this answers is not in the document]
alaypatel07: thanks, resolving since this is fixed at source.

pkg/controller/discovery/web/ns.go
```diff
@@ -117,29 +194,69 @@ func (h NsHandler) List(ctx *gin.Context) {
...
+	ctx.Status(http.StatusMethodNotAllowed)
+}
+
+//
+// Has read access.
+func (h *NsHandler) hasRead(ns string) (bool, error) {
+	return h.rbac.Allow(&auth.Review{
+		Resource: auth.Namespace,
+		Verb:     auth.GET,
+		Name:     ns,
```
alaypatel07, line 206: it is weird but it looks like we need to have `Namespace` in the request... 

```
[alpatel@alpatel mig-operator]$ kubectl create -f - -o yaml << EOF
apiVersion: authorization.k8s.io/v1
kind: SubjectAccessReview
spec:
  user: bob
  resourceAttributes:
    group: ""
    resource: namespaces
    verb: "get"
    namespace: bob-destination
    name: bob-destination
  uid: 190531b5-590a-4ea6-98a4-9b45cdfeb094
  extra:
    scopes:
    - user:full
EOF
apiVersion: authorization.k8s.io/v1
kind: SubjectAccessReview
metadata:
  creationTimestamp: null
spec:
  extra:
    scopes:
    - user:full
  resourceAttributes:
    name: bob-destination
    namespace: bob-destination
    resource: namespaces
    verb: get
  uid: 190531b5-590a-4ea6-98a4-9b45cdfeb094
  user: bob
status:
  allowed: true
  reason: 'RBAC: allowed by RoleBinding "admin/bob-destination" of ClusterRole "admin"
    to User "bob"'
[alpatel@alpatel mig-operator]$ kubectl create -f - -o yaml << EOF
apiVersion: authorization.k8s.io/v1
kind: SubjectAccessReview
spec:
  user: bob
  resourceAttributes:
    group: ""
    resource: namespaces
    verb: "get"
    name: bob-destination
  uid: 190531b5-590a-4ea6-98a4-9b45cdfeb094
  extra:
    scopes:
    - user:full
EOF
apiVersion: authorization.k8s.io/v1
kind: SubjectAccessReview
metadata:
  creationTimestamp: null
spec:
  extra:
    scopes:
    - user:full
  resourceAttributes:
    name: bob-destination
    resource: namespaces
    verb: get
  uid: 190531b5-590a-4ea6-98a4-9b45cdfeb094
  user: bob
status:
  allowed: false
[alpatel@alpatel mig-operator]$
```

jortel: Fixed.

pkg/controller/discovery/auth/sar.go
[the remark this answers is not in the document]
alaypatel07: hrm so AFAIK all the clients that use the config with these fields seem to use the same value, might be missing something. Thought it might make refactoring these constants later safer/easier

That said, if it is not convincing, not a merge blocker from my end!

pkg/controller/discovery/auth/rbac.go
```diff
@@ -0,0 +1,324 @@
...
+	tokenClient, err := p.client()
+	if err != nil {
+		return
+	}
+	list := &project.ProjectList{}
+	err = tokenClient.List(context.TODO(), nil, list)
+	if err != nil {
+		return
+	}
+	p.supported = true
```
alaypatel07, line 291: What happens when the API server responds with a transient error? looks like there in that case the user will no use project API even if the project is available but apiserver returned and expected error? 

This does not feel like the right thing to do, it feels like figuring out if projects API is supported or not using discovery API is a safer alternative.

pkg/controller/discovery/web/ns.go
```diff
@@ -117,29 +167,107 @@ func (h NsHandler) List(ctx *gin.Context) {
...
+		Resource:  auth.ALL,
+		Verb:      auth.ALL,
+	})
+}
+
+//
+// Filter the list of namespaces based on token having
+// at least READ access.
+func (h *NsHandler) hasRead(models []*model.Namespace) ([]*model.Namespace, error) {
+	hasAdmin, err := h.hasAdmin()
```
alaypatel07, line 188: can we please put a comment before this function? IMO it will help in re-reading this code in future
```
// check and return fast if user has access to mig-api in all namespaces
```

pkg/controller/discovery/auth/rbac.go
alaypatel07: After the discussion, it was concluded that since RBAC and PBAC structs are instantiated on every request, handling transient api server will not break functionality, it will only fall back to less optimized SAR checks. Resolving because this does not need to change

djwhatle: @alaypatel07 what is PBAC?

pkg/apis/migration/v1alpha1/migtoken_types.go
```diff
@@ -306,9 +306,6 @@ func (r *MigToken) GetTokenReview(client k8sclient.Client) (*v1beta1.TokenReview
 	if err != nil {
 		return nil, err
 	}
-	if cluster == nil {
```
alaypatel07, line 309: +1

alaypatel07: actually -1, need to rebase, so this doesnt get dropped

jortel: Woops.  Done.

pkg/controller/discovery/auth/rbac.go
```diff
@@ -0,0 +1,324 @@
...
+		Group:     m.Group,
+		Namespace: m.Namespace,
+		Name:      m.Name,
+	}
+	for _, resource := range m.Resources {
+		for _, verb := range m.Verbs {
+			mAllowed = true
+			request.Resource = resource
+			request.Verb = verb
+			allowed, err := r.Allow(request)
```
alaypatel07, line 134: are these requests are missing username, uid and scopes?

alaypatel07: they are taken care of in sar() method

PR #567, opened 2020-06-11
Bug 1846521: Add fallback to parse gitVersion when version.Major/Minor not available

djwhatle:
We have found cases where an OCP cluster does not provide the `major` and `minor` fields in the response from the `version` endpoint. Example below.

```
Response Body: {
  "major": "",
  "minor": "",
  "gitVersion": "v1.9.1+a0ce1bc657",
  "gitCommit": "a0ce1bc",
  "gitTreeState": "clean",
  "buildDate": "2018-04-26T16:48:23Z",
  "goVersion": "go1.9.4",
  "compiler": "gc",
  "platform": "linux/amd64"
}
```

These missing fields are breaking our `compat` client, which relies on knowing the cluster version.

In these cases, we can parse the `gitVersion` field as a fallback. We will still use the major/minor fields by default.

---

**Tracking Links**
https://bugzilla.redhat.com/show_bug.cgi?id=1846521
https://issues.redhat.com/browse/MIG-261

Related plugin PR: https://github.com/konveyor/openshift-migration-plugin/pull/65

### DIFF
--- a/pkg/compat/client.go
+++ b/pkg/compat/client.go
@@ -2,6 +2,8 @@ package compat
 
 import (
 	"context"
+	"errors"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -57,13 +59,14 @@ func NewClient(restCfg *rest.Config) (Client, error) {
 	if err != nil {
 		return nil, err
 	}
-	major, err := strconv.Atoi(version.Major)
-	if err != nil {
-		return nil, err
-	}
-	minor, err := strconv.Atoi(strings.Trim(version.Minor, "+"))
-	if err != nil {
-		return nil, err
+	// Attempt parsing version.Major/Minor first, fall back to parsing gitVersion
+	major, err1 := strconv.Atoi(version.Major)
+	minor, err2 := strconv.Atoi(strings.Trim(version.Minor, "+"))
+	if err1 != nil || err2 != nil {
+		major, minor, err = parseGitVersion(version.GitVersion)
+		if err != nil {
+			return nil, err
+		}
 	}
 	nClient := &client{
 		Config:             restCfg,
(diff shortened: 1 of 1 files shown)


djwhatle:
\test

pkg/compat/client.go
```diff
@@ -219,3 +222,31 @@ func (c client) Update(ctx context.Context, in runtime.Object) error {
 	}
 	return c.Client.Update(ctx, obj)
 }
+
+//
+// Parses gitVersion ("v1.11.0+d4cacc0"), returns major, minor, err
+func parseGitVersion(gitVersion string) (int, int, error) {
```
jortel, line 228: Recommend making this a _method_ instead of a package function unless you indent for it to be generally used within the whole package.

pkg/compat/client.go
```diff
@@ -219,3 +222,31 @@ func (c client) Update(ctx context.Context, in runtime.Object) error {
...
+		return 0, 0, errors.New("gitVersion does not match expected format")
+	}
+	versionArr := strings.Split(gitVersion, "v")
+	if len(versionArr) < 2 {
+		return 0, 0, errors.New("gitVersion missing leading 'v'")
+	}
+	majorMinorArr := strings.Split(versionArr[1], ".")
+	if len(majorMinorArr) < 2 {
+		return 0, 0, errors.New("gitVersion major + minor parse failed")
+	}
```
jortel, line 241: Won't the `MatchString()` catch the missing `v` and `.` errors?
I don't see how these checks would be reached.

djwhatle: @jortel I would think yes, but was trying to be extra safe to avoid controller panic if I try to index into something that doesn't exist.

djwhatle: I thought of a way to make this a little bit cleaner.

jmontleon:
This missing version was a bug in 3.9:
https://github.com/openshift/origin/pull/19731

The build output above is dated April 2018. The PR is from May 2018.

pkg/compat/client.go
```diff
@@ -57,14 +59,30 @@ func NewClient(restCfg *rest.Config) (Client, error) {
...
+	minor, err2 := strconv.Atoi(strings.Trim(version.Minor, "+"))
+
+	if err1 != nil || err2 != nil {
+		// gitVersion format ("v1.11.0+d4cacc0")
+		r, _ := regexp.Compile(`v[0-9]+\.[0-9]+`)
+		valid := r.MatchString(version.GitVersion)
+		if !valid {
+			return nil, errors.New("gitVersion does not match expected format")
+		}
+		majorMinorArr := strings.Split(strings.Split(version.GitVersion, "v")[1], ".")
```
djwhatle, line 74: @jortel made some changes in light of your review

1. Removed extra checks on "indexability". I agree, I can't think of a way for code to reach that point. 
1. I don't think this parsing functionality will likely to be needed elsewhere, so I integrated it with the `NewClient()` method

djwhatle:
\test

djwhatle:
I'll hold off on merging this until plugin fixes are ready as well. Removing 1.2.2 target from this PR.

pkg/compat/client.go
```diff
@@ -57,14 +59,30 @@ func NewClient(restCfg *rest.Config) (Client, error) {
...
-	if err != nil {
-		return nil, err
+
+	// Attempt parsing version.Major/Minor first, fall back to parsing gitVersion
+	major, err1 := strconv.Atoi(version.Major)
+	minor, err2 := strconv.Atoi(strings.Trim(version.Minor, "+"))
+
+	if err1 != nil || err2 != nil {
+		// gitVersion format ("v1.11.0+d4cacc0")
+		r, _ := regexp.Compile(`v[0-9]+\.[0-9]+`)
```
sseago, line 69: It's probably better to verify that there's a second "." in the version string (i.e. that there's always a third number), otherwise the "." delimited split below won't pull out just an int for the second value:
`regexp.Compile(`v[0-9]+\.[0-9]+\.`)`

djwhatle: ACK, good catch! I'll change this and re-test on a few sample `gitVersion` strings before merging.

PR #571, opened 2020-06-17
Bug 1847993: requeue rather than fail if error is resource conflict

sseago:


pkg/controller/migmigration/migrate.go
```diff
@@ -78,6 +79,9 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (time.Du
 	err = task.Run()
 	if err != nil {
 		log.Trace(err)
```
jortel, line 81: Even through the logger will not trace _Conflict_, it would be more efficient to put the Log() after.

PR #572, opened 2020-06-20
Disable imagestreams from migration

dymurray:


pkg/settings/migration.go
```diff
@@ -2,18 +2,22 @@ package settings
...
 type Migration struct {
-	FailureRollback bool
+	FailureRollback       bool
+	DisableImageMigration bool
 }
 
 // Load settings.
 func (r *Migration) Load() error {
 	r.FailureRollback = getEnvBool(FailureRollback, false)
+	r.DisableImageMigration = getEnvBool(DisableImageMigration, false)
```
dymurray, line 20: Going to temporarily set this to true so this is hardcoded into the image we are building for this test case.

PR #576, opened 2020-06-22
HF5 Remove CSVs from backup

dymurray:


jmontleon:
Looks good to me for excluding CSVs. I thought we saw subscriptions fail from 3.11 -> 4 as well. I'm not sure if removing CSVs is going to lead to an InstallPlan or Subscription failing, then onto the next for each of the OLM CRs. Not sure how others feel about adding those proactively.

jwmatthews:
I think this needs some more testing.

Let's have someone spin up an OCP 3.11 with OLM tomorrow, install something, then test a migration.

I agree, likely other issues will be teased out we need to exclude even to have a degraded migration complete.

jmontleon:
FWIW when we saw the failed subscription restore cause a rollback before we disabled the behavior I thought it only resulted in a PartiallyFailed restore result in velero. I didn't think that actually caused a failed migration, other than not restoring the incompatible CRs

eriknelson:
>  I thought we saw subscriptions fail from 3.11 -> 4 as well.

I've added OLM to my 3.11 cluster and installed a mongo operator and tried to migrate it with this patch. No CSV error, but exactly as @jmontleon described it failed with the following Subscription error on the restore: https://gist.githubusercontent.com/eriknelson/5b9db423a52e2576b06560f218f25d0d/raw/07efc644b8aa6422d233a0a75c87bea9d597817a/gistfile1.txt

Root issue here is the Subscriptions in a 4.4 cluster are not API compatible with the version shipped with OLM 3.11 Tech Preview. Unsure the correct way to handle, although I'm testing adding subscriptions now to the list of excluded resources.

eriknelson:
Created Prometheus instance w/ operator and ran hotfix #6 excluding Subscriptions as well. Migration succeeded, but prometheus workload failed to launch on target cluster with following error: https://gist.githubusercontent.com/eriknelson/84a924faa6dec4aa20f0ff27aa92a174/raw/85c0a8f3548c4d36e7864dc68f67907358fcdf92/gistfile1.txt
I believe this is still something we want to add to the proposed hotfix #5; it would allow migrations to continue and the application specific errors (prometheus in this case) can be repaired post-migration.

jmontleon:
@eriknelson is this just adding subscription in addition to clusterserviceversion? Do we want to get this in and deliver them another hotfix if so?

PR #578, opened 2020-06-24
skip PV/stage steps for final migration

sseago:


djwhatle:
\test

sseago:
Updated to disable the test which ensures that the stage itinerary is a subset of the final itinerary.

sseago:
\test

PR #579, opened 2020-06-25
Make "skip PVs in final migration" configurable

sseago:
By default, skip PVs in final migration. To do the normal CAM
stage backup/restore step with PVs as well, add this to the
MigPlan (before migrating):
    annotations:
      migration.openshift.io/migratePVsInFinal: "false"


sseago:
\test

PR #583, opened 2020-06-30
Handle deleted/terminating PVC.

### LINKED ISSUE #582: Migration stalls at StagePodsStarted when PVC deleted/terminated.
When a PVC is deleted after a migration has started, it causes the phase=StagePodsStarted to get stuck forever.  The stage pod gets stuck at _ContainerCreating_ forever which is fever detected by the controller.

On OCP 3.9, the PVC is deleted immediately.
On OCP 3.11, the PVC has a finalizer which causes the PVC to be in _Terminated_ state.

If the PV is deleted while the migration is running, the PVC will phase=`Lost`.

jortel:
fixes #582

When checking to see if stage pods are all phase=_Running_, the logic needed to be updated as follows:
When a stage pod is not _Running, determine if all of the PVC's are healthy:
1. Get the PVC.
2. PVC phase=bound
3. PVC (3.11+) is not `Terminating`.  This is determined by looking at the `deletionTimestamp`.

The goal is to detected cases whereby the pod will never start and fail the migration.

### DIFF
--- a/pkg/controller/migmigration/stage.go
+++ b/pkg/controller/migmigration/stage.go
@@ -2,6 +2,7 @@ package migmigration
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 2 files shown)


djwhatle:
Will need cherry-pick into release-1.2.4 after branch is created.

PR #586, opened 2020-07-14
Enable custom k8s resource exclusions

jmontleon:


jmontleon:
Tied to https://github.com/konveyor/mig-operator/pull/379

jmontleon:
From the golang docs it looks like reading a non-existent env var will result in an empty string. Splitting an empty string will result in an empty slice, so you'll just end up with an empty exclude list with no errors or crashes other than what might occur during the from not excluding problem resources.

PR #587, opened 2020-07-14
Bug 1856814 - add namespace list  to job pod as an env var

jmontleon:
https://bugzilla.redhat.com/show_bug.cgi?id=1856814

### DIFF
--- a/pkg/controller/migmigration/hooks.go
+++ b/pkg/controller/migmigration/hooks.go
@@ -235,6 +235,16 @@ func (t *Task) baseJobTemplate(hook migapi.MigPlanHook, migHook migapi.MigHook) 
 						{
 							Name:  strings.ToLower(t.PlanResources.MigPlan.Name + "-" + hook.Phase),
 							Image: migHook.Spec.Image,
+							Env: []corev1.EnvVar{
+								{
+									Name:  "MIGRATION_NAMESPACES",
+									Value: strings.Join(t.PlanResources.MigPlan.Spec.Namespaces, ","),
+								},
+								{
+									Name:  "MIGRATION_PLAN_NAME",
+									Value: t.PlanResources.MigPlan.Name,
+								},
+							},
 						},
 					},
 					RestartPolicy:         "OnFailure",


jmontleon:
I don't know if anyone has ideas on what other information might be useful to add besides these.

eriknelson:
+1 from me, although I dunno if I should be the one acking this through.

djwhatle:
I remember thinking this type of information would be really useful to have in the hook runner container. +1, ACK

djwhatle:
\test

djwhatle:
Wow that CI finished way too fast. Anyways, simple change, so merge away.

pkg/controller/migmigration/hooks.go
```diff
@@ -235,6 +235,16 @@ func (t *Task) baseJobTemplate(hook migapi.MigPlanHook, migHook migapi.MigHook)
...
 							Name:  strings.ToLower(t.PlanResources.MigPlan.Name + "-" + hook.Phase),
 							Image: migHook.Spec.Image,
+							Env: []corev1.EnvVar{
+								{
+									Name:  "MIGRATION_NAMESPACES",
+									Value: strings.Join(t.PlanResources.MigPlan.Spec.Namespaces, ","),
+								},
+								{
+									Name:  "MIGRATION_NAME",
+									Value: t.PlanResources.MigPlan.Name,
```
djwhatle, line 245: @jmontleon this looks wrong. Shouldn't this be `meta.name`?

djwhatle: This is also the `MIGRATION_PLAN_NAME`, not the `MIGRATION_NAME`. Maybe we should update to reflect the data we're providing.

jmontleon: It works correctly. It's also how the hook name has been generated all along:
https://github.com/konveyor/mig-controller/blob/master/pkg/controller/migmigration/hooks.go#L170

```
$ oc get po hooktest-prebackup-8hjzx-znlq7 -o yaml | grep MIGRATION -A1
    - name: MIGRATION_NAMESPACES
      value: registry-images
    - name: MIGRATION_NAME
      value: hooktest

$ oc get migplan
NAME       AGE
hooktest   26h
```

Not objection to changing the env var name.

jmontleon: unless I'm doing it wrong and misunderstanding Meta.Name doesn't work
image

jmontleon:
Updated `MIGRATION_NAME` to `MIGRATION_PLAN_NAME`

pkg/controller/migmigration/hooks.go
djwhatle: Ahh, my bad. I forgot realize TypeMeta and ObjectMeta and inlined on the MigPlan struct so they're available from top-level like that. Spec isn't inlined.

```
type MigPlan struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
```

PR #588, opened 2020-07-14
Enable custom k8s resource exclusions

sseago:
Jira: https://issues.redhat.com/browse/MIG-273
Corresponding mig-operator PR: https://github.com/konveyor/mig-operator/pull/379

### DIFF
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/deckarep/golang-set"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	"github.com/pkg/errors"
 	velero "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
@@ -38,7 +39,8 @@ func (t *Task) ensureInitialBackup() (*velero.Backup, error) {
 		return nil, err
 	}
 	newBackup.Labels[InitialBackupLabel] = t.UID()
-	newBackup.Spec.ExcludedResources = excludedInitialResources
+	newBackup.Spec.IncludedResources = toStringSlice(includedInitialResources.Difference(toSet(t.PlanResources.MigPlan.Status.ExcludedResources)))
+	newBackup.Spec.ExcludedResources = toStringSlice(excludedInitialResources.Union(toSet(t.PlanResources.MigPlan.Status.ExcludedResources)))
 	delete(newBackup.Annotations, QuiesceAnnotation)
 	err = client.Create(context.TODO(), newBackup)
 	if err != nil {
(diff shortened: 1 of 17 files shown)


sseago:
This will need an operator PR for the CRD change.

jmontleon:
https://github.com/konveyor/mig-operator/pull/379/files

sseago:
@jmontleon Should we put the migplan CRD change into that operator PR or should we create a separate one? This is to add MigPlan.Status.ExcludedResources.

sseago:
@jmontleon CRD change is in https://github.com/konveyor/mig-controller/pull/588 
This PR should work without error without 379, since the only thing it depends on from that PR is the env var, and if it's not set, we don't add any exclusions.

pkg/controller/migplan/migplan_controller.go
```diff
@@ -362,3 +371,16 @@ func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
 
 	return nil
 }
+
+// Update Status.ExcludedResources based on
+// "EXCLUDE_RESOURCES" environment variable
+func (r *ReconcileMigPlan) ensureExcludedResourceList(plan *migapi.MigPlan) error {
+	var excludedResources []string
+	resourcesEnv := os.Getenv("EXCLUDE_RESOURCES")
```
jortel, line 379: All of the controller settings (and environment varables) are encapsulated in the _settings_ package.  

Let's add `ExcludedResources []string` [here](https://github.com/konveyor/mig-controller/blob/master/pkg/settings/plan.go#L14).

Which is loaded are parsed into the list [here](https://github.com/konveyor/mig-controller/blob/master/pkg/settings/plan.go#L21)

And used:
```
excludedResources := Settings.Plan.ExcludedResources
```

pkg/controller/migplan/migplan_controller.go
```diff
@@ -218,6 +220,13 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	// Set excluded resources on Status.
+	err = r.ensureExcludedResourceList(plan)
```
jortel, line 224: After seeing it implemented, I'm not convinced that _ensure_ is the best fit.  Seems like it may imply more than it's doing since we typically use _ensure_ when affecting the outside world.  I don't feel strongly.

Perhaps something like `setExcludedResources()`.

pkg/controller/migmigration/backup.go
```diff
@@ -48,6 +48,27 @@ func (t *Task) ensureInitialBackup() (*velero.Backup, error) {
 	return newBackup, nil
 }
 
+// Combine MigPlan excluded resources with velero excluded resources
+func (t *Task) appendMigPlanExcludes(veleroExcludedResources []string) []string {
```
jortel, line 52: I think this can be simplified using set logic.  The Go pattern I have seen for this is using a `map[string]bool` as a `set` (the value can be anything).
1. Add all of the resources in the velero list.
2. Add all of the resources in the plan list.
3. Return the map keys.

pkg/controller/migplan/migplan_controller.go
sseago: I can use set. Keep in mind that eventually this set method may be merging the EXCLUDE_RESOURCES env var and a future Spec.ExcludedResources slice. If set still works for that, then I can change to set here.

djwhatle:
We'll need to cherry-pick this PR to release-1.2.4 once it merges.

sseago:
@jortel FYI, I'm going to have to add something to this tomorrow. In addition to merging the MigPlan excludes list with the velero excludes list, I also need to *subtract* the MigPlan excludes list from the velero *Includes* list. If persistentvolumes is on both the excludes and includes list, velero complains (for obvious reasons). It should be pretty straightforward to fix this, though.

pkg/controller/migplan/migplan_controller.go
```diff
@@ -362,3 +369,11 @@ func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
 
 	return nil
 }
+
+// Update Status.ExcludedResources based on
+// "EXCLUDED_RESOURCES" environment variable
```
jortel, line 374: Based on settings.

sseago:
\test

Issue #558, opened 2020-06-04
NFS server accessibility validation running when not needed (NFS copy)


djwhatle:
## **Problem Description:**
Migrating from NFS (src) to gp2 (target) using **NFS copy,** it isn't necessary for us to run a validation on NFS accessibility from the target cluster, but a warning condition is still being raised.

Reported by Sergio Regidor.

**PVC Migration Selections**
```
pvc:
      accessModes:
      - ReadWriteOnce
      name: nginx-html
      namespace: nginx-pv
    selection:
      action: copy
      copyMethod: filesystem
      storageClass: gp2
    storageClass: nfs
    supported:
      actions:
      - copy
      copyMethods:
      - filesystem
      - snapshot
```

**Warning Condition**
```
  - category: Warn
    lastTransitionTime: "2020-06-04T17:14:13Z"
    message: NFS servers [172.30.184.84] not accessible on the destination cluster.
    status: "True"
    type: NfsNotAccessible
```

---

**Full MigPlan**
```
kind: MigPlan
metadata:
  annotations:
    openshift.io/touch: 9b65071a-a688-11ea-a1b5-0a580a81020c
  creationTimestamp: "2020-06-04T17:14:04Z"
  generation: 3
  labels:
    controller-tools.k8s.io: "1.0"
  name: nginx-pv-migplan-1591290259
  namespace: openshift-migration
  resourceVersion: "71654"
  selfLink: /apis/migration.openshift.io/v1alpha1/namespaces/openshift-migration/migplans/nginx-pv-migplan-1591290259
  uid: f72c2199-6d25-4f29-8512-136e58199ad4
spec:
  destMigClusterRef:
    name: host
    namespace: openshift-migration
  migStorageRef:
    name: automatic
    namespace: openshift-migration
  namespaces:
  - nginx-pv
  persistentVolumes:
  - capacity: 1Mi
    name: pvc-bd52bd24-a686-11ea-a19e-0e9d0082032d
    pvc:
      accessModes:
      - ReadWriteOnce
      name: nginx-logs
      namespace: nginx-pv
    selection:
      action: copy
      copyMethod: filesystem
      storageClass: gp2
    storageClass: nfs
    supported:
      actions:
      - copy
      copyMethods:
      - filesystem
      - snapshot
  - capacity: 50Mi
    name: pvc-bd5c1f75-a686-11ea-a19e-0e9d0082032d
    pvc:
      accessModes:
      - ReadWriteOnce
      name: nginx-html
[...]

PR #590, opened 2020-07-16
Fix NFS PVC validation for move action only

aufi:
NFS accessibility validation is not needed when PVC data should be copied to the destination (and not moved).

Updated PVC validation not to be executed on copy action.

Fixes https://github.com/konveyor/mig-controller/issues/558

### DIFF
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -1206,7 +1206,8 @@ func (r *NfsValidation) validate() error {
 	server := map[string]bool{}
 	for i := range r.Plan.Spec.PersistentVolumes.List {
 		pv := &r.Plan.Spec.PersistentVolumes.List[i]
-		if pv.NFS == nil {
+		// Skip non-NFS PVs and PVs which data should be copied
+		if pv.NFS == nil || pv.Selection.Action == "copy" {
 			continue
 		}
 		if passed, found := server[pv.NFS.Server]; found {


djwhatle:
Hey @aufi, thanks for the fix! I requested more info on the reproducer steps there.

pkg/controller/migplan/validation.go
```diff
@@ -1206,7 +1206,8 @@ func (r *NfsValidation) validate() error {
 	server := map[string]bool{}
 	for i := range r.Plan.Spec.PersistentVolumes.List {
 		pv := &r.Plan.Spec.PersistentVolumes.List[i]
-		if pv.NFS == nil {
+		// Skip non-NFS PVs and PVs which data should be copied
+		if pv.NFS == nil || pv.Selection.Action == "copy" {
```
djwhatle, line 1210: Fix looks sane, nice and simple. @aufi have you had a chance to test this yet?

djwhatle:
\test

pkg/controller/migplan/validation.go
aufi: Sure, it worked. I dumped MigPlan content for both cases (with/without the change), https://gist.github.com/aufi/cb914d64a7aa0cbe8c3b57302dda7a2c. I didn't capture screencast video, but I could make it if needed. (only issue I had with testing was not reliable local execution of validations)

djwhatle: YAML looks good. Adding my ACK 👍

djwhatle:
@aufi ready to merge this?

aufi:
@djwhatle From my side, ready

PR #591, opened 2020-07-16
Add client metrics posted to Prometheus.

jortel:
Add client metrics posted to Prometheus.


pkg/controller/migmigration/stage.go
```diff
@@ -42,7 +42,6 @@ type PodStartReport struct {
 	reasons []string
 	// all pods started.
 	started bool
-
```
jortel, line 45: Changed by `make fmt` Missed by an earlier PR.

pkg/compat/metrics.go
```diff
@@ -0,0 +1,126 @@
...
+
+//
+// Report `delete` API call.
+func (m *Reporter) Delete(client Client, object api.Object) {
+	m.report(client, Delete, object)
+}
+
+//
+// Determine the call context.
+func (m *Reporter) context() (component, function string) {
```
djwhatle, line 86: What's a practical example of this `context()`? E.g. what will the return values be.

djwhatle:
Regarding an enable/disable switch via `settings`, I think it would make sense to just have this on by default. I don't believe there will be any real performance impact to our controller from keeping these metrics updated.

djwhatle:
@jortel one final follow-up question: do you have a feel for how much of the controller code is _not_ using the custom client? E.g. how many calls will go untracked.

djwhatle:
Resolves #594

jortel:
> @jortel one final follow-up question: do you have a feel for how much of the controller code is _not_ using the custom client? E.g. how many calls will go untracked.

All of the source/destination cluster API calls use the custom client.  The client associated with each _Reconciler_ is the cached client obtained from the `Manager` and is only used for _local_ API calls not related to source/destination cluster reconciliation.  This would be mainly `Mig*` resources and associated secrets.

pkg/compat/metrics.go
jortel: > What's a practical example of this `context()`? E.g. what will the return values be.

The component (which controller) and the package qualified function name making the API call.

PR #595, opened 2020-07-22
Use `liberr` from controller common pkg

djwhatle:
## Summary
Changed all instances of `log.Trace(err)` + `return`. Using liberr.wrap allows the common logger to unwrap the errors and provide context from tip of the call stack.

```
// previous
log.Trace(err)
return err
```

```
// replacement
return liberr.Wrap(err)
```

Note: I did not make changes in the `discovery` pkg, as I believe @jortel agreed that these changes should be made at a later time with other discovery controller refactoring.

## Todo
- [x] Add common logger pkg as replacement for current logger so that wrapped errors will be handled correctly.
- [x] Remove mig-controller logging pkg
- [x] Remove extra comments in PR
- [x] Investigate new deps that appeared around mapset
- [x] In `Reconcile()` do `log.Trace(err)` instead of wrap and return.
- [x] In functions only called by controller-runtime, make sure to not wrap errors since our custom error won't be usable by controller-runtime
- [x] Run manual sanity checks to make sure errors are returned in appropriate manner


pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -123,7 +120,7 @@ func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Re
 		if errors.IsNotFound(err) {
 			return reconcile.Result{}, nil
 		}
-		log.Trace(err)
+		log.Trace(err) // TODO - handle with liberr
```
djwhatle, line 123: @jortel what is the appropriate action to take in places like this where the err isn't immediately returned? Just drop the `log.Trace()`?

pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -207,14 +204,12 @@ func (r *ReconcileMigCluster) setupRemoteWatch(cluster *migapi.MigCluster) error
...
-			log.Trace(err)
-			return err
+			return liberr.Wrap(err)
 		}
 	} else {
 		restCfg, err = cluster.BuildRestConfig(r.Client)
 		if err != nil {
-			log.Trace(err)
-			return err
+			return liberr.Wrap(err)
```
djwhatle, line 212: @jortel is this generally in line with what you were expecting from this PR? A switch from `log.Trace()` over to `liberr.Wrap()`?

pkg/controller/migcluster/migcluster_controller.go
jortel: In the main reconcile loop, I think we should leave this as is.  Same for anywhere else in the code that is Log.Trace(err) but not returning the error.

pkg/controller/migcluster/migcluster_controller.go
jortel: Yes, exactly.

djwhatle:
Note to self: need to add common logger pkg as replacement for current logger.

pkg/controller/migmigration/migrate.go
```diff
@@ -85,7 +85,7 @@ func (r *ReconcileMigMigration) migrate(migration *migapi.MigMigration) (time.Du
 		if errors.IsConflict(err) {
 			return FastReQ, nil
 		}
-		log.Trace(err)
+		log.Trace(err) // TODO - handle with liberr
 		task.fail(MigrationFailed, []string{err.Error()})
 		return task.Requeue, nil
```
jortel, line 90: This should be:
```
- log.Trace(err)
- return task.Requeue, nil
+ return task.Requeue, liberr.Wrap(err)
```

djwhatle: We want to return an err where we previously did not? Wouldn't this perhaps throw off some logic in the caller?

pkg/controller/migmigration/stage.go
```diff
@@ -409,7 +392,7 @@ func (t *Task) ensureStagePodsStarted() (report PodStartReport, err error) {
 			if err != nil {
 				healthy = false
 				if !errors.IsNotFound(err) {
-					log.Trace(err)
+					err = liberr.Wrap(err)
```
jortel, line 395: No change here.

pkg/controller/migmigration/migrate.go
djwhatle: We agreed that in this case, we should just leave the log.Trace where it is.

pkg/controller/migmigration/stage.go
djwhatle: 👍

pkg/controller/migmigration/backup.go
```diff
@@ -5,7 +5,8 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/deckarep/golang-set"
+	mapset "github.com/deckarep/golang-set"
```
djwhatle, line 8: @jortel I'm lost as to why this got auto-aliased, I did not make this change. Any ideas?

pkg/controller/migmigration/migmigration_controller.go
```diff
@@ -21,11 +21,12 @@ import (
...
 	"time"
 
+	liberr "github.com/konveyor/controller/pkg/error"
+	"github.com/konveyor/controller/pkg/logging"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
-	"github.com/konveyor/mig-controller/pkg/logging"
 	migref "github.com/konveyor/mig-controller/pkg/reference"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
djwhatle, line 29: I also did not add this alias...

djwhatle:
@jortel PR marked as ready for review. Please take a look.  I think I took care of all of the items we discussed.

pkg/controller/migmigration/backup.go
jortel: No idea.

jortel:
Places missed?:
https://github.com/konveyor/mig-controller/blob/ddff7f5dff6169294b81197de0d713ddb23b25ea/pkg/controller/migmigration/stage.go#L169
https://github.com/konveyor/mig-controller/blob/ddff7f5dff6169294b81197de0d713ddb23b25ea/pkg/controller/migmigration/stage.go#L197

djwhatle:
@jortel I'm not sure I can safely return a different value (an actual err instead of nil) without changing behavior.

djwhatle:
\test

djwhatle:
@pranavgaikwad @fbladilo any idea what's up with CI?

djwhatle:
\test

pkg/controller/migcluster/remote_watch.go
```diff
@@ -34,8 +35,7 @@ func StartRemoteWatch(r *ReconcileMigCluster, config remote.ManagerConfig) error
 
 	mgr, err := manager.New(config.RemoteRestConfig, manager.Options{})
 	if err != nil {
-		log.Error(err, "[rWatch] Unable to set up remote watcher controller manager")
```
djwhatle, line 37: Removed some custom error messaging from RemoteWatch code

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -20,8 +20,7 @@ import (
 	"context"
 	"time"
 
-	"github.com/konveyor/mig-controller/pkg/logging"
-
+	"github.com/konveyor/controller/pkg/logging"
```
djwhatle, line 23: Switched to common logger

djwhatle:
@jortel ran some sanity checks after final changes, 2 migrations ran without issue. Will plan to merge this on Monday in case you want to take another look.

djwhatle:
\test

PR #604, opened 2020-07-28
Add Analytics controller

jmontleon:
**Jira**: [MIG-277](https://issues.redhat.com/browse/MIG-277)

* [x] Add Plan totals in addition to the NS totals
* [x] Add Operator PR to include the new CR
* [x] Validation(s) (Check if the plan referenced exists...)
* [x] Add Documentation to the Operator Repo
* [x] Exclude excluded resources from counts if possible
* [x] Add PV Count fields per NS and Plan in CR
* [x] Exclude resources excluded by GVK incompatibility if possible
* [X] Add Annotation/Label with associated migplan
* [X] Add Progress
* [X] Add Lists/Totals for Incompatible / Excluded Resources
* [X]  Investigate and implement image size total


jmontleon:
Current example of output:

jmontleon:
I'll look at removing the dep that's pulling in the pile of stuff. I think I got stuck because of the ImageStreamList schema stuff, but I believe it's taken care of here so it's possible it's low effort for me to get rid of all of it. https://github.com/konveyor/mig-controller/blob/98952cc05c61fd85cbef11943e64548dee0df915/pkg/imagescheme/register.go

jmontleon:
gone

jmontleon:
Operator PR https://github.com/konveyor/mig-operator/pull/396

jmontleon:
I had success against a QE 3.7 cluster, testing backward compatibility. Of particular concern was PV Capacity since I had to copy/rewrite a function from a newer version of the k8s api library than we're using to get the quantities, but it appears to be functioning.

```
metadata:
  creationTimestamp: "2020-07-30T15:09:11Z"
  generation: 2
  name: test
  resourceVersion: "73875619"
  selfLink: /apis/migration.openshift.io/v1alpha1/namespaces/openshift-migration/miganalytics/test
  uid: 515b404c-c6d0-45d6-92db-2f7bb594b5ae
spec:
  analyzeImageCount: true
  analyzeK8SResources: true
  analyzePVCapacity: true
  plan:
    name: test37
status:
  analytics:
    k8sResourceTotal: 29
    namespaces:
    - k8sResourceTotal: 29
      k8sResources:
      - count: 1
        kind: endpoints
        version: v1
      - count: 10
        kind: events
        version: v1
      - count: 1
        kind: persistentvolumeclaims
        version: v1
      - count: 1
        kind: pods
        version: v1
      - count: 9
        kind: secrets
        version: v1
      - count: 3
        kind: serviceaccounts
        version: v1
      - count: 1
        kind: services
        version: v1
      - count: 1
        group: extensions
        kind: deployments
        version: v1beta1
      - count: 1
        group: extensions
        kind: replicasets
        version: v1beta1
      - count: 1
        group: apps
        kind: deployments
        version: v1beta1
      namespace: mediawiki
      pvCapacity: 1Gi
    plan: test37
    pvCapacity: 1Gi
  conditions:
  - category: Required
    lastTransitionTime: "2020-07-30T15:09:22Z"
    message: The analytic is ready.
    status: "True"
    type: Ready
  observedGeneration: 2
```

jmontleon:
@djwhatle would appreciate another look when you have time. Functionally I think it's close. There's probably some cleanup to do at the least though.

jmontleon:
Feature Documentation for anyone wishing to test
https://github.com/konveyor/mig-operator/blob/91d15fc7be4f79ee53c9014b0ba85e04d57c6b81/docs/usage/Analytics.md#analytics

Custom operator/image for testing is available from this operator source:
```
apiVersion: operators.coreos.com/v1
kind: OperatorSource
metadata:
  name: migration-operator
  namespace: openshift-marketplace
spec:
  type: appregistry
  endpoint: https://quay.io/cnr
  registryNamespace: jmontleon
  displayName: "Migration Operator"
  publisher: "ocp-migrate-team@redhat.com"

```

djwhatle:
Thanks for the docs Jason, I'll plan to do another round of review on this today.

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -0,0 +1,582 @@
...
+}
+
+func (r *ReconcileMigAnalytic) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	var err error
+	log.Reset()
+
+	// Fetch the MigAnalytic instance
+	analytic := &migapi.MigAnalytic{
+		Spec: migapi.MigAnalyticSpec{
+			MaxListImages: 25,
```
djwhatle, line 109: I would recommend:


1. moving "25" into a named const so we don't have magic numbers, and so that we can change this in one place if changes are needed in future
2. not defaulting in this way, let's keep consistent across our `Get` operations and just initialize an empty object before running `Get`

I feel that this is an unusual way to apply a default. I would recommend just grabbing the const value you create from wherever you need it later on instead  of baking it into the object as a default.

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -0,0 +1,582 @@
...
+	if err != nil {
+		log.Trace(err)
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	// Analyze
+	if !analytic.Status.HasBlockerCondition() {
+		err = r.analyze(analytic)
+	} else {
+		log.Info("Requeuing MigAnalytic, associated plan not ready.", "miganalytic", analytic.Name)
```
djwhatle, line 160: I suggest:

1. Removing this log statement. We only log phase change and errors.
2. Moving this info onto a new MigAnalytic status condition and use "postponed" language 

```
# Example from MigMigration
	// Postpone
	requeueAfter := time.Second * 10
	for position, uid := range pending {
		if uid == migration.UID {
			requeueAfter = time.Second * time.Duration(position*10)
			break
		}
	}
	migration.Status.SetCondition(migapi.Condition{
		Type:     Postponed,
		Status:   True,
		Category: Critical,
		Message:  fmt.Sprintf(PostponedMessage, requeueAfter/time.Second),
	})
```

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -0,0 +1,582 @@
...
+		analytic.Status.Analytics.ImageCount += ns.ImageCount
+		analytic.Status.Analytics.ImageSizeTotal.Add(ns.ImageSizeTotal)
+		analytic.Status.Analytics.PVCapacity.Add(ns.PVCapacity)
+		analytic.Status.Analytics.PVCount += ns.PVCount
+		analytic.Status.Analytics.PercentComplete = (i + 1) * 100 / len(plan.Spec.Namespaces)
+
+		err = r.Update(context.TODO(), analytic)
+		if err != nil {
+			log.Trace(err)
+			return err
```
djwhatle, line 267: Anywhere in this PR outside of 
```
func (r *ReconcileMigAnalytic) Reconcile(request reconcile.Request) (reconcile.Result, error)
```
We should be returning _all_ errors using with liberr.Wrap(). As an example:

```
# Replace this...
if err != nil {
   log.Trace(err)
   return err
}

# With this:
if err != nil {
   return liberr.Wrap(err)
}
```

Only in the root `Reconcile()` methods should we still be doing log.Trace()

More info: https://github.com/konveyor/mig-controller/pull/595

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -0,0 +1,582 @@
...
+		}
+		log.Info("Completed analysis of namespace.", "namespace", namespace)
+	}
+	log.Info("Completed analysis of plan.", "plan", plan.Name)
+	return nil
+}
+
+func (r *ReconcileMigAnalytic) analyzeK8SResources(dynamic dynamic.Interface, resources []*metav1.APIResourceList, namespace *migapi.MigAnalyticNamespace, excludedResources []string, incompatible v1alpha1.Incompatible) error {
```
djwhatle, line 275: Can we get a comment summary of what all is happening in this method? Let's consider decomposing this func to bring down line lengths.

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -0,0 +1,582 @@
...
+
+					dockerImageMetadata := docker10.DockerImage{}
+					err = json.Unmarshal(image.DockerImageMetadata.Raw, &dockerImageMetadata)
+					if err != nil {
+						log.Trace(err)
+						return err
+					}
+
+					namespace.ImageCount += 1
+					namespace.ImageSizeTotal.Add(resource.MustParse((strconv.Itoa(int(dockerImageMetadata.Size)/MiB) + MiBSuffix)))
```
djwhatle, line 396: I think we should try to decompose `analyzeImages()` into the major pieces bring down line lengths.

pkg/apis/migration/v1alpha1/miganalytic_types.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	kapi "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// MigAnalyticSpec defines the desired state of MigAnalytic
+type MigAnalyticSpec struct {
```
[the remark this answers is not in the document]
djwhatle: @jmontleon after thinking on this for a while longer, due to the way that k8s allows us to look up resources, it will respond to

```
oc get miganalytics
```

Which makes sense and is inline with desired naming. I have no idea what would happen with miganalysis on the other hand. Does k8s know how to handle plurals that end in 's'? A mystery for the ages. 

My **updated** vote is to keep as-is

pkg/apis/migration/v1alpha1/miganalytic_types.go
```diff
@@ -0,0 +1,110 @@
...
+	Plan                *kapi.ObjectReference `json:"plan"`
+	AnalyzePVCapacity   bool                  `json:"analyzePVCapacity"`
+	AnalyzeImageCount   bool                  `json:"analyzeImageCount"`
+	AnalyzeK8SResources bool                  `json:"analyzeK8SResources"`
+	ListImages          bool                  `json:"listImages,omitempty"`
```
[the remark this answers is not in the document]
jmontleon: seems if omitempty makes it not mandatory. I didn't want to make listimages mandatory...

pkg/apis/migration/v1alpha1/miganalytic_types.go
```diff
@@ -0,0 +1,110 @@
...
+	ExcludedK8SResourceTotal     int                            `json:"excludedK8SResourceTotal"`
+	IncompatibleK8SResourceTotal int                            `json:"incompatibleK8SResourceTotal"`
+	PVCapacity                   resource.Quantity              `json:"pvCapacity"`
+	PVCount                      int                            `json:"pvCount"`
+	ImageCount                   int                            `json:"imageCount"`
+	ImageSizeTotal               resource.Quantity              `json:"imageSizeTotal"`
+	Images                       []MigAnalyticNamespaceImage    `json:"images,omitempty"`
+	K8SResources                 []MigAnalyticNamespaceResource `json:"k8sResources,omitempty"`
+	ExcludedK8SResources         []MigAnalyticNamespaceResource `json:"excludedK8SResources,omitempty"`
+	IncompatibleK8SResources     []MigAnalyticNamespaceResource `json:"incompatibleK8SResources,omitempty"`
```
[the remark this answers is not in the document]
jmontleon: IncompatibleK8SResources holds an array of incompatible resources with count for each (from GVK comparison). It does not recalculate GVK, it's using the list on the plan, which is why we requeue if the plan is not ready, so we can ensure the list is there.

ExcludedK8sResources holds an array of resources and their count from the excludedresources list.

pkg/controller/miganalytic/miganalytics_controller.go
jmontleon: Agree with the constant. I missed that while cleaning up. It's also not the '50' I put in the docs. If you enable listImages and there isn't a default value set like this the default will be 0. And how do you differentiate a user set value of 0 and a default of 0. Reading about how to set non-zero defaults in Go this is the pattern that was described.

pkg/apis/migration/v1alpha1/miganalytic_types.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// MigAnalyticNamespaceResource defines the observed state of MigAnalyticNamespaceResource
```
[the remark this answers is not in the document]
eriknelson: For some reason this was hidden in the github UI and I had to go searching for my ping, was this hidden because it resolved, or there's still an open question?

Can you give an example of the format you're asking about?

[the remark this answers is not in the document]
jmontleon: I don't think it's resolved. It was hidden for me. I just saw you left a comment and was trying to find it...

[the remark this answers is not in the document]
eriknelson: My 2 cents, +1 for consistency for sure. Would `groupKind` or `groupVersion` try to consolidate the two into a single field? Those are confusing to me without knowing the background around these suggestions. Also, what are the pros for the 2nd two over the first?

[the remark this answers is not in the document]
jmontleon: I think the primary benefit for those suggestions was to condense the number of lines of output.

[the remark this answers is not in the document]
eriknelson: > the number of lines of output

Is this really an optimization people care about?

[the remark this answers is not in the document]
ibolton336: > Seems like there were some differing opinions on the format this should be presented in, I think this format is OK because it gives the user all of the info and the UI can mutate this info however it wants present it to the user.
> 
> @ibolton336 @eriknelson can you guys use the info in this format OK?

Is this replacing the incompatibleNamespaces list that is  currently pulled from the MigPlan.status.incompatibleNamespaces array?

pkg/controller/miganalytic/miganalytics_controller.go
jmontleon: Moved to a constant. Still want to set a non-zero default. This is the only way I know how unless someone has a better idea.

pkg/controller/miganalytic/miganalytics_controller.go
jmontleon: ya, it grew a bit more than I expected when I added a the list of incompatible and excluded resources.

It's basically comparing GVKs against the two lists that will cause them to be exempted from the migration and placing them in the right list:
```
  excludedResources:
  - imagetags
  - templateinstances
  - clusterserviceversions
  - packagemanifests
  - subscriptions
  - servicebrokers
  - servicebindings
  - serviceclasses
  - serviceinstances
  - serviceplans
  incompatibleNamespaces:
  - gvks:
    - group: apps
      kind: deployments
      version: v1beta1
    - group: extensions
      kind: replicasets
      version: v1beta1
```

resulting in something like:
```
    - excludedK8SResourceTotal: 355
      excludedK8SResources:
      - count: 355
        group: packages.operators.coreos.com
        kind: packagemanifests
        version: v1
...
      k8sResources:
      - count: 2
        kind: endpoints
        version: v1
      - count: 2
        kind: persistentvolumeclaims
        version: v1
...
```

The incompatibleResourcess is empty/omitted in the example above. I don't have a cluster stood up at the moment with any incompatible GVKs, but they would appare similarly

jmontleon: pushed some of it out to separate functions, added comments, hopefully it's more readable now.

pkg/controller/miganalytic/miganalytics_controller.go
jmontleon: This isn't working right if the value is absent from the CR completely. Removed it.

pkg/apis/migration/v1alpha1/miganalytic_types.go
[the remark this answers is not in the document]
jmontleon: The reason for leaving it optional is I mostly view it as a way of troubleshooting if a user is for example having difficulty finding/understanding which image(s) are causing an unexpectedly large result.

jmontleon:
@djwhatle I think I touched on most/all of the issues you were concerned with. Please let me know if you see any incorrectly or insufficiently adjusted.

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -159,6 +159,8 @@ func (m *MigCluster) BuildRestConfig(c k8sclient.Client) (*rest.Config, error) {
 		Host:            m.Spec.URL,
 		BearerToken:     string(secret.Data[SaToken]),
 		TLSClientConfig: tlsClientConfig,
+		Burst:           1000,
```
djwhatle, line 162: Where did these burst/QPS values come from? Did we do some testing?

pkg/apis/migration/v1alpha1/miganalytic_types.go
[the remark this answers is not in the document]
djwhatle: +1

pkg/apis/migration/v1alpha1/migcluster_types.go
djwhatle: What is the observed effect of these new values vs. defaults?

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -0,0 +1,586 @@
...
+	log.Info("Beginning analysis of plan.", "migplan", plan.Name)
+
+	for i, namespace := range plan.Spec.Namespaces {
+		log.Info("Beginning analysis of namespace.", "namespace", namespace)
+		ns := migapi.MigAnalyticNamespace{
+			Namespace: namespace,
+		}
+
+		if analytic.Spec.AnalyzeK8SResources {
+			err := r.analyzeK8SResources(dynamic, resources, &ns, plan.Status.ExcludedResources, plan.Status.Incompatible)
```
djwhatle, line 224: @jmontleon request we keep line length at <100 to match controller conventions. May need to do a bit more breaking up of methods.

pkg/apis/migration/v1alpha1/migcluster_types.go
jmontleon: @jortel do you know where these values came from? I copied them from the non-admin branched after you pointed me to them.

As for the observed effect, it causes the queries to take a fraction of the time they do without due to elimination of rate limiting.

jmontleon: For reference: https://github.com/konveyor/mig-controller/blob/non-admin/pkg/apis/migration/v1alpha1/migcluster_types.go#L174

djwhatle:
@jmontleon still a lot of places in MigAnalytics controller where line length > 100.

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -0,0 +1,589 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func (r *ReconcileMigAnalytic) analyzeImages(client k8sclient.Client, namespace *migapi.MigAnalyticNamespace, listImages bool, listImagesLimit int) error {
+	imageStreamList := imagev1.ImageStreamList{}
+	internalRegistry, err := GetRegistryInfo(4, 4, client)
```
djwhatle, line 339: Why is `(4, 4, client)` being passed in here? It seems like `GetRegistryInfo()` is parsing a k8s major/minor version but this looks like an OpenShift version. Why are we hard-coding 4,4?

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -0,0 +1,589 @@
...
+	dynamic, err := dynamic.NewForConfig(client.RestConfig())
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	analytic.Status.Analytics.Plan = plan.Name
+	log.Info("Beginning analysis of plan.", "migplan", plan.Name)
+
+	for i, namespace := range plan.Spec.Namespaces {
+		log.Info("Beginning analysis of namespace.", "namespace", namespace)
```
djwhatle, line 218: @jmontleon I recommend we remove this log statement. We convey all needed info to users (except for unhandled errs) via `Status.Conditions`  in mig-controller.

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -0,0 +1,589 @@
...
+		analytic.Status.Analytics.ImageSizeTotal.Add(ns.ImageSizeTotal)
+		analytic.Status.Analytics.PVCapacity.Add(ns.PVCapacity)
+		analytic.Status.Analytics.PVCount += ns.PVCount
+		analytic.Status.Analytics.PercentComplete = (i + 1) * 100 / len(plan.Spec.Namespaces)
+
+		err = r.Update(context.TODO(), analytic)
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		log.Info("Completed analysis of namespace.", "namespace", namespace)
```
djwhatle, line 259: Recommend we remove these "Completed" messages as well. If we want to tell a user which namespaces were completed, I would advise including them as a `Condition` on the `Status.Conditions` array.

djwhatle: This is in an effort to keep the mig-controller logs uncluttered, except for when critical errors occur.

pkg/controller/miganalytic/miganalytics_controller.go
jmontleon: It shouldn't be hard coded. I need to fix this. The code was copied from openshift-velero-plugin. I think I was missing a GetServerVersion function and hard coded it. I honestly don't even remember it. I'll fix it.

https://github.com/konveyor/openshift-velero-plugin/blob/76906dcc99cf3f5007c7d074461b822b86a580e5/velero-plugins/common/shared.go#L88-L122

jmontleon: Should be fixed. Tested quickly and seems to still be working as expected.

pkg/apis/migration/v1alpha1/migcluster_types.go
jmontleon:
```

Comment (jmontleon) [reply pkg/apis/migration/v1alpha1/migcluster_types.go:162]:
```
The number of client requests or API calls that are sent to the API server is determined by the Queries per second (QPS) value and the number of concurrent requests that can be processed by the API server is determined by the maxRequestsInFlight setting. The number of requests the client can make in excess of the QPS rate depends on the burst value, this is helpful for applications that are bursty in nature and can make irregular number of requests. 
```
https://docs.openshift.com/container-platform/3.11/scaling_performance/host_practices.html

PR #607, opened 2020-07-29
Add round trip time metric to compat client

### LINKED ISSUE #606: Add request round trip time counter to metrics
Expose a counter metric summing total RTT for each type of request issued through compat client to get a better idea of which of our API requests are taking the most time. Right now we have request counts, but this may not be an effective check for which requests are actually slow.

This would be a second metric in addition to the count metric.

Follow-on to https://github.com/konveyor/mig-controller/pull/591

djwhatle:
**Summary**

- Renames `mtc_client` to `mtc_client_count`
- Introduces `mtc_client_rtt`

Provides cumulative milliseconds spent waiting for each type of request to return from APIserver.

Closes #606

### DIFF
--- a/pkg/compat/metrics.go
+++ b/pkg/compat/metrics.go
@@ -77,32 +77,32 @@ type Reporter struct {
 
 //
 // Report `get` API call.
-func (m *Reporter) Get(client Client, object api.Object, count float64) {
-	m.report(client, Get, object, count)
+func (m *Reporter) Get(client Client, object api.Object, increment float64) {
+	m.report(client, Get, object, increment)
 }
 
 //
 // Report `list` API call.
-func (m *Reporter) List(client Client, object api.Object, count float64) {
-	m.report(client, List, object, count)
+func (m *Reporter) List(client Client, object api.Object, increment float64) {
+	m.report(client, List, object, increment)
 }
 
 //
 // Report `create` API call.
-func (m *Reporter) Create(client Client, object api.Object, count float64) {
-	m.report(client, Create, object, count)
+func (m *Reporter) Create(client Client, object api.Object, increment float64) {
+	m.report(client, Create, object, increment)
 }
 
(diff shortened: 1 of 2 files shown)


djwhatle:
@jortel any further thoughts on this? OK with merging?

jortel:
One question though - Why 2 separate Reporters instead of 1 Reporter w/ 2 counters and add the _elapsed_ to each Method?

djwhatle:
@jortel a single prometheus metric can only have one counter, not sure if that answers your question

pkg/compat/metrics.go
```diff
@@ -25,15 +26,38 @@ const (
...
+// Global reporters.
+var RequestCountMetrics *Reporter
+var RequestRTTMetrics *Reporter
 
 func init() {
-	Metrics = &Reporter{
+	RequestCountMetrics = &Reporter{
+		counter: promauto.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "mtc_client_count",
```
djwhatle, line 44: I think I need to make this name more descriptive. The consumer of this data will not have our code for context.

pkg/compat/metrics.go
```diff
@@ -25,15 +26,38 @@ const (
...
+				Method,
+			}),
+	}
+
+	RequestRTTMetrics = &Reporter{
 		counter: promauto.NewCounterVec(
 			prometheus.CounterOpts{
-				Name: "mtc_client",
-				Help: "MTC client API metrics.",
+				Name: "mtc_client_rtt",
```
djwhatle, line 59: similarly, I think we need to be more descriptive with what this metric conveys to match convention in openshift metric naming, `mtc_client_request_rtt`

djwhatle:
@jortel I combined the reporters into one as you suggested, I think it made for a good improvement.

PR #618, opened 2020-08-04
Remove "EnsureLabelsDeleted" phase to speed up migration process

### LINKED ISSUE #608: Optimize  annotations.go `deleteLabels()`
**Summary**

- `deleteLabels()` iterates every possible GVR for in every namespace in migration and runs a discovery client `list()` followed by discovery client `update()` on all `list()` results.
- `deleteLabels()` will (presumably) start completely over in the next reconcile if it encounters a conflict error. If there are enough resources, the possibility of one resource having a conflictError is high, meaning this _could stall forever._
- `deleteLabels()` (aka `EnsureLabelsDeleted`) has been observed taking hours to complete on complex migrations. 

[Link to relevant code](https://github.com/konveyor/mig-controller/blob/master/pkg/controller/migmigration/annotation.go#L450)

**Solutions to consider**

- Do we need to remove labels on every possible GVK? Can we just run removal over a subset of GVKs that we actually label?
- Can we add label retries for `conflictErrors`?  This may allow us to get through this operation on the first try
- Can we drop this function altogether? Does it matter if we leave labels around?

```
// deleteLabels will delete all migration.openshift.io/migrated-by labels
// from the application upon successful completion
func (t *Task) deleteLabels() error {
	client, GVRs, err := gvk.GetGVRsForCluster(t.PlanResources.DestMigCluster, t.Client)
	if err != nil {
		return liberr.Wrap(err)
	}

	listOptions := k8sclient.MatchingLabels(map[string]string{
		MigratedByLabel: string(t.Owner.UID),
	}).AsListOptions()

	for _, gvr := range GVRs {
		for _, ns := range t.destinationNamespaces() {
			list, err := client.Resource(gvr).Namespace(ns).List(*listOptions)
			if err != nil {
				return liberr.Wrap(err)
			}
			for _, r := range list.Items {
				labels := r.GetLabels()
				delete(labels, MigratedByLabel)
				r.SetLabels(labels)
				_, err = client.Resource(gvr).Namespace(ns).Update(&r, metav1.UpdateOptions{})
				if err != nil {
					// The part touches the application on the destination side, fail safe to ensure
					// this won't block the migration
					if k8serror.IsMethodNotSupported(err) {
						continue
					}
					return liberr.Wrap(err)
				}
			}
		}
	}

	return nil
}
```

djwhatle:
Closes #608 

Removing `deleteLabels()` and leaving this `migrated-by` label in place will:

- Speed up migration by ~33% according to tracing figures
- Lead to each resource that is restored to have a `migrated-by=<migmigration_uid>` label leftover, which may be useful for debugging purposes.

### DIFF
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -6,11 +6,8 @@ import (
 
 	liberr "github.com/konveyor/controller/pkg/error"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
-	"github.com/konveyor/mig-controller/pkg/gvk"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	k8serror "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -442,44 +439,6 @@ func (t *Task) deleteNamespaceLabels(client k8sclient.Client, namespaceList []st
 			return liberr.Wrap(err)
 		}
 	}
-	return nil
-}
-
-// deleteLabels will delete all migration.openshift.io/migrated-by labels
-// from the application upon successful completion
-func (t *Task) deleteLabels() error {
-	client, GVRs, err := gvk.GetGVRsForCluster(t.PlanResources.DestMigCluster, t.Client)
-	if err != nil {
-		return liberr.Wrap(err)
-	}
-
-	listOptions := k8sclient.MatchingLabels(map[string]string{
-		MigratedByLabel: string(t.Owner.UID),
-	}).AsListOptions()
-
-	for _, gvr := range GVRs {
-		for _, ns := range t.destinationNamespaces() {
-			list, err := client.Resource(gvr).Namespace(ns).List(*listOptions)
-			if err != nil {
-				return liberr.Wrap(err)
-			}
-			for _, r := range list.Items {
-				labels := r.GetLabels()
-				delete(labels, MigratedByLabel)
-				r.SetLabels(labels)
-				_, err = client.Resource(gvr).Namespace(ns).Update(&r, metav1.UpdateOptions{})
-				if err != nil {
-					// The part touches the application on the destination side, fail safe to ensure
-					// this won't block the migration
-					if k8serror.IsMethodNotSupported(err) {
-						continue
-					}
-					return liberr.Wrap(err)
-				}
-			}
-		}
-	}
-
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


jwmatthews:
I am not aware of a need/driver at present that prohibits us from leaving the label.
Change makes sense to me, agree could be an advantage for debugging.

Would like to ensure we test the concern @sseago raised, ensure we can migrate A->B, then again B->C and the controller updates label as needed.

jmontleon:
Can it be a feature that's disabled by default in case someone objects to the labels? In that they can enable it if it bothers them enough to take a 33% performance hit?

sseago:
@jmontleon Yes, I had just suggested this in slack. I don't know whether we want to make the "optional, disabled by default" change now or whether we would just consider re-adding it like this in the future if someone complains about the labels, though. It would be analogous to the feature we added which allows for automatically deleting and cleaning up resources on failure. It's an option, but it's disabled by default because doing this prevents debugging, as well as deleting migrated resources in cases where the failure is a "false positive" -- i.e. something non-fatal triggered the error but the migration was essentially completed.

djwhatle:
> Can it be a feature that's disabled by default in case someone objects to the labels? In that case they can enable it if it bothers them enough to take a 33% performance hit?

I acknowledge that it's possible folks will want this, but given:
 - how slow the deleteLabels function is for real-world migrations
 - compounding slowness on large migrations: if a single update request fails with a `ConflictError` for instance, this routine will **start over entirely** without consideration of what work is already done

I'm leaning towards removal for now and prioritization of other performance work. If we decide we want to make this optional in the future, we would need to:

1. Add a settings switch to turn this feature on/off
2. Document the settings switch and the downsides of enabling this
3. Improve the function to allow better error handling (ConflictError should trigger retries instead of total restart)

Because of the additional work associated with keeping `deleteLabels()` available behind a switch, I strongly feel we should punt this until requested by users and continue to prioritize perf enhancement work in the meantime.

sseago:
@djwhatle That approach seems fine to me. I'm not certain that the "starting over" aspect is as bad as you suggest, since on retry it queries everything again, but only those labeled resources are returned, so any resources that we succeeded in removing labels for won't show up again. But the overall performance concerns are valid, and I don't know that there's a real customer need to offer this option right now.

djwhatle:
>  I'm not certain that the "starting over" aspect is as bad as you suggest, since on retry it queries everything again, but only those labeled resources are returned, so any resources that we succeeded in removing labels for won't show up again.

You're completely right, I overlooked this before. Now that you've pointed this out, I don't think it's possible that this step will ever stall forever, but it will still potentially waste a lot of time due to lack of retries (still have to query every NS -> every GVK)

djwhatle:
\test

djwhatle:
Tested manually and worked. Merging.

See https://github.com/konveyor/mig-controller/issues/619 for future improvement ideas on adding deleteLabels() back if deemed important to users.

PR #625, opened 2020-08-11
Introducing labels to help identify parent resources 

### LINKED ISSUE #610: add a way to co-relate migmigraion CR to a migplan
Couple of exploratory options to achieve the same goal:

1. The controller adds a label
2. Find a way to query via a field selector

### LINKED ISSUE #614: Backup/Restore obj should have queryable labels with plan & migration name for debug
`Backup` and `Restore` objects have the label `migmigration=<uid>` on it.
It was very confusing to me that this value was actually the uid from the
migmigration and not the name of the `MigMigration`, because they are both UUIDs.
I'm open to reopening discussion about what the UI should use for this name,
but I think that conversation remains out of scope for this issue.
I didn't understand why the label did not match until I realized it was the uid field.
I would propose we:

1. make that label explicit about which field it's referencing,

2. add a label that contains the name of the related `MigMigration` so it
can be queried via label selector.

3. add a label that contains the name of the related `MigPlan`

All of this will help the debug experience, see

[the enhancements repo](https://github.com/konveyor/enhancements/pull/2) for the
dedicated enhancement discussion around the debug experience.

pranavgaikwad:
Fixes #610 
Fixes #614 

This PR introduces  new labels `migmigration-name` and `migplan-name`. The value of each of the labels is the name of the associated parent resource. 

* MigPlan 

```yml
apiVersion: migration.openshift.io/v1alpha1
kind: MigPlan
metadata:
  name: migplan
  [...]
spec:
  [...]
```

* MigMigration

```yml
apiVersion: migration.openshift.io/v1alpha1
kind: MigMigration
metadata:
  labels:
    migration.openshift.io/migplan-name: migplan
  name: migmigration
  uid: 47c31a36-3ffc-421f-82ea-97f89256ba61
  [...]
spec:
  migPlanRef:
    name: migplan
  [...]
status:
[...]
```
* Backup

```yml
apiVersion: velero.io/v1
kind: Backup
metadata:
  labels:
    migmigration: 47c31a36-3ffc-421f-82ea-97f89256ba61
    migration.openshift.io/migmigration-name: migmigration
    migration.openshift.io/migplan-name: migplan
  [...]
spec:
  [...]
status:
  [...]
```

* Restore

```yml
apiVersion: velero.io/v1
kind: Restore
metadata:
  labels:
    migmigration: 47c31a36-3ffc-421f-82ea-97f89256ba61
    migration.openshift.io/migmigration-name: migmigration
    migration.openshift.io/migplan-name: migplan
    [...]
  [...]
spec:
  [...]
status:
  [...]
```

### DIFF
--- a/pkg/controller/migmigration/migmigration_controller.go
+++ b/pkg/controller/migmigration/migmigration_controller.go
@@ -328,6 +328,9 @@ func (r *ReconcileMigMigration) setOwnerReference(migration *migapi.MigMigration
 
 // Ensures that the labels required to assist debugging are present on migmigration
 func (r *ReconcileMigMigration) ensureDebugLabels(migration *migapi.MigMigration) error {
+	if migration.Spec.MigPlanRef == nil {
+		return nil
+	}
 	if migration.Labels == nil {
 		migration.Labels = make(map[string]string)
 	} else {
--- a/pkg/controller/migmigration/migmigration_controller.go
+++ b/pkg/controller/migmigration/migmigration_controller.go
@@ -135,6 +135,13 @@ func (r *ReconcileMigMigration) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	// Ensure parent labels are present on migmigration
+	err = r.ensureParentLabels(migration)
+	if err != nil {
+		log.Trace(err)
+		return reconcile.Result{Requeue: true}, err
+	}
+
 	// Set values.
 	log.SetValues("migration", migration.Name)
 
@@ -316,5 +323,23 @@ func (r *ReconcileMigMigration) setOwnerReference(migration *migapi.MigMigration
 			UID:        plan.UID,
 		})
 
+	return nil
+}
+
+// Ensures that the labels required to assist debugging are present on migmigration
+func (r *ReconcileMigMigration) ensureParentLabels(migration *migapi.MigMigration) error {
+	if migration.Labels == nil {
+		migration.Labels = make(map[string]string)
+	}
+	if value, exists := migration.Labels[migapi.ParentLabel]; exists {
+		if value == migration.Spec.MigPlanRef.Name {
+			return nil
+		}
+	}
+	migration.Labels[migapi.ParentLabel] = migration.Spec.MigPlanRef.Name
+	err := r.Update(context.TODO(), migration)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


pranavgaikwad:
\test

eriknelson:
Should we prefix these labels with something like `migrations.openshift.io/migmigration-name`?

djwhatle:
We own the namespace `migration.openshift.io` (there is no 's')

pranavgaikwad:
\test

eriknelson:
@djwhatle I generally abstain from acking controller PRs because I don't feel like I have the source knowledge to be able to do that.

With that said, LGTM.

pkg/controller/migmigration/migmigration_controller.go
```diff
@@ -318,3 +325,22 @@ func (r *ReconcileMigMigration) setOwnerReference(migration *migapi.MigMigration
...
 	return nil
 }
+
+// Ensures that the labels required to assist debugging are present on migmigration
+func (r *ReconcileMigMigration) ensureDebugLabels(migration *migapi.MigMigration) error {
+	if migration.Labels == nil {
+		migration.Labels = make(map[string]string)
+	} else {
+		if value, exists := migration.Labels[MigPlanDebugLabel]; exists {
+			if value == migration.Spec.MigPlanRef.Name {
```
djwhatle, line 335: +1  on only performing an `Update` if necessary.

pkg/controller/migmigration/migmigration_controller.go
```diff
@@ -318,3 +325,22 @@ func (r *ReconcileMigMigration) setOwnerReference(migration *migapi.MigMigration
...
+				return nil
+			}
+		}
+	}
+	migration.Labels[MigPlanDebugLabel] = migration.Spec.MigPlanRef.Name
```
djwhatle, line 340: May want to do a nil check on `MigPlanRef` at the start of this method, as the `migPlanRef` field can be nil due to being a ptr with `omitempty` set.

https://github.com/konveyor/mig-controller/blob/90e9e8d9b983a149b4c2b4cb14ce75fc2c3c5aa0/pkg/apis/migration/v1alpha1/migmigration_types.go#L32

pkg/controller/migmigration/migmigration_controller.go
```diff
@@ -318,3 +325,22 @@ func (r *ReconcileMigMigration) setOwnerReference(migration *migapi.MigMigration
...
+	err := r.Update(context.TODO(), migration)
```
djwhatle, line 341: @jortel can you think of any complications that may arise as a result of running an `Update` on the object we're reconciling before we arrive at validations? Want to make sure this won't throw off anything with observedVersion etc.

PR #628, opened 2020-08-12
Bug 1867597: Add support for pulling stage pod + registry images from "migration-cluster-config" configmap

djwhatle:
**Merge in sync with**
 - mig-operator PR [400](https://github.com/konveyor/mig-operator/pull/400)
 - openshift-migration-plugin PR [69](https://github.com/konveyor/openshift-migration-plugin/pull/69)

**Summary**
 - Adds ability to customize `STAGE_IMAGE` with configmap value
 - Adds ability to customize `REGISTRY_IMAGE` with configmap value
 - Annotates Stage Restore with `migration.openshift.io/stage-pod-image=<STAGE_IMAGE>`

image

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -224,9 +224,7 @@ func (r *MigPlan) ListMigrations(client k8sclient.Client) ([]*MigMigration, erro
 
 // Registry label for controller-created migration registry resources
 const (
-	MigrationRegistryLabel        = "migration-registry"
-	MigrationRegistryDefaultImage = "registry:2"
-	MigrationRegistryImageEnvVar  = "MIGRATION_REGISTRY_IMAGE"
+	MigrationRegistryLabel = "migration-registry"
 )
 
 // Build a credentials Secret as desired for the source cluster.
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
 	"reflect"
 	"regexp"
 	"sort"
@@ -231,13 +230,13 @@ const (
 )
 
 // Return the image reference for the migration registry (defaults to "registry:2")
-func migRegistryImageRef() string {
-	envImage := os.Getenv(MigrationRegistryImageEnvVar)
-	if envImage == "" {
-		return MigrationRegistryDefaultImage
-	}
-	return envImage
-}
+// func migRegistryImageRef() string {
+// 	envImage := os.Getenv(MigrationRegistryImageEnvVar)
+// 	if envImage == "" {
+// 		return MigrationRegistryDefaultImage
+// 	}
+// 	return envImage
+// }
 
 // Build a credentials Secret as desired for the source cluster.
 func (r *MigPlan) BuildRegistrySecret(client k8sclient.Client, storage *MigStorage) (*kapi.Secret, error) {
@@ -293,7 +292,7 @@ func (r *MigPlan) EqualsRegistrySecret(a, b *kapi.Secret) bool {
 }
 
 // Build a Registry ImageStream as desired for the source cluster.
-func (r *MigPlan) BuildRegistryImageStream(name string) (*imagev1.ImageStream, error) {
+func (r *MigPlan) BuildRegistryImageStream(name string, registryImage string) (*imagev1.ImageStream, error) {
 	labels := r.GetCorrelationLabels()
 	labels[MigrationRegistryLabel] = string(r.UID)
 	labels["app"] = name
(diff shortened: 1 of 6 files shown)


djwhatle:
@jmontleon we already have an env var based setting in place for registry image. Do we need anything further or is this sufficient? 

https://github.com/konveyor/mig-controller/blob/master/pkg/apis/migration/v1alpha1/migplan_types.go#L233-L240

pkg/controller/migmigration/stage.go
```diff
@@ -255,7 +293,11 @@ func (t *Task) ensureStagePodsFromTemplates() error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	stagePods := BuildStagePods(t.stagePodLabels(), t.getPVCs(), &podTemplates, resourceLimitMapping)
+	stagePodImage, err := t.getSrcStagePodImage()
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+	stagePods := BuildStagePods(t.stagePodLabels(), t.getPVCs(), &podTemplates, resourceLimitMapping, stagePodImage)
```
djwhatle, line 300: @sseago please verify that I'm using the correct stage pod image here

pkg/controller/migmigration/stage.go
```diff
@@ -340,13 +382,17 @@ func (t *Task) ensureStagePodsFromRunning() error {
...
+		return liberr.Wrap(err)
+	}
 	for _, ns := range t.sourceNamespaces() {
 		podList := corev1.PodList{}
 		err := client.List(context.TODO(), k8sclient.InNamespace(ns), &podList)
 		if err != nil {
 			return liberr.Wrap(err)
 		}
-		stagePods.merge(BuildStagePods(t.stagePodLabels(), t.getPVCs(), &podList.Items, resourceLimitMapping)...)
+		stagePods.merge(BuildStagePods(t.stagePodLabels(), t.getPVCs(), &podList.Items, resourceLimitMapping, stagePodImage)...)
```
djwhatle, line 395: @sseago please verify that I'm using the correct stage pod image here (srcCluster)

pkg/controller/migmigration/restore.go
```diff
@@ -86,6 +86,13 @@ func (t *Task) ensureStageRestore() (*velero.Restore, error) {
 		return nil, liberr.Wrap(err)
 	}
 	newRestore.Labels[StageRestoreLabel] = t.UID()
+
+	stagePodImage, err := t.getDestStagePodImage()
+	if err != nil {
+		return nil, liberr.Wrap(err)
+	}
+	newRestore.Annotations[StageRestorePodImage] = stagePodImage
```
djwhatle, line 94: @sseago please verify that I'm using the correct stage pod image here (dest cluster)

sseago:
@djwhatle I'm not sure how that registry image env var relates to the one in the new configmap, but the main issue is that the env var is configured for the *controller*, and the configmap approach will use the value configured for the cluster in which we're launching it. We had said that for the stage pod image we couldn't guarantee that the value would be the same across all clusters, especially in the disconnected case. If the same holds true for the registry image, then we'll need to pull the value from the configmap (and probably remove the env var).

djwhatle:
> Ignore this comment if this is still a draft and it's not done yet, but the stage pod image is currently passwed into BuildStagePods and ignored. Both `buildStagePodFromPod` and `buildStagePod` still use the hard-coded image string.

Not sure how I missed this, guess my `find` operation didn't work as intended. Will fix.

djwhatle:
> @djwhatle I'm not sure how that registry image env var relates to the one in the new configmap, but the main issue is that the env var is configured for the _controller_, and the configmap approach will use the value configured for the cluster in which we're launching it. We had said that for the stage pod image we couldn't guarantee that the value would be the same across all clusters, especially in the disconnected case. If the same holds true for the registry image, then we'll need to pull the value from the configmap (and probably remove the env var).

I'll try to get this in. It may be less straightforward since we don't have `task` available from where the stage DC is created to my knowledge.

djwhatle:
\test-with-operator 400

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -304,23 +294,23 @@ func (r *MigPlan) BuildRegistryImageStream(name string) (*imagev1.ImageStream, e
...
 }
 
 // Update a Registry ImageStream as desired for the specified cluster.
-func (r *MigPlan) UpdateRegistryImageStream(imagestream *imagev1.ImageStream) error {
+func (r *MigPlan) UpdateRegistryImageStream(imagestream *imagev1.ImageStream, registryImage string) error {
 	imagestream.Spec = imagev1.ImageStreamSpec{
 		LookupPolicy: imagev1.ImageLookupPolicy{Local: false},
 		Tags: []imagev1.TagReference{
 			imagev1.TagReference{
 				Name: "2",
```
djwhatle, line 307: @sseago is it OK for "2" to be hard-coded here? I think this is in reference to `registry:2`.

djwhatle:
@sseago @jmontleon just concluded a test on this. I verified that:

 - stage pod image is set to restic-restore-helper per operator setting
 - registry pod image is set to konveyor/registry:latest per operator setting

We should merge these PRs together next week.

sseago:
When I started testing this in my environment today (along with the operator and plugin PRs), one of my clusters that I wasn't testing with still had my old operator without this configmap. All migplans that referenced this migcluster without the new configmap fail reconcile:
```
  - category: Critical
    lastTransitionTime: "2020-08-18T18:53:55Z"
    message: 'Reconcile failed: [configmaps "migration-cluster-config" not found].
      See controller logs for details.'
    status: "True"
    type: ReconcileFailed
```
I think we need to add validation to the migcluster to check for the presence of the configmap (and possibly the registry and image entries), and put the migcluster into a nonready state when this occurs.

djwhatle:
@sseago as discussed out of band, will consider a future feature where MTC version match is evaluated at beginning of MigPlan reconcile.

Filed new issue https://github.com/konveyor/mig-controller/issues/633

sseago:
@djwhatle Yes. We don't need the ConfigMap validation here if we're going to add a more general version matching validation soon.

PR #630, opened 2020-08-13
Update Tiltfile to keep image layers for 24h

### LINKED ISSUE #629: `make tilt` target is slow due to rebuilding all layers on each run.
**Problem:**
Tilt will prune image layers after each `make tilt` build + k8s apply new image, but this is slow (2+ mins on my machine) due to the large size of builds. 

**Proposed solution:**
We should keep image layers around for 24h to allow faster iteration during dev when integration testing with mig-operator is needed.

djwhatle:
Speeds up `make tilt` to make integration testing with mig-operator and mig-ui quicker.

Closes #629

### DIFF
--- a/tools/tilt/Tiltfile
+++ b/tools/tilt/Tiltfile
@@ -6,6 +6,8 @@ settings.update(read_json(
     default={},
 ))
 
+docker_prune_settings(max_age_mins=1440)
+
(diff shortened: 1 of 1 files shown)


djwhatle:
I thought this made a difference due to prior testing, but in my testing it makes no difference. Closing.

PR #639, opened 2020-08-24
Plan tree view endpiont.

jortel:
Plan tree view endpoint.
- This involves adding models and endpoints for velero resources.
- Added `selfLink` to all (discovery api) REST resources.
- Inventory MigPlan as regular resource instead of like the MigCluster.  This ensure behavior that is more consistent with other resources.

Unrelated:
- Fixed _concurrent access to map_ issue in the `DataSource`.


pkg/controller/discovery/container/container.go
```diff
@@ -64,10 +64,10 @@ func (r *Container) GetDs(cluster *model.Cluster) (*DataSource, bool) {
 //
 // Add a cluster to the container.
 // Build/update a DataSource for the cluster as needed.
-func (r *Container) Add(cluster *migapi.MigCluster) error {
+func (r *Container) Add(cluster *migapi.MigCluster, collections ...Collection) error {
 	build := func() *DataSource {
-		r.mutex.RLock()
-		defer r.mutex.RUnlock()
+		r.mutex.Lock()
```
djwhatle, line 69: Couldn't remember if this was still needed, glad you added it.

PR #645, opened 2020-08-25
Generated files missed in some earlier PR.

jortel:
Simply ran `make manager` on master.


djwhatle:
Just flipping the switch between `darwin` and `linux` in the generated files :) ACK

PR #649, opened 2020-08-26
Use lib itinerary

jortel:
Update migration controller (task) to use _itinerary_ pacakge in controller common.


djwhatle:
Our itinerary requirements have diverged between MTV and MTC since this PR was proposed, so PR in its current form isn't going to be mergeable. I am closing this due to inactivity but not ruling out the possibility of sharing code in the future here if it makes sense.

PR #651, opened 2020-08-27
Remove all remote watches

djwhatle:
_Bound to mig-operator_ https://github.com/konveyor/mig-operator/pull/427
_UI requirements issue_: https://github.com/konveyor/mig-ui/issues/1003

**Summary**
Eliminates remote watches.

**Goals** 
- Allow for many idle MigPlans with no perf penalty (no reconciles triggered due to various remote events)
- Reduce cluster load from APIserver requests issued during spurious reconciles 
- Make CAM more responsive when hooked up to clusters with a lot going on (less in the reconcile queue)

**Changes**
 - Removes remote watches
 - Adds "spec.refresh" `bool` to MigPlan, MigStorage, MigCluster to provide trigger for manual refresh at any time if desired
    - UI and CLI user can use this to refresh on demand as needed
    - MigPlan refresh will also trigger MigStorage and MigCluster refreshes 
 - Adds `Refresh` and `EnsureRefresh` phases to itinerary to ensure all Mig CRs are additionally reconciled "just-in-time" at the beginning of migration
 - Uses timed requeues during active migrations to check up on progress
 - Continues to use local watches on Mig* kinds

**Side-effects**
 - Mig* resources will no longer be reconciling all the time and keeping MigPlans updated, ensuring Registry is running, etc. These things are ensured "just-in-time" for migration, or whenever the user manually triggers updates with `spec.refresh: true`
 - Remote watch memory consumption is eliminated. This is a large portion of our overall consumption.

**UI changes made**
 - https://github.com/konveyor/mig-ui/pull/1014
 - On load, the PV discovery page of MigPlan triggers `migplan.spec.refresh: true` in the background and poll for this to flip back to `migplan.spec.refresh: false`. The UI should show a spinner until this finishes. 

---
image

---

### Benchmarks 

**Test Scenario 1** - rocket-chat migration, no quiesce

|Scenario|Remote Watches?|CAM Memory|Requests|Total Request Latency|
|----|----|----|----|----|
|1 - Simple Migration|Yes (master)|167.66 MiB (100%)|421 (100%)|2.34 sec (100%)|
|1 - Simple Migration|None (this PR)|114.49 MiB (68.2%)|271 (64%) |1.54 sec (65%)|

---

**[MASTER] 1 - Simple Migration - Remote Watches Enabled (167.66 MiB memory, 100%)**
image

**[THIS PR] 1 - Simple Migration - Remote Watches Removed (114.49 MiB memory, 68.2%)**
image

---


pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -64,6 +64,16 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to cloud providers.
+	err = c.Watch(
+		&ProviderSource{
```
djwhatle, line 69: @jortel when this updates every 30 seconds, does it always update the MigStorage? Trying to figure out if this will be triggered reconciles all the time still.

pkg/controller/migplan/migplan_controller.go
```diff
@@ -276,6 +283,81 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	var storage *migapi.MigStorage
+	var err error
+
+	// Abort if items to be refreshed are not ready
+	if plan.Status.HasAnyCondition(
+		SourceClusterNotReady, InvalidSourceClusterRef,
+		DestinationClusterNotReady, InvalidDestinationClusterRef,
+		StorageNotReady, InvalidStorageRef,
+	) {
+		plan.Spec.Refresh = false
```
djwhatle, line 303: In the case that I can't perform the refresh due to bad k8s object refs etc., I'm giving up on it.

djwhatle:
- Change "InProgress" label to condition
 - Make spec.refresh go to false immediately on plan, replaced by "InProgress" condition
 - "InProgress" condition will not be an "Unready" condition, but upon completion of Refresh plan may go unready
 - Make sure registry will be ready in time for first backup

vconzola:
@djwhatle A couple of UX related questions here. (1) How much additional time will this add to PV discovery? It's already pretty slow. Can we get any sort of progress indication on the whole discovery process beyond the indeterminite spinner we're showing right now? (2) Not sure I understand the need for a refresh except right before plan execution, but if it's needed that's fine. My question is, should refresh be done on a one by one basis, or do we want the user to be able to be able to refresh multiple clusters/repos/plans at once?

djwhatle:
@vconzola, cc @ibolton336 (Ian, thought you might want to get looped into this discussion)

---

**1. PV discovery** runs as soon as the MigPlan is created with namespaces. This will not change due to this PR. The PV discovery process will not be slower. Changes applied by the user in general will get acted upon more quickly since there will be less work in the queue in front of the user request. 

This responsiveness improvement will be especially noticeable when multiple plans are present or if there is a lot of cluster activity (e.g. a real-world cluster). This PR is really focused on reducing work the controller is doing when no one asked the controller to do it, aka "spurious reconciles".

---

**2. Manual Refresh** is needed only in these situations: 

a) When the user re-enters the MigPlan wizard PV selection editing step *after* initial creation.

If a user  already completed the MigPlan wizard once and want to come back and make changes, the UI will need to trigger a refresh in the background and let the user wait for results. PVs will be discovered right after initial creation, and again any time the `spec.refresh=true` is set on the MigPlan.

sseago:
Regarding the manual refresh on the PV selection page, the important part is when the user clicks into the PV selection wizard step, the UI needs to set refresh=true and then wait for PV discovery to complete again before displaying the form to the user. The reason for this is that remote cluster changes will no longer automatically refresh the CAM controller's CRs -- specifically for this step, this means that if storage classes are added or removed on the src or target clusters, that won't get reflected on the MigCluster (and thus the MigPlan's PV list) until refresh is triggered on the migplan, but we must have up-to-date PV discovery+storageclasses for this wizard step.

djwhatle:
@jortel ready for another round of review, made the changes we discussed today. I don't think we want to remove suspended right now because MigPlan controller is watching MigMigration, and we don't want PV discovery to run all throughout migration IMO.

pkg/controller/migplan/migplan_controller.go
```diff
@@ -276,6 +283,84 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
...
+			Status:   True,
+			Category: Advisory,
+			Message:  RefreshInProgressMessage,
+			Durable:  true,
+		})
+		plan.Spec.Refresh = false
+	}
+	return nil
+}
+
```
djwhatle, line 363: The bulk of the refresh logic happens here.

vconzola:
@djwhatle @sseago cc: @ibolton336 So, if I'm reading this correctly no display layer changes are needed in the mig plan wizard. Only, display layer change is to add a kebab action to a plan that kicks off a manual refresh. Correct?

djwhatle:
@vconzola @ibolton336  that sounds about right. These are the only changes I know of:

_(this comment contains most up-to-date info given ongoing development of PR, refer to this as source of truth)_

1.  Need to set `migplan.spec.refresh=true` when re-entering the "edit" migplan wizard for an existing plan. Wait on `migplan.spec.refresh=true` and `migplan.status.conditions[RefreshInProgress].IsAbsent()` to indicate PV discovery is done before showing user PV migration options.

2.  Need kebab menus on Plan, Storage, Clusters to include "Refresh" that will kick the appropriate resources `*.spec.refresh` (this option is available on MigCluster, MigStorage, MigPlan)

3.  It _could_ make sense to have the "Stage" and "Migrate" kebab options to perform one final refresh and wait for results before allowing the user to press the real "Create MigMigration" button. This would allow for the user to realize if something in the plan has gone sour before they run the migration. If we don't do this, the first step in the migration `EnsureRefreshed` will catch the issue, but then the user will have to use "cancel" to stop the migration.

djwhatle:
I think that because of checks every ~30s `RemoteClusterSource` in migcluster_controller.go and  `ProviderSource` in migstorage_controller.go, it will very rarely be the case that anything will go unready during the `EnsureRefreshed` phase. I don't think it's worth changing the layout of the migmigration controller to accommodate auto-cancel functionality here.

I think the only thing left to do now is UI changes.

pkg/controller/migstorage/migstorage_controller.go
djwhatle: I figured this out. It only enqueues a reconcile when there's an issue, meaning this shouldn't cause problems.

djwhatle:
UI changes for this PR are being implemented in https://github.com/konveyor/mig-ui/pull/1014

djwhatle:
Reminder to self: check on MigAnalytics, ensure still working as expected.

djwhatle:
Rebased on master to take advantage of some fixes.

vconzola:
@ibolton336 @djwhatle I'm catching up on some old GitHub issues/PRs. Has the Refresh action been added to a plan's kebab menu? Does whichever of you is implementing it need a mockup?

djwhatle:
@vconzola the way this was implemented, there were no view-layer changes made. The refresh just happens in the background at various user actions.

vconzola:
@djwhatle Sweet! No UI is usually the best UI.

djwhatle:
@jortel @alaypatel07 I ran cam-e2e-qe against the latest commits in this branch and saw only a few TCs fail. I will investigate the cause of failure on these TCs and try to fix. I think the CRD TC failure has already been fixed in master so will just need a rebase. 

**Failed TCs**
```
EXECUTION RESULT: FAILURE (smoke/ocp-34703-crd.yml)
EXECUTION RESULT: FAILURE (smoke/ocp-28021-istags.yml)
EXECUTION RESULT: FAILURE (smoke/ocp-33378-max-pvs-equalquota.yml)
```

**All TCs**
```
EXECUTION RESULT: OK (smoke/ocp-24797-mongodb.yml)
EXECUTION RESULT: OK (smoke/ocp-24686-project.yml)
EXECUTION RESULT: OK (smoke/ocp-29073-internalis.yml)
EXECUTION RESULT: OK (smoke/ocp-24706-basicvolmig.yml)
EXECUTION RESULT: FAILURE (smoke/ocp-34703-crd.yml)
EXECUTION RESULT: FAILURE (smoke/ocp-28021-istags.yml)
EXECUTION RESULT: OK (smoke/ocp-26160-max-pvs.yml)
EXECUTION RESULT: OK (smoke/ocp-24872-mssql.yml)
EXECUTION RESULT: OK (smoke/ocp-24997-configmap.yml)
EXECUTION RESULT: OK (smoke/ocp-32834-pvc-terminating.yml)
EXECUTION RESULT: OK (smoke/ocp-34821-increase-data.yml)
EXECUTION RESULT: OK (smoke/ocp-25022-cronjob-quiesced.yml)
EXECUTION RESULT: OK (smoke/ocp-31339-pvc-not-connected.yml)
EXECUTION RESULT: OK (smoke/ocp-31309-quotanoattach.yml)
EXECUTION RESULT: OK (smoke/ocp-27335-velero-version.yml)
EXECUTION RESULT: OK (smoke/ocp-25021-cronjob.yml)
EXECUTION RESULT: OK (smoke/ocp-24659-mysql.yml)
EXECUTION RESULT: OK (smoke/ocp-25212-initcont.yml)
EXECUTION RESULT: OK (smoke/ocp-29871-resquota.yml)
EXECUTION RESULT: OK (smoke/ocp-24871-datagrid.yml)
EXECUTION RESULT: OK (smoke/ocp-33393-bz1845569.yml)
EXECUTION RESULT: OK (smoke/ocp-24769-cakephp.yml)
EXECUTION RESULT: OK (smoke/ocp-33379-maxns-equalquota.yml)
EXECUTION RESULT: OK (smoke/ocp-25986-maxpods.yml)
EXECUTION RESULT: OK (smoke/ocp-31358-restartrestic-newer-than-310.yml)
EXECUTION RESULT: OK (smoke/ocp-32189-cronjob-with-pvc.yml)
EXECUTION RESULT: FAILURE (smoke/ocp-33378-max-pvs-equalquota.yml)
EXECUTION RESULT: OK (smoke/ocp-24995-role.yml)
EXECUTION RESULT: OK (smoke/ocp-26032-maxns.yml)
EXECUTION RESULT: OK (smoke/ocp-24787-redis.yml)
EXECUTION RESULT: OK (smoke/ocp-34824-decrease-data.yml)
EXECUTION RESULT: OK (smoke/ocp-25090-jobs.yml)
EXECUTION RESULT: OK (smoke/ocp-25000-sets.yml)
EXECUTION RESULT: OK (smoke/ocp-24730-django.yml)
EXECUTION RESULT: OK (smoke/ocp-33375-maxpods-equalquota.yml)
EXECUTION RESULT: OK (smoke/ocp-28967-migplan-naming.yml)
EXECUTION RESULT: OK (smoke/ocp-32077-empty-pvc.yml)
```

Thanks to Sergio for his help getting the test suite running against my cluster, it's pretty easy to do once you have the right set of commands! I'll plan to document the steps during this sprint.

pkg/controller/migmigration/task.go
```diff
@@ -95,6 +97,8 @@ var StageItinerary = Itinerary{
 	Steps: []Step{
 		{phase: Created},
 		{phase: Started},
+		{phase: Refresh},
+		{phase: EnsureRefreshed},
```
alaypatel07, line 101: while it is good to ensure refresh at the start of the Itinerary, would it probably make more sense to refresh before every task is run? 

It's just the refreshing here that does not have much say if a lot of time is spent between this phase and the velero backup and restore phases.

alaypatel07 (approved):
Added a few nits as questions, we have already discussed them in our sync up, but I had the comments in pending state since before that. feel free to resolve them as you see fit. Since this has see aggressive testing, all of the questions here are not blockers, if they are good ideas I would expect them to be submitted as follow up PRs. With that LGTM. Thanks for sanity testing this with QE test cases Derek :)

config/crds/migration_v1alpha1_migstorage.yaml
```diff
@@ -72,6 +72,8 @@ spec:
               type: object
             backupStorageProvider:
               type: string
+            refresh:
+              type: boolean
```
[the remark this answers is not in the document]
djwhatle: @sseago do you recall if there's a reason why we need MigStorage to be refreshed? I thought this might have been put in after a conversation we had.

One benefit of updating the migstorage would be that migstorage validations will be re-run.

In reality though, I'm not sure if it's necessary to refresh the migstorage because migstorage has a background check for storage connectivity that should catch problems.

```
	// Watch for changes to cloud providers.
	err = c.Watch(
		&ProviderSource{
			Client:   mgr.GetClient(),
			Interval: time.Second * 30},
		&handler.EnqueueRequestForObject{})
	if err != nil {
		return err
	}
```

pkg/controller/migmigration/task.go
djwhatle: Interesting thought. The effect of that would be that every MigMigration reconcile that runs a migration phase would also enqueue reconciles on the MigPlan, MigClusters, and MigStorage. 

I think this could actually be useful, although I hesitate to say we should **wait** on the refresh to finish successfully before continuing like we do at the beginning. I'll talk it over with @jortel as well and get his thoughts.

config/crds/migration_v1alpha1_migstorage.yaml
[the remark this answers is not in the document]
alaypatel07: @djwhatle 

> a MigStorage is responsible for local+remote BSLs and VSLs.

I may be missing something but if MigStorage creates remote BSLs/VSLs how does it get the cluster info where it needs to connect to? I need to dig on this one little more, will get back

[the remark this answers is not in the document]
djwhatle: @alaypatel07 I was wrong, MigStorage does not handle this BSLs and VSLs. So I think the main argument in favor of updating the MigStorage would simply be to re-run the validations. I think we could remove this.

pkg/controller/migmigration/task.go
djwhatle: I agree that there will be some time gap between starting the migration and the initial backup, but I am aware of any common scenarios where we really need MigCluster and MigStorage reconcile to run again afterwards. The BSLs and VSLs set up should still be around and it seems unlikely they became invalid if they were OK a few minutes ago. The MigCluster connection is being monitored by a ProviderSource watch mechanism on an ongoing basis. 

```
	// Watch for changes to cloud providers.
	err = c.Watch(
		&ProviderSource{
			Client:   mgr.GetClient(),
			Interval: time.Second * 30},
		&handler.EnqueueRequestForObject{})
```

From this discussion I'm realizing a flaw in this PR as it stands today: a refresh on one MigPlan will eventually trigger reconciles on all other MigPlans connected to the same MigStorage or MigClusters.  I'm trying to think of ways to overcome this.

config/crds/migration_v1alpha1_migcluster.yaml
```diff
@@ -47,6 +47,8 @@ spec:
               type: boolean
             isHostCluster:
               type: boolean
+            refresh:
+              type: boolean
```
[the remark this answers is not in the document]
djwhatle: Ah, you bring up a good point: I may need to remove the condition then re-add it for an accurate time readout with a condition.

pkg/controller/migmigration/task.go
djwhatle: Idea that came to me on this for folks to shoot holes in: what if we _only_ run heavy MigPlan processes (PV discovery, GVK diff) when a refresh is  requested? Otherwise we do a lightweight reconcile that just runs validations quickly.

djwhatle: Idea from Jeff: since MigPlan already watches MigMigrations, we can modify the MigPlan watch predicate for MigMigrations to allow reconcile events to come thorough every time the phase changes on MigMigration.

https://github.com/konveyor/mig-controller/blob/c9218dd717b295f8f2e817ed466d0576924606d1/pkg/controller/migplan/predicate.go#L164-L190

djwhatle: Probably want to stop skipping storage validation while suspended https://github.com/konveyor/mig-controller/blob/6c28e299c8086ed7754ab6c2d8a760aecefd93f7/pkg/controller/migplan/storage.go#L21-L24

djwhatle: @alaypatel07 now thinking we will pull the StorageClass info out of MigCluster and into MigPlan PV discovery routine so that we don't have to refresh MigCluster and can avoid having as much refresh code.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -64,15 +64,18 @@ type MigPlanSpec struct {
...
 type MigPlanStatus struct {
 	UnhealthyResources
 	Conditions
-	Incompatible      `json:",inline"`
-	ObservedDigest    string   `json:"observedDigest,omitempty"`
-	ExcludedResources []string `json:"excludedResources,omitempty"`
+	Incompatible       `json:",inline"`
+	ObservedDigest     string         `json:"observedDigest,omitempty"`
+	ExcludedResources  []string       `json:"excludedResources,omitempty"`
+	SrcStorageClasses  []StorageClass `json:"srcStorageClasses,omitempty"`
```
djwhatle, line 77: There is enough code in mig-ui that relies on data meeting the interface defined by `StorageClass[]` and pulled off a kube resource that I think it's too much work to convert us over to using the discovery service for StorageClasses at this time. 

I will make a separate PR adding StorageClasses to discovery service so that this can be fixed with less effort in the future if prioritized.

djwhatle:
Since I made substantial changes to the PR, I re-ran tests. The results appear ~the same as last time, but the 3 failing TCs are different this time. I'll re-run the mongoDB and mssql tests individually to make sure that this is OK.

```
EXECUTION RESULT: FAILURE (smoke/ocp-24797-mongodb.yml)
EXECUTION RESULT: FAILURE (smoke/ocp-28021-istags.yml)
EXECUTION RESULT: FAILURE (smoke/ocp-24872-mssql.yml) 
```

PR #659, opened 2020-09-08
Registering pprof http handler

### LINKED ISSUE #652: Expose pprof memory profiling data for consumption by must-gather
**Summary**:
There are cases where we would like to analyze memory profiling data as measured by pprof on user clusters. It is possible to set up a pprof endpoint that exposes this data.

We should make this setting controllable from mig-operator and allow for gathering with must-gather as shown in the diagram below. Conprof may be an interesting future area of exploration.

It would be nice if must-gather did the work of generating a PDF of the memory consumption chart for us.

This would need changes in mig-controller, mig-operator, and must-gather.

image

pranavgaikwad:
This PR imports pprof to register its default memory profile http handler as a side-effect.

The memory profile will be automatically available at `localhost:2112/debug/pprof/heap` endpoint

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"net/http"
+	_ "net/http/pprof"
 	"os"
 
 	"github.com/konveyor/mig-controller/pkg/apis"


djwhatle (approved):
Approved. I confirmed this doesn't trample our metrics endpoint at `:2112/metrics`.

cmd/manager/main.go
```diff
@@ -18,6 +18,7 @@ package main
 
 import (
 	"net/http"
+	_ "net/http/pprof"
```
djwhatle, line 21: Is there a reason why we are aliasing this import as `_` ?

PR #665, opened 2020-09-15
Filter velero pod list on currently-running (non-terminating) pods

sseago:
We need to filter out non-running pods since this pod list is used to exec into, and if the pod isn't running, the exec will always fail. I hit this in my dev environment where I had a node in a bad state, so the velero pod on that node was stuck in terminating, but the replacement velero pod was up and running witthout errors. Migrations should have succeeded, but were failing because of the exec-into-dead-pod failures.

### DIFF
--- a/pkg/controller/migmigration/pod.go
+++ b/pkg/controller/migmigration/pod.go
@@ -150,23 +150,7 @@ func (t *Task) findVeleroPods(cluster *migapi.MigCluster) ([]corev1.Pod, error) 
 	if err != nil {
 		return nil, liberr.Wrap(err)
 	}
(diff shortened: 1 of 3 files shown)


djwhatle:
@sseago these changes look sane, do you mind connecting an associated GH issue or Jira, or just linking to an existing one? Would like to attach some context on what this change is needed for.

djwhatle:
@sseago should we merge this? Just want to double check what problem this is fixing so we have record of it.

djwhatle (approved):

djwhatle:
\test

djwhatle:
@sseago feel free to merge this .

PR #675, opened 2020-09-16
[BUG 1879424]: update stage pod running check and add destination stage pod check

alaypatel07:
This PR
- updates if the stage pod is healthy by first checking if
  all containers are ready. The pod can go into Running phase but one
  or more containers can be unready, which is now eliminated

- adds a check on destination stage pods, if during restore,
  stage pods are not ready, Restic will fail after a timeout.
  This eliminates the way by failing faster

### DIFF
--- a/pkg/controller/migmigration/task.go
+++ b/pkg/controller/migmigration/task.go
@@ -45,7 +45,8 @@ const (
 	EnsureStagePodsFromTemplates    = "EnsureStagePodsFromTemplates"
 	EnsureStagePodsFromOrphanedPVCs = "EnsureStagePodsFromOrphanedPVCs"
 	StagePodsCreated                = "StagePodsCreated"
-	StagePodsFailed                 = "StagePodsFailed"
+	SourceStagePodsFailed           = "SourceStagePodsFailed"
+	DestinationStagePodsFailed      = "DestinationStagePodsFailed"
 	RestartRestic                   = "RestartRestic"
 	ResticRestarted                 = "ResticRestarted"
 	QuiesceApplications             = "QuiesceApplications"
@@ -365,12 +366,12 @@ func (t *Task) Run() error {
 			return liberr.Wrap(err)
 		}
 	case StagePodsCreated:
-		report, err := t.ensureStagePodsStarted()
+		report, err := t.ensureSourceStagePodsStarted()
 		if err != nil {
 			return liberr.Wrap(err)
 		}
 		if report.failed {
-			t.fail(StagePodsFailed, report.reasons)
+			t.fail(SourceStagePodsFailed, report.reasons)
 			break
 		}
 		if report.started {
@@ -520,6 +521,16 @@ func (t *Task) Run() error {
 		if restore == nil {
 			return errors.New("Restore not found")
 		}
+		report, err := t.ensureDestinationStagePodsStarted()
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		if report.failed {
+			t.fail(DestinationStagePodsFailed, report.reasons)
+			t.Requeue = NoReQ
+			break
+		}
+		t.Log.Info("all destination stage pods are ready")
 		completed, reasons := t.hasRestoreCompleted(restore)
 		if completed {
 			t.setResticConditions(restore)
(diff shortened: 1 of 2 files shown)


alaypatel07:
I haven't tested this yet, will do it tonight

alaypatel07:
\test all

pranavgaikwad:
@alaypatel07 I don't think mig-controller creates stage pods on destination at all. They are part of Velero backup itself, the stage pods get recreated on destination from the Restore

sseago:
@pranavgaikwad The controller doesn't directly create destination stage pods, but failing stage pods can still stall or fail a migration. Since there's not a separate "stage pod creation" step, though -- they're created *during* the StageRestoreCreated phase, which is fine as long as we're certain that the conditions for which we fail on stage pods aren't triggered during a transient not-yet-ready stage of the migration. That would require some testing to verify. For example, is it possible that the PVCs are pending for a small slice of time before the pod is running? If so, could that trigger this failure? There is the potential for a race condition that would fail migrations in hard-to-reproduce ways if we're not careful with this.

sseago (approved):

pkg/controller/migmigration/task.go
```diff
@@ -520,6 +521,21 @@ func (t *Task) Run() error {
...
 		}
+		report, err := t.ensureDestinationStagePodsStarted()
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		if report.failed {
+			t.fail(DestinationStagePodsFailed, report.reasons)
+			break
+		}
+		if report.started {
```
pranavgaikwad, line 532: @alaypatel07 Usage of `t.next()` is not appropriate after ensure `DestinationStagePodsStarted()` call. It should be only called once per Phase to increment the Phase counter by 1.

pranavgaikwad (changes requested):
I tested these changes, they are looking good, except one ask to remove the `t.next()` call

pkg/controller/migmigration/task.go
```diff
@@ -520,6 +521,18 @@ func (t *Task) Run() error {
...
 		}
+		report, err := t.ensureDestinationStagePodsStarted()
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		if report.failed {
+			t.fail(DestinationStagePodsFailed, report.reasons)
+			break
+		}
+		if !report.started {
```
pranavgaikwad, line 532: @alaypatel07 `!report.started` looks like a typo. I think it would be sane to just remove this check alltogether. Adding a Requeue here doesn't make sense since it will always go forward to `hasRestoreCompleted()` call

pranavgaikwad (changes requested):

pkg/controller/migmigration/task.go
dymurray: Yeah I'm a little confused about this as well. Is this because we already checked if it failed? So `!report.Started` means that its completed?

alaypatel07: @pranavgaikwad yes, since the report.started and report.failed are complementary, I think we can just check the failed and remove this check. However, on the failed, I will requeue so that this gets tried the next time

alaypatel07: `report.failed == !report.started == true` I'll collapse the two conditions and make it into one, and pull out the log entry

alaypatel07: fixed

dymurray (approved):
LGTM

pkg/controller/migmigration/task.go
```diff
@@ -520,6 +521,16 @@ func (t *Task) Run() error {
 		if restore == nil {
 			return errors.New("Restore not found")
 		}
+		report, err := t.ensureDestinationStagePodsStarted()
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		if report.failed {
+			t.fail(DestinationStagePodsFailed, report.reasons)
```
alaypatel07, line 529: think if this need to be re-worked and not fail here but put a warning

pranavgaikwad (approved):
LGTM

djwhatle (approved):

Issue #663, opened 2020-09-15
Allow user to control Rollback of a migration at will or on failure


djwhatle:
**Jira**: https://issues.redhat.com/browse/MIG-249

As a User of MTC, I want the ability to rollback a migration, either because a partial failure has been detected or because I want to consider this migration a 'test'.

_As part of a rollback:_

 - Remove the migrated pieces on the destination cluster
 - Unmount shared remote volumes on destination
 - Scale the applications running source cluster back to their original replica counts
 - Remount shared remote volumes

djwhatle:
**Relevant mig-controller files:**
 - [task.go](https://github.com/konveyor/mig-controller/blob/master/pkg/controller/migmigration/task.go) - the core of the migration state machine. 
 - 

**Relevant past PRs:**
 - [Bug 1844638: Add settings switch controlling whether FailedItinerary (auto-rollback) gets used](https://github.com/konveyor/mig-controller/pull/560)

**Original Implementation PRs related to "cancel" and "rollback"**
 - [Added cancelling feature](https://github.com/konveyor/mig-controller/pull/452)
 - [App backscaling on demand](https://github.com/konveyor/mig-controller/pull/458)
 - [Allow deletion of backup and restore after migration is cancelled](https://github.com/konveyor/mig-controller/pull/465)
 - [Ignore cancelled final migration in validation step](https://github.com/konveyor/mig-controller/pull/476)

djwhatle:
Hey @aufi! I was talking with @jortel earlier today about this issue to nail down the required changes .  Here's a summary of requirements we are proposing:

---

-  Add a new `Rollback` itinerary
-  Remove the global "rollback on failure" switch that I implemented in response to BZ 1844638 (see above comment). Instead, add a new `spec.rollback` field to the MigMigration CR
-  Remove (almost) rollback specific steps from `CancelItinerary` and `FailedItinerary`. These steps will be moved to the new `RollbackItinerary`. 

```
var CancelItinerary = Itinerary{
	Name: "Cancel",
	Steps: []Step{
		{phase: Canceling},
		{phase: DeleteBackups},
		{phase: DeleteRestores},
		{phase: EnsureStagePodsDeleted, all: HasStagePods},
		{phase: EnsureAnnotationsDeleted, any: HasPVs | HasISs},
		{phase: Canceled},
		{phase: Completed},
	},
}

var FailedItinerary = Itinerary{
	Name: "Failed",
	Steps: []Step{
[...]

PR #677, opened 2020-09-21
Update migration to support rollback on demand

aufi:
Updating Itineraries and related code to allow rollback control based on MigMigration CR spec.

Fixes https://github.com/konveyor/mig-controller/issues/663

Related to https://github.com/konveyor/mig-operator/pull/473
Related to https://github.com/konveyor/openshift-velero-plugin/pull/58

## QA steps

Assuming working OCP3 (source cluster) and OCP4 (destination cluster) environments with MTC >= 1.4 and some application on the source cluster.

* Create a MigPlan for an application/namespace which should be migrated
* Execute the migration (via UI or creating migmigration CR), wait until it finishes
* Check destination cluster in the choosen namespace for the migrated application objects, it should be present
* Execute migration rollback (via UI when available or creating migmigration CR with spec.rollback: true), wait until it finishes
* Check the app's namespace on the destination cluster, objects should be removed
* Check the app's namespace on the source cluster, objects should be present and app should be working (unquiesced)
* Execute migration rollback again (via UI when available or creating migmigration CR with spec.rollback: true), wait until it finishes
* Check destination cluster in the choosen namespace for the migrated application objects, it should be present and app working

Example usage of the new migration on demand mechanism - https://youtu.be/T1hAY9ObAG0


config/samples/mig-migration.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
 spec:
   # [!] Set 'cancel: true' will cancel the migration
   cancel: false
```
aufi, line 10: @djwhatle Just out of curiosity, shouldn't there be ```canceled```?

pkg/controller/migmigration/validation.go
```diff
@@ -170,7 +170,8 @@ func (r ReconcileMigMigration) validateFinalMigration(plan *migapi.MigPlan, migr
 			}
 		}
 	}
-	if hasCondition {
+	// Allow perform Rollback on finished Plan.
+	if hasCondition && !migration.Spec.Rollback {
```
aufi, line 174: @djwhatle I'm not sure this is the right approach, but in order to allow run Rollback on completed migration, I needed to modify hasFinalMigration validation.

pkg/controller/migmigration/task.go
```diff
@@ -251,7 +261,7 @@ func (t *Task) Run() error {
 
 	// Run the current phase.
 	switch t.Phase {
-	case Created, Started:
+	case Created, Started, Rollback, RollbackComplete: //TODO: what actions needs to be done in Rollback-like steps?
```
aufi, line 264: @djwhatle based on itineraries, I created these two steps (Rollback and RollbackComplete), currently I do see what to put into these steps. Could you give more context on what additional actions need to be executed on the Rollback scenario? Thanks

config/samples/mig-migration.yaml
djwhatle: You're correct :) we need an update to migsamples.

djwhatle:
@aufi I can't remember if the questions you posed above were answered on our last call. Let me know if you still need info.

aufi:
@djwhatle Yes, questions above were answered, thanks. Currently open question is how to migrate https://github.com/konveyor/openshift-migration-plugin/pull/71 to openshift-velero-plugin which has https://github.com/konveyor/openshift-velero-plugin/blob/master/velero-plugins/common/restore.go file but that seems to work only with registry labels.

pkg/controller/migmigration/task.go
```diff
@@ -178,9 +173,22 @@ var FailedItinerary = Itinerary{
...
+	Steps: []Step{
+		{phase: Rollback},
+		{phase: DeleteBackups},
+		{phase: DeleteRestores},
+		{phase: EnsureStagePodsDeleted, all: HasStagePods},
+		{phase: EnsureAnnotationsDeleted, any: HasPVs | HasISs},
 		{phase: DeleteMigrated},
 		{phase: EnsureMigratedDeleted},
 		{phase: UnQuiesceApplications, all: Quiesce},
+		{phase: RollbackComplete},
```
djwhatle, line 191: Unless I'm missing the purpose of the `RollbackComplete` phase, it seems like we could drop this and go straight from `UnQuiesce...` to `Completed`.

pkg/controller/migmigration/validation.go
djwhatle, line 174: +1, I think this makes sense. We need to override this logic that blocks another migration from happening after a completed final migration in order to do rollback.

pkg/settings/migration.go
```diff
@@ -1,19 +0,0 @@
-package settings
```
djwhatle, line 1: +1 on removing this part of settings, no longer needed since rollback will always be on-demand when user requests it.

pkg/controller/migmigration/restore.go
```diff
@@ -271,7 +279,7 @@ func (t *Task) deleteMigrated() error {
 	}
 
 	listOptions := k8sclient.MatchingLabels(map[string]string{
-		MigratedByLabel: string(t.Owner.UID),
+		MigPlanLabel: string(t.PlanResources.MigPlan.UID),
```
djwhatle, line 282: +1 on labeling restored resources with the MigPlan UID so that we can remove everything that's been migrated so far.

djwhatle (changes requested):
Great work on this. I requested one small change to remove a phase that seems to be unneeded. Let me know what you think.

aufi:
Thanks for review, updated, restored objects have both migmigration and migplan references now. Wxample:
```
    migration.openshift.io/migrated-by-migmigration: 4c2cfcf0-408d-4a07-88c2-4b6687a449f9
    migration.openshift.io/migrated-by-migplan: 0a34500f-6745-4f15-be67-14babede937e
```

djwhatle (approved):

pkg/controller/migmigration/task.go
djwhatle: @aufi I still think this should be fixed.

aufi: Sorry I missed/skipped this before, updated!

pkg/controller/migmigration/task.go
```diff
@@ -164,9 +161,6 @@ var CancelItinerary = Itinerary{
 		{phase: DeleteRestores},
 		{phase: EnsureStagePodsDeleted, all: HasStagePods},
```
djwhatle, line 162: @aufi please note that once we fix Stage pod cleanup to run 1) during rollback, 2) at the beginning of migration, we should remove this stage pod removal step from the `CancelItinerary`.

So once you have a solution for https://github.com/konveyor/mig-controller/issues/700 we can strip out this phase here.

pkg/controller/migmigration/task.go
```diff
@@ -164,9 +161,6 @@ var CancelItinerary = Itinerary{
 		{phase: DeleteRestores},
```
djwhatle, line 161: Once we have a solution for cleaning up pending Velero resources at migration start (https://github.com/konveyor/mig-controller/issues/703) we can also strip out the `DeleteBackups` and `DeleteRestores` phases from the `CancelItinerary`.

pkg/controller/migmigration/task.go
```diff
@@ -164,9 +161,6 @@ var CancelItinerary = Itinerary{
...
 		{phase: EnsureStagePodsDeleted, all: HasStagePods},
 		{phase: EnsureAnnotationsDeleted, any: HasPVs | HasISs},
```
djwhatle, line 163: @aufi it looks like rollback is supposed to run `EnsureAnnotationsDeleted` already, so do you think we can remove this? Please check what runs in the phase to see if `EnsureAnnotationsDeleted` will fail during rollback due to having missing context (e.g. if we are using a MigMigration UID instead of a MigPlan UID to lookup resources we want to delete annotations on.

aufi: Annotations are based on an object relation to Namespace where it belongs (no labels, even correlationLabels dependent on MigMigration/MigPlan's UUIDs, are used)

PR #681, opened 2020-09-21
Improve mig registry health check by failing fast

alaypatel07:
The PR does the following things:

- Updates Registry kind from DeploymentConfig to Deployment.
- Removes Imagestream for registry
- Deletes Registry deployment if migstorage is unready.
- Makes sure that registry deployment is not deleted during the course of migration.
- Adds Liveliness and Readiness Probes to the registry deployment for the health check.
- Adds a custom registry health watch with 2 instances in the migplan controller:
- Watch for changes to registry pod health for MigPlans that have running migrations
- Watch for changes to registry pod health for MigPlans that have failed migrations
- Adds a new state Ensure Registry Health in the Reconciliation of Migplan


alaypatel07:
verified the probes with @shubham-pampattiwar's env
block all traffic by
```
# iptables -S | grep 10\.129\.2
-A INPUT -s 10.129.2.89/32 -p tcp -j DROP
-A INPUT -d 10.129.2.89/32 -p tcp -j DROP
-A FORWARD -s 10.129.2.89/32 -p tcp -j DROP
-A FORWARD -d 10.129.2.89/32 -p tcp -j DROP
-A OUTPUT -s 10.129.2.89/32 -p tcp -j DROP
-A OUTPUT -d 10.129.2.89/32 -p tcp -j DROP
```

The registry pod fails

```
  - containerID: cri-o://f5dcb0c27dc5083d89621399a3d1614d44f2e348cff1ef9b08e8d7106ed95342
    image: quay.io/konveyor/registry:release_2.7
    imageID: quay.io/konveyor/registry@sha256:51c04d1f1f19c79072597f16d5b8e83e4728e25d04abe6171688e02059578f9a
    lastState:
      terminated:
        containerID: cri-o://f5dcb0c27dc5083d89621399a3d1614d44f2e348cff1ef9b08e8d7106ed95342
        exitCode: 2
        finishedAt: "2020-09-22T00:19:21Z"
        reason: Error
        startedAt: "2020-09-22T00:18:51Z"
    name: registry
    ready: false
    restartCount: 4
```

Events show readiness and liveness probe failing

```
Events:
  Type     Reason     Age                    From                                   Message
  ----     ------     ----                   ----                                   -------
  Normal   Pulled     3m52s (x3 over 4h47m)  kubelet, ip-10-0-135-100.ec2.internal  Container image "quay.io/konveyor/registry:release_2.7" already present on machine
  Normal   Created    3m52s (x3 over 4h47m)  kubelet, ip-10-0-135-100.ec2.internal  Created container registry
  Normal   Started    3m52s (x3 over 4h47m)  kubelet, ip-10-0-135-100.ec2.internal  Started container registry
  Normal   Killing    3m52s (x2 over 4m22s)  kubelet, ip-10-0-135-100.ec2.internal  Container registry failed liveness probe, will be restarted
  Warning  Unhealthy  3m32s (x8 over 4m42s)  kubelet, ip-10-0-135-100.ec2.internal  Liveness probe failed: Get http://10.129.2.89:5000/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
  Warning  Unhealthy  3m23s (x9 over 4m43s)  kubelet, ip-10-0-135-100.ec2.internal  Readiness probe failed: Get http://10.129.2.89:5000/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -373,6 +382,22 @@ func (r *MigPlan) UpdateRegistryDeployment(storage *MigStorage, deployment *apps
...
+									Path: "/",
+									Port: intstr.IntOrString{IntVal: 5000},
+								},
+							},
+						},
+						ReadinessProbe: &kapi.Probe{
+							Handler: kapi.Handler{
+								HTTPGet: &kapi.HTTPGetAction{
+									Path: "/",
+									Port: intstr.IntOrString{IntVal: 5000},
```
alaypatel07, line 397: s/intstr.IntOrString{IntVal: 5000}/intstr.IntOrString{IntVal: 5001}

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -373,6 +382,22 @@ func (r *MigPlan) UpdateRegistryDeployment(storage *MigStorage, deployment *apps
...
+								},
+							},
+						},
+						ReadinessProbe: &kapi.Probe{
+							Handler: kapi.Handler{
+								HTTPGet: &kapi.HTTPGetAction{
+									Path: "/",
```
alaypatel07, line 396: see this comment https://github.com/konveyor/mig-controller/issues/626#issuecomment-686721794, specifically

```
# Note 2 : using path `:5001/debug/health` would be more 
#         informative to user if it outputs acceptable HTTP codes for probes to interpret.
```

change the path to /debug/health and port according to the following comment

pkg/apis/migration/v1alpha1/migplan_types.go
shubham-pampattiwar: Done

pkg/apis/migration/v1alpha1/migplan_types.go
shubham-pampattiwar: Updated the path

shubham-pampattiwar:
Steps performed to verify the working of probes for registry deployment:

shubham-pampattiwar:
Steps performed to verify the working of probes for registry deployment via `EgressNetworkPolicy`:
1. Deploy CAM, setup replication repo and create a mig-plan
2. Now switch off the controller deployment (as currently the registry get deleted if its unhealthy)
3. Create an `EgressNetworkPolicy` to restrict the outgoing traffic for the registry pod, following is a sample yaml for reference
```
apiVersion: network.openshift.io/v1
kind: EgressNetworkPolicy
metadata:
  name: migration-egress-network-policy
spec:
  egress:
    - type: Allow
      to:
        cidrSelector: 10.128.0.0/14
    - type: Allow
      to:
        cidrSelector: 172.30.0.0/16
    - type: Deny
      to:
        cidrSelector: 0.0.0.0/0
```
4. This `EgressNetworkPolicy` will switch off the communication of pod with the Internet and the thus the registry pod won't be able to communicate with the storage provider (AWS s3 in our case), following are the `curl -i localhost:5001/debug/health` request responses from inside the registry pod (We can see that the response code are getting changes from `200` to `503`)
```
sh-4.4$  while true; do curl -i localhost:5001/debug/health; sleep 2; done
HTTP/1.1 200 OK                                                                                                                                                            
Content-Length: 2                                                                                                                                                            
Content-Type: application/json; charset=utf-8                                                                                                                                
Date: Wed, 23 Sep 2020 17:16:53 GMT 
HTTP/1.1 503 Service Unavailable                                                                                                                                  [66/1910]
Content-Length: 270
Content-Type: application/json; charset=utf-8
Date: Wed, 23 Sep 2020 17:16:57 GMT
{"storagedriver_s3":"s3aws: RequestError: send request failed\ncaused by: Get https://shubbam-6109f5e9711c8c58131acdd2f490f451.s3.amazonaws.com/?max-keys=1\u0026prefix=s3-sh
ubham-registry-343a2a71-cea4-4851-99f7-4df1dc3ccbfa%2F: dial tcp 52.217.84.156:443: i/o timeout"}HTTP/1.1 503 Service Unavailable
Content-Length: 270
Content-Type: application/json; charset=utf-8
Date: Wed, 23 Sep 2020 17:16:59 GMT
{"storagedriver_s3":"s3aws: RequestError: send request failed\ncaused by: Get https://shubbam-6109f5e9711c8c58131acdd2f490f451.s3.amazonaws.com/?max-keys=1\u0026prefix=s3-sh
ubham-registry-343a2a71-cea4-4851-99f7-4df1dc3ccbfa%2F: dial tcp 52.217.84.156:443: i/o timeout"}HTTP/1.1 503 Service Unavailable
Content-Length: 270
Content-Type: application/json; charset=utf-8
Date: Wed, 23 Sep 2020 17:17:01 GMT  
```
5. Finally we can check out the pod events:
```
Events:
  Type     Reason     Age                From                                   Message
  ----     ------     ----               ----                                   -------
  Normal   Scheduled  14m                default-scheduler                      Successfully assigned openshift-migration/registry-518cbdbc-035b-4270-aa29-40bcba21f11e-6kpqz-f9cc7cf4sr7 to ip-10-0-135-100.ec2.internal
  Warning  Unhealthy  17s (x3 over 37s)  kubelet, ip-10-0-135-100.ec2.internal  Readiness probe failed: HTTP probe failed with statuscode: 503
  Normal   Pulled     16s (x2 over 14m)  kubelet, ip-10-0-135-100.ec2.internal  Container image "quay.io/konveyor/registry:release_2.7" already present on machine
  Normal   Created    16s (x2 over 14m)  kubelet, ip-10-0-135-100.ec2.internal  Created container registry
  Normal   Started    16s (x2 over 14m)  kubelet, ip-10-0-135-100.ec2.internal  Started container registry
  Warning  Unhealthy  16s (x3 over 36s)  kubelet, ip-10-0-135-100.ec2.internal  Liveness probe failed: HTTP probe failed with statuscode: 503
  Normal   Killing    16s                kubelet, ip-10-0-135-100.ec2.internal  Container registry failed liveness probe, will be restarted
```
image

alaypatel07:
@shubham-pampattiwar awesome work!

I think you can do this without step 2
> Now switch off the controller deployment (as currently the registry get deleted if its unhealthy)

Try to start the egress policy on the source cluster. With that only the source registry pod will loose access, but the destination mig-controller will have access to the internet. This will not drive the migstorage in unready state.

alaypatel07:
looking at source master-config.yaml I found this
```
networkConfig:
  clusterNetworks:
  - cidr: 10.1.0.0/16
    hostSubnetLength: 9
  externalIPNetworkCIDRs:
  - 0.0.0.0/0
  networkPluginName: redhat/openshift-ovs-subnet
  serviceNetworkCIDR: 172.30.0.0/16
```

which mean if you use the following egress policy you should be able to block internet access for pods in the source namespace 

```
apiVersion: network.openshift.io/v1
kind: EgressNetworkPolicy
metadata:
  name: migration-egress-network-policy
spec:
  egress:
    - type: Allow
      to:
        cidrSelector: 10.1.0.0/16
    - type: Allow
      to:
        cidrSelector: 172.30.0.0/16
    - type: Deny
      to:
        cidrSelector: 0.0.0.0/0
```

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,143 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"time"
+)
+
+
+// A registry health checker which routinely checks whether registries are healthy or not
+type registryHealth struct {
+	hostClient client.Client
+	srcClient  client.Client
```
alaypatel07, line 22: is this needed?

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"context"
+	"github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
+	"github.com/konveyor/mig-controller/pkg/compat"
+	health "github.com/konveyor/mig-controller/pkg/health"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+	event2 "sigs.k8s.io/controller-runtime/pkg/event"
```
alaypatel07, line 12: can probably remove the alias on this import?

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,143 @@
...
+			for _, registryPod := range registryPodsDest.Items{
+				if health.ContainerUnhealthy(registryPod) || health.UnhealthyStatusConditions(registryPod) {
+					r.enqueue(plan)
+					continue
+				}
+			}
+		}
+
+	}
+
```
alaypatel07, line 126: probably should add a `TODO: need see if this go routine should be stopped and returned`

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (r *registryHealth) getRegistryPods(plan v1alpha1.MigPlan, registryClient compat.Client) (corev1.PodList, error){
+
+	registryDeployment, err := plan.GetRegistryDeployment(registryClient)
```
alaypatel07, line 131: +1

alaypatel07: is this deployment used for anything accept listing the pods? Can you reduce this function to just the pod list call?

pkg/health/pods.go
```diff
@@ -116,8 +116,8 @@ func PodsUnhealthy(client k8sclient.Client, options *k8sclient.ListOptions) (*[]
 	unhealthy := []unstructured.Unstructured{}
 	for _, pod := range podList.Items {
 		if pod.Status.Phase != corev1.PodFailed &&
-			!containerUnhealthy(pod) &&
-			!unhealthyStatusConditions(pod) {
+			!ContainerUnhealthy(pod) &&
+			!UnhealthyStatusConditions(pod) {
```
alaypatel07, line 120: while re-using code is good, I expected you to create a new function instead of modifying this one. Also, is the `UnhealthyStatusConditions` required?

In the new implementation, I think the following logic will work too:

1. for a pod in all pods, 
2.    for a container in pod.Container
3.         if !container.Ready, return this pods as unhealthy pods


You can get away with not having the extra logic that `ContainerUnhealthy` has

pkg/health/pods.go
```diff
@@ -134,7 +134,7 @@ func PodsUnhealthy(client k8sclient.Client, options *k8sclient.ListOptions) (*[]
 	return &unhealthy, nil
 }
 
-func unhealthyStatusConditions(pod corev1.Pod) bool {
+func UnhealthyStatusConditions(pod corev1.Pod) bool {
```
alaypatel07, line 137: please drop the change, look above comment for reasons

pkg/health/pods.go
```diff
@@ -148,7 +148,7 @@ func unhealthyStatusConditions(pod corev1.Pod) bool {
 	return false
 }
 
-func containerUnhealthy(pod corev1.Pod) bool {
+func ContainerUnhealthy(pod corev1.Pod) bool {
```
alaypatel07, line 151: same as above

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,143 @@
...
+			return
+		}
+
+		for _, plan := range planList{
+			//TODO avoid race condition check
+
+
+			srcCluster, err := plan.GetSourceCluster(r.hostClient)
+			if err != nil {
+				log.Trace(err)
```
alaypatel07, line 78: what will happen if you hit this error and the go routine returns? Would it create a situation where we are not watching? Would you want to see if the SourceCluster is in ready state before trying to get the Client on Line 88?

alaypatel07 (commented):
Overall this looks in the directions that we want to go, added some questions please take a look

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -416,19 +343,19 @@ func (r *MigPlan) UpdateRegistryDC(storage *MigStorage, deploymentconfig *appsv1
...
-			"deploymentconfig": name,
-		},
-		Strategy: appsv1.DeploymentStrategy{Resources: kapi.ResourceRequirements{}},
-		Template: &kapi.PodTemplateSpec{
+
+	deployment.Spec = appsv1.DeploymentSpec{
+		Replicas: pointer.Int32Ptr(1),
+		Selector: metav1.SetAsLabelSelector(map[string]string{
+			"app":        name,
+			"deployment": name,
```
alaypatel07, line 362: @shubham-pampattiwar you can add "migplan: r.uid" here to get the label on the pods

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -416,19 +343,19 @@ func (r *MigPlan) UpdateRegistryDC(storage *MigStorage, deploymentconfig *appsv1
...
+		}),
+		Template: kapi.PodTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
 				CreationTimestamp: metav1.Time{},
 				Labels: map[string]string{
-					"app":              name,
-					"deploymentconfig": name,
+					"app":        name,
+					"deployment": name,
```
alaypatel07, line 369: add this here as well https://github.com/konveyor/mig-controller/pull/681/commits/782c73e99b23ced4d99347c98ee541a9cfc922c1#r497020064

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,143 @@
...
+
+
+// A registry health checker which routinely checks whether registries are healthy or not
+type registryHealth struct {
+	hostClient client.Client
+	srcClient  client.Client
+	Interval   time.Duration
+	handler    handler.EventHandler
+	queue      workqueue.RateLimitingInterface
+	predicates []predicate.Predicate
```
alaypatel07, line 26: probably worth adding a `podSelector` field that we can pass from the instantiation. we can use this to watch other dependencies like velero and restic

pkg/controller/migplan/registry_health.go
alaypatel07: Also please add a TODO before the return, once an error occurs, we loose the watch for until the time mig-controller pod restarts

pkg/controller/migplan/registry_health.go
shubham-pampattiwar: Removed the `srcClient` attribute

pkg/controller/migplan/registry_health.go
shubham-pampattiwar: Done

pkg/controller/migplan/registry_health.go
shubham-pampattiwar: Added the TODO

pkg/controller/migplan/registry_health.go
shubham-pampattiwar: Added a new label to registry deployment as suggested (`migplan`:`<migplan name>`) and used this label to directly get the registry pods

pkg/health/pods.go
shubham-pampattiwar: Done, updated the PR

pkg/health/pods.go
shubham-pampattiwar: Done

pkg/health/pods.go
shubham-pampattiwar: Done

pkg/controller/migplan/registry_health.go
shubham-pampattiwar: Added a Cluster ready check before moving onto the registry pods, also added a TODO

pkg/apis/migration/v1alpha1/migplan_types.go
shubham-pampattiwar: Done

pkg/apis/migration/v1alpha1/migplan_types.go
shubham-pampattiwar: Done

pkg/controller/migplan/registry_health.go
shubham-pampattiwar: Added a new attribute `podLabels map[string]string`

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -360,13 +361,15 @@ func (r *MigPlan) UpdateRegistryDeployment(storage *MigStorage, deployment *apps
 		Selector: metav1.SetAsLabelSelector(map[string]string{
 			"app":        name,
 			"deployment": name,
+			"miglan": r.Name,
```
alaypatel07, line 364: typo :)

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -360,13 +361,15 @@ func (r *MigPlan) UpdateRegistryDeployment(storage *MigStorage, deployment *apps
...
 		}),
 		Template: kapi.PodTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
 				CreationTimestamp: metav1.Time{},
 				Labels: map[string]string{
 					"app":        name,
 					"deployment": name,
+					"miglan": r.Name,
```
alaypatel07, line 372: s/miglan/migplan

pkg/controller/migplan/registry_health.go
```diff
@@ -84,60 +84,63 @@ func (r *registryHealth) run() {
...
-			if err != nil {
-				log.Trace(err)
-				return
-			}
-			destClient, err := destCluster.GetClient(r.hostClient)
-			if err != nil {
-				log.Trace(err)
-				return
-			}
+			if srcCluster.Status.IsReady() {
```
alaypatel07, line 87: it is a good practice in golang to have the least amount of indentation. 

I would refactor this as:
1. create a function that take migcluster as output and returnes if plan sure be enqueued or not
2. in the function, if cluster is not ready return immediately, if not process it. Code will be readable, least amount of indents and we can unit test this function

pkg/controller/migplan/registry_health.go
```diff
@@ -84,60 +84,63 @@ func (r *registryHealth) run() {
...
 func (r *registryHealth) getRegistryPods(plan v1alpha1.MigPlan, registryClient compat.Client) (corev1.PodList, error){
 
-	registryDeployment, err := plan.GetRegistryDeployment(registryClient)
-	if err != nil {
-		log.Trace(err)
-		return corev1.PodList{}, err
-	}
 	registryPodList := corev1.PodList{}
-	err = registryClient.List(context.TODO(), k8sclient.MatchingLabels(registryDeployment.GetLabels()), &registryPodList)
+	err := registryClient.List(context.TODO(), k8sclient.MatchingLabels(plan.GetLabels()), &registryPodList)
```
alaypatel07, line 140: I wonder if this will work, we need to list the pod based on "name": plan.Name label selectore. The plan.GetLabels() does not return this

might have to do something like `v1.SetAsLabelSelector(map[string]string{"name": plan.Name})`

pkg/apis/migration/v1alpha1/migplan_types.go
shubham-pampattiwar: Done. thank you !

pkg/apis/migration/v1alpha1/migplan_types.go
shubham-pampattiwar: Done !

pkg/controller/migplan/registry_health.go
shubham-pampattiwar: Thanks for the pointers, added a common function `isRegistryPodUnhealthy`

pkg/controller/migplan/registry_health.go
shubham-pampattiwar: Updated the List call

pkg/controller/migplan/registry.go
```diff
@@ -99,6 +101,47 @@ func (r ReconcileMigPlan) ensureMigRegistries(plan *migapi.MigPlan) error {
...
+			nEnsured++
+		}
+	}
+
+	if nEnsured == 2 {
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     RegistriesHealthy,
+			Status:   True,
+			Category: migapi.Required,
+			Message:  RegistriesHealthyMessage,
```
alaypatel07, line 137: I would like to know if the source registry is unhealthy or destination, which pod(name and namespace) is unhealthy. Its not a blocker for now. So if you want, you can either add a todo or address this now. We would definitely need to address it before the end of next sprint

pkg/controller/migplan/registry.go
```diff
@@ -99,6 +101,47 @@ func (r ReconcileMigPlan) ensureMigRegistries(plan *migapi.MigPlan) error {
...
+		})
+	}
+
```
alaypatel07, line 140: you need to set the `RegistriesHealthy` to `False` here

alaypatel07: or remove the condition, depending on the logic

pkg/controller/migplan/registry.go
shubham-pampattiwar: Done, added the context details

pkg/controller/migplan/registry.go
shubham-pampattiwar: Done

pkg/controller/migplan/migplan_controller.go
```diff
@@ -379,6 +387,48 @@ func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
...
+
+	if len(migrations) == 0 {
+		if plan.Labels == nil {
+			plan.Labels = make(map[string]string)
+		}
+		plan.Labels["migplan-running"] = "false"
+		err = r.Update(context.TODO(), plan)
+		if err != nil {
+			return liberr.Wrap(err)
+		}
```
alaypatel07, line 404: `return nil`
here and remove the else block underneath

pkg/controller/migplan/migplan_controller.go
```diff
@@ -379,6 +387,48 @@ func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
...
+	}else {
+		for _, m := range migrations {
+			if m.Status.HasCondition(migctl.Running) {
+				if plan.Labels == nil {
```
alaypatel07, line 408: instead of if else condition here both having the update the, I would rather keep a counter `runningMigrations := 0` and then increment it for all running migration, then after the for loop, make a single update call `if runningMigrations > 0`

pkg/controller/migplan/migplan_controller.go
shubham-pampattiwar: Done

pkg/controller/migplan/migplan_controller.go
shubham-pampattiwar: Update the flow.

pkg/controller/migplan/migplan_controller.go
```diff
@@ -389,43 +391,46 @@ func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
...
-	if len(migrations) == 0 {
+	//Check if there are any associated migrations that are running
+	for _, m := range migrations {
+		if m.Status.HasCondition(migctl.Running) {
+			runningMigrations++
+		}
+	}
+
+	//No migrations associated with the migplan or there are no running migrations for the Migplan
+	if len(migrations) == 0 || runningMigrations == 0 {
```
alaypatel07, line 408: you only need `runningMigrations == 0` check right? why do you care about the number of migrations?

pkg/controller/migplan/migplan_controller.go
```diff
@@ -389,43 +391,46 @@ func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
...
+			if _, found := plan.Labels[MigplanRunning]; found {
+				delete(plan.Labels, MigplanRunning)
+				needsUpdate = true
+			}
+		}
+		return nil
+	}
+
+	//Migplan is associated with Running migrations
+	if runningMigrations > 0 {
```
alaypatel07, line 419: if is not neded here the above if will always return if `runningMigrations == 0`

alaypatel07: well you need it nevermind, you could use a switch case here, 

```
switch {
    case runningMigrations == 0:
          // set your variables
    case runingMigrations > 0:
          // set vars
}
```

pkg/controller/migplan/migplan_controller.go
```diff
@@ -389,43 +391,46 @@ func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
...
+	//Check if there are any associated migrations that are running
+	for _, m := range migrations {
+		if m.Status.HasCondition(migctl.Running) {
+			runningMigrations++
+		}
+	}
+
+	//No migrations associated with the migplan or there are no running migrations for the Migplan
+	if len(migrations) == 0 || runningMigrations == 0 {
+		if plan.Labels != nil {
```
alaypatel07, line 409: I am curious if this check needed? Labels field should always be present on migplan right?

shubham-pampattiwar: Actually migplan does not have any labels associated, it just has the annotations field.

pkg/controller/migplan/migplan_controller.go
shubham-pampattiwar: I had added that just in case there is a stale running label associated with the migplan.

alaypatel07: if len(migrations) == 0 then the code will never to into this for loop https://github.com/konveyor/mig-controller/pull/681/files#diff-78c6d854c5e343bf8121a5e196e0d5b5R401 and runningMigrations will always be 0 right?

alaypatel07: anyway if you wanna keep it, I am fine, wont break anything IMO just an extra check

pkg/controller/migplan/migplan_controller.go
alaypatel07: I thought even if there are no labels, Labels field will be allocated by the decoder and hence always be non-nil

alaypatel07: let's keep it anyway to safe guard againt panics, +1

pkg/controller/migplan/migplan_controller.go
```diff
@@ -39,7 +39,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-const MigplanRunning = "migplan-running"
+const (
```
alaypatel07, line 42: can you define this consts in v1alpha1/model.go?

Use it there and then import it here as well if you need them

pkg/controller/migplan/migplan_controller.go
```diff
@@ -434,6 +451,50 @@ func (r *ReconcileMigPlan) AddRunningLabelForPlan(plan *migapi.MigPlan) error {
...
+	failedMigrations := 0
+	needsUpdate := false
+	migrations, err := plan.ListMigrations(r)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	//Check if there are any associated migrations that are failed
+	for _, m := range migrations {
+		if m.Status.HasCondition(migctl.PlanNotReady) {
```
alaypatel07, line 465: hrm, is this the right condition for Failed Migration? how about https://github.com/konveyor/mig-controller/blob/4b11788223c4aa5394e0d4881ca0ee8d3e55bc64/pkg/controller/migmigration/validation.go#L19?

pkg/controller/migplan/migplan_controller.go
```diff
@@ -434,6 +451,50 @@ func (r *ReconcileMigPlan) AddRunningLabelForPlan(plan *migapi.MigPlan) error {
...
+			failedMigrations++
+		}
+	}
+
+	switch {
+	//There are no failed migrations associated with the Migplan
+	case failedMigrations == 0:
```
alaypatel07, line 472: how about `failedMigrations == 0 && plan.Labels != nil`?

pkg/controller/migplan/migplan_controller.go
```diff
@@ -434,6 +451,50 @@ func (r *ReconcileMigPlan) AddRunningLabelForPlan(plan *migapi.MigPlan) error {
...
+
+	//Check if there are any associated migrations that are failed
+	for _, m := range migrations {
+		if m.Status.HasCondition(migctl.PlanNotReady) {
+			failedMigrations++
+		}
+	}
+
+	switch {
+	//There are no failed migrations associated with the Migplan
```
alaypatel07, line 471: the general convention is to write comments for switch case inside the case

pkg/controller/migplan/migplan_controller.go
```diff
@@ -434,6 +451,50 @@ func (r *ReconcileMigPlan) AddRunningLabelForPlan(plan *migapi.MigPlan) error {
...
+	}
+
+	//Check if there are any associated migrations that are failed
+	for _, m := range migrations {
+		if m.Status.HasCondition(migctl.PlanNotReady) {
+			failedMigrations++
+		}
+	}
+
+	switch {
```
alaypatel07, line 470: +1 the switch look much much more readable, thanks

pkg/controller/migplan/migplan_controller.go
```diff
@@ -434,6 +451,50 @@ func (r *ReconcileMigPlan) AddRunningLabelForPlan(plan *migapi.MigPlan) error {
...
+	//Migplan is associated with Failed migrations
+	case failedMigrations > 0:
+		if plan.Labels == nil {
+			plan.Labels = make(map[string]string)
+		}
+		plan.Labels[MigplanMigrationFailed] = "true"
+		needsUpdate = true
+	}
+
+	//Plan update call
```
alaypatel07, line 488: comment unnecessary, code seems readable

pkg/controller/migplan/migplan_controller.go
```diff
@@ -434,6 +451,50 @@ func (r *ReconcileMigPlan) AddRunningLabelForPlan(plan *migapi.MigPlan) error {
...
+	if needsUpdate {
```
alaypatel07, line 489: inverse the if to remove the indent level, 

```
        if !needs update{
              return nil
        }
     
	liberr.Wrap(r.Update(context.TODO(), plan))
```

pkg/controller/migplan/registry_health_failed.go
```diff
@@ -0,0 +1,129 @@
+package migplan
+
+import (
+	"github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
```
alaypatel07, line 4: need to order the imports https://godoc.org/golang.org/x/tools/cmd/goimports

pkg/controller/migplan/registry_health_failed.go
```diff
@@ -0,0 +1,129 @@
...
+			return
+		}
+		if planList != nil {
+			for _, plan := range planList {
+				//TODO avoid race condition check
+				srcCluster, err := plan.GetSourceCluster(r.hostClient)
+				if err != nil {
+					log.Trace(err)
+					//TODO Error condition would result in loosing the watch util the pod restarts
+					return
```
alaypatel07, line 72: we talked about removing the returns here so the go routine does not die

pkg/controller/migplan/registry_health_failed.go
```diff
@@ -0,0 +1,129 @@
...
+				}
+
+				if !srcCluster.Status.IsReady() {
+					log.Info("Cannot check registry pod health, cluster is not ready", srcCluster.Name, plan.Name)
+					return
```
alaypatel07, line 77: drop

pkg/controller/migplan/registry_health_failed.go
```diff
@@ -0,0 +1,129 @@
...
+
+				srcClient, err := srcCluster.GetClient(r.hostClient)
+				if err != nil {
+					return
+				}
+
+				destCluster, err := plan.GetDestinationCluster(r.hostClient)
+				if err != nil {
+					log.Trace(err)
+					return
```
alaypatel07, line 88: drop

pkg/controller/migplan/registry_health_failed.go
```diff
@@ -0,0 +1,129 @@
...
+				}
+
+				if !destCluster.Status.IsReady() {
+					log.Info("Cannot check registry pod health, cluster is not ready", destCluster.Name, plan.Name)
+					return
```
alaypatel07, line 93: same for all, log and drop

pkg/controller/migplan/registry_health_failed.go
```diff
@@ -0,0 +1,129 @@
...
+
+				isDestRegistryPodUnhealthy, _, err := isRegistryPodUnHealthy(&plan, destClient)
+				if err != nil {
+					log.Trace(err)
+					return
+				}
+
+				switch {
+				//enqueue a reconcile event when the src registry pod is healthy and the plan is not ready
+				case !isSrcRegistryPodUnhealthy && !plan.Status.HasCondition(RegistriesHealthy):
```
alaypatel07, line 115: you would need this `&& !plan.Status.HadCondition(Ready)` to handle operator opdates right?

pkg/controller/migplan/registry_health_failed.go
```diff
@@ -0,0 +1,129 @@
...
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"time"
+)
+
+// A registry health checker which routinely checks whether registries are healthy or not for migplan's failed migrations
+type registryHealthFailed struct {
```
alaypatel07, line 14: looks like everything for this and `registryHealthRunning` is the same except the switch cases which check for the conditions, can you take in a function pointer to check for cases and reuse the other common code?

pkg/controller/migplan/registry_health_running.go
```diff
@@ -115,21 +115,10 @@ func (r *registryHealth) run() {
...
 					r.enqueue(plan)
 					continue
-
-				//enqueue a reconcile event when the src registry pod is healthy and the plan is not ready
-				case !isSrcRegistryPodUnhealthy && !plan.Status.HasCondition(RegistriesHealthy):
-					r.enqueue(plan)
-					continue
-
 				//enqueue a reconcile event when the destination registry pod is unhealthy and the plan is ready
 				case isDestRegistryPodUnhealthy && plan.Status.HasCondition(RegistriesHealthy):
```
alaypatel07, line 119: you might need to add another condition in with `&&` operator here, see previous comment

pkg/controller/migplan/migplan_controller.go
shubham-pampattiwar: Done

pkg/controller/migplan/migplan_controller.go
shubham-pampattiwar: Updated the conditions

pkg/controller/migplan/migplan_controller.go
shubham-pampattiwar: Done

pkg/controller/migplan/migplan_controller.go
shubham-pampattiwar: Done

pkg/controller/migplan/migplan_controller.go
shubham-pampattiwar: Done

pkg/controller/migplan/registry_health_failed.go
shubham-pampattiwar: Done

pkg/controller/migplan/registry_health_failed.go
shubham-pampattiwar: Done

pkg/controller/migplan/registry_health_failed.go
shubham-pampattiwar: Added

pkg/controller/migplan/registry_health_failed.go
shubham-pampattiwar: Combined the watches.

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -30,12 +35,10 @@ func ListPlans(client k8sclient.Client) ([]MigPlan, error) {
 
 // List `Running` MigPlans
 // Returns and empty list when none found.
-func ListRunningPlans(client k8sclient.Client) ([]MigPlan, error) {
+func ListPlansOnLabels(client k8sclient.Client, labels map[string]string) ([]MigPlan, error) {
```
alaypatel07, line 38: `ListPlanWithLabels`?

pkg/controller/migplan/registry_health_failed.go
alaypatel07: +1 looks great, thanks

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -27,6 +33,21 @@ func ListPlans(client k8sclient.Client) ([]MigPlan, error) {
...
 }
 
+// List `Running` MigPlans
+// Returns and empty list when none found.
+func ListPlansOnLabels(client k8sclient.Client, labels map[string]string) ([]MigPlan, error) {
+	list := MigPlanList{}
+	options := k8sclient.ListOptions{
+		LabelSelector: k8sLabels.SelectorFromSet(labels),
+	}
+	err := client.List(context.TODO(), &options, &list)
```
alaypatel07, line 43: you can probably replace the extra variable use here by

```
    err := client.List(context.TODO(), 
                 &k8sclient.ListOptions{
		    LabelSelector: k8sLabels.SelectorFromSet(labels),
	         }, &list)
```

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -27,6 +33,21 @@ func ListPlans(client k8sclient.Client) ([]MigPlan, error) {
...
+	if err != nil {
```
alaypatel07, line 44: why check for err here just return

```
    return list.Items, err
```

If error, list.Items will be empty

pkg/controller/migplan/migplan_controller.go
```diff
@@ -360,9 +408,118 @@ func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
 	return nil
 }
 
+func (r *ReconcileMigPlan) AddUpdateMigrationStateToPlan(plan *migapi.MigPlan, state string ) error {
```
alaypatel07, line 411: NIT: needs gofmt https://golang.org/cmd/gofmt/

alaypatel07 (commented):
This looks awesome, apart from the small nits in the code, looks like we have everything covered. We should work on bringing the dependent PR in and then merging this one in next 2-3 days

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	"context"
+	k8sLabels "k8s.io/apimachinery/pkg/labels"
```
alaypatel07, line 5: imports

pkg/apis/migration/v1alpha1/v1alpha1_suite_test.go
```diff
@@ -20,6 +20,8 @@ import (
 	golog "log"
 	"os"
 	"path/filepath"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
```
alaypatel07, line 23: imports

pkg/cloudprovider/aws.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/tls"
 	"fmt"
+	"k8s.io/api/apps/v1"
```
alaypatel07, line 7: imports

pkg/cloudprovider/azure.go
```diff
@@ -120,7 +121,7 @@ func (p *AzureProvider) UpdateRegistrySecret(secret, registrySecret *kapi.Secret
 	return nil
 }
 
-func (p *AzureProvider) UpdateRegistryDC(dc *appsv1.DeploymentConfig, name, dirName string) {
+func (p *AzureProvider) UpdateRegistryDC(dc *v1.Deployment, name, dirName string) {
```
alaypatel07, line 124: drop

pkg/cloudprovider/azure.go
```diff
@@ -3,6 +3,7 @@ package cloudprovider
 import (
 	"bytes"
 	"context"
+	"k8s.io/api/apps/v1"
```
alaypatel07, line 6: drop

pkg/cloudprovider/azure.go
```diff
@@ -13,7 +14,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/google/uuid"
 	"github.com/joho/godotenv"
-	appsv1 "github.com/openshift/api/apps/v1"
+	//appsv1 "github.com/openshift/api/apps/v1"
```
alaypatel07, line 17: drop

pkg/cloudprovider/gcp.go
```diff
@@ -68,7 +69,7 @@ func (p *GCPProvider) UpdateRegistrySecret(secret, registrySecret *kapi.Secret)
 	return nil
 }
 
-func (p *GCPProvider) UpdateRegistryDC(dc *appsv1.DeploymentConfig, name, dirName string) {
+func (p *GCPProvider) UpdateRegistryDC(dc *v1.Deployment, name, dirName string) {
```
alaypatel07, line 72: drop

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -460,22 +412,14 @@ func (r *MigPlan) UpdateRegistryDC(storage *MigStorage, deploymentconfig *appsv1
...
-	return nil
+	provider.UpdateRegistryDC(deployment, name, dirName)
 }
 
-// Get an existing registry DeploymentConfig on the specifiedcluster.
-func (r *MigPlan) GetRegistryDC(client k8sclient.Client) (*appsv1.DeploymentConfig, error) {
-	list := appsv1.DeploymentConfigList{}
+// Get an existing registry Deployment on the specified cluster.
+func (r *MigPlan) GetRegistryDeployment(client k8sclient.Client) (*appsv1.Deployment, error) {
+	list := appsv1.DeploymentList{}
```
alaypatel07, line 422: Add a TODO:

```
TODO: We need to convert this from a list call to a get call. The name of the
 deployment is equal to the name of the secret
```

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -492,35 +436,30 @@ func (r *MigPlan) GetRegistryDC(client k8sclient.Client) (*appsv1.DeploymentConf
...
-		len(a.Spec.Template.Spec.Containers) == len(b.Spec.Template.Spec.Containers) &&
-		len(a.Spec.Triggers) == len(b.Spec.Triggers)) {
+		len(a.Spec.Template.Spec.Containers) == len(b.Spec.Template.Spec.Containers)) {
 		return false
 	}
 	for i, container := range a.Spec.Template.Spec.Containers {
 		if !(reflect.DeepEqual(container.Env, b.Spec.Template.Spec.Containers[i].Env) &&
 			reflect.DeepEqual(container.Name, b.Spec.Template.Spec.Containers[i].Name) &&
 			reflect.DeepEqual(container.Ports, b.Spec.Template.Spec.Containers[i].Ports) &&
+			reflect.DeepEqual(container.Image, b.Spec.Template.Spec.Containers[i].Image) &&
```
alaypatel07, line 453: we now also need to compare the probes and fall back on error policy

pkg/controller/migplan/registry.go
```diff
@@ -2,10 +2,13 @@ package migplan
 
 import (
 	"context"
-
+	"fmt"
```
alaypatel07, line 5: new line below

pkg/controller/migplan/registry.go
```diff
@@ -86,7 +100,66 @@ func (r ReconcileMigPlan) ensureMigRegistries(plan *migapi.MigPlan) error {
...
+			Category: migapi.Required,
+			Message:  RegistriesHealthyMessage,
+		})
+	}else {
+		message := fmt.Sprintf(RegistriesUnHealthyMessage, unHealthyPod.Name, unHealthyPod.Namespace, uHealthyClusterName)
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     RegistriesHealthy,
+			Status:   False,
+			Category: migapi.Required,
+			Message:  message,
```
alaypatel07, line 153:
```
         Message: fmt.Sprintf("The Migration registry pod %s/%s is not in a healthy state on cluster %s",
                                                  unHealthyPod.Namespace, unHealthyPod.Name, unHealthyClusterName)
```

alaypatel07:
```
  - category: Required
    lastTransitionTime: "2020-10-13T18:27:08Z"
    message: The Migration registry pod openshift-migration/registry-27e56170-df8e-4a0e-8c83-0316315d7d5d-t6q97-556f58tt6fn
      is not in a healthy state on cluster alpatel-aux-tools-443
    status: "False"
    type: RegistriesHealthy
```

Got this when I tried my source registry pod is crashlooping, trying to fix it and break it during migration again

alaypatel07 (commented):
I have tested the changes in here manually

1. Create migplan
2. Start a migration on the migplan
3. create egressnetwork policy to block internet access on source side, this will make registry pods unready
4. Step three will drive the plan to unread
5. Step four will drive the migmigration in to a stalled state, hence fast failure is achieved

shubham-pampattiwar (approved):

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -416,33 +347,54 @@ func (r *MigPlan) UpdateRegistryDC(storage *MigStorage, deploymentconfig *appsv1
...
+							PeriodSeconds: 5,
+						},
+						ReadinessProbe: &kapi.Probe{
+							Handler: kapi.Handler{
+								HTTPGet: &kapi.HTTPGetAction{
+									Path: "/v2/_catalog",
+									Port: intstr.IntOrString{IntVal: 5000},
+								},
+							},
+							PeriodSeconds: 5,
```
alaypatel07, line 396: @djwhatle and I were wondering if we need to explore timeouts here

alaypatel07: also wondered if we can use this

```
GET /v2/_catalog?n=<integer>
```

pkg/controller/migplan/registry.go
```diff
@@ -214,3 +266,72 @@ func (r ReconcileMigPlan) ensureRegistryService(client k8sclient.Client, plan *m
 
 	return nil
 }
+
+func (r ReconcileMigPlan) getBothClients(plan *migapi.MigPlan) ([]compat.Client, error) {
```
djwhatle, line 270: `getBothClients` is duplicating functionality we already have in mig-controller. Let's drop the `getBothClients` function and construct the two clients in deleteImageRegistryResources as shown below.

See 
```
# from https://github.com/konveyor/mig-controller/blob/master/pkg/apis/migration/v1alpha1/migplan_types.go#L137
plan.GetRefResources(client)
```

```
# use this in deleteImageRegistryResources()

planRefResources, err := r.GetRefResources(r.client)
if err != nil {
  return liberr.wrap(err)
}
srcClient, err := planRefResources.SrcMigCluster.GetClient(r.Client)
if err != nil {
  return liberr.wrap(err)
}
destClient, err := planRefResources.DestMigCluster.GetClient(r.Client)
if err != nil {
  return liberr.wrap(err)
}
clients := []compat.Client{srcClient, destClient}

[...]
```

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,144 @@
...
+	for {
+		time.Sleep(r.Interval)
+		planList, err := v1alpha1.ListPlansWithLabels(r.hostClient, r.planLabels)
+		if err != nil {
+			log.Trace(err)
+			return
+		}
+		if planList != nil {
+			for _, plan := range planList {
+				//TODO avoid race condition check
```
djwhatle, line 71: What work is left to fix these TODOs?

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,144 @@
...
+			return
+		}
+	}
+
+	r.handler.Generic(event, r.queue)
+}
+
+//Run the health checks for registry pods
+func (r *registryHealth) run() {
+	//List all the migplans that are in running state using the hostClient
```
djwhatle, line 58: Recommend shortening this comment paragraph to:
```
Enqueue running MigPlans that have unhealthy registry pods
```

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,144 @@
...
+			log.Trace(err)
+			return
+		}
+		if planList != nil {
+			for _, plan := range planList {
+				//TODO avoid race condition check
+				srcCluster, err := plan.GetSourceCluster(r.hostClient)
+				if err != nil {
+					log.Trace(err)
+					//TODO Error condition would result in loosing the watch util the pod restarts
```
djwhatle, line 75: I assume `util` is meant to be `until`?

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,144 @@
...
+
+				if r.checkPodHealthAndPlanConditionToEnqueue(destRegistryPods, &plan) {
+					r.enqueue(plan)
+					continue
+				}
+			}
+		}
+
+	}
+	//TODO: need see if this go routine should be stopped and returned
```
djwhatle, line 124: What work is left here?

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (r *registryHealth) checkPodHealthAndPlanConditionToEnqueue(podList corev1.PodList, plan *v1alpha1.MigPlan) bool {
```
djwhatle, line 127: Consider dropping `ToEnqueue` from this name.

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,144 @@
...
+
+	podStateUnhealthy, _ := isRegistryPodUnHealthy(podList)
+	enqueue := false
+
+	switch {
+
+	case podStateUnhealthy && plan.Status.HasCondition(RegistriesHealthy) && plan.Status.IsReady():
+		//enqueue a reconcile event when the registry pod is unhealthy and the plan is ready
```
djwhatle, line 135: +1 helpful, concise comments

pkg/controller/migplan/validation.go
```diff
@@ -259,7 +265,11 @@ func (r ReconcileMigPlan) validateStorage(plan *migapi.MigPlan) error {
 			Category: Critical,
 			Message:  StorageNotReadyMessage,
 		})
-		return nil
+		// this is necessary because between there are multiple checks happening
+		// between validateStorage and ensureStorage, if any of those checks are failing
+		// we will never run ensureStorage. This is an attempt to always have the right
+		// state
+		return liberr.Wrap(r.deleteImageRegistryResources(plan))
```
djwhatle, line 272: I feel this comment could be simplified to:

```
Image registry must cleaned up if deleted MigStorage is unready
```

djwhatle (commented):
This registry health check will be great to have in mig-controller. I will continue to review this tomorrow, but added a bunch of comments for now to consider. Please click "expand comments" because I think they are going to get compressed in the GitHub PR interface.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -460,22 +412,15 @@ func (r *MigPlan) UpdateRegistryDC(storage *MigStorage, deploymentconfig *appsv1
...
-			appsv1.DeploymentTriggerPolicy{
-				Type: appsv1.DeploymentTriggerOnConfigChange,
-			},
-		},
 	}
 	provider := storage.GetBackupStorageProvider()
-	provider.UpdateRegistryDC(deploymentconfig, name, dirName)
-
-	return nil
+	provider.UpdateRegistryDC(deployment, name, dirName)
```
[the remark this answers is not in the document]
alaypatel07: Updated in response to Scott's review

pkg/apis/migration/v1alpha1/v1alpha1_suite_test.go
```diff
@@ -40,16 +42,24 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		golog.Fatal(err)
 	}
-
-	if cfg, err = t.Start(); err != nil {
-		golog.Fatal(err)
+	setupControlPlane := os.Getenv("SETUP_CONTROL_PLANE")
```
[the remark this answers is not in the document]
alaypatel07: To be precise, this can be used in non-MacOS testing as well. The fact that control plane can take some time to come up and in my case I was never able to bring it up successfully, restricted me with ability to run unit test that dont depend on control plane

pkg/cloudprovider/aws.go
```diff
@@ -131,15 +128,20 @@ func (p *AWSProvider) UpdateCloudSecret(secret, cloudSecret *kapi.Secret) error
...
 		"access_key":    []byte(secret.Data[AwsAccessKeyId]),
 		"secret_key":    []byte(secret.Data[AwsSecretAccessKey]),
-		"ca_bundle.pem": p.CustomCABundle,
+		"ca_bundle.pem": caBundle,
 	}
 	return nil
 }
 
-func (p *AWSProvider) UpdateRegistryDC(dc *appsv1.DeploymentConfig, name, dirName string) {
+func (p *AWSProvider) UpdateRegistryDC(dc *appsv1.Deployment, name, dirName string) {
```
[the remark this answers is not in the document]
alaypatel07: done

pkg/cloudprovider/azure.go
```diff
@@ -120,7 +120,7 @@ func (p *AzureProvider) UpdateRegistrySecret(secret, registrySecret *kapi.Secret
 	return nil
 }
 
-func (p *AzureProvider) UpdateRegistryDC(dc *appsv1.DeploymentConfig, name, dirName string) {
+func (p *AzureProvider) UpdateRegistryDC(dc *appsv1.Deployment, name, dirName string) {
```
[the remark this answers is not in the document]
alaypatel07: Done

pkg/controller/migmigration/registry.go
```diff
@@ -23,14 +23,18 @@ func (t *Task) getAnnotations(client k8sclient.Client) (map[string]string, error
...
+	registryDC, err := t.PlanResources.MigPlan.GetRegistryDeployment(client)
 	if err != nil {
 		return nil, err
 	}
 	if registryDC == nil {
 		return nil, errors.New("migration registry DeploymentConfig not found")
 	}
 
+	if registryDC.DeletionTimestamp != nil {
+		return nil, errors.New(fmt.Sprintf("DeploymentConfig %s/%s is being garbage collected with deletion timestamp %s", registryDC.Namespace, registryDC.Name, registryDC.DeletionTimestamp))
```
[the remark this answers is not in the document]
alaypatel07: > Is this meant to return an error? What is the consequence to us if a DeletionTimestamp is set but has not yet been acted upon?

When the user has deleted the resource but its awaiting garbage collection from kube, it will have a DeletionTimestamp set, its just to safe guard us against an edge case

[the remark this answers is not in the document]
alaypatel07: will address other comments in this thread

pkg/cloudprovider/gcp.go
```diff
@@ -68,7 +68,7 @@ func (p *GCPProvider) UpdateRegistrySecret(secret, registrySecret *kapi.Secret)
 	return nil
 }
 
-func (p *GCPProvider) UpdateRegistryDC(dc *appsv1.DeploymentConfig, name, dirName string) {
+func (p *GCPProvider) UpdateRegistryDC(dc *appsv1.Deployment, name, dirName string) {
```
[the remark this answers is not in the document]
alaypatel07: Updated

pkg/controller/migplan/migplan_controller.go
```diff
@@ -71,6 +73,32 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 	}
 
+	// Watch for changes to deployment registry referenced by MigPlan that have running migrations
+	err = c.Watch(
+		&registryHealth{
+			hostClient: mgr.GetClient(),
+			planLabels: map[string]string{
+				migapi.MigplanMigrationRunning: "true",
+			},
+			Interval: time.Second * 5},
```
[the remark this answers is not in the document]
alaypatel07: I will update the probe with 3-second timeout and 15 seconds of initial delay. I would still keep this at 5 second, since it will only mean at worst case scenario we will have 2 unnecessary checks. Can discuss this further on a call if this does not sound okay

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -416,33 +347,54 @@ func (r *MigPlan) UpdateRegistryDC(storage *MigStorage, deploymentconfig *appsv1
...
+						Image:                    registryImage,
+						Name:                     "registry",
+						TerminationMessagePolicy: kapi.TerminationMessageFallbackToLogsOnError,
 						Ports: []kapi.ContainerPort{
 							kapi.ContainerPort{
 								ContainerPort: 5000,
 								Protocol:      kapi.ProtocolTCP,
 							},
 						},
+						LivenessProbe: &kapi.Probe{
```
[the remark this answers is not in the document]
alaypatel07: will update with this
```
							PeriodSeconds:       5,
							TimeoutSeconds:      3,
							InitialDelaySeconds: 15,
```

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -551,10 +494,10 @@ func (r *MigPlan) UpdateRegistryService(service *kapi.Service, name string) erro
 			"deploymentconfig": name,
 		},
 	}
-	return nil
 }
 
 // Get an existing registry Service on the specifiedcluster.
+//TODO: We need to convert this from a list call to a get call. The name of the deployment is equal to the name of the secret
```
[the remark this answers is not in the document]
alaypatel07: @djwhatle can we take this as a follow on? This PR is a long-standing one, didn't want to spend more time in an unmerged state unless absolutely critical

pkg/controller/migplan/migplan_controller.go
```diff
@@ -255,9 +296,16 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	// Migration Registry Health check
+	err = r.ensureRegistryHealth(plan)
```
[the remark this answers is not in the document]
alaypatel07: if they are not okay, we will go to unready and fix them into the next reconcile? why would that be a problem?

pkg/controller/migplan/migplan_controller.go
```diff
@@ -360,9 +408,118 @@ func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
...
+}
+
 // Update Status.ExcludedResources based on settings
 func (r *ReconcileMigPlan) setExcludedResourceList(plan *migapi.MigPlan) error {
 	excludedResources := Settings.Plan.ExcludedResources
 	plan.Status.ExcludedResources = excludedResources
 	return nil
 }
+
+func (r ReconcileMigPlan) deleteImageRegistryResourcesForClient(client k8sclient.Client, plan *migapi.MigPlan) error {
```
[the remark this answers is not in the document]
alaypatel07: @djwhatle I think keeping the `ForClient` will help answer questions like why is the `r.Client` not being used here. The `ForClient` explicitly tells me that a different client than the controller cluster client is being passed for a reason that there could be source and destination client

pkg/controller/migplan/registry.go
```diff
@@ -86,7 +101,66 @@ func (r ReconcileMigPlan) ensureMigRegistries(plan *migapi.MigPlan) error {
...
+			Status:   True,
+			Category: migapi.Required,
+			Message:  RegistriesHealthyMessage,
+		})
+	} else {
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     RegistriesHealthy,
+			Status:   False,
+			Category: migapi.Required,
+			Message:  fmt.Sprintf("The Migration registry pod %s/%s is not in a healthy state on cluster %s",
```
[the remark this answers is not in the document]
alaypatel07: @djwhatle I think @jortel has a president of move out of the constant status messages and have some sort of context in them. This came of our logging improvement discussion, doing this way will save us time adding the logging entry everywhere and we could re-use the status condition messages to log instead

pkg/controller/migplan/registry_health.go
```diff
@@ -0,0 +1,144 @@
...
+			for _, plan := range planList {
+				//TODO avoid race condition check
+				srcCluster, err := plan.GetSourceCluster(r.hostClient)
+				if err != nil {
+					log.Trace(err)
+					//TODO Error condition would result in loosing the watch util the pod restarts
+				}
+
+				if !srcCluster.Status.IsReady() {
+					log.Info("Cannot check registry pod health, cluster is not ready", srcCluster.Name, plan.Name)
```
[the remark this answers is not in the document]
alaypatel07: I would actually keep it for debugging if this function is not enqueuing I'd like to know why. May be we can put a log.V(4).Info?

pkg/controller/migplan/registry_health.go
alaypatel07: @djwhatle we probably need to find a way to pass the controller stop channel here so when all the controllers are stopped, we stop this watch source as well. Would like to take that as a follow on.

pkg/controller/migplan/registry_health.go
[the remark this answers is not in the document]
djwhatle: Yeah this seems like a good candidate for v4 (debug) logging

pkg/controller/migplan/registry_health.go
djwhatle: I don't think we ever stop our controllers except for on termination of the process, but maybe you're thinking of some future use-case?

pkg/apis/migration/v1alpha1/migplan_types.go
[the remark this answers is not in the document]
djwhatle: Works for me

pkg/apis/migration/v1alpha1/migplan_types_test.go
```diff
@@ -254,3 +255,58 @@ func TestPersistentVolumes_EndPvStaging(t *testing.T) {
 			},
 		}))
 }
+
+func TestMigPlan_EqualsRegistrySecret(t *testing.T) {
```
djwhatle, line 259: +1 for adding tests

pkg/controller/migplan/migplan_controller.go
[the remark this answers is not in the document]
djwhatle: I'm not sure I understand the need to react so quickly to registry problems, seems like waiting ~30s for the issue to be surfaced by the controller would be acceptable and have less APIserver strain, but I may not have some key piece of info still.

pkg/controller/migplan/migplan_controller.go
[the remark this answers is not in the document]
djwhatle: Yeah actually that sounds fine to me. I just want to make sure this won't cause flapping on registry creation or something.

pkg/controller/migplan/migplan_controller.go
```diff
@@ -360,9 +408,118 @@ func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
...
+		}
+		plan.Labels[planLabel] = "true"
+		needsUpdate = true
+	}
+
+	if !needsUpdate {
+		return nil
+	}
+
+	return liberr.Wrap(r.Update(context.TODO(), plan))
```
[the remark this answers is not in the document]
djwhatle: @shubham-pampattiwar @alaypatel07 I still feel this is not the right place to be doing an update. Wait until the end of the reconcile. The MigPlan will be updated there.

pkg/controller/migplan/migplan_controller.go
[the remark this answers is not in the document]
djwhatle: That makes sense to me, it just differs from the norms in the controller. I think there's some value in having a consistent naming convention for functions that do similar things for codebase readability. If we use this here perhaps we should also consider matching the convention elsewhere.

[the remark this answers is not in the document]
djwhatle: I'll also add that it's very normal for us to be using a passed in client in the migration controller since connecting to remote clusters is very commonplace here. 

We would really need to attach `forClient` to a lot of our functions with this convention.

pkg/controller/migplan/registry.go
```diff
@@ -14,17 +18,34 @@ func (r ReconcileMigPlan) ensureMigRegistries(plan *migapi.MigPlan) error {
...
+	if plan.Status.HasCriticalCondition() {
+		plan.Status.StageCondition(RegistriesEnsured)
+		// this is required because when migstorage is not ready, plan automatically
+		// has a critical condition, hence chances are if migstorage is not ready will
+		// we most likely return from here
+		return liberr.Wrap(r.deleteImageRegistryResources(plan))
+	}
+
+	if plan.Status.HasCondition(Suspended) {
+		// TODO: need to make sure that if plan is suspended due to final migration is complete
```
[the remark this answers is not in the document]
djwhatle: Bump, what's the plan here? Are we thinking fix this or do you want to leave it for next round @alaypatel07 @shubham-pampattiwar

djwhatle (commented):
Looking good. Similar to other big PRs going in right now, I think it would make sense to run the `cam-e2e-qe` test suite against this one and post results to make sure we aren't introducing any regressions / can catch issues early. On this one I'm most concerned that a bug could introduce registry flapping that could leave a migration unable to run due to unreadiness of MigPlan.

I'll forward the info for that along to you @alaypatel07 and @shubham-pampattiwar. Pranav ran this test suite recently so he should be able to provide guidance as well. Make sure to build the controller and deploy it into your cluster before running.

Once we get results posted from the test suite, we just need to coordinate merges. I may need to make a few more changes on remote watch stuff so it may be that this will actually go in first, we'll see.

P.S. If any of the TODOs in here deserve an issue, I'd ask that we file issues as appropriate @alaypatel07

pkg/controller/migplan/migplan_controller.go
```diff
@@ -360,9 +408,118 @@ func (r *ReconcileMigPlan) planSuspended(plan *migapi.MigPlan) error {
...
+			}
+		}
+	}
+
+	planLabel := getLabelForPlan(state)
+
+	switch {
+
+	case migrationStateInstances == 0 && plan.Labels != nil:
+		//There migrations associated with the Migplan for the given state
```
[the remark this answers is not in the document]
djwhatle: Still open question

pkg/controller/migplan/registry.go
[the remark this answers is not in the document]
djwhatle: I'm not sure I understand, seems like the message would be more re-usable if it was a const.

alaypatel07:
@djwhatle I'll get to the test suite today, and follow up on the issues, however, I wonder how much it will impact our dev time if we get into the practice of running test suites for every major PR. It seems that the restore progress reporting and this one does not change the concepts of how we do things in mig-controller, so I wonder how much return on investment it will give us spending time to incorporate that in our development cycle. I guess this begets a wider discussion on maybe a scrum of scrum call or something to make a general guideline of "when it is worth for developers to run QE test suite", for now, I'll test this any way to get this moving forward.

djwhatle:
@alaypatel07 you bring up a good point that we should carefully consider for the future. 

I think on the progress reporting PR, there was very little change to the logic that mig-controller was executing to carry out a migration and I think we could have skipped the tests there, although I think it was valuable in the sense that we are raising dev team awareness of how to run these tests when they're useful, and to validate that the instructions I wrote don't have gaps.

On this PR, I think we are making a lot of logical changes across the codebase to how registries are brought up and pulled down. If there are errors in this it could block migrations. With 600 lines of code of logic changes, it's hard for me to eyeball this PR and say "all of that logic should work" with a high level of confidence, and since the dev team doesn't have time to run manual tests on eachothers PRs right now, I think the burden of running test suite on big enough logic changes is worth it for the ~15-30 minutes it should cost to kick off the test suite and check the result later.

It would be really cool if we could just leave a comment `/test-with-e2e-suite` on the PR and kick off all of the tests to avoid the time lost starting things up manually, although historically the CI tends to need maintenance over time as well.

djwhatle (commented):
Waiting on results from cam-e2e-qe

pkg/controller/migplan/registry.go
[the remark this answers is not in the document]
djwhatle: Will be filed as issue

djwhatle:
Closing in favor of #712 so that we can get this merged with some changes.

PR #698, opened 2020-10-14
MIG-323 - add logging improvements and kube events for status conditions to migstorage_controller

alaypatel07:
This adds logging improvement to migstorage_controller.

The idea is to set a standard between condition messages and
log messages. Since conditions are on an object it has the object
metadata attached to it. The log entry needs the namespace and
the name of the object for which the condition is added.

With this, a developer will be able 
1. to pinpoint the line at which error condition occurred 
2. will be able to get more context on what the code was
    trying to do when the error occurred.

### DIFF
--- a/pkg/apis/migration/v1alpha1/condition.go
+++ b/pkg/apis/migration/v1alpha1/condition.go
@@ -7,7 +7,10 @@ import (
 	"strings"
 	"time"
 
+	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
 )
 
 // Types
@@ -356,4 +359,17 @@ func (r *Conditions) SetReconcileFailed(err error) {
 		Items:    []string{err.Error()},
 	})
 	r.EndStagingConditions()
+}
+
+func (r *Conditions) RecordEvents(obj runtime.Object, recorder record.EventRecorder) {
+	for _, cond := range r.List {
+		eventType := ""
+		switch cond.Category {
+		case Critical, Warn, Error:
+			eventType = kapi.EventTypeWarning
+		default:
+			eventType = kapi.EventTypeNormal
+		}
+		recorder.Event(obj, cond.Reason, eventType, cond.Message)
+	}
 }
--- a/pkg/controller/migstorage/migstorage_controller.go
+++ b/pkg/controller/migstorage/migstorage_controller.go
@@ -26,6 +26,7 @@ import (
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -44,7 +45,7 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileMigStorage{Client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	return &ReconcileMigStorage{Client: mgr.GetClient(), scheme: mgr.GetScheme(), EventRecorder: mgr.GetRecorder("migstorage_controller")}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -95,12 +96,14 @@ var _ reconcile.Reconciler = &ReconcileMigStorage{}
 // ReconcileMigStorage reconciles a MigStorage object
 type ReconcileMigStorage struct {
 	client.Client
+	record.EventRecorder
 	scheme *runtime.Scheme
 }
 
 func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	var err error
 	log.Reset()
+	log.SetValues("storage", request)
 
 	// Fetch the MigStorage instance
 	storage := &migapi.MigStorage{}
@@ -115,6 +118,8 @@ func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Re
 
 	// Report reconcile error.
 	defer func() {
+		log.Info("CR", "conditions", storage.Status.Conditions)
+		storage.Status.Conditions.RecordEvents(storage, r.EventRecorder)
 		if err == nil || errors.IsConflict(err) {
 			return
 		}
@@ -139,7 +144,7 @@ func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Re
 	// Ready
 	storage.Status.SetReady(
 		!storage.Status.HasBlockerCondition(),
-		ReadyMessage)
+		"The storage is ready.")
 
 	// End staging conditions.
 	storage.Status.EndStagingConditions()
--- a/pkg/controller/migstorage/validation.go
+++ b/pkg/controller/migstorage/validation.go
@@ -1,9 +1,11 @@
 package migstorage
 
 import (
+	"fmt"
 	liberr "github.com/konveyor/controller/pkg/error"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	migref "github.com/konveyor/mig-controller/pkg/reference"
+	"path"
 )
 
 // Notes:
@@ -43,19 +45,6 @@ const (
 	False = migapi.False
 )
 
-// Messages
-const (
-	ReadyMessage                   = "The storage is ready."
-	InvalidBSProviderMessage       = "The `backupStorageProvider` must be: (aws|gcp|azure)."
-	InvalidBSCredsSecretRefMessage = "The `backupStorageConfig.credsSecretRef` must reference a `secret`."
-	InvalidBSFieldsMessage         = "The `backupStorageConfig` settings [] not valid."
-	InvalidVSProviderMessage       = "The `volumeSnapshotProvider` must be: (aws|gcp|azure)."
-	InvalidVSCredsSecretRefMessage = "The `volumeSnapshotConfig.credsSecretRef` must reference a `secret`."
-	InvalidVSFieldsMessage         = "The `volumeSnapshotConfig` settings [] not valid."
-	BSProviderTestFailedMessage    = "The Backup storage cloudprovider test failed []."
-	VSProviderTestFailedMessage    = "The Volume Snapshot cloudprovider test failed []."
-)
-
 // Validate the storage resource.
 func (r ReconcileMigStorage) validate(storage *migapi.MigStorage) error {
 	err := r.validateBackupStorage(storage)
@@ -79,7 +68,7 @@ func (r ReconcileMigStorage) validateBackupStorage(storage *migapi.MigStorage) e
 			Status:   True,
 			Reason:   NotSet,
 			Category: Critical,
-			Message:  InvalidBSProviderMessage,
+			Message:  "The `spec.BackupStorageProvider` must be: (aws|gcp|azure).",
 		})
 		return nil
 	}
@@ -93,7 +82,7 @@ func (r ReconcileMigStorage) validateBackupStorage(storage *migapi.MigStorage) e
 			Status:   True,
 			Reason:   NotSupported,
 			Category: Critical,
-			Message:  InvalidBSProviderMessage,
+			Message:  fmt.Sprintf("The `spec.BackupStorageProvider` must be: (aws|gcp|azure), provider %s", storage.Spec.BackupStorageProvider),
 		})
 		return nil
 	}
@@ -105,7 +94,7 @@ func (r ReconcileMigStorage) validateBackupStorage(storage *migapi.MigStorage) e
 			Status:   True,
 			Reason:   NotSet,
 			Category: Critical,
-			Message:  InvalidBSCredsSecretRefMessage,
+			Message:  "The `backupStorageConfig.credsSecretRef` must reference a valid `secret`.",
 		})
 		return nil
 	}
@@ -123,7 +112,8 @@ func (r ReconcileMigStorage) validateBackupStorage(storage *migapi.MigStorage) e
 			Status:   True,
 			Reason:   NotFound,
 			Category: Critical,
-			Message:  InvalidBSCredsSecretRefMessage,
+			Message: fmt.Sprintf("The `backupStorageConfig.credsSecretRef` must reference a valid `secret`, subject %s",
+				path.Join(storage.Spec.BackupStorageConfig.CredsSecretRef.Namespace, storage.Spec.BackupStorageConfig.CredsSecretRef.Name)),
 		})
 		return nil
 	}
@@ -136,8 +126,9 @@ func (r ReconcileMigStorage) validateBackupStorage(storage *migapi.MigStorage) e
 			Status:   True,
 			Reason:   NotSupported,
 			Category: Critical,
-			Message:  InvalidBSFieldsMessage,
-			Items:    fields,
+			Message: fmt.Sprintf("The `backupStorageConfig.credsSecretRef` must reference a valid `secret`, subject %s, []",
+				path.Join(storage.Spec.BackupStorageConfig.CredsSecretRef.Namespace, storage.Spec.BackupStorageConfig.CredsSecretRef.Name)),
+			Items: fields,
 		})
 		return nil
 	}
@@ -151,7 +142,7 @@ func (r ReconcileMigStorage) validateBackupStorage(storage *migapi.MigStorage) e
 				Status:   True,
 				Reason:   TestFailed,
 				Category: Critical,
-				Message:  BSProviderTestFailedMessage,
+				Message:  fmt.Sprintf("The `backupStorageConfig` settings [] not provided in secret %s not valid.", path.Join(secret.Namespace, secret.Name)),
 				Items:    []string{err.Error()},
 			})
 		}
@@ -179,7 +170,7 @@ func (r ReconcileMigStorage) validateVolumeSnapshotStorage(storage *migapi.MigSt
 				Status:   True,
 				Reason:   NotSupported,
 				Category: Critical,
-				Message:  InvalidVSProviderMessage,
+				Message:  fmt.Sprintf("The `volumeSnapshotProvider` must be: (aws|gcp|azure)., provider: %s", storage.Spec.VolumeSnapshotProvider),
 			})
 			return nil
 		}
@@ -191,7 +182,7 @@ func (r ReconcileMigStorage) validateVolumeSnapshotStorage(storage *migapi.MigSt
 				Status:   True,
 				Reason:   NotSet,
 				Category: Critical,
-				Message:  InvalidVSCredsSecretRefMessage,
+				Message:  "The `volumeSnapshotConfig.credsSecretRef` must reference a valid `secret`.",
 			})
 			return nil
 		}
@@ -203,7 +194,9 @@ func (r ReconcileMigStorage) validateVolumeSnapshotStorage(storage *migapi.MigSt
 				Status:   True,
 				Reason:   NotFound,
 				Category: Critical,
-				Message:  InvalidVSCredsSecretRefMessage,
+				Message: fmt.Sprintf("The `volumeSnapshotConfig.credsSecretRef` must reference a `secret`. subject %s", path.Join(
+					storage.Spec.VolumeSnapshotConfig.CredsSecretRef.Namespace,
+					storage.Spec.VolumeSnapshotConfig.CredsSecretRef.Namespace)),
 			})
 			return nil
 		}
@@ -216,7 +209,7 @@ func (r ReconcileMigStorage) validateVolumeSnapshotStorage(storage *migapi.MigSt
 				Status:   True,
 				Reason:   NotSupported,
 				Category: Critical,
-				Message:  InvalidVSFieldsMessage,
+				Message:  fmt.Sprintf("The `volumeSnapshotConfig` settings [] in secret %s not valid.", path.Join(secret.Namespace, secret.Name)),
 				Items:    fields,
 			})
 			return nil
@@ -232,7 +225,7 @@ func (r ReconcileMigStorage) validateVolumeSnapshotStorage(storage *migapi.MigSt
 				Status:   True,
 				Reason:   TestFailed,
 				Category: Critical,
-				Message:  VSProviderTestFailedMessage,
+				Message:  "The Volume Snapshot cloudprovider test failed [].",
 				Items:    []string{err.Error()},
 			})
 		}
--- a/pkg/apis/migration/v1alpha1/migstorage_types.go
+++ b/pkg/apis/migration/v1alpha1/migstorage_types.go
@@ -205,12 +205,12 @@ func (r *MigStorage) GetVolumeSnapshotProvider() pvdr.Provider {
 	return provider
 }
 
-// Get the backup credentials secret.
+// Get the backup credentials secret. If the secret is not found it returns nil
 func (r *MigStorage) GetBackupStorageCredSecret(client k8sclient.Client) (*kapi.Secret, error) {
 	return r.Spec.BackupStorageConfig.GetCredSecret(client)
 }
 
-// Get the backup credentials secret.
+// Get the backup credentials secret. If the secret is not found it returns nil
 func (r *MigStorage) GetVolumeSnapshotCredSecret(client k8sclient.Client) (*kapi.Secret, error) {
 	if r.Spec.VolumeSnapshotProvider != "" {
 		return r.Spec.VolumeSnapshotConfig.GetCredSecret(client)


alaypatel07:
```
$ oc logs -f migration-controller-65c86dcd97-gkw4v -c cam  | grep storage
{"level":"info","ts":1602694266.5340188,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"migstorage-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1602694266.534189,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"migstorage-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1602694266.5342407,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"migstorage-controller","source":{"Client":{"Reader":{"CacheReader":{"Scheme":{}},"ClientRea
der":{}},"Writer":{},"StatusClient":{}},"Interval":30000000000}}
{"level":"info","ts":1602694267.435467,"logger":"kubebuilder.controller","msg":"Starting Controller","controller":"migstorage-controller"}
{"level":"info","ts":1602694267.53573,"logger":"kubebuilder.controller","msg":"Starting workers","controller":"migstorage-controller","worker count":1}
{"level":"info","ts":1602694564.8180778,"logger":"storage|j5gp6","msg":"condition added on openshift-migration/velero-4","category":"Critical","type":"BackupStorageProviderTestFailed","reason":"TestFailed","m
essage":"The Backup storage cloudprovider test failed [].","errors":["AccessDenied: Access Denied\n\tstatus code: 403, request id: D435DDBBA09B3AA8, host id: 2kPycetTuznZT0DOYtXR2QvDxIGrXm1DyjDrUmp3NhVkCmneyt
dLruTU/Y2F9u5mQmnv/abKvLE="]}
```

This is what I get if my velero creds are invalid

alaypatel07:
log
```
{"level":"info","ts":1602694869.2065911,"logger":"storage|2x94s","msg":"condition added on openshift-migration/velero-4","category":"Critical","type":"InvalidBackupStorageCredsSecretRef","reason":"NotFound","
message":"couldnt find secret for backup provider openshift-config/velero-4-nfqxq-","errors":[]}
```
status condition
```
status:
  conditions:
  - category: Critical
    lastTransitionTime: "2020-10-14T17:01:09Z"
    message: couldnt find secret for backup provider openshift-config/velero-4-nfqxq-
    reason: NotFound
    status: "True"
    type: InvalidBackupStorageCredsSecretRef
  observedDigest: bd61c6d6b84c69d0adb43c31b3ea66efb7eb92dd5dae2c9ea77b3c4feb6f0137
```

pkg/controller/discovery/web/web.go
```diff
@@ -70,7 +70,7 @@ func (w *WebServer) buildOrigins() {
 	for _, r := range Settings.Discovery.CORS.AllowedOrigins {
 		expr, err := regexp.Compile(r)
 		if err != nil {
-			Log.Error(
+			Log.ErrorWithTrace(
```
alaypatel07, line 73: Neel to drop this change

pkg/controller/migmigration/restore.go
```diff
@@ -295,7 +295,7 @@ func (t *Task) deleteMigrated() error {
 					if k8serror.IsMethodNotSupported(err) || k8serror.IsNotFound(err) {
 						continue
 					}
-					log.Error(err, fmt.Sprintf("Failed to request delete on: %s", gvr.String()))
+					log.ErrorWithTrace(err, fmt.Sprintf("Failed to request delete on: %s", gvr.String()))
```
alaypatel07, line 298: I'll drop this change

pkg/controller/migstorage/validation.go
```diff
@@ -168,57 +175,61 @@ func (r ReconcileMigStorage) validateVolumeSnapshotStorage(storage *migapi.MigSt
...
-			storage.Status.SetCondition(migapi.Condition{
+			storage.Status.SetAndLogCondition(log, migapi.Condition{
 				Type:     InvalidVSProvider,
 				Status:   True,
 				Reason:   NotSupported,
 				Category: Critical,
 				Message:  InvalidVSProviderMessage,
-			})
+			}, storage.Namespace, storage.Name)
+			log.V(2).Info("migstorage spec.VolumeSnapshotProvider empty", "name", storage.Name, "namespace", storage.Namespace)
```
alaypatel07, line 194: I'll have to drop this

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -107,26 +108,33 @@ func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Re
 	err = r.Get(context.TODO(), request.NamespacedName, storage)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			log.V(2).Info("unable to get migstorage, not found", "migstorage", request.String(), "error", err.Error())
```
djwhatle, line 111: 1. Was this `V(2)` / `V(4)` functionality already built into the logger and we just weren't using it?
2. How does the user control the level of log messages they will see?

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -107,26 +108,33 @@ func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Re
...
 
 	// Report reconcile error.
 	defer func() {
-		if err == nil || errors.IsConflict(err) {
+		switch {
+		case err == nil:
+			log.V(4).Info("migstorage reconcile successful", "migstorage", request.String())
+			return
+		case errors.IsConflict(err):
+			log.Error(fmt.Errorf("unable to get migstorage: %#v", err), "migstorage", request.String())
```
djwhatle, line 125: Might be missing something, but this log message seems to mislead the user into thinking we couldn't `get` the migstorage, when in reality we hit a `ConflictError`.

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -139,7 +147,7 @@ func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Re
 	// Ready
 	storage.Status.SetReady(
 		!storage.Status.HasBlockerCondition(),
-		ReadyMessage)
+		"The storage is ready.")
```
djwhatle, line 150: @jortel I was talking to @alaypatel07 and he mentioned you guys decided to move away from shared message consts to having these messages inline. 

I'd like to understand the reasoning behind this just so I'm aware.

pkg/controller/migstorage/validation.go
```diff
@@ -168,73 +161,79 @@ func (r ReconcileMigStorage) validateVolumeSnapshotStorage(storage *migapi.MigSt
...
 		// Fields
 		fields := provider.Validate(secret)
 		if len(fields) > 0 {
-			storage.Status.SetCondition(migapi.Condition{
+			storage.Status.SetAndLogCondition(log, migapi.Condition{
 				Type:     InvalidVSFields,
 				Status:   True,
 				Reason:   NotSupported,
 				Category: Critical,
-				Message:  InvalidVSFieldsMessage,
```
djwhatle, line 219: I see that all of these const messages are no longer being used. Should we prune the consts that are no longer needed?

pkg/controller/migstorage/validation.go
```diff
@@ -168,73 +161,79 @@ func (r ReconcileMigStorage) validateVolumeSnapshotStorage(storage *migapi.MigSt
...
 		if secret == nil {
-			storage.Status.SetCondition(migapi.Condition{
+			storage.Status.SetAndLogCondition(log, migapi.Condition{
 				Type:     InvalidVSCredsSecretRef,
 				Status:   True,
 				Reason:   NotFound,
 				Category: Critical,
-				Message:  InvalidVSCredsSecretRefMessage,
-			})
+				Message: fmt.Sprintf("VolumeSnapshotConfig secret %s/%s not found",
```
djwhatle, line 202: These messages provide good context, I like that a lot!

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -203,6 +204,33 @@ func (r *Conditions) SetCondition(condition Condition) {
 	}
 }
 
+// Set (add/update) the specified condition to the collection.
+func (r *Conditions) SetAndLogCondition(logger logging.Logger, condition Condition, namespace, name string) {
```
djwhatle, line 208: I'm curious how this interacts with the condition systems. In at least some of our controllers, we _stage_ conditions on every reconcile. 

I'd like to know if this new logging mechanism results in a flurry of log messages on every reconcile, even if no condition changes were really made.

djwhatle (commented):
I added some questions. Thanks for taking the time to help us make logging improvements! I think if we nail this it'll really help with understanding of mig-controller activity.

pkg/apis/migration/v1alpha1/condition.go
jortel, line 208: I'm not convinced we need to log setting the condition but instead could just log all of the `Conditions` object at the end of each reconcile.  That seems simpler and accomplishes the same thing.
That said:
- The `logr.Logger` (interface) should be used instead of `logging.Logger`.
- The logger should be configured at the beginning of the Reconcile() (near log.Reset()) with `, "storage", request)` so that all log entries has that kv pair instead of being passed in.
- Why log all of the condition fields separately?  Instead of just: `logger.Info("...", "condition", condition)`
- I don't think noting that a condition being set has already exists is useful.
- This method should not be re-implementing the SetCondition() logic.  Further, it is not doing it correctly.  The LastTransitionTime should only be updated when the condition has changed.

pkg/controller/migstorage/migstorage_controller.go
jortel, line 111: The `IsNotFound` naturally occurs when the reconcile event has been queued for a deleted resource.  This is worded like an error.  Does logging this really add value?

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -107,26 +108,33 @@ func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Re
 	err = r.Get(context.TODO(), request.NamespacedName, storage)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			log.V(2).Info("unable to get migstorage, not found", "migstorage", request.String(), "error", err.Error())
 			return reconcile.Result{}, nil
 		}
-		log.Trace(err)
+		log.Trace(fmt.Errorf("unable to get migstorage: %#v", err), "migstorage", request.String())
```
jortel, line 114: The `Logger.Trace()` is intended to simply log the error and stack trace.  Logger.Error("Unable to get MigStorage.", err) would be a better fit.
Same comment here re: configuring the Logging with the _storage_ keyvalues so we don't need to explicitly include it in every log message.

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -107,26 +108,33 @@ func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Re
...
+			return
+		case errors.IsConflict(err):
+			log.Error(fmt.Errorf("unable to get migstorage: %#v", err), "migstorage", request.String())
 			return
 		}
 		storage.Status.SetReconcileFailed(err)
 		err := r.Update(context.TODO(), storage)
 		if err != nil {
-			log.Trace(err)
+			log.Trace(fmt.Errorf("unable to update migstorage status: %#v", err), "migstorage", request.String())
```
jortel, line 131: Same here re: configured logger and using Logger.Error().

pkg/controller/migstorage/migstorage_controller.go
jortel: Mainly because it's an extra set of `const` to maintain and unlike all of the other fields are only used in (1) place. Another reason is that the Message _may_ be parameterized and keeping the string and Sprintf() together would be easier to maintain.

pkg/controller/migstorage/validation.go
```diff
@@ -79,7 +67,7 @@ func (r ReconcileMigStorage) validateBackupStorage(storage *migapi.MigStorage) e
 			Status:   True,
 			Reason:   NotSet,
 			Category: Critical,
-			Message:  InvalidBSProviderMessage,
+			Message:  "spec.BackupStorageProvider empty",
```
jortel, line 70: Let's not rewrite all of the messages, please.

pkg/controller/migstorage/validation.go
```diff
@@ -105,55 +93,60 @@ func (r ReconcileMigStorage) validateBackupStorage(storage *migapi.MigStorage) e
...
 	}
 
 	// Secret
 	secret, err := storage.GetBackupStorageCredSecret(r)
 	if err != nil {
-		return liberr.Wrap(err)
+		return liberr.Wrap(fmt.Errorf("error getting migstorage backup provider secret %s/%s: %v",
+			storage.Spec.BackupStorageConfig.CredsSecretRef.Namespace,
+			storage.Spec.BackupStorageConfig.CredsSecretRef.Name,
+			err))
```
jortel, line 107: This is an unanticipated/unrecoverable error which is usually just logged as a stack trace.
I'm a little concerned about rewriting errors.  The fmt.Errorf() probably returns a new error that will not behave as expected for an errors.Is() test.
For this kind of failure, I don't see how the namespace and name is useful.

pkg/controller/migstorage/validation.go
jortel: I kind of disagree.  The condition message seem redundant with the values in the Spec.

jortel: Also, can we please use good grammar in the messages.  Begin with a capital and end with a period?

pkg/controller/migstorage/migstorage_controller.go
jortel: Can we please use good grammar in log messages?  Begin with capital and end with period.

jortel (changes requested):
The only value added by logging conditions is that is provides a history as to how that conditions have changed over time.  As noted, it seems simpler to just log the `Condition` at the end of each reconcile.
Also, as noted, configuring the Logger with kvpair for the resource being reconciled will eliminate the need to include in every log entry explicitly.
Not convinced we the rewriting of all of the condition messages adds value.  Especially since they contain duplicate information from the spec.
I'm concerned about error rewriting using fmt.Errorf().

It seems the goals of this PR could be accomplished with much less change.

pkg/controller/migstorage/validation.go
```diff
@@ -168,73 +161,79 @@ func (r ReconcileMigStorage) validateVolumeSnapshotStorage(storage *migapi.MigSt
...
 		fields := provider.Validate(secret)
 		if len(fields) > 0 {
-			storage.Status.SetCondition(migapi.Condition{
+			storage.Status.SetAndLogCondition(log, migapi.Condition{
 				Type:     InvalidVSFields,
 				Status:   True,
 				Reason:   NotSupported,
 				Category: Critical,
-				Message:  InvalidVSFieldsMessage,
+				Message:  fmt.Sprintf("secret %s/%s has invalid backing store credentials", secret.Namespace, secret.Name),
```
jortel, line 217: The original message wording seemed to describe the problem more clearly and contained a list of the invalid fields.

pkg/controller/migstorage/validation.go
alaypatel07: > I kind of disagree. The condition message seem redundant with the values in the Spec.

The goal of this is to bubble out the state that was processed in the controller. This is useful in scenarios where:

1. For some reason the mig-controller pod is not in running state.
2. User changed the spec, but the condition is still set due to the old spec value.

In the above case, as a user I expect the controller to explicitly tell me which secret it was unable to find

pkg/controller/migstorage/validation.go
alaypatel07: +1 on the list, will find a way to add that

pkg/controller/migstorage/validation.go
alaypatel07: While I do see how the namespace and name can be useful, since this is changing the behavior, I drop this change in favor of the original change anyway

pkg/controller/migstorage/migstorage_controller.go
alaypatel07: +1, I can reword the error message to reflect the update failure

pkg/controller/migstorage/migstorage_controller.go
alaypatel07: +1

pkg/controller/migstorage/migstorage_controller.go
alaypatel07: > Since the MTC operator controls the container environment and by default does not restrict logging level, is this change necessary? How would the logging level get changed resulting in suppression?

I believe turning the flag here to true https://github.com/konveyor/mig-controller/blob/ce7c0b8bc1255f991c94cc4566da176d23aefc35/cmd/manager/main.go#L41 will turn debug level logging on. My understanding on the loggin numbers was different than what I found in the vendored code. it starts with debug level at -1 https://github.com/konveyor/mig-controller/blob/3e8f9fe72c3a5a3fec64b3a0b1bbb0e7101fc544/vendor/go.uber.org/zap/zapcore/level.go#L37

> Any reason to not use good grammar in log messages? Begin with capital and end with period.

I thought that it was a convention in Go to not begin with a Capital. But TIL that's only for errors. I'll refactor this wherever applies.

pkg/apis/migration/v1alpha1/condition.go
alaypatel07: +1. I agree with most points state here, will refactor.

> Further, it is not doing it correctly. The LastTransitionTime should only be updated when the condition has changed.

I am curious why should we not add LastTransitionTime to conditions that are being added? I interpret the transition time of the CR into that condition when the condition is added.

pkg/controller/migstorage/migstorage_controller.go
djwhatle: Alay and I talked about this yesterday, that makes sense to me, 👍  on that change.

pkg/controller/migstorage/validation.go
djwhatle: Maybe instead
```
`spec.backupStorageProvider` is empty. `spec.backupStorageProvider` must be: (aws|gcp|azure).
```
Some way of guiding the user towards the correct input would be good.

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -357,3 +358,9 @@ func (r *Conditions) SetReconcileFailed(err error) {
 	})
 	r.EndStagingConditions()
 }
+
+func (r *Conditions) Log(log logging.Logger) {
+	for _, cond := range r.List {
+		log.Info("updated condition", cond)
```
alaypatel07, line 364: I would prefer this instead in favour of adding the entier condition object for readability of the logs, but not too opinionated if folks think otherwise makes more sense

pkg/controller/migstorage/migstorage_controller.go
alaypatel07: resolving this. feel free to re-open if needed

pkg/apis/migration/v1alpha1/condition.go
djwhatle: Can you give an example of what both formats (single log per condition vs all together) look like so we can get an idea?

alaypatel07:
```
{“info”, “kube-builder”, “all the other default fields”, “updated conditions”, [{“condition1 entire json”}, {“condition2 entire json”}]}
vs
{“info”, “kube-builder”, “all the other default fields”, “updated condition”, {“condition1 entire json”}}
{“info”, “kube-builder”, “all the other default fields”, “updated condition”, {“condition2 entire json”}}
```

I have not built the image but conceptually I am thinking the difference would be as above

djwhatle: I think I tend to agree with you that one log per condition will be more readable.

djwhatle (approved):

pkg/apis/migration/v1alpha1/condition.go
```diff
@@ -357,3 +358,9 @@ func (r *Conditions) SetReconcileFailed(err error) {
 	})
 	r.EndStagingConditions()
 }
+
+func (r *Conditions) Log(log logging.Logger) {
```
jortel, line 362: Why is the method needed?

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -101,27 +101,34 @@ type ReconcileMigStorage struct {
 func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	var err error
 	log.Reset()
+	log.SetValues("storage", request)
```
jortel, line 104: :+1:

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -101,27 +101,34 @@ type ReconcileMigStorage struct {
...
 	// Fetch the MigStorage instance
 	storage := &migapi.MigStorage{}
 	err = r.Get(context.TODO(), request.NamespacedName, storage)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			log.V(-1).Info("Unable to get migstorage, not found.")
 			return reconcile.Result{}, nil
 		}
-		log.Trace(err)
+		log.Error(err, "Unable to get migstorage.")
```
jortel, line 114: How is the stack trace not sufficent?

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -101,27 +101,34 @@ type ReconcileMigStorage struct {
...
 
 	// Report reconcile error.
 	defer func() {
-		if err == nil || errors.IsConflict(err) {
+		storage.Status.Conditions.Log(log)
+		switch {
+		case err == nil:
+			return
+		case errors.IsConflict(err):
+			log.Error(err, "Unable to update migstorage.")
```
jortel, line 125: `Conflict` errors are routine and occur often.  We made a deliberate decision not to log them.  Especially with a stack trace.

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -101,27 +101,34 @@ type ReconcileMigStorage struct {
...
 			return
 		}
 		storage.Status.SetReconcileFailed(err)
 		err := r.Update(context.TODO(), storage)
 		if err != nil {
-			log.Trace(err)
+			log.Error(err, "Unable to update migstorage status.")
```
jortel, line 131: Same here.  How is the trace not sufficient?

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -139,7 +146,7 @@ func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Re
 	// Ready
 	storage.Status.SetReady(
 		!storage.Status.HasBlockerCondition(),
-		ReadyMessage)
+		"The storage is ready.")
```
jortel, line 149: :+1:

pkg/controller/migstorage/migstorage_controller.go
```diff
@@ -152,6 +159,8 @@ func (r *ReconcileMigStorage) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	storage.Status.Conditions.Log(log)
```
jortel, line 162: Why not just something like:
Log.Info("Conditions", "cnd", storage.Status.Conditions)

pkg/controller/migstorage/validation.go
```diff
@@ -123,7 +112,8 @@ func (r ReconcileMigStorage) validateBackupStorage(storage *migapi.MigStorage) e
 			Status:   True,
 			Reason:   NotFound,
 			Category: Critical,
-			Message:  InvalidBSCredsSecretRefMessage,
+			Message: fmt.Sprintf("The `backupStorageConfig.credsSecretRef` must reference a valid `secret`, context %s",
```
jortel, line 115: I know I recommended something like _content_ but after seeing it, I wonder if `key` would be better?

alaypatel07: yeah I like instead of `key` instead of `content` or `context`, +1

pkg/controller/migstorage/migstorage_controller.go
alaypatel07: that would provide one single log entry with all the conditions, I actually like to have a separate entry per condition to make it more readable. 

https://github.com/konveyor/mig-controller/pull/698#discussion_r511091777

pkg/controller/migstorage/migstorage_controller.go
alaypatel07: This needs to be dropped

pkg/controller/migstorage/migstorage_controller.go
alaypatel07: needs to be dropped

pkg/controller/migstorage/migstorage_controller.go
alaypatel07: make sense. will drop this

pkg/controller/migstorage/migstorage_controller.go
jortel: I think a single log entry is desirable because it is simple and more importantly keeps all of the conditions together.  With 6-7 controller logging, it will be easy for the list of conditions to be interlaced with other log entries and hard to read.  Also, if grammatically inspecting/digesting the logs, multiple entries will be more complex to correlate.

alaypatel07:
```
With 6-7 controller logging, it will be easy for the list of conditions to be interlaced with other log entries and hard to read. 
```
With an increase in verbosity, devs will have to find a way to grep for CR's that they are looking for and interlacing logs among the CR is hard to see. if I am looking for storage controller logs I would almost always grep with `storage\|<cr-name>` to construct a timeline, I think with that in mind, a separate per line comment is more readable. I would do it like that, if you feel strongly, I can remove the method and log the entire blob of conditions inline

jortel (approved):
LGTM

shubham-pampattiwar (approved):

djwhatle (approved):
ACK. Event stuff looks cool!

PR #705, opened 2020-10-20
Add StorageClasses to discovery service

djwhatle:
This is related to #651 . The UI changes needed to read StorageClass data from the discovery service instead of from Mig resources are large enough that I won't be able to complete them this sprint. 

This is just a copy-paste of the discovery service files for another cluster-scoped resource (`PVs`) with var names changed. @jortel I can tell you that this builds and seems to work correctly when mig-controller is started, but I haven't been able to validate that it's working beyond that. Not entirely sure how to do so.

### DIFF
--- a/pkg/controller/discovery/container/storageclass.go
+++ b/pkg/controller/discovery/container/storageclass.go
@@ -1,0 +1,157 @@
+package container
+
+import (
+	"context"
+	"time"
+
+	"github.com/konveyor/mig-controller/pkg/controller/discovery/model"
+	v1 "k8s.io/api/storage/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// A collection of k8s StorageClass resources.
+type StorageClass struct {
+	// Base
+	BaseCollection
+}
+
+func (r *StorageClass) AddWatch(dsController controller.Controller) error {
+	err := dsController.Watch(
+		&source.Kind{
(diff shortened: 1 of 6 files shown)


jortel (approved):
LGTM

jortel (changes requested):
LGTM
You need to add the StorageClassHandler to the web/web.go.
Please be sure to test using curl.  You can run the controller with AUTH_OPTIONAL=1 so you wont need to pass the token.

djwhatle:
Thanks for the tips @jortel . I was able to verify that everything is working after making the change to web.go as you mentioned.

```
# Result from OCP 4.5 cluster, as expected
curl http://localhost:8080/namespaces/openshift-migration/clusters/host/storageclasses

{"count":1,"resources":[{"name":"gp2","selfLink":"/namespaces/openshift-migration/clusters/host/storageclasses/gp2","object":{"metadata":{"name":"gp2","selfLink":"/apis/storage.k8s.io/v1/storageclasses/gp2","uid":"89397fad-9897-4e0a-a208-e467b662dd2f","resourceVersion":"11521","creationTimestamp":"2020-10-06T22:18:14Z","annotations":{"storageclass.kubernetes.io/is-default-class":"true"},"ownerReferences":[{"apiVersion":"v1","kind":"clusteroperator","name":"storage","uid":"57005cb1-7d03-40d7-8013-8547c2b8bf1f"}]},"provisioner":"kubernetes.io/aws-ebs","parameters":{"encrypted":"true","type":"gp2"},"reclaimPolicy":"Delete","allowVolumeExpansion":true,"volumeBindingMode":"WaitForFirstConsumer"}}]}


# Result from OCP 3.11 cluster, as expected
curl http://localhost:8080/namespaces/openshift-migration/clusters/o311/storageclasses

{"count":0,"resources":null}
```

jortel (approved):
Well done.

PR #710, opened 2020-10-22
MIG-284 - First phase work for direct image migration controller

sseago:
This creates a controller for MigImageMigration which migrates internal imagestreams
from a set of namespaces on the src cluster to a set of namespaces on the dest cluster.
There's also a secondary controller, created by the first, which manages the migration
of internal images in a single imagestream from the source to destination clusters.

There is not yet any integration with the migplan/migmigration workflow.

These changes should be pushed along with https://github.com/konveyor/mig-operator/pull/478


sseago:
I just realized that I"m not setting owner references on MigImageStreamMigration. This should be owned by the MigImageMigration when the MigImageMigration is creating it. I'll update the PR shortly to include that.

config/crds/migration_v1alpha1_directimagestreammigration.yaml
```diff
@@ -0,0 +1,62 @@
...
+metadata:
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: directimagestreammigrations.migration.openshift.io
+spec:
+  group: migration.openshift.io
+  names:
+    kind: DirectImageStreamMigration
+    plural: directimagestreammigrations
```
alaypatel07, line 12: need to add short name here

alaypatel07: same both CRDs

pkg/apis/migration/v1alpha1/directimagemigration_types.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (r *DirectImageMigration) GetDestinationCluster(client k8sclient.Client) (*MigCluster, error) {
+	return GetCluster(client, r.Spec.DestMigClusterRef)
+}
+
+// GetSourceNamespaces get source namespaces without mapping
+func (r *DirectImageMigration) GetSourceNamespaces() []string {
+	includedNamespaces := []string{}
+	for _, namespace := range r.Spec.Namespaces {
+		namespace = strings.Split(namespace, ":")[0]
```
alaypatel07, line 79: is this because `:` is not allowed in the name of a namespace?

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -498,3 +502,89 @@ func (r *MigCluster) UpdateProvider(provider pvdr.Provider) {
...
+			},
+			&registrySvc)
+		if err != nil {
+			// Return empty registry host but no error; registry not found
+			return "", nil
+		}
+		internalRegistry := registrySvc.Spec.ClusterIP + ":" + strconv.Itoa(int(registrySvc.Spec.Ports[0].Port))
+		return internalRegistry, nil
+	} else {
+		config := kapi.ConfigMap{}
```
alaypatel07, line 545: I think a more authoritative place to get this would be from the apiserver observed config

```
$ oc get openshiftapiservers.operator.openshift.io -oyaml | grep image
            f:imagePolicyConfig: {}
      imagePolicyConfig:
        internalRegistryHostname: image-registry.openshift-image-registry.svc:5000
```

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -498,3 +502,89 @@ func (r *MigCluster) UpdateProvider(provider pvdr.Provider) {
...
+	Subdomain string `json:"subdomain"`
+}
+type imagePolicyConfig struct {
+	InternalRegistryHostname string `json:"internalRegistryHostname"`
+}
+
+// apiServerConfig stores configuration information about the current cluster
+type apiServerConfig struct {
+	ImagePolicyConfig imagePolicyConfig `json:"imagePolicyConfig"`
+	RoutingConfig     routingConfig     `json:"routingConfig"`
```
alaypatel07, line 589:
```
$ oc get openshiftapiservers.operator.openshift.io -oyaml | grep routing -C 2
            f:imagePolicyConfig: {}
            f:projectConfig: {}
            f:routingConfig: {}
            f:servingInfo: {}
            f:storageConfig: {}
--
      projectConfig:
        projectRequestMessage: ""
      routingConfig:
        subdomain: apps.cluster-alpatel-aux-tools-444.alpatel-aux-tools-444.mg.dog8code.com
      servingInfo:
[alpatel@alpatel migsamples]$
```

alaypatel07: Same as above, I think this might be a better place to get the info

pkg/apis/migration/v1alpha1/directimagemigration_types.go
sseago: It's because the namespace mapping is delimited by ":"
If the value is `["ns1", "ns2:ns3"]`, then imagestreams in `ns1` will be migrated to the same dest namespace, `ns1`, and imagestreams in `ns2` will be migrated to `ns3` -- this is the same format used for the namespace list on the migplan.

pkg/apis/migration/v1alpha1/migcluster_types.go
sseago: Currently this is doing it in the same way the plugin handles it. If we do end up needing to make changes, we should probably make similar changes there.
@dymurray We went through several options trying to find the most reliable way of doing this when we added this to the plugin last year. Do you remember whether we considered this other option as well?

config/crds/migration_v1alpha1_directimagestreammigration.yaml
sseago: Sure. I'll add it. Perhaps, separately, we should do the same for our other CRDs, as none of them have short names. I think we'd decided for my new CRDs that I should use "dism" for this one and "dim" for `DirectImageMigration`.

sseago: Done.

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -174,3 +175,28 @@ func GetSecret(client k8sclient.Client, ref *kapi.ObjectReference) (*kapi.Secret
...
+	}
+	object := imagev1.ImageStream{}
+	err := client.Get(
+		context.TODO(),
+		types.NamespacedName{
+			Namespace: ref.Namespace,
+			Name:      ref.Name,
+		},
+		&object)
+	if err != nil {
```
alaypatel07, line 193: I wonder if using a switch will be more readable
```
switch {
    case errs.IsNotFound(err):
        return nil, nil
    case err != nil:
         return nil, err
}
```

pkg/controller/directimagemigration/validate.go
```diff
@@ -0,0 +1,273 @@
...
+	if count == 0 {
+		imageMigration.Status.SetCondition(migapi.Condition{
+			Type:     NsListEmpty,
+			Status:   True,
+			Category: Critical,
+			Message:  NsListEmptyMessage,
+		})
+		return nil
+	}
+	cluster, err := imageMigration.GetSourceCluster(r)
```
alaypatel07, line 236: rename to `srcCluster` ?

pkg/controller/directimagemigration/validate.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+	if cluster == nil || !cluster.Status.IsReady() {
+		return nil
+	}
+	client, err := cluster.GetClient(r)
```
alaypatel07, line 243: rename to `srcClient`?

pkg/controller/directimagemigration/validate.go
```diff
@@ -0,0 +1,273 @@
...
+		} else {
+			return liberr.Wrap(err)
+		}
+	}
+	if len(notFound) > 0 {
+		imageMigration.Status.SetCondition(migapi.Condition{
+			Type:     NsNotFoundOnSourceCluster,
+			Status:   True,
+			Reason:   NotFound,
+			Category: Critical,
```
alaypatel07, line 265: If there are 10 items in the list of namespaces and 1 namespace is not present, I would expect the controller to migrate imagestreams for other 9 namespaces, two questions:

1. Is that a reasonable expectation?
2. If the above answer is yes, looks like the code is not doing that currently because Critical condition will block running the itinerary. Should this be modified?

pkg/apis/migration/v1alpha1/model.go
sseago: I think that's a bit more confusing, and it leaves us dependent on errors.IsNotFound() returning `false` when the input arg is nil, rather than either true or blowing up because it expects input not to be nil. Generally it's best to check for nil before using the value.

pkg/controller/directimagestreammigration/validate.go
```diff
@@ -0,0 +1,306 @@
...
+)
+
+// Statuses
+const (
+	True  = migapi.True
+	False = migapi.False
+)
+
+// Validate the image migration resource
+func (r ReconcileDirectImageStreamMigration) validate(imageStreamMigration *migapi.DirectImageStreamMigration) error {
```
alaypatel07, line 89: A general question about all critical conditions added in this function/sub-functions.

If there is a critical condition added, when will the controller requeue this CR?

It looks like the current code only requeues on unexpected errors, https://github.com/konveyor/mig-controller/pull/710/commits/99fb253d8d8510038f614ae61e3a43b6d5ccf743#diff-775a50c0d704007e87759e873ac5f70e4584cd421209ee5214478461c18b3848R136

If you agree that we need to requeue on critical conditions, there might be two options:
1) Either we make all critical conditions return errors so we requeue
2) we requeue any critical condition present on the CR at the end of the reconcile function.

pkg/controller/directimagemigration/validate.go
sseago: I could do that. We've generally been just using cluster/client when a function only references one, but maybe it's clearer to spell it out here as well.

pkg/controller/directimagemigration/validate.go
sseago: I was following the current behavior of the migplan validation -- if we say "migrate a namespace" and that namespace doesn't exist, then we fail validation before we start, assuming the user made a mistake, possibly a namespace name typo. If we remove "namespace existence" validation and choose to filter instead, that would work, but it changes the behavior in a way that needs to be documented (although, really, none of this is documented now). If I migrate a single namespace and that namespace doesn't exist, does this mean that it's a successful (no-op) migration? Failing on validation before running feels like the "fail early" approach, letting the user know now that a namespace they think they're migrating doesn't exist, rather than silently omitting it, and they find out at the end that something expected to be there isn't.

pkg/controller/directimagestreammigration/validate.go
sseago: That's a question for @djwhatle or @jortel I think.

pkg/apis/migration/v1alpha1/directimagemigration_types.go
alaypatel07: I did not know that, TIL, thanks!

pkg/apis/migration/v1alpha1/model.go
alaypatel07: The `IsNotFound` does return false on nil, and I have seen this to be a pattern adopted across the board.

> Generally it's best to check for nil before using the value.

Unfortunately, I disagree there. For Go specifically, the code is supposed to be organized so you have the least amount of indents, nested ifs are non-starter for me in general

pkg/controller/directimagemigration/validate.go
alaypatel07: yeah, I have been tripped on reading that code, always need to backtrack to where the variable is defined to understand what client the call is made against. Would be nice to have this explicitness

pkg/controller/directimagemigration/validate.go
alaypatel07: I do not mean to suggest silently omitting, we definitely want to show the namespaces that do not exist, but would that block migration of other namespaces if my question?

In general, I am not keeping the practices adopted in migplan controller while reviewing this controller PR. In some cases, the migplan controller is intended to serve the UI and our practices evolved with that need. This new controller is more internal and would set my expectations a little differently

sseago: Right. I mention the migplan controller as the starting point not because we can't change it, but because it's the starting point. If there's a good reason to do differently, we can. It seems to me that failing early and letting the user either remove or fix the nonexistent namespace is the more intuitive approach, but we could handle it differently. If I'm understanding it, the basic part of your suggestion is to remove this from the pre-migration validation and to treat it as a runtime warning or error. Instead of blocking migration completely, we'd let a migration go and (depending on the next decision), either add a warning condition for missing namespaces, or fail the migration over them (but still migrating the others). We could even add a spec flag to determine this behavior, but that may be overkill (i.e. one setting ensures src namespaces in pre-validation, and the other would just warn/fail at the end if they're not there. Then we'd need to decide which behavior would be the default.

sseago: If we do decide to move namespace checks from pre-flight validation to during-migration checks, we'll also need to account for missing src namespaces in the "create namespaces" step. We'd want to skip creating dest namespaces if the src namespace doesn't exist.

sseago: @dymurray Which option do you think makes more sense here? Fail early during validation phase if there's a missing src namespace, or skip namespace validation and fail the migration on a missing namespace (but still migrating the others), or just succeeding and adding a warning for missing namespaces?

sseago: @dymurray I guess similar concerns would apply for `DirectVolumeMigration` if the src PV doesn't exist. Does that create a critical condition from validation, or do you migrate everything else and warn/fail on it?

pkg/apis/migration/v1alpha1/model.go
sseago: I can change this. @djwhatle is this a change that makes sense for the other mig-controller calls in this same file? If I'm going to use the switch/case approach for the new one, I'd rather make the old ones (GetPlan, etc.) consistent with this.

pkg/apis/migration/v1alpha1/directimagestreammigration_types.go
```diff
@@ -0,0 +1,128 @@
...
+	return GetCluster(client, r.Spec.DestMigClusterRef)
+}
+
+func (r *DirectImageStreamMigration) GetImageStream(c k8sclient.Client) (*imagev1.ImageStream, error) {
+	cluster, err := r.GetSourceCluster(c)
+	if err != nil {
+		return nil, err
+	}
+	if !cluster.Status.IsReady() {
+		return nil, errors.New("Source cluster not ready")
```
alaypatel07, line 83: Probably worth defining the error type for "Source cluster not ready" could see how it can be used at a lot of places in mig-controller code

pkg/controller/directimagestreammigration/copyimage.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/konveyor/openshift-velero-plugin/velero-plugins/imagecopy"
+)
+
+func (t *Task) migrateInternalImages() error {
+	imageStream, err := t.Owner.GetImageStream(t.Client)
+	if err != nil {
+		return err
+	}
+	if imageStream == nil {
+		return errors.New("ImageStream not found")
```
alaypatel07, line 18: I would actually return and apiserver NotFound error here, with object imagestream name and namespace embeded

alaypatel07: well after thinking some more, this does not make much sense, we are catching the NotFound error in GetImageStream function and then creating a new error here. If the aim is to bubble up the NotFound error, I wonder if we should return the original error itself?

sseago: Good point. I was following the prior pattern, which makes sense for a general-purpose "find" method where not finding it isn't an error and `nil` is what you want, but I'm only using this in one place, and here, "not found" is an error, so I should probably skip the checking for IsNotFound there, and then here I won't need to check for nil at all.

pkg/controller/directimagemigration/validate.go
dymurray: I think this should fail validation. Since we know its an invalid configuration we want to catch it as early as possible and prevent the user from migrating.

pkg/apis/migration/v1alpha1/migcluster_types.go
dymurray: I think it makes sense to grab this from the CR. Seems to be the source of truth.

pkg/apis/migration/v1alpha1/model.go
sseago: Relating to discussion below, I'm rewriting this to just propagate the Not Found error, since in the situation we're using it, not finding the imagestream is an error condition

pkg/controller/directimagemigration/validate.go
sseago: Talking to @dymurray and @alaypatel07 we're leaving this as-is for now, but we may need to revisit it once the controller redesign is completed.

pkg/controller/directimagestreammigration/copyimage.go
sseago: Ive updated `GetImageStream` to propagate all errors, including not found errors, so I just removed this nil check and error message return.

pkg/apis/migration/v1alpha1/model.go
alaypatel07: make sense thanks!

pkg/controller/directimagemigration/directimagemigration_controller.go
```diff
@@ -0,0 +1,174 @@
...
+
+	return nil
+}
+
+var _ reconcile.Reconciler = &ReconcileDirectImageMigration{}
+
+// ReconcileDirectImageMigration reconciles a DirectImageMigration object
+type ReconcileDirectImageMigration struct {
+	client.Client
+	record.EventRecorder
```
alaypatel07, line 105: +1

pkg/controller/directimagemigration/imagestream.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+func (t *Task) getDirectImageStreamMigrations() ([]migapi.DirectImageStreamMigration, error) {
+	ismList := migapi.DirectImageStreamMigrationList{}
+	err := t.Client.List(
+		context.TODO(),
+		k8sclient.MatchingLabels(t.Owner.GetCorrelationLabels()),
+		&ismList)
+	if err != nil {
+		return nil, err
```
alaypatel07, line 36: I wonder whats the value in sending a nil return value, worst case if a developer does not guard against that, the code will panic. I would think `return ismList, err` is a better way, the user of this function will simple check for error and iterate on the list. 

Ohh that also reminds me, we need to replace all `return err` with `return liberr.Wrap(err)`?

pkg/controller/directimagemigration/imagestream.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+	return nil
+}
+func (t *Task) buildDirectImageStreamMigration(is imagev1.ImageStream, destNsName string) migapi.DirectImageStreamMigration {
+	labels := t.Owner.DirectImageStreamMigrationLabels(is)
+	imageStreamMigration := migapi.DirectImageStreamMigration{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:       labels,
+			GenerateName: t.Owner.GetName() + "-",
```
alaypatel07, line 83: Is there a reason why we need to use the generate name here and not use the same names as the imagestream?

I think we will have a more deterministic name with imagestream name and the list call here https://github.com/konveyor/mig-controller/pull/710/commits/18c8cb613353eb9db1024ab54811de43a805787c?file-filters%5B%5D=.go#diff-1ae7521ab52a30a2c3580aef36d5e819a554960ce543888a1d96548af27e0adfR50 can be replaced with a get call. (More performant as well I think not by much though)

pkg/controller/directimagemigration/imagestream.go
```diff
@@ -0,0 +1,120 @@
...
+		err := srcClient.List(
+			context.TODO(),
+			k8sclient.InNamespace(srcNsName),
+			&isList)
+		if err != nil {
+			return err
+		}
+		for _, is := range isList.Items {
+			ismList := migapi.DirectImageStreamMigrationList{}
+			err := t.Client.List(
```
alaypatel07, line 59: I wonder if making this a Get call instead makes sense

pkg/controller/directimagemigration/imagestream.go
alaypatel07: it would also be easier to debug, seeing the same imagestream name

sseago: I was talking to @dymurray about this before I did this. There are two problems. First, we'd have to combine name and namespace into the name, which could run into length issues. Second, it would make it impossible to create a second CR -- say you did another migration to another cluster after the first. generateName eliminates these problems, because it isolates the CR to this particular migration. We can't hit name clashes because an already-finished (or even a currently-running) CR already exists with this name.

pkg/controller/directimagemigration/imagestream.go
```diff
@@ -0,0 +1,120 @@
...
+	for i := range dism.OwnerReferences {
+		ref := &dism.OwnerReferences[i]
+		if ref.Kind == t.Owner.Kind {
+			ref.APIVersion = t.Owner.APIVersion
+			ref.Name = t.Owner.Name
+			ref.UID = t.Owner.UID
+			return
+		}
+	}
+	dism.OwnerReferences = append(
```
alaypatel07, line 112: I wonder if there should be more owner references than just the t.Owner?

We could replace this with

```
dism.OwnerReferences = []metav1.OwnerReference{
...
...
}
```

A list with just one entry

pkg/controller/directimagemigration/imagestream.go
sseago: Doesn't the Get call require you to know the name? This is only used to make sure we don't create a new CR if one already exists for this `dim` and for this `is`, using label selectors to grab it. Because we need to use generateName (see above), we don't know the name.

pkg/controller/directimagemigration/imagestream.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (t *Task) createDirectImageStreamMigrations() error {
+	// Get client for source
+	srcClient, err := t.getSourceClient()
+	if err != nil {
+		return err
+	}
+	// Get list namespaces to iterate over
+	for srcNsName, destNsName := range t.Owner.GetNamespaceMapping() {
+		isList := imagev1.ImageStreamList{}
```
alaypatel07, line 49: I wonder if we should separate this list call from this method. Since this is `a one shot` run controller, we would want to have separate phases for:
1. Getting the list of imagestreams we commit to migrating
2. Reading the committed list and create dism resources for them

This is to prevent the scenario where we error out creating one or more dism and during the next reconcile loop we find more resources then the initial reconcile loop

pkg/controller/directimagemigration/imagestream.go
sseago: From what I remember, this seems to be what most of the kubernetes code I've interacted with does. Maybe there are other patterns as well. I know that, in general, if something returns an item and an error, if the error isn't nil, my code calling it can't rely on the item existing, and it will usually be ignored anyway. I guess there's no harm in just returning it, but the inconsistency might cause confusion. As for `liberr.Wrap(err)`, the function that this is returning to (in task.go) does that wrapping, so I didn't think it was needed here.

pkg/controller/directimagemigration/imagestream.go
sseago: Yeah, I guess this is attempting to be more generic than is really necessary (taken from the migmigration controller). It's probably simpler to just assign a single one here.

pkg/controller/directimagestreammigration/validation.go
```diff
@@ -0,0 +1,298 @@
...
+	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
+	migref "github.com/konveyor/mig-controller/pkg/reference"
+	kapi "k8s.io/api/core/v1"
+	k8serror "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// Types
+const (
+	InvalidSourceClusterRef               = "InvalidSourceClusterRef"
```
alaypatel07, line 35: probably needs to be removed?

pkg/controller/directimagemigration/validation.go
```diff
@@ -0,0 +1,265 @@
...
+	NsNotFoundOnSourceCluster             = "NamespaceNotFoundOnSourceCluster"
+
+	Running   = "Running"
+	Failed    = "Failed"
+	Succeeded = "Succeeded"
+)
+
+// Reasons
+const (
+	NotFound    = "NotFound"
```
alaypatel07, line 53: Another general issue, sorry on picking this PR, these conts should probably be in the common-lib and imported into all the controllers, instead of being declared again and again

pkg/controller/directimagemigration/validation.go
```diff
@@ -0,0 +1,265 @@
...
+	NotSet      = "NotSet"
+	NotDistinct = "NotDistinct"
+	NotReady    = "NotReady"
+)
+
+// Categories
+const (
+	Critical = migapi.Critical
```
alaypatel07, line 61: also, why redeclare again? we can use the `migapi.Critical` everywhere Critical can be use right?

pkg/controller/directimagemigration/validation.go
```diff
@@ -0,0 +1,265 @@
...
+	Advisory = migapi.Advisory
+)
+
+// Statuses
+const (
+	True  = migapi.True
```
alaypatel07, line 67: same comment regarding reuing `migapi.True`

pkg/controller/directimagemigration/validation.go
```diff
@@ -0,0 +1,265 @@
...
+	ref := imageMigration.Spec.SrcMigClusterRef
+
+	// Not Set
+	if !migref.RefSet(ref) {
+		imageMigration.Status.SetCondition(migapi.Condition{
+			Type:     InvalidSourceClusterRef,
+			Status:   True,
+			Reason:   NotSet,
+			Category: Critical,
+			Message:  "The directImageMigration.srcMigClusterRef must reference name and namespace for a valid `MigCluster`",
```
alaypatel07, line 99: NIT: I would rather use `spec.srcMigClusterRef` because directImageMigration is already the name of the CR when user does a -oyaml to get these conditions

pkg/controller/directimagemigration/validation.go
```diff
@@ -0,0 +1,265 @@
...
+		})
+		return nil
+	}
+
+	// Not ready
+	if !cluster.Status.IsReady() {
+		imageMigration.Status.SetCondition(migapi.Condition{
+			Type:     SourceClusterNotReady,
+			Status:   True,
+			Category: Critical,
```
alaypatel07, line 127: NIT: the reason field is missing in the condition? I'm not sure if this is intentional...

pkg/controller/directimagestreammigration/validation.go
sseago: I'm guessing you mean the next line -- `InvalidSourceCluster` which doesn't seem to be used anywhere. I"m removing it. The above is still used.

pkg/controller/directimagemigration/validation.go
sseago: Not intentional. Apparently the migplan validation code I pulled this from was missing it. I'll add it here.

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -498,3 +502,89 @@ func (r *MigCluster) UpdateProvider(provider pvdr.Provider) {
...
+	client, err := m.GetClient(c)
+	if err != nil {
+		return "", err
+	}
+	isList := imgapi.ImageStreamList{}
+	err = client.List(
+		context.TODO(),
+		k8sclient.InNamespace("openshift"),
+		&isList)
+	if err == nil && len(isList.Items) > 0 {
```
alaypatel07, line 516: I wonder if this would give undesired results? We want the logic to be something closer to the following right?

1. get the imagestreamlist
2. if err != nil return "", err because you don't deterministically know why the error occurred
3 if len(isList.Items > 0) have the parsing logic

pkg/controller/directimagemigration/directimagemigration_controller.go
```diff
@@ -0,0 +1,174 @@
...
+		},
+		//		&ClusterPredicate{}
+	)
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to DirectImageStreamMigrations
+	err = c.Watch(
+		&source.Kind{Type: &migapi.DirectImageStreamMigration{}},
```
alaypatel07, line 84: hrm, since you are using ownerReferences, you can probably use the built-ins to generate this watch

```
	err = c.Watch(&source.Kind{Type: &migapi. DirectImageStreamMigration{}}, &handler.EnqueueRequestForOwner{
		IsController: true,
		OwnerType:    &migapi. DirectImageMigration{},
	})
```

pkg/controller/directimagemigration/imagestream.go
alaypatel07: +1 on the rework, reads very nice and seems more declarative

alaypatel07 (approved):
I just had a small nit about error handling, and also you might need to rebase. Apart from that this LGTM. 
 
Awesome work, thanks!

pkg/apis/migration/v1alpha1/migcluster_types.go
djwhatle: Question: does this change force us to run the controller on OCP 4?  I doubt `openshiftapiservers.operator.openshift.io` exists on OCP 3

pkg/apis/migration/v1alpha1/migcluster_types.go
djwhatle: Same question about if using the CR to get this info forces us to run on OCP 4. Maybe we are at a point where we should drop OCP 3 support for running controller but want to make sure we are consciously making that decision.

pkg/apis/migration/v1alpha1/model.go
djwhatle: @alaypatel07 
> For Go specifically, the code is supposed to be organized so you have the least amount of indents, nested ifs are non-starter for me in general

Source? I agree on reducing indents, but I haven't seen this declared in the style guide https://golang.org/doc/effective_go.html

I think it's probably situation dependent. I tried to view the changes here but the GitHub review interface doesn't handle 5000+ file changes well.

pkg/controller/directimagestreammigration/validate.go
djwhatle: This logic may differ across each controller. I can't give a generic answer without looking at each controller's reconcile func. The migmigration controller requeues itself when it has a crit condition.

djwhatle (commented):
Added some questions about using CRs that probably don't exist on OCP 3.

pkg/apis/migration/v1alpha1/migcluster_types.go
sseago: So I *think* this logic is the way we want it as-is -- @dymurray I think you wrote this originally for the plugin, so you might be able to comment further here.

Basically if we can determine the value from an existing imagestream, then that's how we determine the internal registry path. If we can't do this, we fall back to the ConfigMap for OCP4+ (although we discussed changing that in the future to a different lookup), and the Cluster IP for OCP3. 

So the question is what happens if the is LIst call errors out. Currently that also goes to the same fallback we use when there aren't any imagestreams in the openshift namespace. I think that's probably the better choice, since it gives us the same sort of result we'd get if there weren't any imagestreams. I don't know that erroring out instead improves anything, since we have a fallback here, but I don't have a strong opinion here. Lets see what @dymurray thinks.

pkg/controller/directimagemigration/directimagemigration_controller.go
sseago: Done.

pkg/apis/migration/v1alpha1/migcluster_types.go
sseago: @djwhatle This is code we'll only run if it's OCP4 (there's a check that MinorVersion is 12 or above) -- the ocp3 fallback is just using the cluster IP, since the equivalent in OCP3 isn't available via API at all. Note that both of these are fallbacks if the primary means of lookup fails -- if there's at least one imagestream in the "openshift" namespace, we parse the registry location from that metadata.

pkg/apis/migration/v1alpha1/migcluster_types.go
sseago: So I don't think we're dropping OCP3 support yet since we're only executing this if we're on OCP4, unless I'm missing something.

pkg/apis/migration/v1alpha1/model.go
sseago: @djwhatle in any case this particular instance went away as a concern, as I'm now propagating all errors, including `IsNotFound` based on review comments elsewhere. But yeah, we should probably be clear on what our recommended style is for this sort of thing.

pkg/apis/migration/v1alpha1/migcluster_types.go
dymurray: @sseago @alaypatel07 I agree with Scott's comments that I prefer falling back rather than erroring out. I understand the perspective that we'd want to bubble the error back since we don't know what caused it but if we have a fallback mechanism for errors or an empty list I think we'd want to attempt that before erroring out completely.

pkg/apis/migration/v1alpha1/model.go
alaypatel07: ack, that concern went away but, the effective_go blog describes in a very neutral way for which I used stronger words. But I am thinking they are trying to express the same thing if you read between the lines :) 

https://golang.org/doc/effective_go.html#if

```
This is an example of a common situation where code must guard against a sequence of error conditions. The code reads well if the successful flow of control runs down the page, eliminating error cases as they arise. Since error cases tend to end in return statements, the resulting code needs no else statements.
``` 

Additionally, if you google "go best practices" you would find multiple talks with points that say "reduce nesting and return fast"

djwhatle (approved):

PR #711, opened 2020-10-22
MIG-284 - Phase 1 implementation of direct PVC migration

dymurray:
TODO: Be more resilient to errors if existing cert secrets are laying around
TODO: Use correlation labels
TODO: Be more resilient with SSH keys by using a secret

RBAC changes: https://github.com/konveyor/mig-operator/pull/477


pkg/controller/migdirect/rsync.go
```diff
@@ -0,0 +1,827 @@
...
+
+	// Create rsync transfer pod on destination
+
+	// Create rsync client pod on source
+}
+
+// Generate SSH keys to be used
+// TODO: Need to determine if this has already been generated and
+// not to regenerate
+func (t *Task) generateSSHKeys() error {
```
djwhatle, line 111: Seems like a good thing here could be to generate these and put them in a secret correlated to the MigDirect. Then if the secret already exists just read from there.

pkg/controller/migdirect/rsync.go
```diff
@@ -0,0 +1,827 @@
...
+			containers = append(containers, corev1.Container{
+				Name:  fmt.Sprintf("rsync-client-%s", vol),
+				Image: "quay.io/konveyor/rsync-transfer:latest",
+				Env: []corev1.EnvVar{
+					{
+						Name:  "RSYNC_PASSWORD",
+						Value: "changeme",
+					},
+				},
+				Command: []string{"rsync", "-aPvvHh", "--delete", "--port", "2222", "--log-file", "/dev/stdout", fmt.Sprintf("/mnt/%s/%s/", ns, vol), fmt.Sprintf("rsync://root@%s/%s", ip, vol)},
```
djwhatle, line 614: I was reading an article the other day that was making an argument for always using long-form CLI flags when scripting _for understandability_ instead of the compressed `-avPvvHh` flag sets that you would use to _save time when typing_ out a command. 

I know we're using this in pvc-migrate and probably want to maintain continuity so I'm not going to say we need to change it here, just an interesting thought I wanted to share.

The equivalent flag set would be:
```
-aPvvHh
--archive --verbose --progress --human-readable --hard-links --vv
```

pkg/controller/migdirect/rsync.go
```diff
@@ -0,0 +1,827 @@
...
+
+func (t *Task) haveRsyncClientPodsCompleted() (bool, error) {
+	// Get client for destination
+	srcClient, err := t.getSourceClient()
+	if err != nil {
+		return false, err
+	}
+	selector := labels.SelectorFromSet(map[string]string{
+		"app":       "migdirect-rsync-transfer",
+		"migdirect": "rsync-client",
```
djwhatle, line 665: Same comment here about using correlation labels instead of generically tying back to all MigDirect CRs

pkg/controller/migdirect/stunnel.go
```diff
@@ -0,0 +1,495 @@
...
+		return err
+	}
+
+	subj := pkix.Name{
+		CommonName:         "openshift.io",
+		Country:            []string{"US"},
+		Province:           []string{"NC"},
+		Locality:           []string{"RDU"},
+		Organization:       []string{"Migration Engineering"},
+		OrganizationalUnit: []string{"IT"},
```
djwhatle, line 243: C'mon guys you know we are in the Eng BU. Sounds like a good plan though to delegate all stunnel cert management tasks over to IT 👍

pvc.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
djwhatle, line 1: I'm guessing this file got committed by mistake

djwhatle (approved):

pkg/controller/migdirect/pvcs.go
```diff
@@ -0,0 +1,76 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (t *Task) getDestinationStorageClass(sourceStorageClass string) string {
+	// TODO: Set default
+	defaultStorageClass := "gp2"
+	if destStorageClass, ok := t.Owner.Spec.StorageClassMapping[sourceStorageClass]; ok {
```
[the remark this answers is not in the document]
dymurray: Thanks, this will become relevant with the migmigration integration so I will look into it. For now this is is actually using the new CR's spec field.

pkg/controller/migdirect/rsync.go
```diff
@@ -0,0 +1,827 @@
...
+	destClient, err := t.getDestinationClient()
+	if err != nil {
+		return false, err
+	}
+
+	pvcMap := t.getPVCNamespaceMap()
+
+	selector := labels.SelectorFromSet(map[string]string{
+		"app":     "migdirect-rsync-transfer",
+		"owner":   "migdirect",
```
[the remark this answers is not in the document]
dymurray: Will update with correlation labels with migmigration integration.

pkg/controller/migdirect/stunnel.go
dymurray: hahah thanks

pvc.yaml
dymurray: Yes it did, thanks.

pkg/controller/migdirect/rsync.go
dymurray: Thanks I agree. I will add as a TODO.

pkg/controller/migdirect/rsync.go
dymurray: @alaypatel07 ^ please include this update with the progress fix

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -0,0 +1,99 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// DirectVolumeMigrationSpec defines the desired state of DirectVolumeMigration
+type DirectVolumeMigrationSpec struct {
+	SrcMigClusterRef            *kapi.ObjectReference   `json:"srcMigClusterRef,omitempty"`
+	DestMigClusterRef           *kapi.ObjectReference   `json:"destMigClusterRef,omitempty"`
+	PersistentVolumeClaims      []*kapi.ObjectReference `json:"persistentVolumeClaims,omitempty"`
+	StorageClassMapping         map[string]string       `json:"storageClassMapping,omitempty"`
```
alaypatel07, line 30: @dymurray do we need to change this? it can be a follow-up PR but just curious

pkg/controller/directvolumemigration/directvolumemigration_controller.go
```diff
@@ -0,0 +1,153 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var log = logging.WithName("direct")
+
+/**
+* USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
```
alaypatel07, line 38: remove these comments?

pkg/controller/directvolumemigration/directvolumemigration_controller.go
```diff
@@ -0,0 +1,153 @@
...
+		return err
+	}
+
+	// Watch for changes to DirectVolumeMigration
+	err = c.Watch(&source.Kind{Type: &migrationv1alpha1.DirectVolumeMigration{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	// TODO(user): Modify this to be the types you create
```
alaypatel07, line 67: we need to determine the list of resources to watch and their filter predicates. This IMO is a must-have for 1.4.0

pkg/controller/directvolumemigration/validation.go
```diff
@@ -0,0 +1,245 @@
...
+	client, err := cluster.GetClient(r)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+	// Check if these PVCs actually exist on the source
+	// cluster
+	notFound := make([]string, 0)
+	for _, specPVC := range allPVCs {
+		// Check if pvc actually exists and is bound on source cluster
+		// TODO: Check if PVC is actually attached. We should
```
alaypatel07, line 219: we need to decide what is a good assumption for 1.4.0? 

Willl we tolerate PVCs that have running pods for data copy through this controller or will we assume that they should always be unattached to the pods(i.e. application pods should be quiesced)

pkg/controller/directvolumemigration/task.go
```diff
@@ -0,0 +1,409 @@
...
+			liberr.Wrap(err)
+		}
+		t.Requeue = NoReQ
+		if err = t.next(); err != nil {
+			return liberr.Wrap(err)
+		}
+	case DestinationNamespacesCreated:
+		// Ensure the namespaces are created
+		// TODO: bad func name
+		err := t.getDestinationNamespaces()
```
alaypatel07, line 189: I doubt how useful this is?

Once the namespace object is created, I am not sure what options you have on the object status to check. I would assume that if the create call did not error in the above phase unless some agent explicitly calls delete on the namespace, it will be created

alaypatel07: hrm actually

```
A namespace can be in one of two phases:

- Active the namespace is in use
- Terminating the namespace is being deleted, and can not be used for new objects
```

You can check for `Active` status of the namespace, so might be useful, still doubtful

pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -0,0 +1,76 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      pvc.Name,
+				Namespace: pvc.Namespace,
+				Labels:    srcPVC.Labels,
+			},
+			Spec: newSpec,
+		}
+		err = destClient.Create(context.TODO(), &destPVC)
+		if k8serror.IsAlreadyExists(err) {
+			t.Log.Info("PVC already exists on destination", "name", pvc.Name)
```
alaypatel07, line 56: should we update this?

pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	for _, pvc := range t.Owner.Spec.PersistentVolumeClaims {
+		// Get pvc definition from source cluster
+
+		srcPVC := corev1.PersistentVolumeClaim{}
+		key := types.NamespacedName{Name: pvc.Name, Namespace: pvc.Namespace}
+		err = srcClient.Get(context.TODO(), key, &srcPVC)
+		if err != nil {
+			return err
```
alaypatel07, line 37: I wonder if we should collect the errors here?

pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -0,0 +1,76 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (t *Task) areSourcePVCsUnattached() error {
+	// This function provides state checking on source PVCs, make sure app is
+	// quiesced
+	return nil
+}
+
+func (t *Task) createDestinationPVCs() error {
```
alaypatel07, line 17: I wonder if this function essentially needs to run on all reconcile loops to make sure the destination PVCs are always present.

pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -0,0 +1,76 @@
...
+	// TODO: Set sane default
+	defaultStorageClass := "gp2"
+	if destStorageClass, ok := t.Owner.Spec.StorageClassMapping[sourceStorageClass]; ok {
+		return destStorageClass
+	}
+	return defaultStorageClass
+}
+
+func (t *Task) getDestinationPVCs() error {
+	// Ensure PVCs are bound and not in pending state
```
alaypatel07, line 74: probably a must have for 1.4.0?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -0,0 +1,827 @@
...
+					"app":     "directvolumemigration-rsync-transfer",
+					"purpose": "rsync",
+					"owner":   "directvolumemigration",
+				},
+				Type: corev1.ServiceTypeClusterIP,
+			},
+		}
+		err = destClient.Create(context.TODO(), &svc)
+		if k8serror.IsAlreadyExists(err) {
+			t.Log.Info("Rsync transfer svc already exists on destination", "namespace", ns)
```
alaypatel07, line 295: probably need to compare and update.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -0,0 +1,827 @@
...
+}
+
+// Create rsync transfer route
+func (t *Task) createRsyncTransferRoute() error {
+	// Get client for destination
+	destClient, err := t.getDestinationClient()
+	if err != nil {
+		return err
+	}
+	pvcMap := t.getPVCNamespaceMap()
```
alaypatel07, line 266: NIT: wonder if this is more readable
```
namespacedPVCMap := t.getNamespacedPVCMap()?
```

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -0,0 +1,827 @@
...
+					TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: 2222},
+				},
+				TLS: &routev1.TLSConfig{
+					Termination: routev1.TLSTerminationPassthrough,
+				},
+			},
+		}
+		err = destClient.Create(context.TODO(), &route)
+		if k8serror.IsAlreadyExists(err) {
+			t.Log.Info("Rsync transfer route already exists on destination", "namespace", ns)
```
alaypatel07, line 322: same as above

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -0,0 +1,827 @@
...
+		} else if err != nil {
+			return err
+		}
+		t.RsyncRoutes[ns] = route.Spec.Host
```
alaypatel07, line 326: also, we do certain validations in pvc-migrate to indicate whether a route is admitted in the system or not. Probably need to add that check(may in the same phase I am not really sure)

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -0,0 +1,827 @@
...
+		pvcList := []pvc{}
+		for _, vol := range vols {
+			pvcList = append(pvcList, pvc{Name: vol})
+		}
+		// Generate template
+		rsyncConf := rsyncConfig{
+			SshUser:   "root",
+			Namespace: ns,
+			PVCList:   pvcList,
+			Password:  "changeme",
```
alaypatel07, line 163: A random password generator must for 1.4.0

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -0,0 +1,827 @@
...
+		}
+		err = yaml.Unmarshal(tpl.Bytes(), &configMap)
+		if err != nil {
+			return err
+		}
+
+		// Create configmap on source + dest
+		// Note: when this configmap changes the rsync pod
+		// needs to restart
+		// Need to launch new pod when configmap changes
```
alaypatel07, line 192: this is where I think versioning of configs is needed, to understand what running pod is carrying what config

pkg/controller/directvolumemigration/stunnel.go
```diff
@@ -0,0 +1,495 @@
...
+		Province:           []string{"NC"},
+		Locality:           []string{"RDU"},
+		Organization:       []string{"Migration Engineering"},
+		OrganizationalUnit: []string{"Engineering"},
+	}
+
+	certTemp := x509.Certificate{
+		SerialNumber:          big.NewInt(2020),
+		Subject:               subj,
+		NotBefore:             time.Now(),
```
alaypatel07, line 249: hrm this needs to be reviewed carefully. The source, destination, and controller clusters can have different times and can trip us in ways that would cause a lot of pain

pkg/controller/directvolumemigration/stunnel.go
```diff
@@ -0,0 +1,495 @@
...
+				Ports: []corev1.ServicePort{
+					{
+						Name:       "stunnel",
+						Protocol:   corev1.ProtocolTCP,
+						Port:       int32(2222),
+						TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: 2222},
+					},
+				},
+				Selector: map[string]string{
+					"app":     "directvolumemigration-rsync-transfer",
```
alaypatel07, line 352: probably worth creatint a global variable for labels on service/pod/route for easier maintenance

pkg/controller/directvolumemigration/stunnel.go
```diff
@@ -0,0 +1,495 @@
...
+			return err
+		}
+		t.Log.Info("stunnel client svc created", "name", clientPod.Name, "namespace", svc.Namespace)
+		err = srcClient.Create(context.TODO(), &clientPod)
+		if k8serror.IsAlreadyExists(err) {
+			t.Log.Info("Stunnel client pod already exists on source", "namespace", clientPod.Namespace)
+		} else if err != nil {
+			return err
+		}
+		t.Log.Info("stunnel client pod created", "name", clientPod.Name, "namespace", clientPod.Namespace)
```
alaypatel07, line 455: I like the logs here, +1

pkg/controller/directvolumemigration/stunnel.go
```diff
@@ -0,0 +1,495 @@
...
+			&k8sclient.ListOptions{
+				Namespace:     ns,
+				LabelSelector: selector,
+			},
+			&pods)
+		if err != nil {
+			return false, err
+		}
+		for _, pod := range pods.Items {
+			if pod.Status.Phase != corev1.PodRunning {
```
alaypatel07, line 488: we would want to iterate on each container and check for Ready state for more deterministic check

pkg/controller/directvolumemigration/stunnel.go
```diff
@@ -0,0 +1,495 @@
...
+			return err
+		}
+		t.Log.Info("stunnel client pod created", "name", clientPod.Name, "namespace", clientPod.Namespace)
+	}
+	return nil
+}
+
+// check if stunnel client pods are running
+func (t *Task) areStunnelClientPodsRunning() (bool, error) {
+	// Get client for destination
```
alaypatel07, line 462: after a certain phase, we probably want this function to be executed on every reconcile loop

pkg/controller/directvolumemigration/task.go
dymurray: Yeah I agree... part of the reason why this function just returns nil. I didn't see much value. I'm good to delete this.

pkg/controller/directvolumemigration/pvcs.go
dymurray: Unsure... there are cases where an update might fail due to immutable fields right? But I agree perhaps we should at least try.

pkg/controller/directvolumemigration/rsync.go
dymurray: Yes either that or regenerate this configmap on every run. Agree this current workflow isn't sustainable

pkg/controller/directvolumemigration/stunnel.go
dymurray: Hmm... hadn't considered that. Let's chat about this.

pkg/controller/directvolumemigration/stunnel.go
dymurray: let's also use correlation labels to make this more consistent with existing workflows.

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
alaypatel07: this is separated out in an issue, we will address, marking resolved for now

pkg/controller/directvolumemigration/validation.go
alaypatel07: we have broken this out in a separate issue. Resolving this for now

alaypatel07 (approved):
LTGM

PR #712, opened 2020-10-23
Improve mig registry health check by failing fast

alaypatel07:
(Replaces #681)

This PR does the following things:

- Updates Registry kind from DeploymentConfig to Deployment.
- Removes Imagestream for registry
- Deletes Registry deployment if migstorage is unready.
- Makes sure that registry deployment is not deleted during the course of migration.
- Adds Liveliness and Readiness Probes to the registry deployment for the health check.
- Adds a custom registry health watch with 2 instances in the migplan controller:
- Watch for changes to registry pod health for MigPlans that have running migrations
- Watch for changes to registry pod health for MigPlans that have failed migrations
- Adds a new state Ensure Registry Health in the Reconciliation of Migplan


djwhatle (approved):
Approved. I will plan to make some changes in a follow-on PR so that registries only launch when a migration is started so that we can allow for many idle migplans from a registry perspective.

This PR has a known issue of delaying initial MigPlan readiness for ~20 seconds which leads to the UI not showing PV discovery results. This will need to be fixed soon, I'll plan to address it today or Monday.

alaypatel07:
+1 thanks @djwhatle

PR #717, opened 2020-10-27
MIG-360 - Registry lazy start

### LINKED ISSUE #716: MIG-360 - Migration registries should only exist during active migrations.
Currently, we create 2 Migration Registry deployments per non-closed MigPlan that exists. This doesn't scale well (registries consume CPU and memory) and blocks us from being able to achieve the goal of allowing the user to have many idle MigPlans.

_Sidenote:_ after merge of #712,  MigPlans will not be marked as ready until Registries are ready. This breaks assumptions the UI makes about MigPlans and causes the PV discovery screen to appear broken. UI workaround for this has been merged in  https://github.com/konveyor/mig-ui/pull/1034 but should be reverted after this issue is resolved.

- [x] 1. Early in MigMigration itinerary, introduce a phase "CreateRegistries" that will delete old registry deployments and create new ones. This phase will _not_ wait for registries to start. 
- [x] 2. While registries are starting up, perform other pre-migration tasks
- [x] 3. Directly before registries are first needed in the migration, introduce a phase "WaitForRegistriesReady" that blocks on registry readiness. When registries are observed as ready, set a durable "RegistriesHealthy" condition indicating registries are healthy
- [x] 4. A validation on MigMigration will see that "RegistriesHealthy" is set and start validating the health of registries on each reconcile. If registries go unready, the MigMigration will also go unready (matches todays behavior). The advantage of doing this ongoing check in a validation (and out of the itinerary) is that if the registries return to readiness, the migration will resume itself. 
   - If a problem is detected, raise an error condition similar to what is raised on the MigPlan today when registries go unhealthy.
- [x] 5. The 'Stage' and 'Final' MigMigration itineraries will have a "DeleteRegistries" phase near the end. "Cancel" and "Rollback" will also include a "DeleteRegistries" phase.

We previously implemented creation of migration registries from the MigCluster and then the MigPlan controller, but since the registries are only used in the MigMigration this solution aims to more directly integrate registry bring-up and tear-down with other migration logic that uses the registry.

**Notes**
This should reduce the max running registries to 1 per cluster, with 0 running most of the time.

I'll plan to copy a large bulk of the logic for this over from #712 to leverage the work Alay and Shubham have already done here. I'm just going to be re-arranging where health checking is done from.

djwhatle:
- Moves startup and teardown of migration registries to be part of the migration flow. 
- During migration after registries are started, registries are continually health checked during migmigration validation.

Closes #716 

_In the video below, I simulate a registry issue by momentarily scaling the registry to 0 and then back to 1. The migration pauses and displays the critical condition messages, then resumes once possible to do so._
migplan-readiness


djwhatle:
cam_e2e_qe tests are in-progress.

pkg/controller/migmigration/registry.go
```diff
@@ -53,3 +60,269 @@ func (t *Task) getAnnotations(client k8sclient.Client) (map[string]string, error
...
+	}
+
+	// Condition
+	ensured := nEnsured == 2 // Both clusters.
+	if ensured {
+		t.Owner.Status.SetCondition(migapi.Condition{
+			Type:     RegistriesCreated,
+			Status:   True,
+			Category: migapi.Required,
+			Message:  RegistriesCreatedMessage,
```
alaypatel07, line 122: I wonder if having incremental messages would be helpful.

1. ensure source registry, report successful
2. ensure destination registry, report successful.

That was if we are stuck on either source or destination it is clear where we are in the process. I know we have this pattern everywhere, but just a suggestion.

djwhatle:
Just checked, health-checking of registry during migration isn't working as intended. Will post a commit with fix soon.

pkg/controller/migmigration/registry.go
djwhatle: I just added something like what you're suggesting in https://github.com/konveyor/mig-controller/pull/717/files#diff-7595a02f0293165f7556b448f21e75188e7f77f2b00a9fd05d999e39d777be1aR276-R281
 
This will let you know if we find a cluster with a missing registry pod and give the user a heads up. See the GIF i posted in the first comment for a demo.

djwhatle:
@alaypatel07 tests are green now, the itinerary test was position sensitive and I need to check for registry health at different places in the stage vs final migration.

sseago (approved):

alaypatel07:
This looks solid, VISIACK

LGTM

alaypatel07 (approved):
Thanks @djwhatle for re-working this

shubham-pampattiwar (approved):
ACK, Thank you @djwhatle !

PR #723, opened 2020-10-29
Fix discovery debug tree filtering on PVBs/PVRs

djwhatle:
Fixes https://github.com/konveyor/mig-ui/issues/1049

This endpoint was returning all PodVolumeRestores (I had 260) instead of just the single PodVolumeRestore that was performed during my migration.

**Before**
```
curl --silent http://localhost:8080/namespaces/openshift-migration/plans/plan-foo/tree | grep -o PodVolumeRestore | wc -l                                                       
     260                    
```

**After**
```
curl --silent http://localhost:8080/namespaces/openshift-migration/plans/plan-foo/tree | grep -o PodVolumeRestore | wc -l
       1
```

### DIFF
--- a/pkg/controller/discovery/web/plan.go
+++ b/pkg/controller/discovery/web/plan.go
@@ -3,14 +3,15 @@ package web
 import (
 	"database/sql"
 	"fmt"
+	"net/http"
+	"strings"
+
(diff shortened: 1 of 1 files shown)


jortel (approved):
LGTM

PR #731, opened 2020-10-30
MIG-323 - Add logging improvements to migcluster_controller

shubham-pampattiwar:
This adds logging improvements and kube events for status conditions to the Migcluster_controller, follows the pattern applied in https://github.com/konveyor/mig-controller/pull/698 for Migstorage_controller.

### DIFF
--- a/pkg/controller/migcluster/migcluster_controller.go
+++ b/pkg/controller/migcluster/migcluster_controller.go
@@ -98,8 +98,9 @@ var _ reconcile.Reconciler = &ReconcileMigCluster{}
 // ReconcileMigCluster reconciles a MigCluster object
 type ReconcileMigCluster struct {
 	k8sclient.Client
-	scheme     *runtime.Scheme
 	record.EventRecorder
+
+	scheme     *runtime.Scheme
 	Controller controller.Controller
 }
 
--- a/pkg/controller/migcluster/migcluster_controller.go
+++ b/pkg/controller/migcluster/migcluster_controller.go
@@ -26,6 +26,7 @@ import (
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
 
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -45,7 +46,8 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) *ReconcileMigCluster {
-	return &ReconcileMigCluster{Client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	return &ReconcileMigCluster{Client: mgr.GetClient(), scheme: mgr.GetScheme(), EventRecorder:
+		mgr.GetRecorder("migcluster_controller")}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -97,12 +99,14 @@ var _ reconcile.Reconciler = &ReconcileMigCluster{}
 type ReconcileMigCluster struct {
 	k8sclient.Client
 	scheme     *runtime.Scheme
+	record.EventRecorder
 	Controller controller.Controller
 }
 
 func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	var err error
 	log.Reset()
+	log.SetValues("cluster", request)
 
 	// Fetch the MigCluster
 	cluster := &migapi.MigCluster{}
@@ -117,6 +121,8 @@ func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Re
 
 	// Report reconcile error.
 	defer func() {
+		log.Info("CR", "conditions", cluster.Status.Conditions)
+		cluster.Status.Conditions.RecordEvents(cluster, r.EventRecorder)
 		if err == nil || errors.IsConflict(err) {
 			return
 		}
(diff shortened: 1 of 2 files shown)


pkg/controller/migcluster/validation.go
```diff
@@ -159,7 +152,7 @@ func (r ReconcileMigCluster) validateSaSecret(cluster *migapi.MigCluster) error
 			Status:   True,
 			Reason:   NotFound,
 			Category: Critical,
-			Message:  InvalidSaSecretRefMessage,
+			Message:  fmt.Sprintf("The `serviceAccountSecretRef` must reference a `secret`, subject %s", path.Join(cluster.Spec.ServiceAccountSecretRef.Namespace, cluster.Spec.ServiceAccountSecretRef.Name)),
```
alaypatel07, line 155: "must reference a valid `secret`"

pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -97,12 +99,14 @@ var _ reconcile.Reconciler = &ReconcileMigCluster{}
 type ReconcileMigCluster struct {
 	k8sclient.Client
 	scheme     *runtime.Scheme
+	record.EventRecorder
```
alaypatel07, line 102: I would put this one line above so all embedded types are together and then have a blank line:

```
    k8sclient.Client
    record.EventRecorder

    scheme *runtime.Scheme
```

pkg/controller/migcluster/validation.go
```diff
@@ -208,13 +201,12 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
...
 			helpText = CaBundleMessage
 		}
-		message := fmt.Sprintf(TestConnectFailedMessage, helpText, err)
 		cluster.Status.SetCondition(migapi.Condition{
 			Type:     TestConnectFailed,
 			Status:   True,
 			Reason:   ConnectFailed,
 			Category: Critical,
-			Message:  message,
+			Message:  fmt.Sprintf("Test connect failed. %s: %s", helpText, err),
```
alaypatel07, line 209: might want to put the error in the errors in the item to keep it consistent

pkg/controller/migcluster/validation.go
```diff
@@ -110,7 +99,7 @@ func (r ReconcileMigCluster) validateURL(cluster *migapi.MigCluster) error {
 			Status:   True,
 			Reason:   Malformed,
 			Category: Critical,
-			Message:  MalformedURLMessage,
+			Message:  "The `spec.URL` is malformed.`",
```
alaypatel07, line 102: Need to drop the extra "`" at the end?

alaypatel07 (approved):
Apart from the small nit, this LGTM thanks

pkg/controller/migcluster/validation.go
jortel: Yes.

pkg/controller/migcluster/validation.go
```diff
@@ -159,7 +148,7 @@ func (r ReconcileMigCluster) validateSaSecret(cluster *migapi.MigCluster) error
 			Status:   True,
 			Reason:   NotFound,
 			Category: Critical,
-			Message:  InvalidSaSecretRefMessage,
+			Message:  fmt.Sprintf("The `serviceAccountSecretRef` must reference a valid `secret`, subject %s", path.Join(cluster.Spec.ServiceAccountSecretRef.Namespace, cluster.Spec.ServiceAccountSecretRef.Name)),
```
jortel: We try to keep line length <= 100.  Mainly to keep the code easy to view in github.

jortel: Also, don't remember the convention you followed in the other PR but thinking that `subject:` may be clearer.  (with colon).

pkg/controller/migcluster/validation.go
```diff
@@ -267,13 +256,14 @@ func (r *ReconcileMigCluster) validateSaTokenPrivileges(cluster *migapi.MigClust
...
 	}
+
 	if !migrationSar.Status.Allowed || !veleroSar.Status.Allowed {
 		cluster.Status.SetCondition(migapi.Condition{
 			Type:     SaTokenNotPrivileged,
 			Status:   True,
 			Reason:   Unauthorized,
 			Category: Critical,
-			Message:  SaTokenNotPrivilegedMessage,
+			Message:  fmt.Sprintf("The `saToken` has insufficient privileges in either `migration.openshift.io` or `velero.io` group."),
```
jortel, line 266: I'm concerned that by listing the API groups is opening a can of worms since it adds a little more information but is not a complete description of the auth needed.  Also, if the group name change, this would be easily missed.  I think it's better to leave the message as it was.

pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -117,6 +122,8 @@ func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Re
 
 	// Report reconcile error.
 	defer func() {
+		log.Info("CR", "conditions", cluster.Status.Conditions)
```
jortel, line 125: Didn't notice this on the last PR but:
Thinking something like this would be clearer:
```
log.Info("Conditions", "list", cluster.Status.Conditions)
```
The "CR" feels cryptic.

jortel (approved):
Few comments for consideration.

pkg/controller/migcluster/migcluster_controller.go
alaypatel07: I am not sure about this one. I agree it feels cryptic in the code, but if you look at the logs entry it produces, it is in the format

```
{... "CR", "cluster", "foo-ns/foo-cluster", "conditions", "[]"}
```

The suggested change will look like,

```
{... "Conditions", "cluster", "foo-ns/foo-cluster", "list", "[]"}
```
The former seems more readable log entry IMO

djwhatle (approved):
ACK

pkg/controller/migcluster/validation.go
shubham-pampattiwar: Done.

pkg/controller/migcluster/validation.go
shubham-pampattiwar: Agreed, updated the PR.

pkg/controller/migcluster/validation.go
shubham-pampattiwar: Updated the `Message`.

alaypatel07 (approved):
ACK

PR #732, opened 2020-10-30
MIG-284 - phase 1 of rsync client pod progress reporting

### LINKED ISSUE #730: Progress reporting for rsync client pod
For the new DirectVolumeMigration Controller, we need the ability to report the progress of the migration operation. At the very minimum, we want to report something like:

- 3/10 rsync client pods are running
- 4/10 rsync clients pods have completed
- 3/10 rsync clients pods have failed

For more granular reporting we want to get logs from the source client pods and get the actual rsync progress. 

- Phase 1: Add a new CR that will

1. poll every 5 seconds to get the pod
2. report whether the pod is running, completed, or failed.

- Phase 2: If the pod is running, stream updates from the client rsync pod logs in the status.

Running the poll in the wait step of DirectVolumeMigration controller phase.

This was rejected because it will complicate the Phase logic. All it really cares about is to wait on the status of each client pod. Hence having an abstraction object in the same cluster as the controller would allow this controller to watch that new object and hide the implementation detail of polling under the new object controller

**Additional context**

alaypatel07:
This depends on #711 

Intended to partially fix #730


pkg/controller/directpvmigrationprogress/directpvmigrationprogress_controller.go
```diff
@@ -0,0 +1,300 @@
...
+
+// ReconcileDirectPVMigrationProgress reconciles a DirectPVMigrationProgress object
+type ReconcileDirectPVMigrationProgress struct {
+	client.Client
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a DirectPVMigrationProgress object and makes changes based on the state read
+// and what is in the DirectPVMigrationProgress.Spec
+// TODO(user): Modify this Reconcile function to implement your Controller logic.  The scaffolding writes
```
dymurray, line 110: remove

pkg/controller/directpvmigrationprogress/directpvmigrationprogress_controller.go
```diff
@@ -0,0 +1,300 @@
...
+		return err
+	}
+
+	// Watch for changes to DirectPVMigrationProgress
+	err = c.Watch(&source.Kind{Type: &migapi.DirectPVMigrationProgress{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	// TODO(user): Modify this to be the types you create
```
dymurray, line 87: Need to set proper watches

pkg/controller/directpvmigrationprogress/directpvmigrationprogress_controller.go
```diff
@@ -0,0 +1,300 @@
...
+
+	if pvProgress.Status.PodPhase == kapi.PodFailed || pvProgress.Status.PodPhase == kapi.PodRunning {
+		// if the condition is Failed or Running poll
+		return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 5}, nil
+	}
+
+	// the pod condition is Complete
+	return reconcile.Result{}, nil
+
+	// at this point assume that the container is running
```
dymurray, line 173: Remove this I imagine?

dymurray (approved):
Looks good to me for the most part, fine if you want to address most of this in a follow up PR.

config/crds/migration_v1alpha1_directpvmigrationprogress.yaml
```diff
@@ -0,0 +1,49 @@
...
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: directpvmigrationprogresses.migration.openshift.io
+spec:
+  group: migration.openshift.io
+  names:
+    kind: DirectPVMigrationProgress
```
djwhatle, line 11: Based on the name, this CR doesn't run the PV migration, it just reflects progress. Which controller is responsible for starting the Direct PV migration?

djwhatle: I just noticed the other controller seems to be `directvolumemigration`, why the name convention change between these two? Seems like we should settle on using `PV` or `Volume` in both if they are related.

alaypatel07: > Based on the name, this CR doesn't run the PV migration, it just reflects progress. Which controller is responsible for starting the Direct PV migration?

@djwhatle https://github.com/konveyor/mig-controller/pull/732/files#diff-78a6bcf4a126e869712d026733ae2193015d729e56cbd9a3c49a14d8a6f6999cR117

alaypatel07:
closing in favor of #734

PR #738, opened 2020-11-03
[MIG-249] Change stage pod cleanup to remove all stale stage pods from migration namespaces

### LINKED ISSUE #700: Orphaned stage pods should be cleaned up at the beginning of each 'Stage', 'Final', 'Rollback' migration
Stage pods may be orphaned on the source or target cluster if a migration fails or is cancelled. This is undesirable and leads to errors on later migration attempts. 

We must _not_ delete stage pods before a user has a chance to inspect the world and decide on the next steps. _After_ the user inspects the world, they may execute:

_'Rollback' itinerary_:
Part of the rollback routine should be deletion of stage pods with a matching MigPlan ref on the source and target cluster. 

_'Stage' and 'Final' itineraries_:
Early in both of these itineraries before any new migration work is done, we should delete all stage pods on the source and target cluster using the same stage pod deletion methods defined for use in rollback.

The benefit of waiting until migration time to perform stage pod deletions is that the user can investigate the failure as needed up until they make the decision to run the next action (stage, migrate, rollback, manual cleanup).

@sseago  and I considered running this deletion at different times (e.g. on migration fail for immediate cleanup), but waiting until the next migration has the benefit of letting the user inspect the world and try to understand why the last attempt failed.

**Additional context**
This is related to the effort of un-quiescing apps. We have some additional complexity to deal with on quiescing apps, but the proposed strategy for that will be written up in a separate issue.

djwhatle:
- Make stage pod cleanup delete all previous stage pods found in namespaces to be migrated, regardless of correlation label
 - Remove the `Prepare` "combo" phase in favor of more explicit individual phases under the `Prepare` step.
 - Remove stage pod cleanup from `FailedItinerary`, we want to leave stage pods around so user can inspect. Stage pod will be auto-cleaned at rollback or next migration when user has decided what to do.

Closes #700


pkg/controller/migmigration/task.go
```diff
@@ -212,7 +214,7 @@ var RollbackItinerary = Itinerary{
 		{Name: DeleteBackups, Step: StepFinal},
 		{Name: DeleteRestores, Step: StepFinal},
 		{Name: DeleteRegistries, Step: StepFinal},
-		{Name: EnsureStagePodsDeleted, Step: StepFinal, all: HasStagePods},
```
djwhatle, line 215: I want to call out that I removed this flag so that rollback will always remove stage pods

pkg/controller/migmigration/task.go
```diff
@@ -316,21 +318,6 @@ func (t *Task) Run() error {
 				return liberr.Wrap(err)
 			}
 		}
-
-	case Prepare:
```
djwhatle, line 320: Deletion of stage pods and annotation cleanup have are still run in the same spot during a migration, but now they are individual phases instead of being bundled together in `Prepare`

pkg/controller/migmigration/stage.go
```diff
@@ -506,7 +506,7 @@ func (t *Task) ensureStagePodsDeleted() error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	options := k8sclient.MatchingLabels(t.PlanResources.MigPlan.GetCorrelationLabels())
+	options := k8sclient.MatchingLabels(t.stagePodIdentifierLabel())
```
djwhatle, line 509: This new `stagePodIdentifier` label is used for lookup of _all_ stage pod regardless of correlation label.

pkg/controller/migmigration/stage.go
```diff
@@ -562,6 +562,16 @@ func (t *Task) ensureStagePodsTerminated() (bool, error) {
 	return true, nil
 }
 
+// Label applied to all stage pods for easy cleanup
+func (t *Task) stagePodIdentifierLabel() map[string]string {
+	return map[string]string{StagePodLabel: migapi.True}
+}
+
+// Build map of all stage pod labels
```
djwhatle, line 570: I decided to leave the correlation labels in place, they could be useful for something I think and no harm in leaving them on a resource we will delete soon anyways.

pkg/controller/migmigration/stage.go
```diff
@@ -147,7 +147,7 @@ func (t *Task) createStagePods(client k8sclient.Client, stagePods StagePodList)
 
 func (t *Task) listStagePods(client k8sclient.Client) (StagePodList, error) {
 	podList := corev1.PodList{}
-	options := k8sclient.MatchingLabels(t.stagePodLabels())
+	options := k8sclient.MatchingLabels(t.stagePodIdentifierLabel())
```
sseago, line 150: Wouldn't the question of "list all stage pods" vs. "list stage pods for this MigPlan" both be needed, depending on context? If we're trying to find stage pods created for this migration for the purpose of using them, shouldn't we use the correlation label? We wouldn't want to rely on an old stage pod created for a different MigPlan (which should have been deleted by now anyway). The "list all stage pods regardless of migplan" would be mainly used for deletion. Also, isn't this context in this function not a single-namespace-scoped list? i.e. it will list all stage pods in all namespaces. We definitely want the migplan filtering there I'd think, or we might end up seeing pods for other migplans.

pkg/controller/migmigration/stage.go
sseago: In this case we would need to filter on namespace here. When we were using the correlation label we didn't need namespace filtering for deletion since the label was enough to know it was one that this operation should delete.

pkg/controller/migmigration/stage.go
```diff
@@ -543,7 +543,7 @@ func (t *Task) ensureStagePodsTerminated() (bool, error) {
 		corev1.PodFailed:    true,
 		corev1.PodUnknown:   true,
 	}
-	options := k8sclient.MatchingLabels(t.PlanResources.MigPlan.GetCorrelationLabels())
+	options := k8sclient.MatchingLabels(t.stagePodIdentifierLabel())
```
sseago, line 546: We probably need namespace filtering here too.

pkg/controller/migmigration/stage.go
djwhatle: Good points. I will make some changes to:

1. Continue using the correlation label based stage pod list where we can
2. For cleanup purposes, have a separate function that constructs a list of stage pods _in the namespaces selected for migration_

djwhatle: @sseago I addressed your comments in the latest change.

pkg/controller/migmigration/stage.go
djwhatle: @sseago I addressed your comments in the latest change.

djwhatle:
@sseago I addressed your comments in the latest change. 

 -  Tried to make it more clear that the new label is specifically for cleanup purposes.
```
func (t *Task) stagePodCleanupLabel() map[string]string {
```
 - Go back to using stage pod correlation labels for non-cleanup tasks
 - Filter stage pod cleanup queries using namespaces included on migplan

djwhatle:
@sseago added support for doing stage pod cleanup in the presence of src:dest namespace mapping.

sseago (approved):
LGTM

pranavgaikwad (approved):

djwhatle:
Wrote up #740 to describe an issue I saw while working on this PR. Our registry pod queries need to be more specific.

PR #741, opened 2020-11-03
Only refresh MigPlan from StartRefresh phase, skip MigCluster/MigStorage refresh.

djwhatle:
MigStorage and MigClusters use watch ProviderSources to continuously monitor health, so I don't think we need to manually poke them for a refresh at migration start time.

https://github.com/konveyor/mig-controller/blob/master/pkg/controller/migstorage/migstorage_controller.go#L68-L76
https://github.com/konveyor/mig-controller/blob/master/pkg/controller/migcluster/migcluster_controller.go#L69-L77

_The problem with performing a refresh on these two at migration start:_ updating  MigCluster/MigStorage will trigger reconciles of _all_ other MigPlans using these resources meaning that the MigPlan reconciler will be tied up for potentially a long time if the user has many several idle MigPlans.

I noticed this problem when running migrations in my stress test env with 50 MigPlans. It doesn't block the migration, but the MigPlan reconciler does get tied up for a while after.

### DIFF
--- a/pkg/controller/migmigration/verify.go
+++ b/pkg/controller/migmigration/verify.go
@@ -109,7 +109,7 @@ func (t *Task) reportHealthCondition() {
 	}
 }
 
-// Start a refresh on MigPlan and attached resources
+// Start a refresh on MigPlan
 func (t *Task) startRefresh() (bool, error) {
 	started := false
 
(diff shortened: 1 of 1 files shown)


alaypatel07 (approved):
ACK

PR #746, opened 2020-11-05
MIG-323 - Add logging improvements to mighook_controller

shubham-pampattiwar:
This adds logging improvements and kube events for status conditions to the mighook_controller, follows the pattern applied in #698 for migstorage_controller.

Essentially, this PR does the following:
- Adds the eventRecorder for status conditions
- Adds end of reconcile logging for status conditions
- Moves all const messages for mighook controller directly into where conditions are set, some messages may be reframed
- Adds additional context/subject if needed to the condition messages
- Adds missing `Reason` field for the Conditions

### DIFF
--- a/pkg/controller/mighook/validation.go
+++ b/pkg/controller/mighook/validation.go
@@ -80,7 +80,7 @@ func (r ReconcileMigHook) validateTargetCluster(hook *migapi.MigHook) error {
 			Status:   True,
 			Reason:   NotSet,
 			Category: Critical,
-			Message:  "Only 'source' and 'destination' are accepted as values for spec.targetCluster.",
+			Message:  "Only `source` and `destination` are accepted as values for spec.targetCluster.",
 		})
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


djwhatle (approved):

pkg/controller/mighook/validation.go
```diff
@@ -90,7 +80,7 @@ func (r ReconcileMigHook) validateTargetCluster(hook *migapi.MigHook) error {
 			Status:   True,
 			Reason:   NotSet,
 			Category: Critical,
-			Message:  InvalidTargetClusterMessage,
+			Message:  "Only 'source' and 'destination' are accepted as values for spec.targetCluster.",
```
alaypatel07, line 83: nit: other places have use ` instead of '

alaypatel07 (approved):
Left a small suggestion! LGTM irrespective of that

alaypatel07 (approved):
ACK

Issue #740, opened 2020-11-03
Seeing "Unhealthy registry pod" critical condition on a non-registry pod


djwhatle:
**Describe the bug**

This critical condition is getting raised on a MigMigration. 
```
Migration Registry Pod nginx-pv-30/stage-nginx-deployment-557dd97bf8-w69hw-s6vx6 is in unhealthy state on cluster host
```
The strange thing about this is... The above is not a registry Pod.

**To Reproduce**
1. Run a stage migration on nginx-pv with mig-controller latest

**Expected behavior**
Migration runs to completion without any pauses

**Additional context**
Expect this is a bug in recent registry health check work done in #681 and #717

djwhatle:
Err is coming from here https://github.com/konveyor/mig-controller/blob/0dcd0b4faa24762f3219170c8cf31332bf77eba8/pkg/controller/migmigration/validation.go#L293-L296

djwhatle:
The problem is that `getRegistryPods` doesn't filter pods correctly. It just selects every pod in any namespace with a migplan label.

https://github.com/konveyor/mig-controller/blob/0dcd0b4faa24762f3219170c8cf31332bf77eba8/pkg/controller/migmigration/validation.go#L312-L326

PR #752, opened 2020-11-06
Refine getRegistryPods query to fetch only the associated registry pods

shubham-pampattiwar:
This PR does the following (fixes https://github.com/konveyor/mig-controller/issues/740):
- Updates the registry deployment such that the `migration-registry` label gets cascaded to registry pods.
- Updates the `getRegistryPods` function to fetch the pods based on the `migration-registry` label and not on the `migplan` label.

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -359,18 +359,18 @@ func (r *MigPlan) UpdateRegistryDeployment(storage *MigStorage, deployment *apps
 	deployment.Spec = appsv1.DeploymentSpec{
 		Replicas: pointer.Int32Ptr(1),
 		Selector: metav1.SetAsLabelSelector(map[string]string{
-			"app":        name,
-			"deployment": name,
-			"migplan":    string(r.UID),
+			"app":                  name,
+			"deployment":           name,
+			"migplan":              string(r.UID),
 			MigrationRegistryLabel: string(r.UID),
 		}),
 		Template: kapi.PodTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
 				CreationTimestamp: metav1.Time{},
 				Labels: map[string]string{
-					"app":        name,
-					"deployment": name,
-					"migplan":    string(r.UID),
+					"app":                  name,
+					"deployment":           name,
+					"migplan":              string(r.UID),
 					MigrationRegistryLabel: string(r.UID),
 				},
 			},
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -362,6 +362,7 @@ func (r *MigPlan) UpdateRegistryDeployment(storage *MigStorage, deployment *apps
 			"app":        name,
 			"deployment": name,
 			"migplan":    string(r.UID),
+			MigrationRegistryLabel: string(r.UID),
 		}),
 		Template: kapi.PodTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
@@ -370,6 +371,7 @@ func (r *MigPlan) UpdateRegistryDeployment(storage *MigStorage, deployment *apps
 					"app":        name,
 					"deployment": name,
 					"migplan":    string(r.UID),
+					MigrationRegistryLabel: string(r.UID),
 				},
 			},
 			Spec: kapi.PodSpec{
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -48,7 +48,7 @@ const (
 	ClosedIndexField = "closed"
 )
 
-// MigPlanHook hold a referene to a MigHook along with the desired phase to run it in
+// MigPlanHook hold a reference to a MigHook along with the desired phase to run it in
 type MigPlanHook struct {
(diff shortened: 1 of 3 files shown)


pkg/controller/migmigration/validation.go
```diff
@@ -316,7 +316,7 @@ func getRegistryPods(plan *migapi.MigPlan, registryClient compat.Client) (corev1
 	registryPodList := corev1.PodList{}
 	err := registryClient.List(context.TODO(), &k8sclient.ListOptions{
 		LabelSelector: k8sLabels.SelectorFromSet(map[string]string{
-			"migplan": string(plan.UID),
+			"migration-registry": string(plan.UID),
```
sseago, line 319: You should probably use the `MigrationRegistryLabel` const here as well.

sseago (approved):
LGTM, with the minor suggestion to use the defined const instead of the string literal in validation.

pkg/controller/migmigration/validation.go
shubham-pampattiwar: Done!

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -359,17 +359,19 @@ func (r *MigPlan) UpdateRegistryDeployment(storage *MigStorage, deployment *apps
 	deployment.Spec = appsv1.DeploymentSpec{
 		Replicas: pointer.Int32Ptr(1),
 		Selector: metav1.SetAsLabelSelector(map[string]string{
-			"app":        name,
-			"deployment": name,
-			"migplan":    string(r.UID),
+			"app":                  name,
+			"deployment":           name,
+			"migplan":              string(r.UID),
+			MigrationRegistryLabel: string(r.UID),
```
pranavgaikwad, line 365: Shouldn't label constant be in the api too?

sseago (approved):

pranavgaikwad (changes requested):

pkg/apis/migration/v1alpha1/migplan_types.go
sseago: It's here: https://github.com/konveyor/mig-controller/blob/master/pkg/apis/migration/v1alpha1/migplan_types.go#L236

But actually, now that I think about it, we should probably remove it from the controller and just reference `migapi.MigrationRegistryLabel` instead of definining it twice.

pkg/controller/migmigration/registry.go
```diff
@@ -21,6 +21,11 @@ const (
 	MigRegistryDirAnnotationKey string = "openshift.io/migration-registry-dir"
 )
 
+// Registry label for controller-created migration registry resources
+const (
+	MigrationRegistryLabel = "migration-registry"
```
sseago, line 26: This can be removed.

pkg/controller/migmigration/validation.go
```diff
@@ -316,7 +316,7 @@ func getRegistryPods(plan *migapi.MigPlan, registryClient compat.Client) (corev1
 	registryPodList := corev1.PodList{}
 	err := registryClient.List(context.TODO(), &k8sclient.ListOptions{
 		LabelSelector: k8sLabels.SelectorFromSet(map[string]string{
-			"migplan": string(plan.UID),
+			MigrationRegistryLabel: string(plan.UID),
```
sseago, line 319: After removing the controller one, this can reference `migapi.MigrationRegistryLabel`

sseago (changes requested):
Actually, a minor change -- we don't need to define the MigrationRegistryLabel const in both api and controller. The API definition can be used in both places.

djwhatle (approved):

pranavgaikwad (approved):
Consider removing redundancy in label values. `migration-registry` and `migplan` both labels have value MigPlan's UID as value. Instead, a better approach would be, add 2 labels on Registry pods: 

```yml
labels:
  migration.openshift.io/migration-registry: true
  migplan: <migplan_uid>
```

djwhatle:
@shubham-pampattiwar I was going to suggest the same as Pranav did above.

djwhatle (approved):
I tested this and verified it solves the linked issue. LGTM.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -231,15 +231,15 @@ func (r *MigPlan) ListMigrations(client k8sclient.Client) ([]*MigMigration, erro
 // Registry
 //
 
-// Registry label for controller-created migration registry resources
+// Registry labels for controller-created migration registry resources
 const (
-	MigrationRegistryLabel = "migration-registry"
+	MigrationRegistryLabel = "migration.openshift.io/migration-registry"
```
sseago, line 236: Are we namespacing labels now too? So far we've only been adding "migration.openshift.io" to annotations. Or is this a new convention we're trying to follow @djwhatle

sseago (approved):
LGTM with the minor question around label naming convention.

pkg/apis/migration/v1alpha1/migplan_types.go
djwhatle: Slowly moving towards namespacing labels, see https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ for some sample labels

pranavgaikwad (approved):

alaypatel07 (approved):
ACK

PR #754, opened 2020-11-09
Enhance Makefile to allow for easier running of local controller + discovery

### LINKED ISSUE #728: Enable local discovery service to communicate with mig-ui
Today, controller devs must choose one of the two:
-  fast iteration time, non-functioning mig-ui (`make run`)
-  slower iteration time, full mig-ui functionality (`docker build, docker push, oc apply`).

mig-controller developers must stop the mig-controller Pod running on the cluster so that they can run a local copy of CAM without interference. Stopping the mig-controller Pod will also stop the discovery service, which will break mig-ui. 

Make it possible for mig-ui to communicate with a locally running discovery service

1. Add `discovery_url_override` variable to mig-operator (may already exist?). This will modify the mig-ui configmap to point at a different discovery endpoint of the user's choosing. The developer should set this to point at their localhost discovery service.

2. Add mig-ui route URL interrogation and auto-config of discovery service CORS whitelist to `make run`
   - `make run` should automatically check `oc get route mig-ui` to find the appropriate CORS URL to whitelist 
   - Apply the mig-ui URL to the CORS whitelist for the discovery service. This is done so that the local discovery service will accept incoming connections from the mig-ui domain name.

djwhatle:
We've had a pretty poor dev experience around running mig-controller locally in situations when mig-ui integration testing is also needed. Up until now you would have needed to build a container image, push it, and swap the image out in your migrationcontroller CR to test mig-controller change in mig-ui. 

I wanted to fix this to enable faster iteration time for controller devs who need to test their changes along with mig-ui

 - `make run` now handles CORS setup for discovery service automatically, enabling mig-ui to be fully functional with locally running mig-controller
 - Two new targets, `make use-local-controller` and `make use-oncluster-controller` will automatically patch the mig-operator `migrationcontroller` CR to make necessary changes for running mig-controller locally or on-cluster.

Closes #728

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,15 +19,33 @@ manager: generate fmt vet
 # Run against the configured Kubernetes cluster in ~/.kube/config after login as mig controller SA
 run: generate fmt vet
 	./hack/controller-sa-login.sh
-	KUBECONFIG=$(KUBECONFIG)-${KUBECONFIG_POSTFIX} go run -tags "${BUILDTAGS}" ./cmd/manager/main.go
+	KUBECONFIG=$(KUBECONFIG)-${KUBECONFIG_POSTFIX} ./hack/start-local-controller.sh
 
+# Same as `make run`, but skips generate, fmt, vet. Useful for quick iteration.
+run-fast:
+	./hack/controller-sa-login.sh
+	KUBECONFIG=$(KUBECONFIG)-${KUBECONFIG_POSTFIX} ./hack/start-local-controller.sh
 
 # Run against the configured Kubernetes cluster in ~/.kube/config, skip login to mig-controller SA
 run-skip-sa-login: generate fmt vet
-	go run -tags "${BUILDTAGS}" ./cmd/manager/main.go
+	./hack/start-local-controller.sh
 
 tilt:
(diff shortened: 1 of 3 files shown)


shawn-hurley:
Might be nice to have a helper function that could tell me what my current setup is (on cluster vs local), maybe add as an echo when it is run just for a warning?

otherwise /lgtm

djwhatle:
Added this warning that will show up if you run mig-controller locally without having the in-cluster controller turned off:

```
./hack/controller-sa-login.sh
Found mig-ui route domain.
Found migration-controller SA. Performing token login...
KUBECONFIG=/Users/dwhatley/.kube/awsconfig-mig-controller ./hack/start-local-controller.sh

[!] WARNING: migrationcontroller CR has '.spec.migration_controller=true' Running a local controller will conflict.
[!] To resolve conflict, run 'make use-local-controller', then re-run this command.
```

pranavgaikwad (approved):
I don't know what magic this PR does. But, I tested this with a locally running UI talking to locally running discovery and controller, it works as expected. The CORS headers are set up properly. 

Screenshot from 2020-11-10 13-06-40

PR #756, opened 2020-11-09
Integrate DVM controller into migmigration workflow

dymurray:
TODO: Transfer pod should be a deployment


pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -22,14 +22,20 @@ import (
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type PVCToMigrate struct {
+	Name               string                            `json:"name,omitempty"`
+	Namespace          string                            `json:"namespace,omitempty"`
```
sseago, line 27: Do we need a TargetNamespace here too, in case the MigPlan has a namespace mapping, or are there other aspects here which are known not to support that yet which will need to be fixed before namespace mappings work?

pkg/controller/migmigration/dvm.go
```diff
@@ -0,0 +1,123 @@
...
+	client, err := t.getDestinationClient()
+	if err != nil {
+		return err
+	}
+	err = client.Create(context.TODO(), dvm)
+	return err
+
+}
+
+func (t *Task) buildDirectVolumeMigration() *migapi.DirectVolumeMigration {
```
sseago, line 34: This needs to set owner references, and the migmigration_controller needs to use this to watch DVMs.

sseago:
For the mig-operator changes, we could either update https://github.com/konveyor/mig-operator/pull/489 or create a separate PR -- either way, all of the operator PRs for this topic branch need to be merged at the same time we merge the topic branch to master.

pkg/controller/migmigration/dvm.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, err
+	}
+	if len(list.Items) > 0 {
+		return &list.Items[0], nil
+	}
+
+	return nil, nil
+}
+
+func (t *Task) hasDirectVolumeMigrationCompleted(dvm *migapi.DirectVolumeMigration) (bool, []string) {
```
dymurray, line 82: After discussing with @pranavgaikwad and @alaypatel07 we determined that the best way for progress reporting to exist in the UI is to read directly from the ``dvm` and `dvmp` resources and to use the existing `dvm` CR status fields to populate this data on the `migmigration` resources.

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
dymurray: Good catch. There is nothing technically that is stopping namespace mappings from working, we just need to tweak the code a bit. I will add support for that here.

pkg/controller/migmigration/task.go
```diff
@@ -406,7 +417,7 @@ func (t *Task) Run() error {
 				}
 			}
 		}
-		t.Requeue = NoReQ
+		t.Requeue = PollReQ
```
dymurray, line 420: @sseago this is me changing NoReQ for DIM. Let me know if you want me to change this.

dymurray:
I am going to tackle namespace mapping in a separate PR

Issue #750, opened 2020-11-06
Fix noisy "unsupported event type" message 


djwhatle:
**Describe the bug**

After recent round of event logging changes, seeing this a lot coming out of mig-controller. 
```
E1106 14:49:11.692555       1 event.go:264] Unsupported event type: ''
```

shubham-pampattiwar:
@djwhatle this is caused if the `Reason` field is not set for a given `Condition` , i.e. we need `Reason` field set for all the `Condition(s)` we are logging, essentially for all the `SetCondition` calls. I have taken this into consideration and added the missing `Reason` fields in the PRs raised after the `migstorage_controller` and `migcluster_controller`, once those PRs are merged, this issue should be fixed.

djwhatle:
Nice, I'll link those PRs then. Thanks @shubham-pampattiwar

alaypatel07:
Team, this should go away, please feel to re-open if it still persists

PR #757, opened 2020-11-10
fixups: fix typo in arguments

alaypatel07:
use cond.Type which is always preset as reason for our conditions


djwhatle (approved):

djwhatle:
@alaypatel07 does this resolve #750 ?

alaypatel07:
@djwhatle Yes, I think it will fix #750. Also note, some of our conditions are missing `Reason` field but all of them have a Type field. In order to make the Type and Reason not be redundant on our conditions, I am sending in our condition Type as event Reason.

shubham-pampattiwar (approved):
ACK, thanks @alaypatel07

Issue #701, opened 2020-10-20
[MIG-249] Apps should be un-quiesced at appropriate times 


djwhatle:
**Is your feature request related to a problem? Please describe.**

When a migration fails, applications can be left in a quiesced state (0 replicas). 

1. **We must allow the user to manually un-quiesce their apps.** If the user wants to _rollback_ all changes made by migration attempts, the rollback steps should restore replica counts on the source cluster _after_ removing the app from the target cluster.

2. **Apps should not be restored in a quiesced state if the first migration attempt failed** If the user wants to run another _stage_ or _final_ migration,  they should be able to leave the source cluster app quiesced and try again. Upon successful migration, the target cluster app should be running, restored to the original replica count.

_Note_: un-quiescing apps can be dangerous. We must not start apps on source+target at 
the same time that mount the same PVC data, or data corruption may occur.

**Describe the solution you'd like**

To solve (1.), we should 
 - modify the rollback itinerary as needed to restore the original replica counts (for deployments, replicasets, etc.) and scheduling rules (for daemonsets only) _only after_ the app has been deleted from the target cluster. 
    -  original replica counts should be available on each resource from the first round of rollback changes.

To solve (2.), we should:
 -  modify [openshift-velero-plugin]( https://github.com/konveyor/openshift-velero-plugin/) to detect when a 'final' restore is being run
 -  swap in the 'stashed' replicacounts for each app resource kind as they are processed by the restore plugin. This is necessary because the 'InitialBackup' which is used in the 'FinalRestore' typically contains definitions of running apps, but in this case we are allowing a migration attempt where the apps were already quiesced on the source cluster.
 - if we see that the replica count is already stashed in the annotation and is >0, we should _skip_ the stashing of replica count step and keep the existing value

**Describe alternatives you've considered**
@sseago and I considered making a manual "un-quiesce only" lightweight rollback as an additional option for the user. This would allow us to avoid adding plugin logic mentioned above, but we thought it would be a better UX if the user doesn't have to know about the un-quiesce intricacies and could just click the migrate button again. 

**Additional context**
I probably missed some details on how to solve (2.), so feel free to ask them on this issue for anything that is unclear.

aufi:
I started work on this issue, however it started look to me, that most of mentioned points were already resolved by current state of the codebase.

- unquiesce operation updates number of replicas before annotation creation, but that's operation on a Map structure, both changes are saved with Update() method. I'm not sure we want make Update() called twice, but it is possible if needed
[...]

PR #758, opened 2020-11-10
Update rollback with unquiesce step being run always

aufi:
Rollback should include unquiesce step always regardless the quiescePods flag on its migmigration CR.

Removing conditional lookup for such flag.

Related to https://github.com/konveyor/mig-controller/issues/701


djwhatle (approved):

djwhatle:
Partial fix for #701

PR #761, opened 2020-11-10
Bug 1896015: Fix PodVolumeBackup & PodVolumeRestore not reflected in plan tree.

### LINKED ISSUE #724: Bug 1896015: Discovery Service requires restart to get latest changes to Plan Tree view
See: https://bugzilla.redhat.com/show_bug.cgi?id=1896015

**Describe the bug**
I ran a few migrations against a single MigPlan while working on #723 . I found that the PodVolumeRestores and PodVolumeBackups will sometimes only show up in the plan tree view if you kill mig-controller, which also kills the discovery service, and let it restart.

The PodVolumeRestores and PodVolumeBackups are also  missing from JSON output of /plan/tree.

**What I expected**
image

**What I saw**
Same as above, but with PodVolumeBackups and PodVolumeRestores missing until I restarted the controller.

**To Reproduce**
1. Run a few migrations against the same MigPlan
2. Check the debug tree view in the UI
3. Notice that the output received from the discovery service is missing PodVolumeBackups that are present on the cluster

jortel:
Both the `PodVolumeBackup` and `PodVolumeRestore` models were not being correctly by their predicates.  The missing cluster PK caused an FK constraint violation.  This was seen in the logs.

fixes #724

### DIFF
--- a/pkg/controller/discovery/container/velero.go
+++ b/pkg/controller/discovery/container/velero.go
@@ -397,7 +397,11 @@ func (r *PodVolumeBackup) Create(e event.CreateEvent) bool {
 	if !cast {
 		return false
 	}
-	backup := model.PodVolumeBackup{}
+	backup := model.PodVolumeBackup{
+		Base: model.Base{
+			Cluster: r.ds.Cluster.PK,
+		},
+	}
 	backup.With(object)
 	r.ds.Create(&backup)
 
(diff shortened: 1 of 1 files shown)


djwhatle (approved):

djwhatle:
Thanks Jeff!

Issue #685, opened 2020-09-23
[MIG-119] Allow showing progress of ongoing phase in MigMigration CR


pranavgaikwad:
**Is your feature request related to a problem? Please describe.**
Currently, an ongoing phase in migration doesn't have a way to show its progress. Implement a new field / condition to allow relaying progress to the MigMigration CR. 
 
**Describe the solution you'd like**
I'd like to see the progress of an ongoing phase in MigMigration CR. This can be simply implemented by adding a new field `progress` in the Running condition of the ongoing phase. 

Few things to consider:
* Velero Backup can have multiple PodVolumeBackup resources associated with it. How would you show progress of all of the associated PVBs in one field? 

**Describe alternatives you've considered**
Instead of showing progress in a field, we can create a new condition specifically for showing progress which can be associated with the ongoing phase. 

**Additional context**

pranavgaikwad:
Closed through #706

PR #762, opened 2020-11-10
[MIG-119] Add: Requeue to Annotation/Label phase

JaydipGabani:
Adding Requeue logic to `AnnotateResource` phase to allow mid-phase updates to `migmigration` CR and break the phase between multiple reconcile. A bit skeptical about having a 3-second interval for a requeue. On my cluster, this interval is way to large to have a requeue. The entire phase takes less than 3 seconds for the migration I carried out.

### DIFF
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -98,49 +98,49 @@ func (t *Task) annotateStageResources() error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	itemsUpdate := 0
+	itemsUpdated := 0
 	// Namespaces
-	itemsUpdate, err = t.labelNamespaces(sourceClient, itemsUpdate)
+	itemsUpdated, err = t.labelNamespaces(sourceClient, itemsUpdated)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	if itemsUpdate > 50 {
+	if itemsUpdated > 50 {
 		t.Requeue = FastReQ
 		return nil
 	}
 	// Pods
-	itemsUpdate, serviceAccounts, err := t.annotatePods(sourceClient, itemsUpdate)
+	itemsUpdated, serviceAccounts, err := t.annotatePods(sourceClient, itemsUpdated)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	if itemsUpdate > 50 {
+	if itemsUpdated > 50 {
 		t.Requeue = FastReQ
 		return nil
 	}
 	// PV & PVCs
-	itemsUpdate, err = t.annotatePVs(sourceClient, itemsUpdate)
+	itemsUpdated, err = t.annotatePVs(sourceClient, itemsUpdated)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	if itemsUpdate > 50 {
+	if itemsUpdated > 50 {
 		t.Requeue = FastReQ
 		return nil
 	}
 	// Service accounts used by stage pods.
-	itemsUpdate, err = t.labelServiceAccounts(sourceClient, serviceAccounts, itemsUpdate)
+	itemsUpdated, err = t.labelServiceAccounts(sourceClient, serviceAccounts, itemsUpdated)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	if itemsUpdate > 50 {
+	if itemsUpdated > 50 {
 		t.Requeue = FastReQ
 		return nil
 	}
 
-	itemsUpdate, err = t.labelImageStreams(sourceClient, itemsUpdate)
+	itemsUpdated, err = t.labelImageStreams(sourceClient, itemsUpdated)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	if itemsUpdate > 50 {
+	if itemsUpdated > 50 {
 		t.Requeue = FastReQ
 		return nil
 	}
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -139,7 +139,7 @@ func (t *Task) annotateStageResources() (bool, error) {
 	if itemsUpdated > 50 {
 		return false, nil
 	}
-	return false, nil
+	return true, nil
 }
 
 // Gets a list of restic volumes and restic verify volumes for a pod
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -2,7 +2,9 @@ package migmigration
 
 import (
 	"context"
+	"fmt"
 	"strings"
+	"time"
 
 	liberr "github.com/konveyor/controller/pkg/error"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
@@ -103,25 +105,40 @@ func (t *Task) annotateStageResources() error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	if t.Requeue == FastReQ {
+		return nil
+	}
 	// Pods
 	serviceAccounts, err := t.annotatePods(sourceClient)
 	if err != nil {
 		return liberr.Wrap(err)
+	}
+	if t.Requeue == FastReQ {
+		return nil
 	}
 	// PV & PVCs
 	err = t.annotatePVs(sourceClient)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	if t.Requeue == FastReQ {
+		return nil
+	}
 	// Service accounts used by stage pods.
 	err = t.labelServiceAccounts(sourceClient, serviceAccounts)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	if t.Requeue == FastReQ {
+		return nil
+	}
 
 	err = t.labelImageStreams(sourceClient)
 	if err != nil {
 		return liberr.Wrap(err)
+	}
+	if t.Requeue == FastReQ {
+		return nil
 	}
 
 	return nil
@@ -164,7 +181,9 @@ func (t *Task) getResticVolumes(client k8sclient.Client, pod corev1.Pod) ([]stri
 
 // Add label to namespaces
 func (t *Task) labelNamespaces(client k8sclient.Client) error {
-	for _, ns := range t.sourceNamespaces() {
+	total := len(t.sourceNamespaces())
+	now := time.Now()
+	for i, ns := range t.sourceNamespaces() {
 		namespace := corev1.Namespace{}
 		err := client.Get(
 			context.TODO(),
@@ -178,6 +197,9 @@ func (t *Task) labelNamespaces(client k8sclient.Client) error {
 		if namespace.Labels == nil {
 			namespace.Labels = make(map[string]string)
 		}
+		if _, exist := namespace.GetLabels()[IncludedInStageBackupLabel]; exist {
+			continue
+		}
 		namespace.Labels[IncludedInStageBackupLabel] = t.UID()
 		err = client.Update(context.TODO(), &namespace)
 		if err != nil {
@@ -187,7 +209,14 @@ func (t *Task) labelNamespaces(client k8sclient.Client) error {
 			"NS annotations/labels added.",
 			"name",
 			namespace.Name)
+
+		if time.Now().Sub(now) == time.Second*3 {
+			t.Requeue = FastReQ
+			t.setProgress([]string{fmt.Sprintf("%v/%v Namespaces labeled", i, total)})
+			return nil
+		}
 	}
+	t.Requeue = NoReQ
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/migmigration/annotation.go
```diff
@@ -184,7 +206,14 @@ func (t *Task) labelNamespaces(client k8sclient.Client) error {
 			"NS annotations/labels added.",
 			"name",
 			namespace.Name)
+
+		if time.Now().Sub(now) == time.Second*3 {
```
djwhatle, line 210: @JaydipGabani to validate how this change works with large migrations, I'd suggest using [kube-kong-operator](https://github.com/konveyor/kube-kong-operator) to deploy a few thousand resources into your namespaces. If you have questions about using kube-kong, @pranavgaikwad should be able to help you get started.

Once we do this, we can see how many reconciles it takes to get through annotation of all resources.

djwhatle (commented):
Would like to see a benchmark of this against a large migration (~thousands of resources) to understand how the chosen elapsed time allowance affects migration speed.

djwhatle (changes requested):
@JaydipGabani feedback from @sseago and @alaypatel07 indicates that folks would be more comfortable with returning after we complete `x annotations` rather than a timing based approach.

Please adapt the PR to this. Maybe we can start with `50` or `100` annotations before requeuing and reporting progress.

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	itemsUpdate := 0
```
djwhatle, line 98: Assume this var name is supposed to be `itemsUpdated`

pkg/controller/migmigration/annotation.go
```diff
@@ -170,22 +192,30 @@ func (t *Task) labelNamespaces(client k8sclient.Client) error {
...
-			return liberr.Wrap(err)
+			return itemsUpdated, liberr.Wrap(err)
 		}
 		log.Info(
 			"NS annotations/labels added.",
 			"name",
 			namespace.Name)
+		itemsUpdated++
+		if itemsUpdated > 50 {
+			t.setProgress([]string{fmt.Sprintf("%v/%v Namespaces labeled", i, total)})
```
djwhatle, line 214: @pranavgaikwad can you confirm this use of `t.setProgress` is correct? It looks like this is wiping out any existing progress and creating a new progress string array. 

In the next function within this same phase the same thing will happen with a different progress message. 
```
t.setProgress([]string{fmt.Sprintf("%v/%v Pod annotations/labels added.", i, total)})
```

JaydipGabani: @djwhatle if `t.setProgress` is used once in a single reconcile it should be fine. And that is what will happen here always since we are requeuing right after we set to progress. thoughts? @pranavgaikwad . We can also pass `progress` array between different functions just like `lastUpdate`, but I am not sure if that is useful. any thoughts?

pkg/controller/migmigration/annotation.go
JaydipGabani: I missed that, changing the name of the variable

pranavgaikwad:
```suggestion
	itemsUpdated := 0
```

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	itemsUpdate := 0
 	// Namespaces
-	err = t.labelNamespaces(sourceClient)
+	itemsUpdate, err = t.labelNamespaces(sourceClient, itemsUpdate)
```
pranavgaikwad, line 100:
```suggestion
	itemsUpdated, err = t.labelNamespaces(sourceClient, itemsUpdated)
```

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
...
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	if itemsUpdate > 50 {
```
pranavgaikwad, line 104:
```suggestion
	if itemsUpdated > 50 {
```

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
...
+		t.Requeue = FastReQ
+		return nil
+	}
 	// Pods
-	serviceAccounts, err := t.annotatePods(sourceClient)
+	itemsUpdate, serviceAccounts, err := t.annotatePods(sourceClient, itemsUpdate)
```
pranavgaikwad, line 109:
```suggestion
	itemsUpdated, serviceAccounts, err := t.annotatePods(sourceClient, itemsUpdated)
```

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
...
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	if itemsUpdate > 50 {
```
pranavgaikwad, line 113:
```suggestion
	if itemsUpdated > 50 {
```

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
...
+		t.Requeue = FastReQ
+		return nil
+	}
 	// PV & PVCs
-	err = t.annotatePVs(sourceClient)
+	itemsUpdate, err = t.annotatePVs(sourceClient, itemsUpdate)
```
pranavgaikwad, line 118:
```suggestion
	itemsUpdated, err = t.annotatePVs(sourceClient, itemsUpdated)
```

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
...
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	if itemsUpdate > 50 {
```
pranavgaikwad, line 122:
```suggestion
	if itemsUpdated > 50 {
```

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
...
+		t.Requeue = FastReQ
+		return nil
+	}
 	// Service accounts used by stage pods.
-	err = t.labelServiceAccounts(sourceClient, serviceAccounts)
+	itemsUpdate, err = t.labelServiceAccounts(sourceClient, serviceAccounts, itemsUpdate)
```
pranavgaikwad, line 127:
```suggestion
	itemsUpdated, err = t.labelServiceAccounts(sourceClient, serviceAccounts, itemsUpdated)
```

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
...
 	if err != nil {
 		return liberr.Wrap(err)
 	}
+	if itemsUpdate > 50 {
```
pranavgaikwad, line 131:
```suggestion
	if itemsUpdated > 50 {
```

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
...
+		t.Requeue = FastReQ
+		return nil
+	}
 
-	err = t.labelImageStreams(sourceClient)
+	itemsUpdate, err = t.labelImageStreams(sourceClient, itemsUpdate)
```
pranavgaikwad, line 136:
```suggestion
	itemsUpdated, err = t.labelImageStreams(sourceClient, itemsUpdated)
```

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
...
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-
+	if itemsUpdate > 50 {
```
pranavgaikwad, line 140:
```suggestion
	if itemsUpdated > 50 {
```

pkg/controller/migmigration/annotation.go
```diff
@@ -160,8 +181,9 @@ func (t *Task) getResticVolumes(client k8sclient.Client, pod corev1.Pod) ([]stri
 }
 
 // Add label to namespaces
-func (t *Task) labelNamespaces(client k8sclient.Client) error {
-	for _, ns := range t.sourceNamespaces() {
+func (t *Task) labelNamespaces(client k8sclient.Client, itemsUpdated int) (int, error) {
```
pranavgaikwad, line 184: Question : I am just curious, I don't think I completely understand the logic. Why pass `itemsUpdated` to each of these individual functions? Can't they just report the number of items they updated?

pranavgaikwad (changes requested):
Requesting explanation on passing `itemsUpdated` down to individual function. I think the concerns should be separated.

pkg/controller/migmigration/annotation.go
JaydipGabani: Passing `itemsUpdated` makes sure that we get an update after every 50th object is processed. We can report the updated items within a function after a fixed number of objects are updated without passing the variable (this was originally the proposed solution), but this approach is implemented based on the feedback I received from the team.

pkg/controller/migmigration/annotation.go
djwhatle: I think this is OK.

eriknelson:
Ah edited this title but actually we don't need to tag PRs with the `[MIG-XXX]` tag, just issues.

djwhatle:
Related to #685

pkg/controller/migmigration/annotation.go
djwhatle: @pranavgaikwad the idea here is to make sure that we process 50 items total per reconcile.

pkg/controller/migmigration/task.go
```diff
@@ -457,8 +457,10 @@ func (t *Task) Run() error {
 		if err != nil {
 			return liberr.Wrap(err)
 		}
-		if err = t.next(); err != nil {
-			return liberr.Wrap(err)
+		if t.Requeue != FastReQ {
```
djwhatle, line 460: @JaydipGabani I don't think we should be using `t.Requeue` as a signal for whether `t.annotateStageResources()` is done.

djwhatle: I suggest adapting `t.annotateStageResources()` to directly return a value indicating  if it has completed all annotations.

```
finished, err := t.annotateStageResources()
// if !finished, allow default FastReQ to be used
// if finished, proceed to next phase with `t.next()`
```

pkg/controller/migmigration/annotation.go
```diff
@@ -95,32 +95,53 @@ func (t *Task) annotateStageResources() error {
...
+	itemsUpdated, err = t.labelImageStreams(sourceClient, itemsUpdated)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-
+	if itemsUpdated > 50 {
+		t.Requeue = FastReQ
+		return nil
+	}
+	t.Requeue = NoReQ
```
djwhatle, line 144: We should not be setting t.Requeue from inside of here, see other comment.

djwhatle: We also don't want to use `NoReQ`, or the process will halt.

djwhatle (changes requested):
Getting closer! Please see comments.

pkg/controller/migmigration/annotation.go
```diff
@@ -90,59 +90,53 @@ type ServiceAccounts map[string]map[string]bool
...
+		return false, liberr.Wrap(err)
 	}
 	if itemsUpdated > 50 {
-		t.Requeue = FastReQ
-		return nil
+		return false, nil
 	}
-	t.Requeue = NoReQ
-	return nil
+	return false, nil
```
djwhatle, line 139: It looks like you are never returning `true`, so the phase will never finish.

pkg/controller/migmigration/annotation.go
```diff
@@ -318,21 +363,26 @@ func (t *Task) annotatePVs(client k8sclient.Client) error {
...
+			return itemsUpdated, liberr.Wrap(err)
 		}
 		log.Info(
 			"PVC annotations/labels added.",
 			"ns",
 			pv.PVC.Namespace,
 			"name",
 			pv.PVC.Name)
+		itemsUpdated++
+		if itemsUpdated > 50 {
```
djwhatle, line 375: Last thing I'd recommend is making this number into a const so that we can change it from one place.

djwhatle (approved):
LGTM. I tried this out and saw a migration complete. I didn't see any annotation progress reflected but probably because I had <50 things to annotate.

pkg/controller/migmigration/annotation.go
```diff
@@ -78,6 +78,9 @@ const (
 	StagePodLabel = "migration.openshift.io/is-stage-pod"
 )
 
+// Bucket limit for number of items annotated in one Reconcile
+const NumberOfItems = 50
```
djwhatle, line 82: Let's name this to something more specific since this is a package level const. 

Perhaps `AnnotationsPerReconcile`

djwhatle (approved):

PR #771, opened 2020-11-12
Only set NoReQ if we're waiting for dim to complete

sseago:


pkg/controller/directimagemigration/task.go
```diff
@@ -169,7 +169,6 @@ func (t *Task) Run() error {
 		if err != nil {
 			return liberr.Wrap(err)
 		}
-		t.Requeue = NoReQ
```
djwhatle, line 172: Ah, this explains the problem

djwhatle (approved):

pkg/controller/directimagemigration/task.go
```diff
@@ -202,6 +201,7 @@ func (t *Task) Run() error {
 			if len(failedISList) > 0 {
 				t.fail(MigrationFailed, failedISList)
 			} else {
+				t.Requeue = PollReQ
```
djwhatle, line 204: PollReQ works here. You can also just not set it and it will stay as the default `FastReQ` I believe.

djwhatle: Actually, not sure what the default is in the DIM controller task.go

PR #773, opened 2020-11-12
Unquiesce apps on destination after final restore

### LINKED ISSUE #701: [MIG-249] Apps should be un-quiesced at appropriate times

djwhatle:
Closes #701 

After further consideration and consultation with @sseago, decided to re-use the logic already in the controller for un-quiescing in the case where an app that is already quiesced is migrated.

**Testing Results**

I saw this PR successfully scale an nginx-pv deployment back to the original replica count when it was quiesced on my src cluster at the start of the final migmigration run.


pkg/controller/migmigration/quiesce.go
```diff
@@ -123,7 +143,7 @@ func (t *Task) quiesceDeploymentConfigs(client k8sclient.Client) error {
 }
 
 // Scales DeploymentConfig back up on source cluster
-func (t *Task) unQuiesceDeploymentConfigs(client k8sclient.Client) error {
+func (t *Task) unQuiesceDeploymentConfigs(client k8sclient.Client, namespaces []string) error {
 	for _, ns := range t.sourceNamespaces() {
```
djwhatle, line 147: I missed this one somehow

aufi:
Works well for me :+1: Tested a quiesced source app which has been unquiesced on the destination cluster during migration (e.g. also daemonsets un-disabled by node selector, cronjob un-suspended).

aufi (approved):

sseago (approved):
LGTM

Issue #775, opened 2020-11-13
Research RFE: investigate migration completion time for 1.4.0


alaypatel07:
**Is your feature request related to a problem? Please describe.**
WIth the amount of features going into 1.4.0, it felt the migrations were slower than usual. I compared a QE test case `ocp-24787-redis.yml` to compare the time of entire migration. For konveyor master it was about 10 mins and for 1.3.0 it was about 6min20 seconds. Both the installations were in the destination cluster. First I tried the master then I uninstalled MTC latest to install 1.3.0 and tried the 1.3.0, if the order if important for reproducing this.

A screenshot of time spent in each step of the pipeline for more debugging.
![image](https://user-images.githubusercontent.com/8192650/99101360-828ed980-25aa-11eb-8501-0272d59c1511.png)

**Describe the solution you'd like**
We should probably bring down the time spent for the entire migration for closer to 1.3.0. If it is not possible we should have a reason why this time is justified. For example, we spend a lot of time cleaning up, hence it is important and will improve the overall migration at scale with small prices to pay upfront.

djwhatle:
There are a few new phases in mig-controller using a 3 second requeue where they should use the default `FastReQ` for transitioning to the next phase. These are quick fixes. I'll post a PR for this soon.

PR #776, opened 2020-11-13
Default to FastReQ unless PollReQ is more appropriate in task.go

djwhatle:
Part of solution to #775 

FastReQ is the default set by `task.init()`. We should use this unless we are waiting on something to finish. 

Removed 13 instances of PollReQ. In the best case this could be a ~39 second improvement (3 seconds per incorrectly used PollReQ)


djwhatle:
Tested this with a migration, completed as expected.

alaypatel07 (approved):
thanks @djwhatle, LTGM

PR #787, opened 2020-11-13
Bubble up progress report to dvm from dvmp

alaypatel07:
This PR add owner references to `dvmp` CR. It also add progress percent and transfer rate to DVM CR status in the runningPods section.

It also refactors the leftover files for consistency in CR name


config/samples/migration_v1alpha1_directpvmigrationprogress.yaml
```diff
@@ -3,7 +3,7 @@ kind: DirectVolumeMigrationProgress
 metadata:
   labels:
     controller-tools.k8s.io: "1.0"
-  name: directpvmigrationprogress-sample
+  name: directvolumemigrationprogress-sample
 spec:
   # Add fields here
   foo: bar
```
dymurray, line 9: Do we want actual spec fields here?

dymurray (approved):

config/samples/migration_v1alpha1_directpvmigrationprogress.yaml
sseago: Users won't create these, though will they? If these are only ever controller-generated, do we need samples?

dymurray: I was told to keep these up to date for users so unsure

sseago (approved):

PR #788, opened 2020-11-16
Direct Migrations

alaypatel07:
This rebases the direct-migration topic branch onto master for merge reviews


pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -22,14 +22,20 @@ import (
...
+}
+
 // DirectVolumeMigrationSpec defines the desired state of DirectVolumeMigration
 type DirectVolumeMigrationSpec struct {
-	SrcMigClusterRef            *kapi.ObjectReference   `json:"srcMigClusterRef,omitempty"`
-	DestMigClusterRef           *kapi.ObjectReference   `json:"destMigClusterRef,omitempty"`
-	PersistentVolumeClaims      []*kapi.ObjectReference `json:"persistentVolumeClaims,omitempty"`
-	StorageClassMapping         map[string]string       `json:"storageClassMapping,omitempty"`
-	CreateDestinationNamespaces bool                    `json:"createDestinationNamespaces,omitempty"`
-	DeleteProgressReportingCRs  bool                    `json:"deleteProgressReportingCRs,omitempty"`
```
alaypatel07, line 32: we probably need this

pkg/controller/migmigration/task.go
```diff
@@ -21,84 +21,94 @@ var NoReQ = time.Duration(0)
...
-	InitialBackupFailed             = "InitialBackupFailed"
-	AnnotateResources               = "AnnotateResources"
-	EnsureStagePodsFromRunning      = "EnsureStagePodsFromRunning"
-	EnsureStagePodsFromTemplates    = "EnsureStagePodsFromTemplates"
-	EnsureStagePodsFromOrphanedPVCs = "EnsureStagePodsFromOrphanedPVCs"
-	StagePodsCreated                = "StagePodsCreated"
-	StagePodsFailed                 = "StagePodsFailed"
-	SourceStagePodsFailed           = "SourceStagePodsFailed"
-	RestartVelero                   = "RestartVelero"
+	CleanStaleResticCRs             = "CleanStaleResticCRs"CleanStaleStagePods                    = "CleanStaleStagePods"
```
alaypatel07, line 28: needs fixing

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
alaypatel07: its as expected!!

pkg/controller/migmigration/task.go
```diff
@@ -386,7 +409,39 @@ func (t *Task) Run() error {
...
+		if dim == nil {
+			return errors.New("DirectImageMigration not found")
+		}
 
+		completed, reasons := dim.HasCompleted()
+		t.Log.Info("is migrations", "name", dim.Name, "completed", completed, "phase", dim.Status.Phase)
+
+		if completed {
+			if len(reasons) > 0 {
+				t.fail(MigrationFailed, reasons)
```
djwhatle, line 436: So the whole migration is considered to be failed if any image migrations fail, is this correct?

pkg/controller/migmigration/task.go
```diff
@@ -574,6 +629,53 @@ func (t *Task) Run() error {
...
 		}
+	case CreateDirectVolumeMigration:
+		err := t.createDirectVolumeMigration()
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		if err = t.next(); err != nil {
+			return liberr.Wrap(err)
+		}
+	case DirectVolumeMigrationStarted:
```
djwhatle, line 640: I would remove this phase.

pkg/controller/migmigration/task.go
```diff
@@ -574,6 +629,53 @@ func (t *Task) Run() error {
...
+		// FIXME: currently a placefiller
+		if err = t.next(); err != nil {
+			return liberr.Wrap(err)
+		}
+	case WaitForDirectVolumeMigrationToComplete:
+		dvm, err := t.getDirectVolumeMigration()
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		// if no dvm, continue to next task
```
djwhatle, line 658: Having no DVM is OK? I assumed this phase would only run if direct migration flags are set and volumes exist.

pkg/controller/migmigration/task.go
```diff
@@ -386,7 +409,39 @@ func (t *Task) Run() error {
...
+		dim, err := t.getDirectImageMigration()
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		if dim == nil {
+			return errors.New("DirectImageMigration not found")
+		}
 
+		completed, reasons := dim.HasCompleted()
+		t.Log.Info("is migrations", "name", dim.Name, "completed", completed, "phase", dim.Status.Phase)
```
djwhatle, line 432: Feels like this info should perhaps be rolled into `t.SetProgress()` instead of just logged on every reconcile

djwhatle (changes requested):
Added some comments.

pkg/controller/migmigration/task.go
```diff
@@ -21,84 +21,97 @@ var NoReQ = time.Duration(0)
...
+	EnsureStageBackup                      = "EnsureStageBackup"
+	StageBackupCreated                     = "StageBackupCreated"
+	StageBackupFailed                      = "StageBackupFailed"
+	EnsureInitialBackupReplicated          = "EnsureInitialBackupReplicated"
+	EnsureStageBackupReplicated            = "EnsureStageBackupReplicated"
+	EnsureStageRestore                     = "EnsureStageRestore"
+	StageRestoreCreated                    = "StageRestoreCreated"
+	StageRestoreFailed                     = "StageRestoreFailed"
+	CreateDirectVolumeMigration            = "CreateDirectVolumeMigration"
+	DirectVolumeMigrationStarted           = "DirectVolumeMigrationStarted"
```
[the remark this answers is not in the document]
djwhatle: Noticed later on that the `DirectVolumeMigrationStarted` phase is a noop, recommend removing

config/crds/migration_v1alpha1_migplan.yaml
```diff
@@ -66,6 +66,10 @@ spec:
                 - serviceAccount
                 type: object
               type: array
+            indirectImageMigration:
```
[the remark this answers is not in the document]
sseago: Yes, this was a deliberate choice because we wanted the direct option to be the default -- not specifying these fields results in "indirectFoo=false", meaning "we want direct migration".

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -703,23 +713,23 @@ func (t *Task) createPVProgressCR() error {
...
 	t.Owner.Status.FailedPods = []*corev1.ObjectReference{}
 	t.Owner.Status.SuccessfulPods = []*corev1.ObjectReference{}
 
 	pvcMap := t.getPVCNamespaceMap()
 	for ns, vols := range pvcMap {
 		for _, vol := range vols {
-			pod := corev1.Pod{}
-			err = srcClient.Get(context.TODO(), types.NamespacedName{
+			dvmp := migapi.DirectVolumeMigrationProgress{}
+			err = destClient.Get(context.TODO(), types.NamespacedName{
```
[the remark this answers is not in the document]
alaypatel07: +1 yes, good catch, will fix

pkg/controller/migmigration/task.go
```diff
@@ -125,10 +138,14 @@ var StageItinerary = Itinerary{
 		{Name: RestartVelero, Step: StepPrepare},
 		{Name: CleanStaleStagePods, Step: StepPrepare},
 		{Name: WaitForStaleStagePodsTerminated, Step: StepPrepare},
-		{Name: CreateRegistries, Step: StepPrepare},
-		{Name: EnsureStagePodsFromRunning, Step: StepStageBackup, all: HasPVs},
-		{Name: EnsureStagePodsFromTemplates, Step: StepStageBackup, all: HasPVs},
-		{Name: EnsureStagePodsFromOrphanedPVCs, Step: StepStageBackup, all: HasPVs},
+		{Name: CreateRegistries, Step: StepPrepare, all: IndirectImage},
+		{Name: EnsureCloudSecretPropagated, Step: StepPrepare},
```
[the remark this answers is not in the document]
alaypatel07: looks like a rebase error, +1, will fix

config/samples/migration_v1alpha1_directpvmigrationprogress.yaml
```diff
@@ -3,7 +3,11 @@ kind: DirectVolumeMigrationProgress
...
     controller-tools.k8s.io: "1.0"
-  name: directpvmigrationprogress-sample
+  name: directvolumemigrationprogress-sample
 spec:
-  # Add fields here
-  foo: bar
+  clusterRef:
+    name: sample-source-3110
+    namespace: openshift-migration
+  podRef:
```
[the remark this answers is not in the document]
sseago: Also, it looks like this missed out on the rename to directvolumemigrationprogress.

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -566,8 +567,21 @@ func (m *MigCluster) GetInternalRegistryPath(c k8sclient.Client) (string, error)
 	}
 }
 
+func (m *MigCluster) SetRegistryPath(c k8sclient.Client) error {
+	newRegistryPath, err := m.GetRegistryPath(c)
+	if err != nil {
+		return err
+	}
+	m.Status.RegistryPath = newRegistryPath
```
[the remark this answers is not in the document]
sseago: @djwhatle We're not using this field on the controller. We explicitly call m.GetRegistryPath() when we need it. I added this field at the request of the UI team so that they could display to the user the registry path that would be used.

pkg/controller/migmigration/task.go
```diff
@@ -163,12 +182,14 @@ var FinalItinerary = Itinerary{
...
 		{Name: CleanStaleStagePods, Step: StepPrepare},
 		{Name: WaitForStaleStagePodsTerminated, Step: StepPrepare},
-		{Name: CreateRegistries, Step: StepPrepare},
+		{Name: CreateRegistries, Step: StepPrepare, all: IndirectImage},
 		{Name: WaitForVeleroReady, Step: StepPrepare},
 		{Name: WaitForResticReady, Step: StepPrepare, any: HasPVs},
-		{Name: WaitForRegistriesReady, Step: StepPrepare},
+		{Name: WaitForRegistriesReady, Step: StepPrepare, all: IndirectImage},
+		{Name: CreateDirectImageMigration, Step: StepDirectImage, all: DirectImage},
 		{Name: EnsureCloudSecretPropagated, Step: StepPrepare},
```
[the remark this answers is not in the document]
sseago: @dymurray You're right. I think this was a merge error. I had originally set this to StepPrepare.

pkg/controller/migmigration/task.go
sseago: That is the way we originally wrote this. It's possible we want to revisit given the recent move to parallel execution. Originally this ran and finished first, so the thinking was a failure here would be following the "fail early" model. @dymurray should we modify this to follow the "Completed with warnings" approach? I guess the same question would apply to dvm.

pkg/apis/migration/v1alpha1/migcluster_types.go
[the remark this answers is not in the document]
djwhatle: Gotcha. Sounds good then.

pkg/controller/migmigration/task.go
sseago: This will run if direct volume migration is selected. If there are no filesystem copy volumes to create, I'm assuming we won't have a DVM, so the "wait for complete" step will always move on.

pkg/controller/directimagemigration/task.go
```diff
@@ -202,6 +201,7 @@ func (t *Task) Run() error {
 			if len(failedISList) > 0 {
 				t.fail(MigrationFailed, failedISList)
 			} else {
+				t.Requeue = FastReQ
```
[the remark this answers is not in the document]
sseago: We can probably remove it. After the problems that were happening with PollReQ/NoReQ appearing to have an effect beyond the current step, I wanted to make sure that wasn't happening here.

pkg/controller/migmigration/task.go
sseago: I think the original purpose here was to provide some sort of feedback to the effect of the dvm process has actually started (vs. the CR only being created), but it's probably unnecessary. The dim process doesn't have an equivalent step. I have one early step where the CR is created, and a later step to wait for completion.

pkg/controller/migmigration/registry.go
```diff
@@ -24,34 +24,36 @@ const (
...
-		return nil, liberr.Wrap(errors.New("migration registry service not found"))
-	}
-	registryDC, err := t.PlanResources.MigPlan.GetRegistryDeployment(client)
-	if err != nil {
-		return nil, err
-	}
-	if registryDC == nil {
-		return nil, liberr.Wrap(errors.New("migration registry DeploymentConfig not found"))
-	}
+	if t.PlanResources.MigPlan.Spec.IndirectImageMigration {
```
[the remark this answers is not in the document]
sseago: @djwhatle you mean just something like "Direct migration doesn't require a migration registry" or something like that?

pkg/controller/migmigration/task.go
```diff
@@ -125,10 +138,14 @@ var StageItinerary = Itinerary{
 		{Name: RestartVelero, Step: StepPrepare},
 		{Name: CleanStaleStagePods, Step: StepPrepare},
 		{Name: WaitForStaleStagePodsTerminated, Step: StepPrepare},
-		{Name: CreateRegistries, Step: StepPrepare},
-		{Name: EnsureStagePodsFromRunning, Step: StepStageBackup, all: HasPVs},
-		{Name: EnsureStagePodsFromTemplates, Step: StepStageBackup, all: HasPVs},
-		{Name: EnsureStagePodsFromOrphanedPVCs, Step: StepStageBackup, all: HasPVs},
+		{Name: CreateRegistries, Step: StepPrepare, all: IndirectImage},
```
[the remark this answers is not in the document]
sseago: @djwhatle That's another merge error. The original commit had `all:IndirectImage` there.

pkg/controller/migmigration/task.go
```diff
@@ -125,10 +138,14 @@ var StageItinerary = Itinerary{
 		{Name: RestartVelero, Step: StepPrepare},
 		{Name: CleanStaleStagePods, Step: StepPrepare},
 		{Name: WaitForStaleStagePodsTerminated, Step: StepPrepare},
```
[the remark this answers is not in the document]
sseago: @djwhatle We don't use stage pods for direct image migration. Doesn't `WaitForStaleStagePodsTerminated` mainly deal with stage pods from prior migrations, though? Prior migrations may or may not have used direct image migration, so it's possible that there are still stale pods.

pkg/controller/migmigration/task.go
dymurray: Yeah Scott's right here. Intention was to update migmigration with some status reporting once it started prior to moving on but sounds like we may have this resolved with Alay's progress reporting changes anyway.

pkg/controller/migmigration/task.go
dymurray: +1 I think this should follow the "Completed with warnings" approach.... though there is a concern that all the resources that get restored which depend upon the volume/image may be borked.

But I suppose that's consistent with our other approach returning "Completed with warnings" so I think thats fine

sseago: I think we're in agreement that we should move to the "Completed With Warnings" behavior, but lets handle that as follow-on issues. I've filed separate issues for DIM and DVM:
https://github.com/konveyor/mig-controller/issues/790
https://github.com/konveyor/mig-controller/issues/789

pkg/controller/migmigration/dvm.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return err
+	}
+	// If already created exit with no error
+	if existingDvm != nil {
+		return nil
+	}
+	dvm := t.buildDirectVolumeMigration()
+	if dvm == nil {
+		return errors.New("failed to build directvolumeclaim list")
```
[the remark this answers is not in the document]
dymurray: I believe this should be an error that should surface to the user only because it means something went catastrophically wrong. This would mean that the controller has indicated we are supposed to be creating a DVM but building up the resource returned nothing. that's my take at least

pkg/controller/migmigration/dvm.go
```diff
@@ -0,0 +1,166 @@
...
+	if len(list.Items) > 0 {
+		return &list.Items[0], nil
+	}
+
+	return nil, nil
+}
+
+// Check if the DVM has completed.
+// Returns if it has completed, why it failed, and it's progress results
+func (t *Task) hasDirectVolumeMigrationCompleted(dvm *migapi.DirectVolumeMigration) (completed bool, failureReasons, progress []string) {
```
dymurray, line 87: There are improvements I'd like to make on this function but with the progress enhancements coming down the pipeline I think this is fine as is

pkg/controller/migmigration/dvm.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (t *Task) getDirectVolumeClaimList() *[]migapi.PVCToMigrate {
+	pvcList := []migapi.PVCToMigrate{}
+	for _, pv := range t.PlanResources.MigPlan.Spec.PersistentVolumes.List {
+		if pv.Selection.Action != migapi.PvCopyAction && pv.Selection.CopyMethod != migapi.PvFilesystemCopyMethod {
+			continue
+		}
+		accessModes := pv.PVC.AccessModes
+		/* TODO: When access mode selection is available override this
```
dymurray, line 126: Do we enable this now? Or wait until UI supports? Talking with @sseago it sounded like we want this supported now since it works in the API. If so I can follow up with a PR.

sseago: @dymurray I think we'd better enable this now (or in a follow-up PR) since otherwise this would be a regression for something supported when using restic.

pkg/controller/migmigration/task.go
```diff
@@ -1213,6 +1366,32 @@ func (t *Task) hasVerify() bool {
...
+
+// Returns true if the IndirectVolumeMigration override on the plan is set (plan is configured not to do direct migration)
+func (t *Task) indirectVolumeMigration() bool {
+	return t.PlanResources.MigPlan.Spec.IndirectVolumeMigration
+}
+
+// Returns true if the IndirectVolumeMigration override on the plan is not set (plan is configured to do direct migration)
+// There must exist a set of direct volumes for this to return true
+func (t *Task) directVolumeMigration() bool {
+	return !t.indirectVolumeMigration() && t.hasDirectVolumes()
```
sseago, line 1387: So one concern I have adding this here is it means we end up generating the direct volume migration list more than once. `hasDirectVolumes` iterates over the whole PV list, figures out which ones are needed, and then figures out access modes, etc. and builds the list, where we use it here to determine whether the list is empty or not. Then we throw it away and buildDirectVolumeMigration calls it again. We then call it again when deciding whether we need to wait on the DVM. I think it would be more efficient to just go to the createDVM step if the boolean is false, and then there we call hasDirectVolumes() and if it's empty, no DVM is created. Then, when we get to the wait step, we don't have to iterate over the PVC list yet again, we can just do the get for the DVM, and if it's not found, we're done -- there's nothing to wait for, so we call `t.next()`.

dymurray: Great points here. This needs to be addressed.

pkg/controller/migmigration/dvm.go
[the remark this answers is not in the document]
alaypatel07: @dymurray since it is an error state that we want to bubble up to user, should we set a `Warn` Condition here?

also @djwhatle do you mind if we take this as follow on since there needs to be some discussion around how to handle this?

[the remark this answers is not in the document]
sseago: @alaypatel07 Yeah -- see the other discussion with @dymurray, but having no dvm probably shouldn't be an error or a warning, since it just means there weren't any filesystem copy PVCs. If the user chooses move or snapshot copy for everything, we won't create a dvm, and that's fine. Likewise the WaitFor... case won't find a DVM so it won't have to wait -- but I think we can defer that fix (and removing this error) to a follow-on PR.

config/samples/migration_v1alpha1_directpvmigrationprogress.yaml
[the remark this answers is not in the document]
alaypatel07: Added

pkg/controller/directimagemigration/task.go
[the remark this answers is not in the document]
alaypatel07: removed this, PTAL

pkg/controller/migmigration/task.go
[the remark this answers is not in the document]
alaypatel07: Removed

pkg/controller/migmigration/task.go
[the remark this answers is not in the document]
alaypatel07: fixed

pkg/controller/migmigration/task.go
dymurray: Opening follow-up issue.

pkg/controller/migmigration/task.go
dymurray: Create follow-up github issue @sseago

pkg/controller/migmigration/dvm.go
dymurray: This will be tackled in a follow-up PR.

dymurray: https://github.com/konveyor/mig-controller/issues/794

pkg/controller/migmigration/task.go
dymurray: https://github.com/konveyor/mig-controller/issues/795

pkg/controller/migmigration/task.go
sseago: Github issue for DIM progress reporting: https://github.com/konveyor/mig-controller/issues/796
@djwhatle When we implement this, should this be instead of the log, or in addition to it?

djwhatle: @sseago I think it makes sense to log when the status of the DIM changes, but I'm not sure logging the full list of progress every 3 seconds is necessary.

sseago: @djwhatle OK, so when I do the follow-on progress reporting update, I'll change this to `t.Log.V(4).Info...` -- making it a debug-only log.

pkg/controller/migmigration/task.go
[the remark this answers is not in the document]
djwhatle: @sseago yes this deals with terminating stage pods from past migrations. I could see us going either way on this, seems safest to keep it around though. Marking as resolved.

djwhatle (approved):

sseago (approved):

dymurray (approved):

PR #798, opened 2020-11-17
Use ubi8/go-toolset:1.14.7 to avoid ratelimit build failures

djwhatle:
Dockerhub introduced some new rate limiting rules that require a paid account to overcome, and quay.io auto-builds are now  failing most of the time.

To overcome this we changed the builder image to `go-toolset:1.14.7`

_Just to give an idea of the current scope of upstream build failures due to this_
image


jmontleon (approved):
ACK

PR #801, opened 2020-11-18
break out hasPVs into two separate functions

alaypatel07:


pkg/controller/migmigration/task.go
```diff
@@ -1296,18 +1296,22 @@ func (t *Task) getPVCs() map[k8sclient.ObjectKey]migapi.PV {
 
 // Get whether the associated plan lists not skipped PVs.
 // First return value is PVs overall, and second is limited to Move or snapshot copy PVs
```
dymurray, line 1298: Fix comment

dymurray (approved):

pkg/controller/migmigration/task.go
```diff
@@ -1313,19 +1313,22 @@ func (t *Task) getPVCs() map[k8sclient.ObjectKey]migapi.PV {
...
-		}
 		if pv.Selection.Action != migapi.PvSkipAction {
-			anyPVs = true
+			return true
+		}
+	}
+	return false
+}
+
+func (t *Task) isAnyPVsUsingSnapshotMethod() bool {
```
sseago, line 1325: This should be renamed to `isAnyPVsUsingMoveOrSnapshotCopyMethod` and the check needs to include `migapi.PvMoveAction ||`

sseago: Also, either change `are` to `is` in the name, or change `PVs` to `PV`.

sseago (changes requested):
Looks good overall, but the refactor lost the check for `move` PVs. We need to include both Move and Snapshot Copy PVs in the stage backup, which means the check as to whether we need a stage backup also needs to include the move PVs.

pkg/controller/migmigration/task.go
alaypatel07: +1 good catch, I updated the PR with suggested changes

djwhatle:
@alaypatel07 @dymurray @sseago what's left on this PR?

alaypatel07:
too late to merge this now

PR #811, opened 2020-11-20
Bug 1883212: Warning message should be given to user indicating empty project migration.

seshapad:
Following snapshots indicate the warning messages introduced by the bug-fix, for empty and non-empty (nginx) namespaces:

### Empty Namespace [Stage]:
image
### Empty Namespace [Migrate]:
image
### NGinx Example Namespace [Stage and Migrate]:
image

Signed-off-by: seshapad <seshapad@in.ibm.com>

### DIFF
--- a/pkg/controller/migmigration/task.go
+++ b/pkg/controller/migmigration/task.go
@@ -335,6 +335,7 @@ func (t *Task) Run() error {
 	t.Requeue = FastReQ
 	t.Log.Info("[RUN]", "stage", t.stage(), "phase", t.Phase)
 
+	// Init method
 	err := t.init()
 	if err != nil {
 		return err
--- a/pkg/controller/migplan/gvk.go
+++ b/pkg/controller/migplan/gvk.go
@@ -1,6 +1,9 @@
 package migplan
 
 import (
+	"fmt"
+	"strings"
+
 	liberr "github.com/konveyor/controller/pkg/error"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	"github.com/konveyor/mig-controller/pkg/gvk"
@@ -23,6 +26,16 @@ func (r ReconcileMigPlan) compareGVK(plan *migapi.MigPlan) error {
 	incompatibleMapping, err := gvkCompare.Compare()
 	if err != nil {
 		err = liberr.Wrap(err)
+	}
+
+	if len(gvkCompare.EmptyNamespaces) == 0 {
+		msg := fmt.Sprintf("Following namespaces are not empty [%s]. Elapsed time %f", strings.Join(gvkCompare.EmptyNamespaces, ", "), gvkCompare.Elapsed)
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     NsEmpty,
+			Status:   True,
+			Category: Warn,
+			Message:  msg,
+		})
 	}
 
 	reportGVK(plan, incompatibleMapping)
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -43,6 +43,7 @@ const (
 	NsNotFoundOnDestinationCluster             = "NamespaceNotFoundOnDestinationCluster"
 	NsLimitExceeded                            = "NamespaceLimitExceeded"
 	NsLengthExceeded                           = "NamespaceLengthExceeded"
+	NsEmpty                                    = "NamespacesEmpty"
 	NsHaveNodeSelectors                        = "NamespacesHaveNodeSelectors"
 	PodLimitExceeded                           = "PodLimitExceeded"
 	SourceClusterProxySecretMisconfigured      = "SourceClusterProxySecretMisconfigured"
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -3,6 +3,7 @@ package gvk
 import (
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/konveyor/controller/pkg/logging"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -59,6 +60,8 @@ type Compare struct {
 	DstDiscovery          discovery.DiscoveryInterface
 	SrcClient             dynamic.Interface
 	CohabitatingResources map[string]*CohabitatingResource
+	EmptyNamespaces       []string
+	Elapsed               float64
 }
 
 // Compare GVKs on both clusters, find incompatible GVKs
@@ -79,16 +82,30 @@ func (r *Compare) Compare() (map[string][]schema.GroupVersionResource, error) {
 		return nil, err
 	}
 
+	nanoToMilli := 1e6
+	start := time.Now()
+	err = r.checkForEmptyNamespaces(preferredSrcResourceList)
+	if err != nil {
+		return nil, err
+	}
+	r.Elapsed = float64(time.Since(start) / nanoToMilli)
+
 	resourcesDiff := r.compareResources(preferredSrcResourceList, dstResourceList)
-	incompatibleGVKs, err := convertToGVRList(resourcesDiff)
+	incompatibleGVRs, err := convertToGVRList(resourcesDiff)
 	if err != nil {
 		return nil, err
 	}
 
 	// Don't report an incompatibleGVK if user settings will skip resource anyways
+	filteredGVRs := r.filterGVRList(incompatibleGVRs)
+
+	return r.collectIncompatibleMapping(filteredGVRs)
+}
+
+func (r *Compare) filterGVRList(gvrList []schema.GroupVersionResource) []schema.GroupVersionResource {
 	excludedResources := toStringSlice(settings.ExcludedInitialResources.Union(toSet(r.Plan.Status.ExcludedResources)))
-	filteredGVKs := []schema.GroupVersionResource{}
-	for _, gvr := range incompatibleGVKs {
+	filteredGVRs := []schema.GroupVersionResource{}
+	for _, gvr := range gvrList {
 		skip := false
 		for _, resource := range excludedResources {
 			if strings.EqualFold(gvr.Resource, resource) {
@@ -96,11 +113,41 @@ func (r *Compare) Compare() (map[string][]schema.GroupVersionResource, error) {
 			}
 		}
 		if !skip {
-			filteredGVKs = append(filteredGVKs, gvr)
+			filteredGVRs = append(filteredGVRs, gvr)
 		}
 	}
 
-	return r.collectIncompatibleMapping(filteredGVKs)
+	return filteredGVRs
+}
+
+// Check for empty namespaces and list them in the compare store
+func (r *Compare) checkForEmptyNamespaces(resourceList []*metav1.APIResourceList) error {
+	gvrList, err := convertToGVRList(resourceList)
+	if err != nil {
+		return err
+	}
+	filteredGVRs := r.filterGVRList(gvrList)
+	nsList := r.Plan.GetSourceNamespaces()
+	for _, ns := range nsList {
+		empty := true
+		for _, gvr := range filteredGVRs {
+			list, err := r.SrcClient.Resource(gvr).Namespace(ns).List(metav1.ListOptions{})
+			if err != nil {
+				return err
+			}
+
+			if len(list.Items) > 0 {
+				empty = false
+				break
+			}
+		}
+
+		if empty {
+			r.EmptyNamespaces = append(r.EmptyNamespaces, ns)
+		}
+	}
+
+	return nil
 }
 
 func toStringSlice(set mapset.Set) []string {


sseago:
So this PR seems to do several things, but I'm not sure it's a net improvement.

The original purpose of the "StageNoOp" warning was to let a user know that they were performing a stage migration on a namespace which didn't have anything to stage -- i.e. it wasn't a mistake to show "success" even though there was no velero backup/restore created, but also to let the user know that for this sort of content to migrate, there's no value in stage, just go to final migration. With this PR we're adding in pod template checking to the same test, which is something we probably don't want to include for stage migration checking (since we'll no longer display the warning if there are no PVs or imagestreams for stage migration, since Deployments, etc. which are only included in final migration are also included in the "artifacts available" check). Also, for final migrations, the pod template check is probably insufficient, since it will flag as "empty" any namespace which doesn't have PVs, Imagestreams, or Pod Templates. The namespace could still have CRs, secrets, etc.

Also, I don't think forcing cancel is what we want here, especially considering the possibility that our "is the user migrating something real?" check may not be 100% accurate.

sseago:
I'm thinking that the right approach here may be more along the lines of leaving the stage migration check/warning as-is but adding a new check (and warning condition) for final migrations, possibly one which makes use of some of our existing health check functionality since I know we're already looking at various resource types. @dymurray is there anything we have already that will give a quick answer to "Is there anything here to migrate?" In any case, I think for now we're probably better off just sticking with the warnings and not forcing cancellation. The downside of not canceling is that a migration with little to do will run to completion fairly quickly. The downside of canceling is that we may be preventing users from migrating actual data for certain edge cases.

seshapad:
@sseago Could I summarize your comments as follows:
1. Remove "cancellation" for empty namespaces in case of both staging and migration.
2. Retain the "stage migration check/warning" as it
3. For the final migration, use existing health checks defined in the mig-controller code which already looks for "Is there anything here to migrate?" (@dymurray could help clarify this).
Have I left anything out here? please clarify.

sseago:
@seshapad Yes, I think that's right. And regarding 3), I'm not entirely sure whether we have anything appropriate. My concern was that "not having PodTemplates" is still a valid use case, so we probably need a more expansive check and it's possible we already have something that could be reused here.

seshapad:
Sure. I will wait for @djwhatle for any pointers he may share on this.

dymurray:
Apologize for the delay in response... but I am not aware of anything that exists today which indicates "is there anything in this namespace to migrate" but it sounds like a function that would be usable for other portions of the migration code.

I will review this PR independent of that, but I think Scott is right it's likely a more expensive check and is worth separating out to a separate check rather than the stage check.

pkg/controller/migmigration/task.go
```diff
@@ -1194,6 +1207,46 @@ func (t *Task) anyFlags(phase Phase) (bool, error) {
...
+	if t.hasStageBackup(anyImageStream, anyPVs, moveSnapshotPVs) {
+		return true, nil
+	}
+
+	client, err := t.getSourceClient()
+	if err != nil {
+		return false, liberr.Wrap(err)
+	}
+
+	podTemplates, err := migpods.ListTemplatePods(client, t.sourceNamespaces())
```
dymurray, line 1240: Would this report empty namespaces for namespaces with deployments that are scaled down to zero?

dymurray:
I agree with Scott. This check should be used for final migrations and this stage check should be left alone. Other than that I have a few questions:

1. I think using `ListTemplatePods` is a good idea to determine if there's anything actually running in the namespace, but I don't think it accurately represents if a namespace is "empty". I think we should change the warning condition to reflect that we are checking for running pods in this case.
2. Is it worth checking for routes/networking objects? Unsure if it's a valid use case but I was trying to think of any reason someone would want a namespace migrated without running pods/objects. 

Other than that I think you are on the right track, please let me know if it's worth setting up a conversation to discuss this further.

pkg/controller/migmigration/task.go
```diff
@@ -995,21 +996,33 @@ func (t *Task) init() error {
...
-	anyPVs, _ := t.hasPVs()
-	if t.stage() && (!anyPVs && !hasImageStreams) {
-		t.Owner.Status.SetCondition(migapi.Condition{
-			Type:     StageNoOp,
-			Status:   True,
-			Category: migapi.Warn,
-			Message:  "Stage migration was run without any PVs or ImageStreams in source cluster. No Velero operations were initiated.",
-			Durable:  true,
-		})
+	if !someArtifactAvailable {
```
dymurray, line 1004: Make this a separate check from the stage check.

sseago:
I think there are various use cases for namespaces that don't have pods, although these are all probably edge cases, so a pod-centric warning message which doesn't fail anything may still be appropriate. Shared CRs could be put in their own namespace (i.e. as we do for openshift-migration, although in that case we also have pods there), or it could be used for images only, etc.

seshapad:
@dymurray @sseago Based on comments you provided here and during the meeting, I am planning below modification to the PR:
1. I will remove "cancellation" of stage and migrate and restore "stage warning" to what it used to be before this change.
2. I will retrieve the miganalytics resource (from within the migrate task) and lookup the analytics (image count, resources, pv etc) it provides for the given namespace and check if they are all zero. If so, I will create a warning (without cancellation) to inform the user that the `so and so resources are not available in the namespace` (Note that I do this only if whatever resources miganalytics compiles turn out to be zero. if any of them are non-zero, I will go by the assumption that the warning is not necessary).

Please let me know if this modification is ok.

sseago:
@seshapad Yes, that sounds like the appropriate way to move forward with this.

seshapad:
> @seshapad Yes, that sounds like the appropriate way to move forward with this.

Thanks @dymurray . I will go ahead with this.

seshapad:
@sseago @dymurray Amended PR now applies the changes discussed here [https://github.com/konveyor/mig-controller/pull/811#issuecomment-740724976] and displays warning based on migrated resource statistics collected by miganalytic:
image
Does this change address the feedback raised previously?

pkg/controller/migmigration/task.go
```diff
@@ -1362,6 +1379,66 @@ func (t *Task) hasImageStreams() (bool, error) {
...
+		log.Trace(err)
+		return false, noMsg, err
+	}
+
+	missingArtifactsInNamespace := make(map[string]string, 0)
+	for _, ns := range analytic.Status.Analytics.Namespaces {
+		missingArtifacts := make([]string, 0)
+		if ns.PVCount == 0 {
+			missingArtifacts = append(missingArtifacts, "Persistent Volumes")
+		}
```
sseago, line 1409: I don't think there's any need to check PV or Image count here, since I think those will be included in the `K8SResourceTotal` -- @djwhatle is this correct?

I think you only need to look at K8SResourceTotal here.

pkg/controller/migmigration/task.go
```diff
@@ -1362,6 +1379,66 @@ func (t *Task) hasImageStreams() (bool, error) {
...
+		if ns.ImageCount == 0 {
+			missingArtifacts = append(missingArtifacts, "Images")
+		}
+		if ns.K8SResourceTotal == 0 || ns.K8SResourceTotal == ns.ExcludedK8SResourceTotal {
```
sseago, line 1413: Also, i don't think you want to compare `K8SResourceTotal` with `ExcludedK8SResourceTotal` -- if a resource is in excluded total, then it won't be added to total. I think the comparison to 0 is all you need here.

pkg/controller/migmigration/task.go
```diff
@@ -1011,6 +1011,23 @@ func (t *Task) init() error {
 			Durable:  true,
 		})
 	}
+
+	someArtifactAvailable, msg, err := t.hasRelevantArtifacts()
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	if !someArtifactAvailable && !t.rollback() {
```
sseago, line 1020: Also add `!t.Stage()` to the comparison here. Stage migrations are already considered above with the `StageNoOp` warning, so we don't want to compare with the mig analytics numbers for stage migrations.

sseago (changes requested):
A few changes -- but we should get confirmation from @djwhatle on the mig analytics question to make sure my understanding here is correct.

pkg/controller/migmigration/task.go
djwhatle: I think @jmontleon would know the answer to this

djwhatle: Looking closer, I think this may be an unintended usage of MigAnalytics, although I could certainly be missing some context.

pkg/controller/migmigration/task.go
```diff
@@ -1362,6 +1379,66 @@ func (t *Task) hasImageStreams() (bool, error) {
 	return false, nil
 }
 
+// Get whether the associated plan has relevant artifacts using the analytics resource.
+func (t *Task) hasRelevantArtifacts() (bool, string, error) {
+	noMsg := ""
+	client, err := t.getDestinationClient()
```
djwhatle, line 1385: This does not seem appropriate, we should be using `t.Client` (controller client) rather than the destination client to interact with MigAnalytics, since the controller client will always provide access to Mig* resources.

pkg/controller/migmigration/task.go
```diff
@@ -1362,6 +1379,66 @@ func (t *Task) hasImageStreams() (bool, error) {
...
+	if err != nil {
+		log.Trace(err)
+		return false, noMsg, err
+	}
+
+	analytic := migapi.MigAnalytic{}
+
+	err = client.Get(context.Background(),
+		k8sclient.ObjectKey{
+			Name:      t.PlanResources.MigPlan.Name,
```
djwhatle, line 1395: This `client.Get()` will fail if the user is doing a CLI driven migration. If I'm remembering correctly, MigAnalytics are only created when doing a UI driven migration.

pkg/controller/migmigration/task.go
```diff
@@ -1362,6 +1379,66 @@ func (t *Task) hasImageStreams() (bool, error) {
...
+			Namespace: t.PlanResources.MigPlan.Namespace,
+		},
+		&analytic)
+	if err != nil {
+		log.Trace(err)
```
djwhatle, line 1400: Instead of doing `log.Trace(err)` anywhere in this PR, we should be wrapping and returning errors.

```
# Convert this...
log.Trace(err)
return false, noMsg, err

# To this
return false, noMsg, liberr.wrap(err)
```

log.Trace() is only used if we are in the top level reconcile() function for one of the controllers.

pkg/controller/migmigration/task.go
```diff
@@ -1362,6 +1379,66 @@ func (t *Task) hasImageStreams() (bool, error) {
 	return false, nil
 }
 
+// Get whether the associated plan has relevant artifacts using the analytics resource.
+func (t *Task) hasRelevantArtifacts() (bool, string, error) {
```
djwhatle, line 1383: Could we change occurrences of the term `Artifacts` to `Resources`? This would be more in line with our terminology throughout the codebase. I don't believe artifacts has an official meaning within the k8s ecosystem.

pkg/controller/migmigration/task.go
```diff
@@ -1362,6 +1379,66 @@ func (t *Task) hasImageStreams() (bool, error) {
...
+		}
+		if ns.ImageCount == 0 {
+			missingArtifacts = append(missingArtifacts, "Images")
+		}
+		if ns.K8SResourceTotal == 0 || ns.K8SResourceTotal == ns.ExcludedK8SResourceTotal {
+			missingArtifacts = append(missingArtifacts, "K8s Resources")
+		}
+
+		if len(missingArtifacts) > 0 {
+			msg := "Missing ["
```
djwhatle, line 1418: I would suggest using golang string formatting tools instead of a for loop here

```
msg := fmt.Sprintf("Missing [%s]", strings.Join(missingArtifacts, ", "))
missingArtifactsInNamespace[ns.Namespace] = msg
```

Output
```
Missing [foo, bar, baz]
```

djwhatle (changes requested):

djwhatle:
@seshapad I am concerned about this PRs usage of MigAnalytics to make its determinations. We can't rely on MigAnalytics being present for a MigPlan. cc @jmontleon

seshapad:
@djwhatle @sseago @dymurray From Dererk's comments, it seems the MigAnalytics approach is not suitable for getting the list of k8s resources. In this scenario, would it be a good idea to obtain the resource list using first principles. For instance:

> func collectResources(client compat.Client) ([]*metav1.APIResourceList, error) {
> 	resources, err := client.ServerResources()
> 	if err != nil {
> 		return nil, err
> 	}
> 
> 	for _, res := range resources {
> 		res.APIResources = namespaced(res.APIResources)
> 		res.APIResources = excludeSubresources(res.APIResources)
> 		res.APIResources = listAllowed(res.APIResources)
> 	}
> 
> 	return resources, nil
> }

> https://github.com/konveyor/mig-controller/pull/811#issuecomment-746492042

pkg/controller/migmigration/task.go
seshapad: Yes. I will make this change.

pkg/controller/migmigration/task.go
jmontleon: We include PV and Image counts in Analytics as @sseago mentioned (unless they are excluded from the migration either by flipping the booleans or adding the appropriate resources to the excludedResources list).

pkg/controller/migmigration/task.go
seshapad: @djwhatle This PR was originally meant to check if there are any resources available in a namespace and if none is available, raise a warning. If the MigAnalytics is not suitable, would the method suggested in [comment](https://github.com/konveyor/mig-controller/pull/811#issuecomment-746547684 ) help?

pkg/controller/migmigration/task.go
seshapad: @sseago This comment and related ones (w.r.t MigAnalytics) might not be applicable following Derek's comments. So, shall I make any changes to this PR based on what we decide based on further discussion?

djwhatle:
@seshapad I was thinking about if there's any way for us to avoid spending extra API requests / migration time to figure this out. What if we were to just examine the completed Velero Backup and pull the "Progress" field contents off to see how many items it actually backed up?  Would that work? @sseago thoughts on if this would accomplish the goals of this fix?

---

If pulling this info off the Velero Backup CR seems unsuitable, I think the "first principles" approach you suggest could definitely work. I think we would only have to run this check once per migration so the additional API calls should only need to happen once. I wonder if we could re-use any of the logic from MigAnalytics controller (e.g. if we relocated some of the logic there into a common package like migapi

```
```

seshapad:
> @seshapad I was thinking about if there's any way for us to avoid spending extra API requests / migration time to figure this out. What if we were to just examine the completed Velero Backup upon completion and pull the "Progress" field contents off to see how many items it actually backed up? Would that work? @sseago thoughts on if this would accomplish the goals of this fix?
> 
> E.g. if totalItems == 0 and phase=Completed || Failed || whatever, raise the condition.
> 
> ```
> status:
>   completionTimestamp: "2020-12-15T15:27:56Z"
>   expiration: "2021-01-14T15:27:52Z"
>   formatVersion: 1.1.0
>   phase: Completed
>   progress:
>     itemsBackedUp: 28
>     totalItems: 28
>   startTimestamp: "2020-12-15T15:27:52Z"
>   version: 1
> ```
> 
> If pulling this info off the Velero Backup CR seems unsuitable, I think the "first principles" approach you suggest could definitely work. I think we would only have to run this check once per migration so the additional API calls should only need to happen once. I wonder if we could re-use any of the logic from MigAnalytics controller (e.g. if we relocated some of the logic there into a common package like migapi
> 
> ```
> 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
> ```
> 
> Not a hard requirement to re-use logic from MigAnalytics if this logic is unique/trivial, but wondering if anything is shared.

Thanks @djwhatle . I will wait for @sseago's comments and share a tentative approach to keep us all on the same page.

sseago:
@djwhatle My initial reaction to your suggestion was that the "what's in the velero backup?" appraoch might be a simpler (and better) approach, but on further thinking, it starts to get messy. The big picture question we're asking is "are we migrating anything?". One velero backup won't tell that. First, we have to look at both initial and stage backups and add up the two totals. But we also have to consider direct migration of PVs and Images, since it's possible that Velero won't actually copy anything because the only thing migrated was images or volumes via direct migration. So we could follow this approach, but we'd need to include both velero backups and DIM and DVM totals. Then the question is which phase of the migration we'd do this at, since we have to look back at four completed phases and consider information from all of them. The "first principles" approach might end up being simpler.

seshapad:
> > @seshapad I was thinking about if there's any way for us to avoid spending extra API requests / migration time to figure this out. What if we were to just examine the completed Velero Backup upon completion and pull the "Progress" field contents off to see how many items it actually backed up? Would that work? @sseago thoughts on if this would accomplish the goals of this fix?
> > E.g. if totalItems == 0 and phase=Completed || Failed || whatever, raise the condition.
> > ```
> > status:
> >   completionTimestamp: "2020-12-15T15:27:56Z"
> >   expiration: "2021-01-14T15:27:52Z"
> >   formatVersion: 1.1.0
> >   phase: Completed
> >   progress:
> >     itemsBackedUp: 28
> >     totalItems: 28
> >   startTimestamp: "2020-12-15T15:27:52Z"
> >   version: 1
> > ```
> > 
> > 
> > If pulling this info off the Velero Backup CR seems unsuitable, I think the "first principles" approach you suggest could definitely work. I think we would only have to run this check once per migration so the additional API calls should only need to happen once. I wonder if we could re-use any of the logic from MigAnalytics controller (e.g. if we relocated some of the logic there into a common package like migapi
> > ```
> > 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
> > ```
> > 
> > 
> > Not a hard requirement to re-use logic from MigAnalytics if this logic is unique/trivial, but wondering if anything is shared.
> 
> Thanks @djwhatle . I will wait for @sseago's comments and share a tentative approach to keep us all on the same page.

@djwhatle Right now, we are raising the warning at the `init` phase of the migration itself.  If we go with the Velero Backup CR approach, we will have to raise the warning condition after the backup phase is done, is it not?

seshapad:
> @djwhatle My initial reaction to your suggestion was that the "what's in the velero backup?" appraoch might be a simpler (and better) approach, but on further thinking, it starts to get messy. The big picture question we're asking is "are we migrating anything?". One velero backup won't tell that. First, we have to look at both initial and stage backups and add up the two totals. But we also have to consider direct migration of PVs and Images, since it's possible that Velero won't actually copy anything because the only thing migrated was images or volumes via direct migration. So we could follow this approach, but we'd need to include both velero backups and DIM and DVM totals. Then the question is which phase of the migration we'd do this at, since we have to look back at four completed phases and consider information from all of them. The "first principles" approach might end up being simpler.

@djwhatle @sseago So...following from both your comments, shall I go by this simple flow:
1. I will add a function (say, `CollectResources()`) which fetches the list of resources:
> migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1" "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
2. Invoke `CollectResources()` with task.go `init()` to check the returned list of resources is non-zero. If zero, I raise warning condition. If not zero, do nothing.

Is this good?

djwhatle:
That sounds reasonable to me. And upon further thinking, I'm not sure migapi is the right place to put this func, probably OK to keep it in the same pkg assuming size is reasonable.

sseago:
@djwhatle I think this all makes sense, but yes if this is only ever called from the controller validation code, then it can just be a non-exported func within the same package.

seshapad:
@djwhatle @sseago I have moved the code to around to api pkg and modified analytic and migration code to use common functions. I will push the commit. Please take a look and let me know if it is appropriate.

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -218,6 +218,87 @@ func (m *MigCluster) BuildRestConfig(c k8sclient.Client) (*rest.Config, error) {
 	return restConfig, nil
 }
 
+// GetResources retrieves a list of resources on a cluster
+func (m *MigCluster) GetResources(c k8sclient.Client) ([]*metav1.APIResourceList, error) {
```
djwhatle, line 222: Mostly already defined in https://github.com/konveyor/mig-controller/blob/master/pkg/gvk/gvk.go#L123

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -218,6 +218,87 @@ func (m *MigCluster) BuildRestConfig(c k8sclient.Client) (*rest.Config, error) {
...
+	for _, res := range resources {
+		if res.Namespaced {
+			filteredList = append(filteredList, res)
+		}
+	}
+
+	return filteredList
+}
+
+func listAllowed(resources []metav1.APIResource) []metav1.APIResource {
```
djwhatle, line 264: https://github.com/konveyor/mig-controller/blob/master/pkg/gvk/gvk.go

pkg/controller/migmigration/task.go
```diff
@@ -1015,6 +1020,22 @@ func (t *Task) init() error {
...
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	if !anyRes && !t.stage() && !t.rollback() {
+		t.Owner.Status.SetCondition(migapi.Condition{
+			Type:     MigNoOp,
+			Status:   True,
+			Category: migapi.Warn,
+			Message:  "Migration was run without any artifacts in source cluster.",
```
djwhatle, line 1034: artifacts => resources

djwhatle (changes requested):
@seshapad requested some changes. Has the performance impact of this PR been measured? I'd be curious to know the elapsed time for this collectResources() call to run on ~10 namespaces. 

From my understanding, this will query at least:
```
# of APIs * # of namespaces included in migration.
```

pkg/controller/migmigration/validation.go
```diff
@@ -36,6 +36,7 @@ const (
 	VeleroFinalRestorePartiallyFailed  = "VeleroFinalRestorePartiallyFailed"
 	DirectImageMigrationFailed         = "DirectImageMigrationFailed"
 	StageNoOp                          = "StageNoOp"
+	MigNoOp                            = "MigrateNoOp"
```
djwhatle, line 39: Suggest renaming this to `MigrateNoOp` to be more specific since `mig` is prefixed to a bunch of stuff.

seshapad:
@djwhatle what is the standard way to measure the timing of the function call in mig controller? Should I simply timestamp the function  and log the difference in time?

djwhatle:
> @djwhatle what is the standard way to measure the timing of the function call in mig controller? Should I simply timestamp the function and log the difference in time?

Here's an example of how to record an elapsed time figure.

https://github.com/konveyor/mig-controller/blob/6c28e299c8086ed7754ab6c2d8a760aecefd93f7/pkg/compat/client.go#L188-L193

```
nanoToMilli = 1e6
start := time.Now()
elapsed := float64(time.Since(start) / nanoToMilli)
# now you can log the elapsed time
```

PR #816, opened 2020-11-20
Adds ability to mark a step 'skipped'

### LINKED ISSUE #815: Add ability to skip a pipeline step
**Describe the bug**
Some steps in the pipeline are evaluated at run-time. Based on the flags, some steps can be skipped even though they were added to the pipeline at the start of the migration. Current pipeline implementation doesn't allow a steps to be marked skipped.

**Expected behavior**
When a step is not run, it should be marked skipped instead of marked 'not started'.

pranavgaikwad:
Fixes #815 

This will help UI show a different class of markers for skipped steps.


eriknelson (approved):
Looks sensible to me, would want someone else from the controller to take a look. @djwhatle ?

djwhatle (approved):

PR #817, opened 2020-11-23
Add progress statistics to failed / partially failed Velero entities

pranavgaikwad:
Adds progress statistics to failed Velero objects. 

This is useful for cases when a Backup / PodVolumeBackup / PodVolumeRestore fails after some amount of work is done.

### DIFF
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -223,9 +223,11 @@ func getPodVolumeBackupsProgress(pvbList *velero.PodVolumeBackupList) (progress 
 				getPVBDuration(&pvb))
 		case velero.PodVolumeBackupPhaseFailed:
 			msg = fmt.Sprintf(
-				"PodVolumeBackup %s/%s: Failed%s",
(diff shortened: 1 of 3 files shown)


pkg/controller/migmigration/backup.go
```diff
@@ -306,16 +308,27 @@ func (t *Task) hasBackupCompleted(backup *velero.Backup) (bool, []string) {
 			backup.Namespace,
 			backup.Name)
 		reasons = append(reasons, message)
+		itemsBackedUp, totalItems := getBackupStats(backup)
+		message = fmt.Sprintf(
+			"%s %d out of estimated total of %d objects backed up%s",
```
dymurray, line 313: Maybe dumb question, do we need a space at the end for %s and the duration? Or is that taken into account?

dymurray: https://github.com/konveyor/mig-controller/blob/4e103fd65be2c2babf7974172647d3674385c817/pkg/controller/migmigration/backup.go#L190

Answered my own question.

dymurray (approved):

PR #808, opened 2020-11-19
Add status.phaseDescription to DVM CR

alaypatel07:
This will need an operator change

Depends on #799

PR #819, opened 2020-11-23
Add cleanup phase to DVM controller task start

### LINKED ISSUE #785: [MIG-284] Add cleanup of direct mig resources to prepare phase
In case there are leftover resources from a DVM we need to ensure they are properly cleaned up prior to migration run or we could be using the wrong set of certs/keys on migration.

dymurray:
TODO: Add extra label for runtime cleanup as opposed to stale cleanup

Fixes #785

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -784,6 +784,110 @@ func (t *Task) deleteRsyncResources() error {
 	return nil
 }
 
+func (t *Task) waitForRsyncResourcesDeleted() (error, bool) {
+	srcClient, err := t.getSourceClient()
+	if err != nil {
+		return err, false
+	}
+	destClient, err := t.getDestinationClient()
+	if err != nil {
+		return err, false
+	}
+	err, deleted := t.areRsyncResourcesDeleted(srcClient)
+	if err != nil {
+		return err, false
+	}
+	if !deleted {
+		return nil, false
+	}
+	err, deleted = t.areRsyncResourcesDeleted(destClient)
+	if err != nil {
+		return err, false
+	}
+	if !deleted {
+		return nil, false
+	}
+	return nil, true
+}
+
+func (t *Task) areRsyncResourcesDeleted(client compat.Client) (error, bool) {
+	selector := labels.SelectorFromSet(map[string]string{
+		"app": "directvolumemigration-rsync-transfer",
+	})
+	pvcMap := t.getPVCNamespaceMap()
+	for ns, _ := range pvcMap {
+		podList := corev1.PodList{}
+		cmList := corev1.ConfigMapList{}
+		svcList := corev1.ServiceList{}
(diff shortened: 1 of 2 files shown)


pkg/controller/directvolumemigration/task.go
```diff
@@ -174,6 +179,30 @@ func (t *Task) Run() error {
...
+		// one label for both is the wrong approach.
+		err := t.deleteRsyncResources()
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		t.Requeue = NoReQ
+		if err = t.next(); err != nil {
+			return liberr.Wrap(err)
+		}
+	case WaitForStaleRsyncResourcesTerminated:
```
pranavgaikwad, line 194: @dymurray can we merge both cases in one case? looks like both are doing exactly same thing. 
```suggestion
	case WaitForStaleRsyncResourcesTerminated, WaitForRsyncResourcesTerminated:
```

pkg/controller/directvolumemigration/task.go
```diff
@@ -330,6 +359,18 @@ func (t *Task) Run() error {
 		if err = t.next(); err != nil {
 			return liberr.Wrap(err)
 		}
+	case WaitForRsyncResourcesTerminated:
```
pranavgaikwad, line 362: possible to merge this above?

pranavgaikwad (changes requested):

pkg/controller/directvolumemigration/task.go
dymurray: Yes... duh. Thanks

pranavgaikwad (approved):
ACK

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -784,8 +784,113 @@ func (t *Task) deleteRsyncResources() error {
...
+			return nil, false
+		}
+	}
+	return nil, true
+
+}
+
 func (t *Task) findAndDeleteResources(client compat.Client) error {
 	// Find all resources with the app label
+	// TODO: This label set should include a DVM run-specific UID.
```
alaypatel07, line 893: nit: probably worth creating an issue if not already

pkg/controller/directvolumemigration/task.go
```diff
@@ -174,6 +179,18 @@ func (t *Task) Run() error {
 		if err = t.next(); err != nil {
 			return liberr.Wrap(err)
 		}
+	case CleanStaleRsyncResources:
+		// TODO Need to add some labels during DVM run to differentiate
```
alaypatel07, line 183: note: with #808 we made a call to try the DVM migration only once and in the future allowing a user-configurable policy on the number of retries. We should probably have the notion which retries number the certs belong to in the label selectors in order to avoid conflicts. but +1 on the TODO

alaypatel07 (approved):
VISIACK, with two comments none requesting changes!

PR #828, opened 2020-12-02
Store SSH key in secret for DVM

### LINKED ISSUE #784: [MIG-284] SSH keys needs to be stored in a secret
To provide idempotence and resiliency, the SSH keys being used for DVM should be kept in a secret rather than in memory at runtime.

dymurray:
Fixes #784

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -148,10 +148,6 @@ func (t *Task) generateSSHKeySecret() error {
 		return err
 	}
 
-	t.SSHKeys = &sshKeys{
-		PublicKey:  &privateKey.PublicKey,
-		PrivateKey: privateKey,
-	}
 	publicRsaKey, err := ssh.NewPublicKey(&privateKey.PublicKey)
 	if err != nil {
 		return err
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -104,14 +104,17 @@ func (t *Task) areRsyncTransferPodsRunning() (bool, error) {
 	// Create rsync client pod on source
 }
 
-// Generate SSH keys to be used
-// TODO: Need to determine if this has already been generated and
-// not to regenerate
-func (t *Task) generateSSHKeys() error {
-	// Check if already generated
-	if t.SSHKeys != nil {
-		return nil
-	}
+func (t *Task) getSSHKeySecret() (*corev1.Secret, error) {
+	secret := corev1.Secret{}
+
+	key := types.NamespacedName{Name: "directvolumemigration-ssh-keys", Namespace: migapi.OpenshiftMigrationNamespace}
+	err := t.Client.Get(context.TODO(), key, &secret)
+	return &secret, err
+}
+
+// Generate SSH keys to be used and ensure secret is created in controller
+// namespace
+func (t *Task) generateSSHKeySecret() error {
 	// Private Key generation
 	privateKey, err := rsa.GenerateKey(rand.Reader, 4096)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/directvolumemigration/rsync.go
```diff
@@ -352,12 +391,10 @@ func (t *Task) createRsyncTransferPods() error {
 
 	// Generate pubkey bytes
 	// TODO: Use a secret for this so we aren't regenerating every time
-	publicRsaKey, err := ssh.NewPublicKey(t.SSHKeys.PublicKey)
+	pubKeyBytes, err := t.getPubKeyBytes()
```
alaypatel07, line 394: is there any plan to mount the secret directly instead of populating `pubKeyBytes` as an environment [variable](https://github.com/konveyor/mig-controller/pull/828/files#diff-7299a3c4e600e2733571d1d39c64bbfe6c7e448a6bb15aa472514a173d3095c0R515)

The problem I see is:
1. Assume a dvm run failed in a way that would leave the pods but delete the secret
2. The pod yaml has data from old secret but the controller creates a new secret in the next run.
3. The controller will not patch the pod yaml with this new secret data.


Had a secret been mounted on the pod directly, if the secret is patched the pods will see the results in around 30 seconds max (I think)

I also think the above mentioned failure scenario is very rare to occur, but just calling out the thats I had that could fail this

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -124,11 +148,26 @@ func (t *Task) generateSSHKeys() error {
...
 	}
-	return nil
+
+	pubKeyBytes := ssh.MarshalAuthorizedKey(publicRsaKey)
+	secret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: migapi.OpenshiftMigrationNamespace,
+			Name:      "directvolumemigration-ssh-keys",
+			Labels: map[string]string{
+				"app": "directvolumemigration-rsync-transfer",
```
alaypatel07, line 162: +1 This label means our pre-existing deletion code will work in garbage collecting this secret after a successful migration.

alaypatel07 (commented):
I think this fixes most failures cases, I added a small question for another rare failure scenario.

I am quickly trying these changes. Will in few mins

alaypatel07 (approved):
Tested this e2e, works, LGTM, thanks @dymurray

pkg/controller/directvolumemigration/rsync.go
dymurray: Yeah interesting point... that is certainly possible as a corner case.

For this to work we'd need to create the secrets in each namespace which comes with it's own set of possible corner cases if one namespace has the secret and the other doesn't between the two clusters.

I think we should perhaps open an issue to track this corner case but moving it to one secret per namespace also scares me so I'd like to think further on it.

pkg/controller/directvolumemigration/rsync.go
dymurray: @alaypatel07 actually.... we won't clean this up with this PR. I took the approach of "create it if it doesn't exist, otherwise use it". So subsequent DVMs would use the same key.

If you think that's a bad idea we need to add openshift-migration NS to the cleanup code which is simple

alaypatel07 (changes requested):
We need more changes being discussed in the slack thread with @dymurray

djwhatle:
@dymurray what are the remaining changes to be made?

dymurray:
After discussing with Alay we are not merging this for 1.4 due to the need to make this more dynamic than reusing the same key. Will target it for 1.4.1. Keeping PR open.

djwhatle:
@dymurray what's the status on this? Should it be merged at this point?

Issue #822, opened 2020-11-30
[MIG-284] Use a randomly generated password for rsync connections


dymurray:
We need to change the implementation of rsync to instead use a randomly generated password that is stored in a secret rather than a hardcoded one.

I don't see much of an issue using the hard-coded one for now but it's an improvement we should make.

PR #836, opened 2020-12-03
Create secret to hold rsync password and read from it

JaydipGabani:
resolves: #822

This PR creates a random password, stores the same in a secret, and reads the password from the secret.

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -6,6 +6,7 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"fmt"
+	random"math/rand"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	"github.com/konveyor/mig-controller/pkg/compat"
 	routev1 "github.com/openshift/api/route/v1"
@@ -19,6 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"text/template"
+	"time"
 )
 
 type pvc struct {
@@ -143,6 +145,15 @@ func (t *Task) createRsyncConfig() error {
 		return err
 	}
 
+	err = t.createRsyncPassword(destClient)
+	if err != nil {
+		return err
+	}
+	password, err := t.getRsyncPassword(destClient)
+	if err != nil {
+		return err
+	}
+
 	// Create rsync configmap/secret on source + destination
 	// Create rsync secret (which contains user/pass for rsync transfer pod) in
 	// each namespace being migrated
@@ -159,7 +170,7 @@ func (t *Task) createRsyncConfig() error {
 			SshUser:   "root",
 			Namespace: ns,
 			PVCList:   pvcList,
-			Password:  "changeme",
+			Password:  string(password),
 		}
 		var tpl bytes.Buffer
 		temp, err := template.New("config").Parse(rsyncConfigTemplate)
@@ -217,7 +228,7 @@ func (t *Task) createRsyncConfig() error {
 				},
 			},
 			Data: map[string][]byte{
-				"RSYNC_PASSWORD": []byte("changeme"),
+				"RSYNC_PASSWORD": password,
 			},
 		}
 		err = srcClient.Create(context.TODO(), &srcSecret)
@@ -235,7 +246,7 @@ func (t *Task) createRsyncConfig() error {
 				},
 			},
 			Data: map[string][]byte{
-				"credentials": []byte("root:changeme"),
+				"credentials": []byte("root:" + string(password)),
 			},
 		}
 		err = destClient.Create(context.TODO(), &destSecret)
@@ -565,6 +576,65 @@ func (t *Task) getRsyncRoute(namespace string) (string, error) {
 	return route.Spec.Host, nil
 }
 
+func (t *Task) createRsyncPassword(destClient compat.Client) error {
+	var letters = []byte("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+	random.Seed(time.Now().UnixNano())
+	password := make([]byte, 6)
+	for i := range password {
+		password[i] = letters[random.Intn(len(letters))]
+	}
+
+	secret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: migapi.OpenshiftMigrationNamespace,
+			Name:      "rsync-pass",
+			Labels: map[string]string{
+				"app": "directvolumemigration-rsync-transfer",
+			},
+		},
+		Data: map[string][]byte{
+			"password": password,
+		},
+	}
+	err := destClient.Create(context.TODO(), &secret)
+	if k8serror.IsAlreadyExists(err) {
+		t.Log.Info("Secret already exists on destination", "namespace", migapi.OpenshiftMigrationNamespace)
+	} else if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (t *Task) getRsyncPassword(destClient compat.Client) ([]byte, error){
+	rsyncSecret := corev1.Secret{}
+	key := types.NamespacedName{Name: "rsync-pass", Namespace: migapi.OpenshiftMigrationNamespace}
+	err := destClient.Get(context.TODO(), key, &rsyncSecret)
+	if err != nil {
+		return nil, err
+	}
+	return rsyncSecret.Data["password"], nil
+}
+
+func (t *Task) deleteRsyncPassword(client compat.Client) error {
+
+	secret := corev1.Secret{}
+	err := client.Get(
+		context.TODO(),
+		types.NamespacedName{Name: "rsync-pass", Namespace: migapi.OpenshiftMigrationNamespace},
+		&secret)
+	if k8serror.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	err = client.Delete(context.TODO(), &secret, k8sclient.PropagationPolicy(metav1.DeletePropagationBackground))
+	if err != nil && !k8serror.IsNotFound(err) {
+		return err
+	}
+	return nil
+}
+
 // Create rsync client pods
 func (t *Task) createRsyncClientPods() error {
 	// Get client for destination
@@ -574,6 +644,14 @@ func (t *Task) createRsyncClientPods() error {
 	}
 
 	pvcMap := t.getPVCNamespaceMap()
+	destClient, err := t.getDestinationClient()
+	if err != nil {
+		return err
+	}
+	password, err := t.getRsyncPassword(destClient)
+	if err != nil {
+		return err
+	}
 	for ns, vols := range pvcMap {
 		// Get stunnel svc IP
 		svc := corev1.Service{}
@@ -607,7 +685,7 @@ func (t *Task) createRsyncClientPods() error {
 				Env: []corev1.EnvVar{
 					{
 						Name:  "RSYNC_PASSWORD",
-						Value: "changeme",
+						Value: string(password),
 					},
 				},
 				TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
@@ -764,6 +842,11 @@ func (t *Task) deleteRsyncResources() error {
 	}
 
 	err = t.findAndDeleteResources(destClient)
+	if err != nil {
+		return err
+	}
+
+	err = t.deleteRsyncPassword(destClient)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/directvolumemigration/rsync.go
```diff
@@ -565,6 +576,65 @@ func (t *Task) getRsyncRoute(namespace string) (string, error) {
...
+	random.Seed(time.Now().UnixNano())
+	password := make([]byte, 6)
+	for i := range password {
+		password[i] = letters[random.Intn(len(letters))]
+	}
+
+	secret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: migapi.OpenshiftMigrationNamespace,
+			Name:      "rsync-pass",
```
dymurray, line 590: `directvolumemigration-rsync-pass` please. Just to keep inline with the other naming

dymurray (dismissed):
Just a nit on the naming but other than that looks good!

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -6,6 +6,7 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"fmt"
+	random"math/rand"
```
alaypatel07, line 9: nit: space needed, `random "math/rand"`

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -143,6 +145,15 @@ func (t *Task) createRsyncConfig() error {
 		return err
 	}
 
+	err = t.createRsyncPassword(destClient)
+	if err != nil {
+		return err
+	}
+	password, err := t.getRsyncPassword(destClient)
```
alaypatel07, line 152: I wonder if this should be the other way around? i.e. first try to get the password from the secret, if the password does not exist, generate the password and create secret?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -565,6 +576,65 @@ func (t *Task) getRsyncRoute(namespace string) (string, error) {
...
+	key := types.NamespacedName{Name: "rsync-pass", Namespace: migapi.OpenshiftMigrationNamespace}
+	err := destClient.Get(context.TODO(), key, &rsyncSecret)
+	if err != nil {
+		return nil, err
+	}
+	return rsyncSecret.Data["password"], nil
+}
+
+func (t *Task) deleteRsyncPassword(client compat.Client) error {
+
```
alaypatel07, line 619: nit: remove the blank space

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -565,6 +576,65 @@ func (t *Task) getRsyncRoute(namespace string) (string, error) {
...
+}
+
+func (t *Task) getRsyncPassword(destClient compat.Client) ([]byte, error){
+	rsyncSecret := corev1.Secret{}
+	key := types.NamespacedName{Name: "rsync-pass", Namespace: migapi.OpenshiftMigrationNamespace}
+	err := destClient.Get(context.TODO(), key, &rsyncSecret)
+	if err != nil {
+		return nil, err
+	}
+	return rsyncSecret.Data["password"], nil
```
alaypatel07, line 615: if the password does not exist, this could panic?  probably worth hiding this behind a check

```
    if pass, ok := rsyncSecret.Data["password"]; ok {
        return pass, nill
    }
    return "", nil
```

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -565,6 +576,65 @@ func (t *Task) getRsyncRoute(namespace string) (string, error) {
...
+}
+
+func (t *Task) deleteRsyncPassword(client compat.Client) error {
+
+	secret := corev1.Secret{}
+	err := client.Get(
```
alaypatel07, line 621: This get is just for delete right? you can probably get rid of the get, create the secret in line 620 with name and namespace and issue a delete, if it does not exist, you'll get a not found error

Example here https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#example-Client--Delete

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -565,6 +576,65 @@ func (t *Task) getRsyncRoute(namespace string) (string, error) {
...
+			Name:      "rsync-pass",
+			Labels: map[string]string{
+				"app": "directvolumemigration-rsync-transfer",
+			},
+		},
+		Data: map[string][]byte{
+			"password": password,
+		},
+	}
+	err := destClient.Create(context.TODO(), &secret)
```
alaypatel07, line 599: we probably want this in the controller cluster and not in the destination cluster. You can do `t.Client.Create(...)` instead

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -574,6 +644,14 @@ func (t *Task) createRsyncClientPods() error {
 	}
 
 	pvcMap := t.getPVCNamespaceMap()
+	destClient, err := t.getDestinationClient()
+	if err != nil {
+		return err
+	}
+	password, err := t.getRsyncPassword(destClient)
```
alaypatel07, line 651: Same comment as above, we want this in the controller cluster, not in the destination cluster `t.getRayncPassword(t.Client)`

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -565,6 +576,65 @@ func (t *Task) getRsyncRoute(namespace string) (string, error) {
...
+
+	secret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: migapi.OpenshiftMigrationNamespace,
+			Name:      "rsync-pass",
+			Labels: map[string]string{
+				"app": "directvolumemigration-rsync-transfer",
+			},
+		},
+		Data: map[string][]byte{
```
alaypatel07, line 595: A question: should we use basic auth secret instead of Opaque.

https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret

alaypatel07 (changes requested):
@JaydipGabani overall idea looks good, added a few questions might want to consider changes on!

pkg/controller/directvolumemigration/rsync.go
dymurray: read through this kind of quickly the first time as well, with travis disabled I realized I took it for granted on the formatting bits.

@JaydipGabani you may want to consider using something like `gofmt` and run it prior to submitting the PRs. Theres some nice go plugins as well that auto adds these spaces/tabs in the right spots when you save the file.

pkg/controller/directvolumemigration/rsync.go
dymurray: +1

pkg/controller/directvolumemigration/rsync.go
dymurray: Yup thanks. I made this same mistake when I was first only testing with destination cluster as controller cluster. Good catch.

pkg/controller/directvolumemigration/rsync.go
dymurray: Yes, I agree. Basic auth is better.

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: @alaypatel07 Does the function `createRsyncConfig()` is invoked over multiple reconcile in the same phase? if it is then it would make sense to first try to get the password from the secret and if it doesn't exist then create a secret. But if it is not, then the secret would get deleted after migration, so it would not exist in the namespace. In that case, creating a secret first would make sense. Let me know if I am misinterpreting your comment here.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -145,11 +145,12 @@ func (t *Task) createRsyncConfig() error {
 		return err
 	}
 
-	err = t.createRsyncPassword(destClient)
+	err = t.createRsyncPassword()
 	if err != nil {
 		return err
 	}
-	password, err := t.getRsyncPassword(destClient)
+	time.Sleep(time.Second)
```
JaydipGabani, line 152: Not sure on waiting for a Second or a few hundred milliseconds.

alaypatel07: @JaydipGabani I think we should not have a sleep in the middle of a reconcile loop. Lets get on a call to talk about this

alaypatel07: I would recommend you change the implementation here as follows:

1. get rsync secret, if you find it, grab the password and use it
2. if the secret is not found, create the password, create the secret and exit the reconcile loop with immediate requeue.

That will remove the sleep in the middle of the reconcile loop.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -131,16 +133,28 @@ func (t *Task) generateSSHKeys() error {
 	return nil
 }
 
-func (t *Task) createRsyncConfig() error {
+func (t *Task) createRsyncConfig() (bool, error) {
```
dymurray, line 136: Why the extra boolean return?

pkg/controller/directvolumemigration/task.go
```diff
@@ -244,13 +244,15 @@ func (t *Task) Run() error {
 			return liberr.Wrap(err)
 		}
 	case CreateRsyncConfig:
-		err := t.createRsyncConfig()
+		created, err := t.createRsyncConfig()
```
dymurray, line 247: Reading the code I'm not sure if this boolean gives us anything worthwhile. If `err` is `nil` then `created` is `true`. I would just keep it as a func that just returns an error unless I'm missing something.

pkg/controller/directvolumemigration/rsync.go
dymurray: It looks like if `err` is `nil` the boolean is always true, and it's always `false` if err is not nil. So it feels redundant

dymurray (changes requested):

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: @dymurray  no it's not, we are returning `false` and no error when the password is not found. Extra boolean confirm whether the secret with the password is created or not. @alaypatel07 suggested that we first look for a secret if the secret is not found create one and requeue, if the secret is found read the password from it and execute the rest of the logic. This was required because creating a secret and reading from the same secret needed some delay in between.

pkg/controller/directvolumemigration/task.go
JaydipGabani: Boolean is only to determine if the secret exists or not

pkg/controller/directvolumemigration/rsync.go
dymurray: I see.... so `createRsyncConfig` now returns a bool that determines whether we can safely continue on to the next task. Is that acccurate?

JaydipGabani: Yes.

pkg/controller/directvolumemigration/task.go
dymurray: NIT: rename to `continue` then since this bool returns true when are okay to continue to the next step. It could be false and also create some of the rsync config.

dymurray (approved):

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -565,6 +578,69 @@ func (t *Task) getRsyncRoute(namespace string) (string, error) {
...
+		t.Log.Info("Secret already exists on host", "name", "directvolumemigration-rsync-pass", "namespace", migapi.OpenshiftMigrationNamespace)
+	} else if err != nil {
+		return "", err
+	}
+	return string(password), nil
+}
+
+func (t *Task) getRsyncPassword() (string, error) {
+	rsyncSecret := corev1.Secret{}
+	key := types.NamespacedName{Name: "directvolumemigration-rsync-pass", Namespace: migapi.OpenshiftMigrationNamespace}
```
alaypatel07, line 613: question: what would happen if two dvm CR are running at the same time, both of them will use the same password? Is this intentional or a mistake?

alaypatel07 (approved):
VISICK, added a question, but other parts of the code look good to me, thanks @JaydipGabani

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: This was not intentional. @dymurray @alaypatel07 is creating one secret per dvm for this better?

dymurray: I think eventually we will want to do that yes. Today we can reasonably expect only one DVM to be created and running at a time.

I also don't see a huge deal with reusing the same password since its all secured with certs anyway

Issue #834, opened 2020-12-03
[MIG-284] DVM fails on RWO volumes if pods aren't scheduled on the same node


dymurray:
For some storage providers, RWO volumes *must* be mounted by pods that are scheduled on the same node. For stage pod workflow we did this by copying the original application pod's affinity values and applying them to the stage pod we create for that volume.

However, in the DVM case we operate on a per-pvc basis without any knowledge of the application pod itself. It's possible for the rsync client pod to be scheduled on a node different from the app pod as seen below:
```
$ oc get pods directvolumemigration-rsync-transfer-mysql -o yaml |grep -i nodeName
  nodeName: ip-172-18-8-191.ec2.internal
$ oc get pods mysql-1-kq8xc -o yaml |grep -i nodeName
  nodeName: ip-172-18-6-152.ec2.internal
```

We need to update DVM to check if a PVC is being used by a running pod and copy that pods affinity and/or node name to be used when scheduling the rsync client pod.

sseago:
Implementation suggestion:
As a one-time step before creating any pods, iterate over *running* pods (in each namespace), any pods that have PVCs, create a mapping from PVC namespaced name to nodeName, then when creating src transfer pods, set nodeName based on the map.
￼`transferPod.NodeName = pvcNodeMap[pvcNamespacedName]`
￼

dymurray:
Ahh thanks Scott, actually makes things easier.

PR #845, opened 2020-12-08
Schedule Rsync client pods on the same nodes as the volumes

shubham-pampattiwar:
This PR does the following:
- Adds a new function `getPVCNodeNameMap()` which returns a PVCNamespacedName : NodeName map
- Schedules Rsync client pods on the same nodes as the persistent volume claims
- Fixes https://github.com/konveyor/mig-controller/issues/834

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -641,6 +641,36 @@ func (t *Task) deleteRsyncPassword() error {
 	return nil
 }
 
+//Returns a map of PVCNamespacedName to the pod.NodeName
+func (t *Task) getPVCNodeNameMap() (map[string]string, error) {
+	nodeNameMap := map[string]string{}
+	pvcMap := t.getPVCNamespaceMap()
+	nsPodList := corev1.PodList{}
+	srcClient, err := t.getSourceClient()
+	if err != nil {
+		return nil, err
+	}
+
+	for ns, _ := range pvcMap {
+		err = srcClient.List(context.TODO(), k8sclient.InNamespace(ns), &nsPodList)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	for _, pod := range nsPodList.Items {
+		if pod.Status.Phase == corev1.PodRunning {
+			for _, vol := range pod.Spec.Volumes {
+				if vol.PersistentVolumeClaim != nil {
+					pvcNsName := pod.ObjectMeta.Namespace + "/" + vol.PersistentVolumeClaim.ClaimName
+					nodeNameMap[pvcNsName] = pod.Spec.NodeName
+				}
+			}
+		}
+	}
+	return nodeNameMap, nil
+}
+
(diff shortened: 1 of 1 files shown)


dymurray (approved):

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -648,6 +684,7 @@ func (t *Task) createRsyncClientPods() error {
 					RestartPolicy: corev1.RestartPolicyNever,
 					Volumes:       volumes,
 					Containers:    containers,
+					NodeName: pvcNodeMap[ns + "/" + vol],
```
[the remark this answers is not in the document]
sseago: @dymurray Yeah, that was my thinking. IF it's not there, we're setting NodeName to "", which is the same as if we'd checked for existence and didn't set it.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -565,6 +565,36 @@ func (t *Task) getRsyncRoute(namespace string) (string, error) {
...
+		return nil, err
+	}
+
+	for ns, _ := range pvcMap {
+		err = srcClient.List(context.TODO(), k8sclient.InNamespace(ns), &nsPodList)
+		if err != nil {
+			return nil, err
+		}
+	}
+
```
sseago, line 584: Won't this just end up with the pods for the final namespace included? I think the pod list loop should be inside the above loop.

pkg/controller/directvolumemigration/rsync.go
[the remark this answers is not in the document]
sseago: The indent should match, though -- putting pvcNodeMap right under containers.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -584,6 +614,12 @@ func (t *Task) createRsyncClientPods() error {
 		trueBool := true
 		runAsUser := int64(0)
 
+		pvcNodeMap, err := t.getPVCNodeNameMap()
```
sseago, line 617: This produces a list across namespaces, so this should be called outside the `for ns, vols` loop.

sseago (changes requested):
Looks good overall, just a couple changes needed for this to work properly when more than one namespace is involved (mentioned in comments).

dymurray:
Good catch Scott

pkg/controller/directvolumemigration/rsync.go
shubham-pampattiwar: Thanks for pointing it out. I had earlier done the same but was trying something else and missed out the multiple ns scenario. Updated the PR.

pkg/controller/directvolumemigration/rsync.go
shubham-pampattiwar: Done, Updated the PR

dymurray (approved):
LGTM after change

sseago (approved):

PR #849, opened 2020-12-09
remove old CRD's and regenerate

shawn-hurley:
Fixing up go mod change.

/cc @sseago


djwhatle (approved):
VISACK

sseago:
So if we want to keep the duplicated CRDs, then this is fine as-is. If we want only one copy of each yaml, then we need to figure out which one to remove (and the best way to do it to make sure that `make manifests` ends up with one copy).

djwhatle:
> So if we want to keep the duplicated CRDs, then this is fine as-is. If we want only one copy of each yaml, then we need to figure out which one to remove (and the best way to do it to make sure that `make manifests` ends up with one copy).

I think we want to avoid having duplicated CRDs

djwhatle (changes requested):
Let's figure out if we can remove the duplicated CRDs

sseago (approved):
LGTM

Makefile
```diff
@@ -51,19 +51,19 @@ use-oncluster-controller:
 
 # Install CRDs into a cluster
 install: manifests
-	kubectl apply -f config/crds
+	kubectl apply -f config
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 deploy: manifests
-	kubectl apply -f config/crds
+	kubectl apply -f crds
```
djwhatle, line 58: I don't have a `crds` dir after running make manifests. Not sure what this is doing.

djwhatle:
```
make deploy
/Users/dwhatley/code/go/bin/controller-gen "crd:trivialVersions=true" crd rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crds/bases 
kubectl apply -f crds
error: the path "crds" does not exist
make: *** [deploy] Error 1

```

sseago: Oh, wait, this is wrong. It should be "config" and not "crds".

sseago: config/crds no longer has the CRDs. They're now one level up in "config".

Makefile
```diff
@@ -51,19 +51,19 @@ use-oncluster-controller:
...
 	kustomize build config/default | kubectl apply -f -
 
 # Provide CRDs that work back to k8s 1.11
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
 # Generate manifests e.g. CRD, Webhooks
 manifests:
-	${CONTROLLER_GEN} ${CRD_OPTIONS} crd rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crds/bases output:crd:dir=config/crds
+	${CONTROLLER_GEN} ${CRD_OPTIONS} crd rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crds/bases
```
djwhatle, line 66: @shawn-hurley to preserve the previous directory structure, can we add some commands here?


```
manifests:                                                                         
  ${CONTROLLER_GEN} ${CRD_OPTIONS} crd rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crds/bases 
  mv config/migration.openshift.io*yaml config/crds                                
  rm -rf config/crds/bases
```

sseago: From what @shawn-hurley told me earlier, the yaml files under config/crd/bases may or may not always be the same as the files under config, although it's not clear that, as an intermediate output step we need to keep them once the final crds are created under config.

djwhatle: My code snippet above is based on the assumption that we do not need the files in `bases` after the final output is produced.

shawn-hurley:
Merging fixup to prepare for tomorrow, addressed @djwhatle concerns and impleted the change asked for

PR #851, opened 2020-12-09
Allow another final migration after successful rollback

djwhatle:
Implementation of:

```
1. Pull the list of Migrations associated with a Plan
2. Sort the Migrations in reverse order according to creation timestamp (completionTimestamp would be better but maybe not available)
3. Iterate over timestamp sorted migrations (newest to oldest)
   a. If a successful rollback migration is seen, break out of the loop and mark the plan as migrateable
   b. If a successful final migration is seen, break out of the loop and mark the plan as rollbackable
   c. If neither of the above are seen, mark the plan as migrateable
```

### DIFF
--- a/pkg/controller/migmigration/validation.go
+++ b/pkg/controller/migmigration/validation.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"sort"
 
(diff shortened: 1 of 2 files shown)


dymurray (approved):
VISIACK

Issue #840, opened 2020-12-04
[MIG-249] DVM controller must label PVCs appropriately for rollback


djwhatle:
**Is your feature request related to a problem? Please describe.**
I don't think rollback will be able to remove PVCs created by DVM controller due to lack of labels on the PVCs.

See https://github.com/konveyor/openshift-velero-plugin/pull/58 for info on labels needed

 - MigPlan correlation label
 - MigMigration correlation label

**Describe the solution you'd like**
Appropriate labels should be added to the PVCs created by DVM controller. Once this is in place, Rollback should be able to delete PVCs from the destination cluster using the labels above.

dymurray:
With the fix for this, please also update the labels set on DVM resources to use a common func to set the labels to ensure we aren't missing labels on any individual component (see https://github.com/konveyor/mig-controller/blob/master/pkg/controller/directvolumemigration/rsync.go#L233).

PR #852, opened 2020-12-10
Update DVM PVC check to include determination whether PVCs are bound and usable

### LINKED ISSUE #829: [MIG-284] Update DVM PVC check to include determination whether PVCs are bound and usable
Today `GetDestinationPVCs` is a no-op phase, we need to update this to include a check of whether the PVC is bound or not. If the PVC is pending, do not continue.

This check should be moved to after the creation of the transfer pods.

JaydipGabani:
This PR checks the state of the PVCs on the destination cluster to make sure PVCs are in the bound state before completing DVM.

Fixes #829

### DIFF
--- a/pkg/controller/directvolumemigration/pvcs.go
+++ b/pkg/controller/directvolumemigration/pvcs.go
@@ -100,35 +100,12 @@ func (t *Task) getDestinationPVCs() (bool, error) {
 			return false, err
 		}
 		for _, pvc := range pvcList.Items {
-			t.Log.Info("mounted PVC", pvc.Name, pvc.Status.Phase)
-			if pvc.Status.Phase != corev1.ClaimBound {
+			if pvc.Status.Phase == corev1.ClaimLost {
+				t.Log.Info("The PVC in lost state", pvc.Name, pvc.Status.Phase)
+			} else if pvc.Status.Phase != corev1.ClaimBound {
 				return false, nil
 			}
 		}
 	}
-
-	//selector = labels.SelectorFromSet(map[string]string{
-	//	"app":     "directvolumemigration-rsync-transfer",
-	//	"owner":   "directvolumemigration",
-	//	"purpose": "rsync",
-	//})
-	//for ns, _ := range pvcMap {
-	//	pods := corev1.PodList{}
-	//	err = destClient.List(
-	//		context.TODO(),
-	//		&k8sclient.ListOptions{
-	//			Namespace:     ns,
-	//			LabelSelector: selector,
-	//		},
-	//		&pods)
-	//	if err != nil {
-	//		return false, err
-	//	}
-	//	podToPVC := make(map[string]string)
-	//	for _, pod := range pods.Items {
-	//		t.Log.Info("mounted PVC", "mount", pod.Spec.Volumes)
-	//
-	//	}
-	//}
 	return true, nil
 }
--- a/pkg/controller/directvolumemigration/pvcs.go
+++ b/pkg/controller/directvolumemigration/pvcs.go
@@ -107,7 +107,7 @@ func (t *Task) areDestinationPVCsBound() (bool, error) {
 						Type:     PVCNotBoundOnDestinationCluster,
 						Status:   True,
 						Reason:   migapi.NotReady,
-						Category: Critical,
+						Category: migapi.Warn,
 						Message:  fmt.Sprintf("PVC %s of %s namespace is in %s state", pvc.Name, pvc.Namespace, pvc.Status.Phase),
 						Durable:  true,
 					})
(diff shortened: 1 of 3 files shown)


pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -38,6 +40,7 @@ func (t *Task) createDestinationPVCs() error {
 		}
 
 		newSpec := srcPVC.Spec
+		srcPVC.Labels["owner"] = "directvolumemigration"
```
JaydipGabani, line 43: @dymurray @sseago Is adding this label fine? This is to get PVCList to check if each PVC included in the migration is mounted or not. I am open to suggestion for other label or annotation, I just need a selector.

dymurray: @JaydipGabani Please sync up with @shubham-pampattiwar who is working on https://github.com/konveyor/mig-controller/issues/840

I've asked him to update the code to set a common label for all components in DVM so that we aren't using strings everywhere. I think this will solve your issue.

pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -72,7 +75,35 @@ func (t *Task) createDestinationPVCs() error {
...
+				Namespace:     ns,
+				LabelSelector: selector,
+			},
+			&pvcList)
+		if err != nil {
+			return false, err
+		}
+		for _, pvc := range pvcList.Items {
+			if pvc.Status.Phase == corev1.ClaimLost {
+				t.Log.Info("The PVC in lost state", pvc.Name, pvc.Status.Phase)
```
dymurray, line 102: Hmm this seems wrong. We are logging the PVC is in `ClaimLost` state which will break the rsync transfer but returning `true, nil` right?

dymurray: Ideally this is data that is bubbled back up to the user

pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -72,7 +75,35 @@ func (t *Task) createDestinationPVCs() error {
 	return nil
 }
 
-func (t *Task) getDestinationPVCs() error {
+func (t *Task) getDestinationPVCs() (bool, error) {
```
dymurray, line 78: func name should change to `areDestinationPVCsBound()` or something that indicates why we are returning a bool

dymurray (changes requested):

pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -72,7 +77,41 @@ func (t *Task) createDestinationPVCs() error {
...
+				LabelSelector: selector,
+			},
+			&pvcList)
+		if err != nil {
+			return false, err
+		}
+		for _, pvc := range pvcList.Items {
+			if pvc.Status.Phase != corev1.ClaimBound {
+				t.Owner.Status.SetCondition(migapi.Condition{
+					Type:     Running,
```
alaypatel07, line 105: Is this enough?

What if the PVCs are not bound for 2 hours? or 2 days?

Keeping this in running forever is not ideal. Can we raise a critical condition after say 10 mins into this condition?

alaypatel07 (commented):
The changes look sane but added a question from the User Experience point of view

pkg/controller/directvolumemigration/pvcs.go
JaydipGabani: Yes, I agree. Done

alaypatel07:
Removing from the description and adding it here.

Final state of the status of dmv field if the PVCs are in unbound state:

```
status:
  conditions:
    - category: Critical
      durable: true
      lastTransitionTime: '2020-12-17T21:16:06Z'
      message: PVC example1 of mongo namespace is in Pending state
      reason: DestinationPVCsCreated
      status: 'True'
      type: PVCNotBoundOnDestinationCluster
```

pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -72,7 +78,47 @@ func (t *Task) createDestinationPVCs() error {
...
+						Durable:  true,
+					})
+				}
+				return false, nil
+			}
+		}
+	}
+	conditions := t.Owner.Status.Conditions.FindConditionByCategory(Critical)
+	if len(conditions) > 0 {
+		t.Owner.Status.DeleteCondition(Critical)
```
alaypatel07, line 121: NIT: should we delete only the condition whose type is `PVCNotBoundOnDestinationCluster` ? to avoid deleting other Critical conditions by mistake?

alaypatel07 (approved):
LGTM!

dymurray (approved):

PR #853, opened 2020-12-10
fixing CRD condition generation such that conditions can be empty

shawn-hurley:


sseago (approved):
This fixes the problem for me.

dymurray (approved):

PR #859, opened 2020-12-14
Apply appropriate labels on PVCs to facilitate DVM rollback

shubham-pampattiwar:
The PR does the following:
- Adds two new attributes `PlanResources` and `MigrationUID` to the DVM task struct.
- Adds a new function `GetMigrationForDVM` based on owner reference.
- Adds `MigMigrationLabel` and `MigPlanLabel` on destination PVCs to facilitate rollback.
- Adds a new function to fetch common dvm labels `getDVMLabels()`.
- Converts labels strings to const strings.
- Fixes https://github.com/konveyor/mig-controller/issues/840

### DIFF
--- a/pkg/apis/migration/v1alpha1/model.go
+++ b/pkg/apis/migration/v1alpha1/model.go
@@ -2,7 +2,8 @@ package v1alpha1
 
 import (
 	"context"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	liberr "github.com/konveyor/controller/pkg/error"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"strconv"
 
 	imagev1 "github.com/openshift/api/image/v1"
@@ -134,7 +135,7 @@ func GetStorage(client k8sclient.Client, ref *kapi.ObjectReference) (*MigStorage
 
 // Get a referenced Migration for DVM.
 // Return nil if the reference cannot be resolved.
-func GetMigrationForDVM(client k8sclient.Client, owners []v1.OwnerReference) (*MigMigration, error) {
+func GetMigrationForDVM(client k8sclient.Client, owners []metav1.OwnerReference) (*MigMigration, error) {
 	if len(owners) == 0 {
 		return nil, nil
 	}
@@ -145,22 +146,15 @@ func GetMigrationForDVM(client k8sclient.Client, owners []v1.OwnerReference) (*M
 			Name:      migrationName,
 			Namespace: OpenshiftMigrationNamespace,
 		}, &migrationObject)
-	//objectList := MigMigrationList{}
-	//err := client.List(context.TODO(), &k8sclient.ListOptions{
-	//	LabelSelector: k8sLabels.SelectorFromSet(dvmLabels),
-	//}, &objectList)
 
 	if err != nil {
 		if errors.IsNotFound(err) {
 			return nil, nil
 		} else {
-			return nil, err
+			return nil, liberr.Wrap(err)
 		}
 	}
-
-	if &migrationObject == nil {
-		return nil, nil
-	}
+	
 	return &migrationObject, nil
 }
 
--- a/pkg/apis/migration/v1alpha1/model.go
+++ b/pkg/apis/migration/v1alpha1/model.go
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	"context"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"strconv"
 
 	imagev1 "github.com/openshift/api/image/v1"
@@ -129,6 +130,38 @@ func GetStorage(client k8sclient.Client, ref *kapi.ObjectReference) (*MigStorage
 	}
 
 	return &object, err
+}
+
+// Get a referenced Migration for DVM.
+// Return nil if the reference cannot be resolved.
+func GetMigrationForDVM(client k8sclient.Client, owners []v1.OwnerReference) (*MigMigration, error) {
+	if len(owners) == 0 {
+		return nil, nil
+	}
+	migrationName := owners[0].Name
+	migrationObject := MigMigration{}
+	err := client.Get(context.TODO(),
+		k8sclient.ObjectKey{
+			Name:      migrationName,
+			Namespace: OpenshiftMigrationNamespace,
+		}, &migrationObject)
+	//objectList := MigMigrationList{}
+	//err := client.List(context.TODO(), &k8sclient.ListOptions{
+	//	LabelSelector: k8sLabels.SelectorFromSet(dvmLabels),
+	//}, &objectList)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil, nil
+		} else {
+			return nil, err
+		}
+	}
+
+	if &migrationObject == nil {
+		return nil, nil
+	}
+	return &migrationObject, nil
 }
 
 // List MigStorage
--- a/pkg/controller/directvolumemigration/migrate.go
+++ b/pkg/controller/directvolumemigration/migrate.go
@@ -2,6 +2,7 @@ package directvolumemigration
 
 import (
 	"fmt"
+	liberr "github.com/konveyor/controller/pkg/error"
 	"time"
 
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
@@ -10,6 +11,32 @@ import (
 )
 
 func (r *ReconcileDirectVolumeMigration) migrate(direct *migapi.DirectVolumeMigration) (time.Duration, error) {
+
+	// Ready
+	migration, err := direct.GetMigrationForDVM(r)
+	if err != nil {
+		return 0, liberr.Wrap(err)
+	}
+	if migration == nil {
+		log.Info("Migration not found for DVM", "name", direct.Name)
+		return 0, liberr.Wrap(err)
+	}
+
+	plan, err := migration.GetPlan(r)
+	if err != nil {
+		return 0, liberr.Wrap(err)
+	}
+	if !plan.Status.IsReady() {
+		log.Info("Plan not ready.", "name", migration.Name)
+		return 0, liberr.Wrap(err)
+	}
+
+	// Resources
+	planResources, err := plan.GetRefResources(r)
+	if err != nil {
+		return 0, liberr.Wrap(err)
+	}
+
 	// Started
 	if direct.Status.StartTimestamp == nil {
 		direct.Status.StartTimestamp = &metav1.Time{Time: time.Now()}
@@ -22,8 +49,10 @@ func (r *ReconcileDirectVolumeMigration) migrate(direct *migapi.DirectVolumeMigr
 		Owner:            direct,
 		Phase:            direct.Status.Phase,
 		PhaseDescription: direct.Status.PhaseDescription,
+		PlanResources:    planResources,
+		MigrationUID:     string(migration.UID),
 	}
-	err := task.Run()
+	err = task.Run()
 	if err != nil {
 		if errors.IsConflict(err) {
 			return FastReQ, nil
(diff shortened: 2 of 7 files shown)


pkg/controller/directvolumemigration/migrate.go
```diff
@@ -10,6 +11,32 @@ import (
 )
 
 func (r *ReconcileDirectVolumeMigration) migrate(direct *migapi.DirectVolumeMigration) (time.Duration, error) {
+
+	// Ready
+	migration, err := direct.GetMigrationForDVM(r)
```
djwhatle, line 16: +1 on using the cached client for this.

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	"context"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
djwhatle, line 5: Let's alias this as `metav1` instead of just `v1` so that it's clear what the alias refers to.

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -131,6 +132,38 @@ func GetStorage(client k8sclient.Client, ref *kapi.ObjectReference) (*MigStorage
...
+		return nil, nil
+	}
+	migrationName := owners[0].Name
+	migrationObject := MigMigration{}
+	err := client.Get(context.TODO(),
+		k8sclient.ObjectKey{
+			Name:      migrationName,
+			Namespace: OpenshiftMigrationNamespace,
+		}, &migrationObject)
+	//objectList := MigMigrationList{}
```
djwhatle, line 148: Let's remove these comments if the code is unused.

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -131,6 +132,38 @@ func GetStorage(client k8sclient.Client, ref *kapi.ObjectReference) (*MigStorage
...
+	//err := client.List(context.TODO(), &k8sclient.ListOptions{
+	//	LabelSelector: k8sLabels.SelectorFromSet(dvmLabels),
+	//}, &objectList)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil, nil
+		} else {
+			return nil, err
```
djwhatle, line 157: Let's use `liberr.wrap(err)` so that the full stack trace will be returned down the call chain when this error is surfaced later.

pkg/apis/migration/v1alpha1/model.go
```diff
@@ -131,6 +132,38 @@ func GetStorage(client k8sclient.Client, ref *kapi.ObjectReference) (*MigStorage
...
+		}
+	}
+
+	if &migrationObject == nil {
```
djwhatle, line 161: I think we can drop this check, won't we return the same thing without this in place? 

If `&migrationObject` is nil then `return &migrationObject, nil` on 164 will resolve to `return nil, nil`

pkg/controller/directvolumemigration/migrate.go
```diff
@@ -10,6 +11,32 @@ import (
...
+	if err != nil {
+		return 0, liberr.Wrap(err)
+	}
+	if migration == nil {
+		log.Info("Migration not found for DVM", "name", direct.Name)
+		return 0, liberr.Wrap(err)
```
djwhatle, line 22: I think @dymurray mentioned that the DVM should be allowed to run even if there is no associated MigMigration, right? Sometimes you guys do separate DVM operators for testing/debugging purposes. 

It looks like the way this is written, lack of a linked migmigration will cause the DVM to stop.

This should be written such that it still work when a user creates a DVM separate from a migmigration.

pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -42,12 +42,22 @@ func (t *Task) createDestinationPVCs() error {
 		newSpec.AccessModes = pvc.TargetAccessModes
 		newSpec.VolumeName = ""
 
+		//Add src labels and rollback labels
+		pvcLabels := srcPVC.Labels
+		if pvcLabels == nil {
```
djwhatle, line 47: This looks good.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -74,12 +74,10 @@ func (t *Task) areRsyncTransferPodsRunning() (bool, error) {
 	}
 
 	pvcMap := t.getPVCNamespaceMap()
+	dvmLabels := t.getDVMLabels()
+	dvmLabels["purpose"] = Rsync
```
djwhatle, line 78: Suggestion: maybe rename the `Rsync` const to `RsyncPurpose` or something that gives more context of what is contained within, since this is a pkg level var.

djwhatle (changes requested):

djwhatle:
@shubham-pampattiwar have you been able to verify that rollback cleans up the DVM resources with your changes in place?

pkg/controller/directvolumemigration/task.go
```diff
@@ -52,6 +52,31 @@ const (
...
+	DirectVolumeMigrationRsyncPass          = "directvolumemigration-rsync-pass"
+	DirectVolumeMigrationStunnelTransfer    = "directvolumemigration-stunnel-transfer"
+	// Identifies associated migplan
+	// to allow migplan restored resources rollback
+	// The value is Task.PlanResources.MigPlan.UID
+	MigPlanLabel = "migration.openshift.io/migrated-by-migplan" // (migplan UID)
+	// Identifies the resource as migrated by us
+	// for easy search or application rollback.
+	// The value is the Task.UID().
+	MigMigrationLabel = "migration.openshift.io/migrated-by-migmigration" // (migmigration UID)
```
djwhatle, line 74: I would suggest renaming this to `MigratedByMigrationLabel`

pkg/controller/directvolumemigration/task.go
```diff
@@ -52,6 +52,31 @@ const (
...
+	DirectVolumeMigrationRsyncTransferSvc   = "directvolumemigration-rsync-transfer-svc"
+	DirectVolumeMigrationRsyncTransferRoute = "directvolumemigration-rsync-transfer-route"
+	DirectVolumeMigrationStunnelConfig      = "directvolumemigration-stunnel-config"
+	DirectVolumeMigrationStunnelCerts       = "directvolumemigration-stunnel-certs"
+	DirectVolumeMigrationRsyncPass          = "directvolumemigration-rsync-pass"
+	DirectVolumeMigrationStunnelTransfer    = "directvolumemigration-stunnel-transfer"
+	// Identifies associated migplan
+	// to allow migplan restored resources rollback
+	// The value is Task.PlanResources.MigPlan.UID
+	MigPlanLabel = "migration.openshift.io/migrated-by-migplan" // (migplan UID)
```
djwhatle, line 70: I would suggest renaming this to `MigratedByPlanLabel`

pkg/controller/directvolumemigration/task.go
```diff
@@ -52,6 +52,31 @@ const (
...
+	// Identifies the resource as migrated by us
+	// for easy search or application rollback.
+	// The value is the Task.UID().
+	MigMigrationLabel = "migration.openshift.io/migrated-by-migmigration" // (migmigration UID)
+	Rsync             = "rsync"
```
djwhatle, line 75: It feels like these last 3 consts are different from the rest of the consts above in style. Is there a reason for this difference?

shubham-pampattiwar:
> @shubham-pampattiwar nice work on the PR so far.
> 
> Have you been able to verify that rollback cleans up the DVM resources with your changes in place?
@djwhatle Yes I tried the rollback operation after all my changes for the django app, the running pods, svc and routes were not present on the destination cluster namespace as well as the pvc was deleted. Is there anything else that I need to check ?

djwhatle:
> > @shubham-pampattiwar nice work on the PR so far.
> > Have you been able to verify that rollback cleans up the DVM resources with your changes in place?
> 
> @djwhatle Yes I tried the rollback operation after all my changes for the django app, the running pods, svc and routes were not present on the destination cluster namespace as well as the pvc was deleted. Is there anything else that I need to check ?

That sounds good!

pkg/apis/migration/v1alpha1/model.go
dymurray: +1 looks right to me

pkg/controller/directvolumemigration/migrate.go
dymurray: Correct, we still want to support running DVM outside of the context of migmigration controller

pkg/apis/migration/v1alpha1/model.go
shubham-pampattiwar: Done

pkg/apis/migration/v1alpha1/model.go
shubham-pampattiwar: Done

pkg/apis/migration/v1alpha1/model.go
shubham-pampattiwar: Done, updated the PR

pkg/controller/directvolumemigration/migrate.go
shubham-pampattiwar: Update the PR.

pkg/controller/directvolumemigration/task.go
```diff
@@ -454,3 +476,13 @@ func (t *Task) getDestinationClient() (compat.Client, error) {
 	}
 	return client, nil
 }
+
+// Get DVM labels for the migration
+func (t *Task) getDVMLabels() map[string]string {
```
dymurray, line 481: NIT: consider renaming `buildDVMLabels`.

dymurray:
Changes look sane to me. Give me a chance to finish testing but code looks good.

pkg/controller/directvolumemigration/migrate.go
```diff
@@ -22,6 +54,8 @@ func (r *ReconcileDirectVolumeMigration) migrate(direct *migapi.DirectVolumeMigr
 		Owner:            direct,
 		Phase:            direct.Status.Phase,
 		PhaseDescription: direct.Status.PhaseDescription,
+		PlanResources:    planResources,
+		MigrationUID:     string(migration.UID),
```
djwhatle, line 58: This seems like it would fail if there was no connected migration. @shubham-pampattiwar have you been able to test with an isolated DVM?

djwhatle (changes requested):
I saw that we are referencing `migration.UID` where we may not have a valid reference to a migration, thinking that may fail.

pkg/controller/directvolumemigration/migrate.go
djwhatle: Specifically, if migration is nil, wouldn't trying to run `string(migration.UID)` cause a panic?

dymurray: `migration` won't ever be `nil` right? L15 sets it to an empty object. My initial readthrough assumed this would set UID to an empty string. @shubham-pampattiwar can you confirm?

dymurray: Can test this on my end as well.

pkg/controller/directvolumemigration/task.go
shubham-pampattiwar: Done !

pkg/controller/directvolumemigration/migrate.go
shubham-pampattiwar: Thank you @djwhatle for pointing it out, I was not getting a panic in instantiating task object due to the precise reasons stated by @dymurray But some changes were needed in labeling the destination PVCs. I have updated the PR with the latest changes, please take another look. With these latest changes, I have tested a DVM Migration, DVM Rollback as well as an independent DVM Migrate Operation.

djwhatle (approved):
Assuming we've tested with the latest code revision, LGTM!

dymurray (approved):
Tested on my end worked for me! Opening a separate issue to handle canceled migrations

Issue #780, opened 2020-11-13
[MIG-284] Validation to verify networking environment is suitable for DVM/DIM use case


dymurray:
Upon migplan creation, we want to perform some basic validation check to confirm that if the user has selected Direct migration that the migration won't totally fail.

** Release 1.4.0**
- [ ] Document a way to deploy, test, and verify a direct migration is possible from 3.11->4.6 using service catalog application
- [ ] Support validation of exposed registry route for migclusters

** Release 1.4.z**
- [ ] Design document for preflight validation integration with DVM/DIM
- [ ] Implement preflight validation direct migration requirements are satisfied

sseago:
The goal here is to attempt to perform an actual PV and image copy with some small test volume and image. One possible approach would be a new task phase for this, which could be incorporated as an optional step at the beginning of a migration, or even as a standalone "validate" itinerary, which would be launched by creating a migmigration with `validateOnly=true` or something similar.

dymurray:
For docker-registry svcs with v2 support, we can validate the route is valid by checking `/v2/` endpoint with a valid token and confirm it returns a 200.

```
$ curl -k https://docker-registry-default.apps.dymurray-ocp3.mg.dog8code.com/v2/ -H "Authorization: bearer UCbSDHG4n4riixjkBHvWP9GTCuAvNIl_CbigKg4fJno"
{}
```

We will want to fail migcluster validation if we do not get a 200 response from this endpoint.

djwhatle:
@dymurray is this for release-1.4.0, or release-1.4.z? I see both labels assigned.

PR #867, opened 2020-12-16
Add Validation for Exposed Registry Route

shubham-pampattiwar:
This PR does the following:
- Adds a new `MigCluster` validation called `validateRegistryRoute`
- This validation checks if the exposed image registry route added by the user for a cluster is valid or not
- The validation is done based on https://github.com/konveyor/mig-controller/issues/780#issuecomment-741869357
- Fixes https://github.com/konveyor/mig-controller/issues/780

### DIFF
--- a/pkg/apis/migration/v1alpha1/model.go
+++ b/pkg/apis/migration/v1alpha1/model.go
@@ -154,7 +154,7 @@ func GetMigrationForDVM(client k8sclient.Client, owners []metav1.OwnerReference)
 			return nil, liberr.Wrap(err)
 		}
 	}
-	
+
 	return &migrationObject, nil
 }
 
--- a/pkg/controller/migcluster/validation.go
+++ b/pkg/controller/migcluster/validation.go
@@ -258,7 +258,7 @@ func (r ReconcileMigCluster) validateRegistryRoute(cluster *migapi.MigCluster) e
 				Status:   True,
 				Reason:   RouteTestFailed,
 				Category: Critical,
-				Message:  fmt.Sprintf("Exposed registry route is invalid, Error : %#v", err),
+				Message:  fmt.Sprintf("Exposed registry route is invalid, Error : %#v", err.Error()),
 				Items:    []string{err.Error()},
 			})
 			return nil
--- a/pkg/controller/migcluster/validation.go
+++ b/pkg/controller/migcluster/validation.go
@@ -258,8 +258,19 @@ func (r ReconcileMigCluster) validateRegistryRoute(cluster *migapi.MigCluster) e
 				Status:   True,
 				Reason:   RouteTestFailed,
 				Category: Critical,
-				Message:  fmt.Sprintf("Exposed registry route is invalid, response received : %#v", res),
+				Message:  fmt.Sprintf("Exposed registry route is invalid, Error : %#v", err),
 				Items:    []string{err.Error()},
+			})
+			return nil
+		}
+
+		if res.StatusCode != 200 {
+			cluster.Status.SetCondition(migapi.Condition{
+				Type:     InvalidRegistryRoute,
+				Status:   True,
+				Reason:   RouteTestFailed,
+				Category: Critical,
+				Message:  fmt.Sprintf("Exposed registry route connection test failed, Response code received: %#v", res.StatusCode),
 			})
 			return nil
 		}
--- a/pkg/controller/migcluster/validation.go
+++ b/pkg/controller/migcluster/validation.go
@@ -228,9 +228,6 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
 
 // Validate the Exposed registry route
 func (r ReconcileMigCluster) validateRegistryRoute(cluster *migapi.MigCluster) error {
-	if cluster.Spec.IsHostCluster {
-		return nil
-	}
 
 	if cluster.Status.HasCriticalCondition() {
 		return nil
--- a/pkg/controller/migcluster/validation.go
+++ b/pkg/controller/migcluster/validation.go
@@ -2,7 +2,9 @@ package migcluster
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
+	"net/http"
 	"net/url"
 	"path"
 	"strings"
@@ -20,6 +22,7 @@ const (
 	InvalidURL           = "InvalidURL"
 	InvalidSaSecretRef   = "InvalidSaSecretRef"
 	InvalidSaToken       = "InvalidSaToken"
+	InvalidRegistryRoute = "InvalidRegistryRoute"
 	TestConnectFailed    = "TestConnectFailed"
 	SaTokenNotPrivileged = "SaTokenNotPrivileged"
 )
@@ -31,12 +34,13 @@ const (
 
 // Reasons
 const (
-	NotSet        = "NotSet"
-	NotFound      = "NotFound"
-	ConnectFailed = "ConnectFailed"
-	Malformed     = "Malformed"
-	InvalidScheme = "InvalidScheme"
-	Unauthorized  = "Unauthorized"
+	NotSet          = "NotSet"
+	NotFound        = "NotFound"
+	ConnectFailed   = "ConnectFailed"
+	Malformed       = "Malformed"
+	RouteTestFailed = "RouteTestFailed"
+	InvalidScheme   = "InvalidScheme"
+	Unauthorized    = "Unauthorized"
 )
 
 // Statuses
@@ -68,6 +72,12 @@ func (r ReconcileMigCluster) validate(cluster *migapi.MigCluster) error {
 
 	// Token privileges
 	err = r.validateSaTokenPrivileges(cluster)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	// Exposed registry route
+	err = r.validateRegistryRoute(cluster)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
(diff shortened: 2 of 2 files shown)


pkg/apis/migration/v1alpha1/model.go
```diff
@@ -154,7 +154,7 @@ func GetMigrationForDVM(client k8sclient.Client, owners []metav1.OwnerReference)
 			return nil, liberr.Wrap(err)
 		}
 	}
-	
+
```
alaypatel07, line 157: nit: intentional? should rely on `gofmt`

pkg/controller/migcluster/validation.go
```diff
@@ -72,6 +76,12 @@ func (r ReconcileMigCluster) validate(cluster *migapi.MigCluster) error {
 		return liberr.Wrap(err)
 	}
 
+	// Exposed registry route
+	err = r.validateRegistryRoute(cluster)
```
alaypatel07, line 80: question: is this enough? should we have a logic like this in place?

https://github.com/konveyor/mig-controller/blob/cf705e01f0dac48c3a557d61d98776b54b6c3420/pkg/controller/migstorage/provider.go#L40

alaypatel07 (commented):
Added two questions, otherwise LGTM

pkg/controller/migcluster/validation.go
```diff
@@ -216,6 +226,50 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
...
+func (r ReconcileMigCluster) validateRegistryRoute(cluster *migapi.MigCluster) error {
+	if cluster.Spec.IsHostCluster {
+		return nil
+	}
+
+	if cluster.Status.HasCriticalCondition() {
+		return nil
+	}
+
+	if cluster.Spec.ExposedRegistryPath != "" {
```
djwhatle, line 239: Just curious:

1. Are we expecting that `cluster.Spec.ExposedRegistryPath` will always be a bare domain name without protocol attached? 
2. Will this work if a non-standard port is in use for registry?

pkg/controller/migcluster/validation.go
```diff
@@ -216,6 +226,50 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
 	return nil
 }
 
+// Validate the Exposed registry route
+func (r ReconcileMigCluster) validateRegistryRoute(cluster *migapi.MigCluster) error {
+	if cluster.Spec.IsHostCluster {
```
djwhatle, line 231: I suppose the registry route for the host cluster is not user-provided based on this?

pkg/controller/migcluster/validation.go
```diff
@@ -216,6 +226,50 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
...
+
+		req.Header.Set("Authorization", "bearer "+token)
+
+		res, err := client.Do(req)
+		if err != nil {
+			cluster.Status.SetCondition(migapi.Condition{
+				Type:     InvalidRegistryRoute,
+				Status:   True,
+				Reason:   RouteTestFailed,
+				Category: Critical,
```
djwhatle, line 263: I noticed this is a critical condition. We don't always require a registry route to perform migrations. 

I guess the idea here is that _if_ a user supplies a value then we will block MigCluster from being ready (and therefore block all associated migrations), but they can remove `Spec.ExposedRegistryPath` if they can't figure out the registry route and want to do an indirect migration.

pkg/controller/migcluster/validation.go
sseago: The expectation is that it is a host/domain name with an optional port, although I think at one point we added some code that strips off "http[s]://" if it's there since the an early iteration of the UI was adding it in. The path is expected to be used to create a docker image reference `docker://<exposedRegistryPath>/<namespace>/<imagename>:<tag>`
For the internal registry, this will normally have ":5000:" appended. For an exposed route using a normal OpenShift passthrough route, there will be no port specified (I think the actual port in use is 443). If the route is actually created with a non-standard (external) port, then that would be included in the registry path.

pkg/controller/migcluster/validation.go
sseago: The UI provides no way for specifying it, and for most purposes it's not needed there since we just use the internal registry path. There is one scenario where the host cluster does require it, though: If the controller is run via `make run` and the host cluster is either a source or destination cluster, then an ExposedRegistryPath must be provided, since the controller running locally won't be able to access cluster-internal service ports. This will require the user to manually set this parameter on the MigCluster (since the UI doesn't allow editing the host migcluster). This isn't a supported scenario, since we don't support customers running `make run`, but it's relevant for developers (who can `oc edit` the cluster easily enough). Back to the topic of this PR, though, the above is fine if we don't need to run this validation for this non-standard developer-only scenario. Alternatively, we could replace `if cluster.Spec.IsHostCluster` with `if cluster.Spec.ExposedRegistryPath != "" `

sseago: Just noticing that below here we don't do anything if the exposed registry path is empty, so we may want to just eliminate the IsHostCluster test.

pkg/controller/migcluster/validation.go
sseago: We don't require a registry route, which is why this PR doesn't fail validation of the MigCluster if the registry route is absent. I think it's reasonable to expect that *if* a user provides a route that it be valid, and if it's not valid, make the cluster non-ready. If the user is doing indirect image migration, the remedy for this critical condition would be to clear the registry route entry on the MigCluster.

pkg/controller/migcluster/validation.go
dymurray: Can you explain further Alay?

pkg/controller/migcluster/validation.go
dymurray: Should we add validation the protocol is unspecified? Below we also automatically assume the protocol is `https`... isn't it possible for the registry to not be behind `https`?

pkg/controller/migcluster/validation.go
```diff
@@ -216,6 +226,47 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
...
+		req, err := http.NewRequest("GET", url, nil)
+
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+
+		req.Header.Set("Authorization", "bearer "+token)
+
+		res, err := client.Do(req)
+		if err != nil {
```
dymurray, line 255: This should not only check for error, but it should also be performing a check on the status code to ensure it returns a 200.

This is failing for me if I do a valid URL but an invalid route to a registry.

pkg/controller/migcluster/validation.go
```diff
@@ -216,6 +226,47 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
...
+			cluster.Status.SetCondition(migapi.Condition{
+				Type:     InvalidRegistryRoute,
+				Status:   True,
+				Reason:   RouteTestFailed,
+				Category: Critical,
+				Message:  fmt.Sprintf("Exposed registry route is invalid, response received : %#v", res),
```
dymurray, line 261: Let's change this message. `res` is going to be `nil` in the case of an error. put `err` here.

then below if it fails a status code check lets show the status code to the user

dymurray (changes requested):

pkg/controller/migcluster/validation.go
shubham-pampattiwar: Thanks for the catch. Updated the PR.

pkg/controller/migcluster/validation.go
shubham-pampattiwar: Done.

pkg/controller/migcluster/validation.go
sseago: @dymurray So the protocol specified in the generated dockerImageReference is "docker://" -- I'm not quite sure what the internal docker rules are on when to use ssl vs. not there. But logically there is no protocol attached to the exposed path since the path is inserted directly after "docker://" in the generate string.

sseago: The input value is a "host[:port]" value, not a URL with embedded protocol.

sseago: But since our test uses https/http rather than "docker://", we need to infer a protocol from the path. If it's possible that there will be an "http://" exposed registry path, we may need to try http when https fails. I don't know that I've run into a scenario where the internal registry uses http, so this may not be a scenario we care about, so assuming that "https://" works may be sufficient here.

dymurray (approved):
Worked for me. Intention is to change the UI wording to explicitly mention we should be supplying a route "host" rather than a URL to not include the prefix

pkg/controller/migcluster/validation.go
alaypatel07: @dymurray, @shubham-pampattiwar, and I met to discuss this and decided that this is currently not needed.

pkg/controller/migcluster/validation.go
```diff
@@ -216,6 +226,58 @@ func (r ReconcileMigCluster) testConnection(cluster *migapi.MigCluster) error {
...
 
+// Validate the Exposed registry route
+func (r ReconcileMigCluster) validateRegistryRoute(cluster *migapi.MigCluster) error {
+
+	if cluster.Status.HasCriticalCondition() {
+		return nil
+	}
+
+	if cluster.Spec.ExposedRegistryPath != "" {
+		url := "https://" + cluster.Spec.ExposedRegistryPath + "/v2/"
```
alaypatel07, line 237: nit: if we are changing the naming in UI, shall we change this to `ExpectedRegistryHost` as well?

alaypatel07 (approved):
Added a small nit. Only apply after we have made UI changes. 

VISIACK

djwhatle (approved):

pkg/controller/migcluster/validation.go
sseago: Host might be confusing too, though. But ExpectedRegistryHostAndOptionalPort is probably too long.

sseago (approved):

PR #868, opened 2020-12-17
warn a user if dvm controller is blocked because of dependency pods

### LINKED ISSUE #839: [MIG-284]: no warning is given to user if a pod owned by dvm controller spends a lot of time in pending state
**Describe the bug**
We expect the dvm controller to raise a warning to the user if the transfer pod, or stunnel pod on the source or detination take a lot of time to go into a running state. Ideally, if the pod is in a pending state for 5 mins, we need to raise a warning to the user in the UI.

**To Reproduce**
1. Run a DVM migration
2. After the run is successful, delete the underlying ebs volume from aws console
3. Run dvm again, you'll see the pod on the destination is in a pending state
4. DVM and migmigration are stuck without warnings

**Expected behavior**
For 1.4.0, I expect the system to raise a warning.

alaypatel07:
This adds the following:
- warning to dvm controllers for pending transfer and stunnel pods
- bubbles up the warnings to migmigration from the dvm controller

fixes #839

### DIFF
--- a/pkg/apis/migration/v1alpha1/condition.go
+++ b/pkg/apis/migration/v1alpha1/condition.go
@@ -223,7 +223,7 @@ func (r *Conditions) SetCondition(condition Condition) {
 	condition.staged = true
 	found := r.find(condition.Type)
 	if found == nil {
-		condition.LastTransitionTime = metav1.NewTime(time.Now())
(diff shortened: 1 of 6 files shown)


dymurray (approved):
LGTM

alaypatel07:
image

Tested this, and it created the warning as in the above image.

Once I fix my pvc, the same migration goes to success

PR #869, opened 2020-12-18
dev guide - Add details on condition staging and durable conditions

djwhatle:
Adding some details on condition staging from a slack conversation I had with @jortel a while back.

image


alaypatel07 (approved):
+1 very helpful thank you @djwhatle. I'll merge this end of the day is this is still open

PR #870, opened 2020-12-22
read the pod configurations from configmap

### LINKED ISSUE #863: [MIG-284] Ability to configure resource requirements for rsync pods
For 1.4.0 we want to expose at a minimum the ability to provide an operator configuration setting (i.e. `rsync_transfer_resource_cpu`) that allows the user to customize the pod resource requirements around:
* cpu
* memory
* bandwidth
and these settings will be applied to all pods. If the settings are not configured, we will set no resource requirements on the pods.

For 1.4.z+ we need to implement this on a per-volume basis.

See https://github.com/konveyor/mig-controller/pull/861 for guidance

JaydipGabani:
This PR requires operator change (needs values in `migration-controller` configMap), The PR for the same can be found https://github.com/konveyor/mig-operator/pull/533. 

This PR reads the respective values from `migration-controller` configMap and configures the rsync transfer, rsync client, and stunnel pods.

Fixes #863

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -364,7 +364,7 @@ func (t *Task) createRsyncTransferPods() error {
 	if err != nil {
 		return err
 	}
-	limits, requests, err := getPodConfig(t.Client, []string{"TRANSFER_POD_CPU_LIMIT", "TRANSFER_POD_MEMORY_LIMIT", "TRANSFER_POD_CPU_REQUEST", "TRANSFER_POD_MEMORY_REQUEST"})
+	limits, requests, err := getPodConfig(t.Client, "TRANSFER_POD_CPU_LIMIT", "TRANSFER_POD_MEMORY_LIMIT", "TRANSFER_POD_CPU_REQUEST", "TRANSFER_POD_MEMORY_REQUEST")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/directvolumemigration/rsync.go
```diff
@@ -553,6 +561,37 @@ func (t *Task) createRsyncTransferPods() error {
...
 }
 
+func getPodConfig(client k8sclient.Client, keys []string) (corev1.ResourceList, corev1.ResourceList, error) {
+	podConfigMap := &corev1.ConfigMap{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: "migration-controller", Namespace: migapi.OpenshiftMigrationNamespace}, podConfigMap)
+	if err != nil {
+		return nil, nil, err
+	}
+	limits := corev1.ResourceList{}
+	if _, exists := podConfigMap.Data[keys[0]]; exists {
```
sseago, line 571: A slice of keys is passed in to this func, but then inside the func we're assuming that the first key is a CPU limit, the second is a memory limit, etc. This feels like a fragile way of handling it. Since this is used for both transfer and client pod, one possibility would be to just pass those strings in as individual function params -- replace `keys []string` with `cpu_limit, memory_limit, cpu_request, memory_request string` or something like that.

sseago (commented):
Also, a more general question. It looks like if the configMap values aren't found, then either limits or requests end up as empty structs. Is this equivalent to not setting them, or do we need to conditionalize the param setting to omit them if they're empty?

pkg/controller/directvolumemigration/rsync.go
dymurray: +1 please take Scott's approach or pass in a struct instead

alaypatel07: +1

alaypatel07 (changes requested):
ACK, the concept looks good but need to modify the code for few nits, PTAL @JaydipGabani

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: Good catch! Updating code in regards to this

sseago (approved):

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -360,6 +361,10 @@ func (t *Task) createRsyncTransferPods() error {
 	if err != nil {
 		return err
 	}
+	limits, requests, err := getPodConfig(t.Client, "TRANSFER_POD_CPU_LIMIT", "TRANSFER_POD_MEMORY_LIMIT", "TRANSFER_POD_CPU_REQUEST", "TRANSFER_POD_MEMORY_REQUEST")
```
dymurray, line 364: NIT: use constants here

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -706,6 +751,11 @@ func (t *Task) createRsyncClientPods() error {
 		return err
 	}
 
+	limits, requests, err := getPodConfig(t.Client, "CLIENT_POD_CPU_LIMIT", "CLIENT_POD_MEMORY_LIMIT", "CLIENT_POD_CPU_REQUEST", "CLIENT_POD_MEMORY_REQUEST")
```
dymurray, line 754: NIT: constants since these are reused

pkg/controller/directvolumemigration/stunnel.go
```diff
@@ -338,6 +338,10 @@ func (t *Task) createStunnelClientPods() error {
 		return err
 	}
 
+	limits, requests, err := getPodConfig(t.Client, "STUNNEL_POD_CPU_LIMIT", "STUNNEL_POD_MEMORY_LIMIT", "STUNNEL_POD_CPU_REQUEST", "STUNNEL_POD_MEMORY_REQUEST")
```
dymurray, line 341: NIT: use constants

dymurray (approved):
Changes look sane to me, consider using some constants to pass into `getPodConfig`

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: +1

pkg/controller/directvolumemigration/stunnel.go
JaydipGabani: +1

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: +1

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -553,6 +576,43 @@ func (t *Task) createRsyncTransferPods() error {
 	return nil
 }
 
+func getPodConfig(client k8sclient.Client, cpu_limit string, memory_limit string, cpu_request string, memory_request string, ) (corev1.ResourceList, corev1.ResourceList, error) {
```
alaypatel07, line 579: is there function better named as `getPodResourceLists`?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -553,6 +576,43 @@ func (t *Task) createRsyncTransferPods() error {
...
+	if err != nil {
+		return nil, nil, err
+	}
+	limits := corev1.ResourceList{
+		corev1.ResourceMemory: resource.MustParse("1Gi"),
+		corev1.ResourceCPU:    resource.MustParse("1"),
+	}
+	if _, exists := podConfigMap.Data[cpu_limit]; exists {
+		cpu := resource.MustParse(podConfigMap.Data[cpu_limit])
+		if _, exists := podConfigMap.Data[memory_limit]; exists {
```
alaypatel07, line 591: why is this nested if? should the if here and if 2 lines above be independent?

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: +1

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: The idea was to check for both the values and then set, but now looking at it again I think we can remove nested if.

alaypatel07 (approved):
LGTM

PR #874, opened 2021-01-05
Allow configuring Rsync options through MigrationController CR

### LINKED ISSUE #872: [MIG-284] Allow configureable Rsync options through MigrationController CR
Add the ability to configure Rsync options used by DVM.  This allows users to use their own options as needed in their environments.

For 1.4.0, we can allow setting global Rsync options through _MigrationController_ CR. The options will be applicable globally for all volumes and will override the default options used by the DVM controller. Some options will never be overridden, for example, `-Pvvv` as these options are used to reliably find the progress information. The idea is to allow users to provide performance related options only. 

An alternative is to configure Rsync options on a per volume basis in the MigPlan Wizard. The UI can provide templated options in a dropdown to avoid human errors. However, this would be a long term solution in 1.4.z releases.

pranavgaikwad:
Fixes #872 

This PR allows configuring different Rsync options through _MigrationController_ CR

Following default options will **never** be overridden as they are required for progress reporting and logging purposes:
- --human-readable
- --log-file

Following options are controlled through variables in _MigrationController_ CR:

| Option                  	| Type   	| Default                                          	| Operator Flag       	|
|-------------------------	|--------	|--------------------------------------------------	|---------------------	|
| --bwlimit=<>            	| int    	| Not Set                                          	| rsync_opt_bwlimit   	|
| --archive               	| bool   	| True                                             	| rsync_opt_archive   	|
| --partial               	| bool   	| True                                             	| rsync_opt_partial   	|
| --delete                	| bool   	| True                                             	| rsync_opt_delete    	|
| --hard-links            	| bool   	| True                                             	| rsync_opt_hardlinks 	|
| --info=<>               	| string 	| COPY2,DEL2,REMOVE2,SKIP2,FLIST2,PROGRESS2,STATS2 	| rsync_opt_info      	|
| Any other extra options 	| string 	| Empty                                            	| rsync_opt_extras    	|

Please review accompanying operator PR : https://github.com/konveyor/mig-operator/pull/540

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -6,8 +6,14 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"fmt"
+	random "math/rand"
+	"strings"
+	"text/template"
+	"time"
+
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	"github.com/konveyor/mig-controller/pkg/compat"
+	migsettings "github.com/konveyor/mig-controller/pkg/settings"
 	routev1 "github.com/openshift/api/route/v1"
 	"golang.org/x/crypto/ssh"
 	"gopkg.in/yaml.v2"
@@ -17,10 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	random "math/rand"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
-	"text/template"
-	"time"
 )
 
 type pvc struct {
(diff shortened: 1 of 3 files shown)


dymurray (approved):
LGTM

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -678,6 +681,47 @@ func (t *Task) getPVCNodeNameMap() (map[string]string, error) {
...
+	if rsyncOptions.BwLimit != -1 {
+		rsyncOpts = append(rsyncOpts,
+			fmt.Sprintf("--bwlimit=%d", rsyncOptions.BwLimit))
+	}
+	if rsyncOptions.Archive {
+		rsyncOpts = append(rsyncOpts, "--archive")
+	}
+	if rsyncOptions.Delete {
+		rsyncOpts = append(rsyncOpts, "--delete")
+		// --delete option does not work without --recursive
```
alaypatel07, line 704: +1, nice catch

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -706,6 +750,7 @@ func (t *Task) createRsyncClientPods() error {
 		return err
 	}
 
+	t.Log.Info(fmt.Sprintf("Using Rsync options %s", strings.Join(t.getRsyncOptions(), " ")))
```
alaypatel07, line 753: qq: should we also have the command echo'ed in the bash of the client pod?

alaypatel07 (approved):
Added a small question, otherwise this LGTM

ACK

pranavgaikwad:
@alaypatel07 The entire Rsync command will now be logged in DVM controller. The volume name in the command is enough info to correlate it with the Rsync client pod.

PR #876, opened 2021-01-06
[MIG-363] Add a "migrationcontroller" label bsl/vsl to identify controller

sseago:
Add a migrationcontroller label to backupstoragelocation and
volumesnapshotlocation to identify the controlling controller.

### DIFF
--- a/pkg/controller/migplan/storage.go
+++ b/pkg/controller/migplan/storage.go
@@ -8,6 +8,8 @@ import (
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	velero "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	kapi "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/controller/migplan/storage.go
```diff
@@ -130,10 +133,31 @@ type PlanStorage struct {
...
+	migrationController := &unstructured.Unstructured{}
+	gvk := schema.GroupVersionKind{
+		Kind:    "migrationcontroller",
+		Group:   "migration.openshift.io",
+		Version: "v1alpha1",
+	}
+	migrationController.SetGroupVersionKind(gvk)
+	migrationControllerName := types.NamespacedName{
+		Namespace: migapi.VeleroNamespace,
+		Name:      "migration-controller",
```
dymurray, line 148: If the user changes the name of this CR this feature won't work correct?

sseago: I had assumed that the CR name was controlled by the operator. Is that not the case?

sseago: I could pull the list and just take the UID of the first one, but it might be better to just not annotate if the operator-created one isn't there.

dymurray: The user creates this from OLM generally, which defaults to this name but it's not the operator that does it.

Grabbing the first one in the list could work since there should only be one... but I'm not sure on the best way to proceed. 

Maybe we get the list, if there is more than one we just don't set it?

sseago: I've updated this to just pull the UID from the first one found.

dymurray (approved):

djwhatle (approved):
LGTM

PR #879, opened 2021-01-07
Bug 1906724: Delete moved NFS PVs during rollback

djwhatle:
https://bugzilla.redhat.com/show_bug.cgi?id=1906724

This PR will rollback (delete) NFS PV resources that were moved over to the target cluster during a migration so that the next migration attempt can succeed.

For some additional context: PVs are normally removed by their provisioner when we delete the corresponding PVC (already handled in rollback). In the NFS move case, we create PVs along with PVCs. This is a special case since PVs are cluster-scoped and prior to this PR, rollback didn't handle cluster-scoped resources.

---

I spoke with @sseago on this and he suggested that we should only delete dest PVs during rollback when these 3 are true:

1. The PV was migrated by our migplan
2. The PV is NFS
3. The PV has `ReclaimPolicy=Retain`

---

I've seen success with `migrate -> rollback -> migrate` using this PR.

image

### DIFF
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -69,7 +69,7 @@ func (r *Compare) Compare() (map[string][]schema.GroupVersionResource, error) {
 		return nil, err
 	}
 
-	dstResourceList, err := collectResources(r.DstDiscovery)
+	dstResourceList, err := collectNamespacedResources(r.DstDiscovery)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/gvk/gvk.go
```diff
@@ -119,8 +119,8 @@ func toSet(strSlice []string) mapset.Set {
 	return mapset.NewSetFromSlice(interfaceSlice)
 }
 
-// collectResources collects all namespaced scoped apiResources from the cluster
-func collectResources(discovery discovery.DiscoveryInterface) ([]*metav1.APIResourceList, error) {
+// collectNamespacedResources collects all namespace-scoped apiResources from the cluster
+func collectNamespacedResources(discovery discovery.DiscoveryInterface) ([]*metav1.APIResourceList, error) {
```
djwhatle, line 123: I made a few changes to the `collectResources` and `GetGVRsForCluster` funcs to more accurately reflect that they are only retrieving NS scoped resources.

sseago (approved):
LGTM

PR #882, opened 2021-01-11
Checking desired number of objects before returning "True" in phases

### LINKED ISSUE #875: DVM: Check status of current phase before returning "true" to move to next phase
The DVM controller returns `true` in some edge cases where returning `true` is not the correct approach. For instance [here](https://github.com/konveyor/mig-controller/blob/c7801d91c2c4e70b6ba2b73d8bffcd83da46ee36/pkg/controller/directvolumemigration/stunnel.go#L493), the function `areStunnelClientPodsRunning` will return `true` even if there are 0 stunnel pods running. 

**Expected behavior**
In such cases, the DVM controller should put up a warning depicting the expected behavior (in the above example, it should warn that there are no stunnel pods running instead of returning `true`) 

**Instances found**

https://github.com/konveyor/mig-controller/blob/master/pkg/controller/directvolumemigration/namespace.go#L24 (Can be discarded)
https://github.com/konveyor/mig-controller/blob/master/pkg/controller/directvolumemigration/rsync.go#L72
https://github.com/konveyor/mig-controller/blob/master/pkg/controller/directvolumemigration/stunnel.go#L493

JaydipGabani:
This PR checks for the desired number of objects before moving to the next phase wherever needed.

Resolves #875

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -110,8 +110,7 @@ func (t *Task) areRsyncTransferPodsRunning() (bool, error) {
 		if err != nil {
 			return false, err
 		}
-		verified := t.verifyNumberOfObjects(len(pods.Items), 1)
-		if !verified {
+		if len(pods.Items) != 1 {
 			return false, nil
 		}
 		for _, pod := range pods.Items {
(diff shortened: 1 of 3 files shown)


alaypatel07 (approved):
LGTM

dymurray (approved):

PR #883, opened 2021-01-11
Bug 1908270 - Limit route prefix to 63 characters

dymurray:


dymurray:
https://github.com/konveyor/mig-operator/pull/547

cmd/manager/main.go
```diff
@@ -86,6 +87,10 @@ func main() {
 		log.Error(err, "unable to add OpenShift route APIs to scheme")
 		os.Exit(1)
 	}
+	if err := operatorv1.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "unable to add OpenShift operator APIs to scheme")
```
djwhatle, line 91: Is this only available in OCP 4?

dymurray: Yes. This section will only run on the destination cluster which is guaranteed to be OCP 4.x.

dymurray: @djwhatle is there a concern putting OCP4.x specific apis to the manager scheme?

djwhatle: Right now our controller works on OCP 3.x and 4.x, so this seems like it would be a breaking change.

sseago: @djwhatle @dymurray Is this running on the destination cluster or the controller cluster? Controller can definitely be 3.x. Also, at one point things still had to work with dest being 3.x. Is that no longer the case? Is it now acceptable for a 3.x dest cluster to break completely?

dymurray:
I am closing this as discussed with Derek and Scott we are going to go with a simpler change for Release 1.4.0 and adding some additional validation with intention to make this more robust in a future release

Issue #865, opened 2020-12-16
[MIG-284] Direct migration resources are not cleaned up when migration is canceled


dymurray:
We added support for rollback but a cancel does not stop the DVM/DIM from running once it has started and does not go and clean up the direct migration resources.

PR #884, opened 2021-01-11
Cleanup Direct migration Resources when the Cancel operation is invoked

shubham-pampattiwar:
This PR does the following:
- Adds two new phases in `Cancel` Itinerary - `DeleteDirectVolumeMigrationResources` and `DeleteDirectImageMigrationResources`.
- The phase `DeleteDirectVolumeMigrationResources` fetches DVM and DVMP for a particular migration and deletes it.
- The phase `DeleteDirectImageMigrationResources` fetches DIM and DISM for a particular migration and deletes it.
- These two phases are only invoked when the `Cancel` operation is performed for a Direct Migration.
- Fixes https://github.com/konveyor/mig-controller/issues/865.
- Also, adds correlation labels on DVMP.

### DIFF
--- a/pkg/apis/migration/v1alpha1/directvolumemigration_types.go
+++ b/pkg/apis/migration/v1alpha1/directvolumemigration_types.go
@@ -117,9 +117,3 @@ func (r *DirectVolumeMigration) HasErrors() bool {
 func init() {
 	SchemeBuilder.Register(&DirectVolumeMigration{}, &DirectVolumeMigrationList{})
 }
-
-func (r *DirectVolumeMigration) DirectVolumeMigrationLabels() map[string]string {
-	labels := r.GetCorrelationLabels()
-	labels["directvolumemigration"] = string(r.UID)
-	return labels
-}
--- a/pkg/apis/migration/v1alpha1/directvolumemigration_types.go
+++ b/pkg/apis/migration/v1alpha1/directvolumemigration_types.go
@@ -117,3 +117,9 @@ func (r *DirectVolumeMigration) HasErrors() bool {
 func init() {
 	SchemeBuilder.Register(&DirectVolumeMigration{}, &DirectVolumeMigrationList{})
 }
+
+func (r *DirectVolumeMigration) DirectVolumeMigrationLabels() map[string]string {
+	labels := r.GetCorrelationLabels()
+	labels["directvolumemigration"] = string(r.UID)
+	return labels
+}
--- a/pkg/controller/migmigration/directimagemigration.go
+++ b/pkg/controller/migmigration/directimagemigration.go
@@ -19,7 +19,6 @@ package migmigration
 import (
 	"context"
 	"fmt"
-	"github.com/konveyor/mig-controller/pkg/compat"
 	"k8s.io/apimachinery/pkg/labels"
 
 	liberr "github.com/konveyor/controller/pkg/error"
@@ -43,13 +42,12 @@ func (t *Task) getDirectImageMigration() (*migapi.DirectImageMigration, error) {
 	return nil, nil
 }
 
-func (t *Task) getDirectImageStreamMigrationForDIM(dim *migapi.DirectImageMigration, destClient compat.Client) (*migapi.DirectImageStreamMigration, error) {
+func (t *Task) getDirectImageStreamMigrationForDIM(dim *migapi.DirectImageMigration) (*migapi.DirectImageStreamMigration, error) {
 
 	dimLabels := dim.GetCorrelationLabels()
-	dimLabels["directimagemigration"] = string(dim.UID)
 	selector := labels.SelectorFromSet(dimLabels)
 	dismList := migapi.DirectImageStreamMigrationList{}
-	err := destClient.List(context.TODO(),
+	err := t.Client.List(context.TODO(),
 		&k8sclient.ListOptions{
 			LabelSelector: selector,
 		},
@@ -111,11 +109,6 @@ func (t *Task) setDirectImageMigrationWarning(dim *migapi.DirectImageMigration) 
 }
 
 func (t *Task) deleteDirectImageMigrationResources() error {
-	// fetch the destination cluster client
-	destClient, err := t.getDestinationClient()
-	if err != nil {
-		return liberr.Wrap(err)
-	}
 
 	// fetch the DIM
 	dim, err := t.getDirectImageMigration()
@@ -123,22 +116,24 @@ func (t *Task) deleteDirectImageMigrationResources() error {
 		return liberr.Wrap(err)
 	}
 
-	// fetch the corresponding DISM instance
-	dism, err := t.getDirectImageStreamMigrationForDIM(dim, destClient)
-	if err != nil {
-		return liberr.Wrap(err)
-	}
+	if dim != nil {
+		// fetch the corresponding DISM instance
+		dism, err := t.getDirectImageStreamMigrationForDIM(dim)
+		if err != nil {
+			return liberr.Wrap(err)
+		}
 
-	// delete the DISM instance
-	err = destClient.Delete(context.TODO(), dism)
-	if err != nil {
-		return liberr.Wrap(err)
-	}
+		// delete the DISM instance
+		err = t.Client.Delete(context.TODO(), dism)
+		if err != nil {
+			return liberr.Wrap(err)
+		}
 
-	// delete the DIM instance from destination cluster
-	err = destClient.Delete(context.TODO(), dim)
-	if err != nil {
-		return liberr.Wrap(err)
+		// delete the DIM instance from destination cluster
+		err = t.Client.Delete(context.TODO(), dim)
+		if err != nil {
+			return liberr.Wrap(err)
+		}
 	}
 
 	return nil
--- a/pkg/controller/migmigration/directimagemigration.go
+++ b/pkg/controller/migmigration/directimagemigration.go
@@ -120,24 +120,10 @@ func (t *Task) deleteDirectImageMigrationResources() error {
 		return liberr.Wrap(err)
 	}
 
-	if dim != nil {
-		// fetch the corresponding DISM instance
-		dism, err := t.getDirectImageStreamMigrationForDIM(dim)
-		if err != nil {
-			return liberr.Wrap(err)
-		}
-
-		// delete the DISM instance
-		err = t.Client.Delete(context.TODO(), dism)
-		if err != nil {
-			return liberr.Wrap(err)
-		}
-
-		// delete the DIM instance from destination cluster
-		err = t.Client.Delete(context.TODO(), dim)
-		if err != nil {
-			return liberr.Wrap(err)
-		}
+	// delete the DIM instance
+	err = t.Client.Delete(context.TODO(), dim)
+	if err != nil {
+		return liberr.Wrap(err)
 	}
 
 	return nil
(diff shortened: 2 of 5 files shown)


pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -117,3 +117,9 @@ func (r *DirectVolumeMigration) HasErrors() bool {
 func init() {
 	SchemeBuilder.Register(&DirectVolumeMigration{}, &DirectVolumeMigrationList{})
 }
+
+func (r *DirectVolumeMigration) DirectVolumeMigrationLabels() map[string]string {
+	labels := r.GetCorrelationLabels()
+	labels["directvolumemigration"] = string(r.UID)
```
djwhatle, line 123: Are lines 122 and 123 doing the same thing?


Correlation label looks like this:

```
// Build label (key, value) used to correlate CRs.
// Format: <kind>: <uid>.  The <uid> should be the ObjectMeta.UID
func CorrelationLabel(r interface{}, uid types.UID) (key, value string) {
	return labelKey(r), string(uid)
}
```

`"directvolumemigration": "<uid>"`

pkg/controller/migmigration/directimagemigration.go
```diff
@@ -41,6 +43,25 @@ func (t *Task) getDirectImageMigration() (*migapi.DirectImageMigration, error) {
 	return nil, nil
 }
 
+func (t *Task) getDirectImageStreamMigrationForDIM(dim *migapi.DirectImageMigration, destClient compat.Client) (*migapi.DirectImageStreamMigration, error) {
+
+	dimLabels := dim.GetCorrelationLabels()
+	dimLabels["directimagemigration"] = string(dim.UID)
```
djwhatle, line 49: Again, I'm confused as to what line 49 is adding on top of line 48 here.

pkg/controller/migmigration/directimagemigration.go
```diff
@@ -88,3 +109,37 @@ func (t *Task) setDirectImageMigrationWarning(dim *migapi.DirectImageMigration)
 		})
 	}
 }
+
+func (t *Task) deleteDirectImageMigrationResources() error {
+	// fetch the destination cluster client
+	destClient, err := t.getDestinationClient()
```
djwhatle, line 115: We should not assume that the DIM resource lives on the target cluster, as this will not always be the case (e.g. if controller is running on src cluster). We should instead use the controller clusters client, `t.Client`

pkg/controller/migmigration/dvm.go
```diff
@@ -76,6 +79,25 @@ func (t *Task) getDirectVolumeMigration() (*migapi.DirectVolumeMigration, error)
 	return nil, nil
 }
 
+func (t *Task) getDirectVolumeMigrationProgressForDVM(dvm *migapi.DirectVolumeMigration, destClient compat.Client) (*migapi.DirectVolumeMigrationProgress, error) {
+
+	dvmLabels := dvm.GetCorrelationLabels()
+	dvmLabels["directvolumemigration"] = string(dvm.UID)
```
djwhatle, line 85: Same question about why we are modifying what comes out of `GetCorrelationLabels`

pkg/controller/migmigration/dvm.go
```diff
@@ -172,3 +194,38 @@ func (t *Task) getDirectVolumeClaimList() *[]migapi.PVCToMigrate {
 	}
 	return nil
 }
+
+func (t *Task) deleteDirectVolumeMigrationResources() error {
+
+	// fetch the destination cluster client
+	destClient, err := t.getDestinationClient()
```
djwhatle, line 201: Switch to `t.Client` for reason above.

pkg/controller/migmigration/task.go
```diff
@@ -231,6 +233,8 @@ var CancelItinerary = Itinerary{
 		{Name: DeleteRestores, Step: StepCleanupVelero},
 		{Name: DeleteRegistries, Step: StepCleanupHelpers},
 		{Name: DeleteHookJobs, Step: StepCleanupHelpers},
+		{Name: DeleteDirectVolumeMigrationResources, Step: StepCleanupHelpers, all: DirectVolume},
```
djwhatle, line 236: These look good 👍

pkg/controller/migmigration/task.go
```diff
@@ -949,6 +953,22 @@ func (t *Task) Run() error {
 				return liberr.Wrap(err)
 			}
 		}
+	case DeleteDirectVolumeMigrationResources:
```
djwhatle, line 956: This looks good.

djwhatle (changes requested):

djwhatle:
fyi @shubham-pampattiwar I modified the title to remove the jira ticket. You only need the ticket ID on the github issue, that's what our bot is looking for.

pkg/controller/migmigration/directimagemigration.go
```diff
@@ -41,6 +42,24 @@ func (t *Task) getDirectImageMigration() (*migapi.DirectImageMigration, error) {
...
+		&k8sclient.ListOptions{
+			LabelSelector: selector,
+		},
+		&dismList)
+
+	if err != nil {
+		return nil, liberr.Wrap(err)
+	}
+
+	return &dismList.Items[0], nil
```
shawn-hurley, line 60: can't Items be an empty list?

pkg/controller/migmigration/dvm.go
```diff
@@ -78,6 +80,24 @@ func (t *Task) getDirectVolumeMigration() (*migapi.DirectVolumeMigration, error)
...
+		&k8sclient.ListOptions{
+			LabelSelector: selector,
+		},
+		&dvmpList)
+
+	if err != nil {
+		return nil, liberr.Wrap(err)
+	}
+
+	return &dvmpList.Items[0], nil
```
shawn-hurley, line 98: Items can be empty I think this will panic as well

shawn-hurley (changes requested):
we handle nil return that will panic in the case it should return nil I think.

pkg/controller/migmigration/directimagemigration.go
shubham-pampattiwar: Thanks for the catch @shawn-hurley , updated the code.

pkg/controller/migmigration/dvm.go
shubham-pampattiwar: Done, updated the code/

pkg/controller/migmigration/directimagemigration.go
shubham-pampattiwar: Done, updated the code.

pkg/controller/migmigration/dvm.go
shubham-pampattiwar: Done

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
shubham-pampattiwar: Thank you @djwhatle , I have made the appropriate changes.

pkg/controller/migmigration/dvm.go
shubham-pampattiwar: Done, removed the redundant code.

shawn-hurley (approved):

pkg/controller/migmigration/directimagemigration.go
```diff
@@ -41,6 +42,28 @@ func (t *Task) getDirectImageMigration() (*migapi.DirectImageMigration, error) {
...
+			LabelSelector: selector,
+		},
+		&dismList)
+
+	if err != nil {
+		return nil, liberr.Wrap(err)
+	}
+
+	if len(dismList.Items) > 0 {
+		return &dismList.Items[0], nil
```
sseago, line 61: Shouldn't we be cleaning up all DISM resources, not just the first? There will be one for every imagestream in the migrated namespaces.

pkg/controller/migmigration/directimagemigration.go
```diff
@@ -88,3 +111,34 @@ func (t *Task) setDirectImageMigrationWarning(dim *migapi.DirectImageMigration)
...
+
+	// fetch the DIM
+	dim, err := t.getDirectImageMigration()
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	if dim != nil {
+		// fetch the corresponding DISM instance
+		dism, err := t.getDirectImageStreamMigrationForDIM(dim)
```
sseago, line 125: As above, this should return a list of DISM resources rather than just one.

sseago (changes requested):

pkg/controller/migmigration/dvm.go
```diff
@@ -78,6 +80,28 @@ func (t *Task) getDirectVolumeMigration() (*migapi.DirectVolumeMigration, error)
...
+		&k8sclient.ListOptions{
+			LabelSelector: selector,
+		},
+		&dvmpList)
+
+	if err != nil {
+		return nil, liberr.Wrap(err)
+	}
+
+	if len(dvmpList.Items) > 0 {
```
dymurray, line 98: This should be cleaning up all DVMs with the label selector.

pkg/controller/migmigration/dvm.go
```diff
@@ -197,3 +221,34 @@ func (t *Task) getDirectVolumeClaimList() *[]migapi.PVCToMigrate {
...
+
+	// fetch the DVM
+	dvm, err := t.getDirectVolumeMigration()
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	if dvm != nil {
+		// fetch the corresponding DVMP instance
+		dvmp, err := t.getDirectVolumeMigrationProgressForDVM(dvm)
```
dymurray, line 235: This logic may be unnecessary. I believe we set an owner ref on the DVMP to the DVM. This means we should be able to just delete the DVM and rely on the garbage collector to get the dvmps

pkg/controller/migmigration/dvm.go
shubham-pampattiwar: Done, updated the code.

pkg/controller/migmigration/dvm.go
shubham-pampattiwar: Thanks for the tip @dymurray , I have updated the code just to delete the DVM and DIM, this surely cascades the delete all to all the subsequent DVMPs and DISMs, I tested this for 2 PVs and 2 IS in a migplan.

pkg/controller/migmigration/directimagemigration.go
shubham-pampattiwar: Updated the delete logic.

pkg/controller/migmigration/directimagemigration.go
shubham-pampattiwar: Done

dymurray (approved):
VISIACK

sseago (approved):

djwhatle (approved):
Code looks good. Nice work @shubham-pampattiwar. If you see success in a manual test with this feel free to merge.

shubham-pampattiwar:
> Code looks good. Nice work @shubham-pampattiwar. If you see success in a manual test with this feel free to merge.

Thank you @djwhatle , just a added few `nil` checks to the PR before deleting DVM/DIM. I tested out the cancel operation successfully for the following scenarios:
- 1 PV and 1 IS (Cancel operation triggered after Direct Migration resources were created)
- 2 PV, 2 IS and 2 NS (Cancel operation triggered after Direct Migration resources were created)
- 1 PV and 1 IS (Cancel operation triggered after DIM was created and DVM was not)

Going ahead with the merge.

PR #885, opened 2021-01-13
BUG 1908214: carry fsGroup/supp-group id if there are pods running mounted on a pvc

alaypatel07:
This avoids breaking the source application pod on gp2

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -830,19 +830,27 @@ func (t *Task) getfsGroupMapForNamespace() (map[string][]PVCWithSecurityContext,
 		if err != nil {
 			return nil, err
 		}
-		for _, claimName := range pvcs {
-			for _, p := range podList.Items {
-				if !isClaimUsedByPod(claimName, &p) {
-					continue
+
+		// for each namespace, have a pvc->SCC map to look up in the pvc loop later
+		// we will use the scc of the last pod in the list mounting the pvc
+		pvcSecurityContextMapForNamespace := map[string]PVCWithSecurityContext{}
+		for _, pod := range podList.Items {
+			for _, vol := range pod.Spec.Volumes {
+				if vol.PersistentVolumeClaim != nil {
+					pvcSecurityContextMapForNamespace[vol.PersistentVolumeClaim.ClaimName] = PVCWithSecurityContext{
+						name:               vol.PersistentVolumeClaim.ClaimName,
+						fsGroup:            pod.Spec.SecurityContext.FSGroup,
+						supplementalGroups: pod.Spec.SecurityContext.SupplementalGroups,
+						seLinuxOptions:     pod.Spec.SecurityContext.SELinuxOptions,
+					}
 				}
-				pvcSecurityContextMap[ns] = append(pvcSecurityContextMap[ns], PVCWithSecurityContext{
-					name:               claimName,
-					fsGroup:            p.Spec.SecurityContext.FSGroup,
-					supplementalGroups: p.Spec.SecurityContext.SupplementalGroups,
-					seLinuxOptions:     p.Spec.SecurityContext.SELinuxOptions,
-				})
-				// get the first lucky pod's fsgroup to avoid selection problem.
-				break
+			}
+		}
+
+		for _, claimName := range pvcs {
+			pss, exists := pvcSecurityContextMapForNamespace[claimName]
+			if exists {
+				pvcSecurityContextMap[ns] = append(pvcSecurityContextMap[ns], pss)
 			}
 			// pvc not used by any pod
 			pvcSecurityContextMap[ns] = append(pvcSecurityContextMap[ns], PVCWithSecurityContext{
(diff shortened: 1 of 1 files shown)


alaypatel07:
A follow-up for converting this to a non-privileged implementation pending but this is the bare minimum that should make the 1.4.0 cut.

alaypatel07:
Also, I will feel better if this has unit tests, I will make that call tomorrow morning just in case anyone is reviewing

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -886,9 +966,17 @@ func (t *Task) createRsyncClientPods() error {
 					Requests: requests,
 				},
 			})
+			//var fsGroup *int64
```
shawn-hurley, line 969: Left over testing code? otherwise can you explain in the comment some more?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -792,6 +792,79 @@ func (t *Task) getRsyncOptions() []string {
...
+		srcClient, err := t.getSourceClient()
+		if err != nil {
+			return nil, err
+		}
+		podList := &corev1.PodList{}
+		err = srcClient.List(context.TODO(), &k8sclient.ListOptions{Namespace: ns}, podList)
+		if err != nil {
+			return nil, err
+		}
+		for _, claimName := range pvcs {
```
shawn-hurley, line 833: What if the claim is in [Lost](https://github.com/kubernetes/api/blob/cb95709d38de6602153fd4f5c05b036ba88256f5/core/v1/types.go#L592) status

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -792,6 +792,79 @@ func (t *Task) getRsyncOptions() []string {
...
+			for _, p := range podList.Items {
```
shawn-hurley, line 834: nit: premature optimization, make a map from the pods list to `map[vol.PersistentVolumeClaim.ClaimName]Pod or to PodSecurityContext and then you can look up in the map instead of doing multiple loops?

shawn-hurley (changes requested):
Some suggestions the handling of a lost PVC I think is the only one I would consider a blocker.

pkg/controller/directvolumemigration/rsync.go
alaypatel07: left over needs to be gone!!

pkg/controller/directvolumemigration/rsync.go
alaypatel07: talked with shawn on slack we will take this issue up in 1.4.z

pkg/controller/directvolumemigration/rsync.go
alaypatel07: +1 this optimization is agreed upon especially since we dont have cache enabled for remote clusters. will produce o(n) + o(m) on number of pods(n) and number of pvc(m)

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -792,6 +792,86 @@ func (t *Task) getRsyncOptions() []string {
...
+		if err != nil {
+			return nil, err
+		}
+		podList := &corev1.PodList{}
+		err = srcClient.List(context.TODO(), &k8sclient.ListOptions{Namespace: ns}, podList)
+		if err != nil {
+			return nil, err
+		}
+
+		// for each namespace, have a pod->SCC map to look up in the pvc loop later
```
shawn-hurley, line 834: nit: s/pod->SCC map/PVC -> SCC map

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -792,6 +792,86 @@ func (t *Task) getRsyncOptions() []string {
...
+		podSecurityContextMap := map[string]PVCWithSecurityContext{}
```
alaypatel07, line 835: s/podSecurityContextMap/pvcSecurityContextMap

pkg/controller/directvolumemigration/rsync.go
shawn-hurley: Maybe also add a statement on why we just always overwrite the map value, say a two pods have the same claim, we would overwrite the claim name with the second pod. NIT: comment explaining

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -829,7 +913,10 @@ func (t *Task) createRsyncClientPods() error {
 		// Get stunnel svc IP
 		svc := corev1.Service{}
 		key := types.NamespacedName{Name: DirectVolumeMigrationRsyncTransferSvc, Namespace: ns}
-		srcClient.Get(context.TODO(), key, &svc)
+		err := srcClient.Get(context.TODO(), key, &svc)
+		if err != nil {
```
shawn-hurley, line 917: +1

shawn-hurley (approved):

dymurray (approved):

jmontleon (approved):
VISIACK

PR #747, opened 2020-11-05
Bug 1871059: Adding ability to wait for init containers to start and finish

shawn-hurley:
* Updating container status check to check the container state not readiness probe
* Adding warn condition on StageReport
* Handle stage pod not able to pull image.

PR #889, opened 2021-01-14
Bug 1907828: Adding claim health checking for a waiting pods

shawn-hurley:
when a pod is in a waiting / ContainerCreating state, it can stay there because of a PVC issue (not bound, deleted, not found) This was initially introduced here: #583 and was regressed in #747


dymurray (approved):

pkg/controller/migmigration/stage.go
```diff
@@ -487,6 +487,12 @@ func (t *Task) stagePodReport(client k8sclient.Client) (report PodStartReport, e
 		for _, c := range pod.Status.InitContainerStatuses {
 			// If the init contianer is waiting, then nothing can happen.
 			if c.State.Waiting != nil {
+				// If the pod has unhealthy claims, we will fail the migration
```
alaypatel07, line 490: +1

alaypatel07 (approved):
VISIACK, thanks @shawn-hurley

PR #890, opened 2021-01-14
Bug 1912246 - Add checksum flag to rsync when verify is specified

dymurray:


pkg/controller/directvolumemigration/rsync.go
```diff
@@ -948,6 +955,9 @@ func (t *Task) createRsyncClientPods() error {
 			})
 			rsyncCommand := []string{"rsync"}
 			rsyncCommand = append(rsyncCommand, t.getRsyncOptions()...)
+			if vol.verify {
+				rsyncCommand = append(rsyncCommand, "--checksum")
```
alaypatel07, line 959: I have a question here: if a pod already exists without the checksum command on the source side, https://github.com/konveyor/mig-controller/pull/890/files#diff-7299a3c4e600e2733571d1d39c64bbfe6c7e448a6bb15aa472514a173d3095c0R1015 we would not be creating one. So we are promissing user to run with checksum but dont actual run it be checksum in that case?

alaypatel07: this will work for 90% of the case I think but might need a followup to fix it for good

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -26,6 +26,7 @@ type PVCToMigrate struct {
 	*kapi.ObjectReference `json:",inline"`
 	TargetStorageClass    string                            `json:"targetStorageClass"`
 	TargetAccessModes     []kapi.PersistentVolumeAccessMode `json:"targetAccessModes,omitEmpty"`
+	Verify                bool                              `json:"verify,omitEmpty"`
```
alaypatel07, line 29: this will require a mig-operator PR?

alaypatel07 (changes requested):
Needs a rebase and https://github.com/dymurray/mig-controller/pull/3

dymurray:
https://github.com/konveyor/mig-operator/pull/553

alaypatel07 (approved):
ACK

PR #892, opened 2021-01-15
Revert "Update DVM PVC check to include determination whether PVCs ar…

alaypatel07:
…e bound and usable (#852)"

This reverts commit 8b5de2ef

The original PR assumes that we add migration labels to pvc. There are two problems with it:

1. PVC's could be pre-created in customer env and not having the labels, migration will be stuck if that is the case
2. It is not right to add labels like `app: xxx`, `owner: yyy` to pvc, they might have their own labels with same key that can conflict

Given the above, we should revert the change and fix it cleanly post 1.4.0

### DIFF
--- a/pkg/controller/directvolumemigration/pvcs.go
+++ b/pkg/controller/directvolumemigration/pvcs.go
@@ -2,15 +2,10 @@ package directvolumemigration
 
 import (
 	"context"
-	"fmt"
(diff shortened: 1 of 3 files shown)


JaydipGabani (approved):
+1 on reverting the changes and fixing it post 1.4.0

dymurray (approved):

PR #897, opened 2021-01-18
Bug 1908270: Add namespace length validation and subdomain override

dymurray:
Operator PR: https://github.com/konveyor/mig-operator/pull/552

### DIFF
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -57,7 +57,7 @@ const (
 	RegistryImageKey      = "REGISTRY_IMAGE"
 	StagePodImageKey      = "STAGE_IMAGE"
 	RsyncTransferImageKey = "RSYNC_TRANSFER_IMAGE"
-	RsyncTransferImageKey = "CLUSTER_SUBDOMAIN"
+	ClusterSubdomainKey   = "CLUSTER_SUBDOMAIN"
 )
 
 // MigClusterSpec defines the desired state of MigCluster
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -189,7 +189,7 @@ func (m *MigCluster) GetClusterSubdomain(c k8sclient.Client) (string, error) {
 		return "", liberr.Wrap(err)
 	}
 	clusterSubdomain, ok := clusterConfig.Data[ClusterSubdomainKey]
-	if !ok {
+	if !ok || clusterSubdomain == "" {
 		return "", liberr.Wrap(errors.Errorf("configmap key not found: %v", ClusterSubdomainKey))
 	}
 	return clusterSubdomain, nil
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -57,6 +57,7 @@ const (
 	RegistryImageKey      = "REGISTRY_IMAGE"
 	StagePodImageKey      = "STAGE_IMAGE"
 	RsyncTransferImageKey = "RSYNC_TRANSFER_IMAGE"
+	RsyncTransferImageKey = "CLUSTER_SUBDOMAIN"
 )
 
 // MigClusterSpec defines the desired state of MigCluster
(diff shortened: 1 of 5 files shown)


pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -175,6 +176,25 @@ func (m *MigCluster) GetRsyncTransferImage(c k8sclient.Client) (string, error) {
...
+	}
+	clusterConfig := &corev1.ConfigMap{}
+	clusterConfigRef := types.NamespacedName{Name: ClusterConfigMapName, Namespace: VeleroNamespace}
+	err = client.Get(context.TODO(), clusterConfigRef, clusterConfig)
+	if err != nil {
+		return "", liberr.Wrap(err)
+	}
+	clusterSubdomain, ok := clusterConfig.Data[ClusterSubdomainKey]
+	if !ok {
+		return "", liberr.Wrap(errors.Errorf("configmap key not found: %v", ClusterSubdomainKey))
```
djwhatle, line 193: For anyone else reading this about to suggest we not return an err here, the err is ignored so it's OK

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -352,6 +353,32 @@ func (t *Task) createRsyncTransferRoute() error {
...
+		// Get cluster subdomain if it exists
+		cluster, err := t.Owner.GetDestinationCluster(t.Client)
+		if err != nil {
+			return err
+		}
+		// Ignore error since this is optional config and won't break
+		// anything if it doesn't exist
+		subdomain, _ := cluster.GetClusterSubdomain(t.Client)
+
+		// This is a backdoor setting to help guarantee DVM can still function if a
```
djwhatle, line 365: Big +1 for adding these comments here

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -645,6 +672,19 @@ func (t *Task) getRsyncRoute(namespace string) (string, error) {
...
+	admitted := false
+	for _, ingress := range route.Status.Ingress {
+		for _, condition := range ingress.Conditions {
+			if condition.Type == routev1.RouteAdmitted {
+				admitted = true
+				break
+			}
+		}
+	}
+	if !admitted {
```
djwhatle, line 685: How soon does this run after route creation? Is there room for a race condition here where the route controller could be slow to react and we fail a migration due to that?

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -175,6 +176,25 @@ func (m *MigCluster) GetRsyncTransferImage(c k8sclient.Client) (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+	clusterConfig := &corev1.ConfigMap{}
+	clusterConfigRef := types.NamespacedName{Name: ClusterConfigMapName, Namespace: VeleroNamespace}
+	err = client.Get(context.TODO(), clusterConfigRef, clusterConfig)
+	if err != nil {
+		return "", liberr.Wrap(err)
+	}
+	clusterSubdomain, ok := clusterConfig.Data[ClusterSubdomainKey]
```
alaypatel07, line 191: probably need to check if data is not nil to avoid panics

pkg/controller/directvolumemigration/rsync.go
alaypatel07: +1 yes, this is not something we should fail at. Probably follow the same pattern if the route is not admitted within 10 mins, raise a warning like we do for pending pods that will be bubbled up to the UI

pkg/controller/migplan/validation.go
```diff
@@ -247,10 +249,46 @@ func (r ReconcileMigPlan) validateNamespaces(plan *migapi.MigPlan) error {
 		})
 		return nil
 	}
+	namespaces := r.validateNamespaceLengthForDVM(plan)
+	if len(namespaces) > 0 {
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     NsLengthExceeded,
+			Status:   True,
+			Reason:   LengthExceeded,
+			Category: Warn,
```
alaypatel07, line 258: should we make this critical?

if length >59 and configmap does not have a domain, migplan has a critical condition and is not ready?

dymurray: I preferred not to prevent the migplan from running since this case should be rare... but I suppose it's fair to say we should be raising a critical condition since the route will likely fail to get created

pkg/controller/directvolumemigration/rsync.go
dymurray: +1 good point. Will add a new task to the itinerary.

pkg/controller/migplan/validation.go
dymurray: We agreed to keep this as a warning for now on scrum. I confirmed it is displayed in the UI at the final screen of the wizard.

pkg/controller/directvolumemigration/task.go
```diff
@@ -258,6 +260,36 @@ func (t *Task) Run() error {
 		if err = t.next(); err != nil {
 			return liberr.Wrap(err)
 		}
+	case EnsureRsyncRouteAdmitted:
+		admitted, reasons, err := t.areRsyncRoutesAdmitted()
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		if admitted {
+			t.Requeue = NoReQ
```
djwhatle, line 269: So if the route is admitted we are not requeuing, how does the next reconcile get triggered? I expected a FastReQ here, am I missing something?

djwhatle (approved):

pkg/controller/directvolumemigration/task.go
djwhatle: Answer: no predicates in DVM, so echo reconcile causes requeue

alaypatel07 (approved):
VISIACK

PR #903, opened 2021-01-20
Bug 1918503: Prevent panic if migmigration object is deleted while DVM runs

dymurray:


pkg/controller/directvolumemigration/migrate.go
```diff
@@ -34,7 +38,7 @@ func (r *ReconcileDirectVolumeMigration) migrate(direct *migapi.DirectVolumeMigr
 	}
 	err = task.Run()
 	if err != nil {
-		if errors.IsConflict(err) {
+		if k8serrors.IsConflict(err) {
```
alaypatel07, line 41: Will have to use unwrap here. Please wait for #902 to get in

dymurray: Done

alaypatel07 (approved):
ACK

PR #904, opened 2021-01-20
Adding github actions for automation for BZ automation

shawn-hurley:
* Adding BZ automation for when a PR is created for a BZ, it will be attached to the BZ
* Verifies that the BZ is in the valid state to be work (one of "New, "Assigned", "Post")
* Will alert the user with comments on the PR to tell them if the association was successful
* Adding BZ automation for when a PR is merged.
* Maps branch to target releases to determine if the fix is for the bug or not.
  * example merging bz fix into master then cherry picking will only mark bz modified on the cherry-pick
  * Will handle if more then 1 PR is attached that is open, it will not mark the BZ as modified until all PR's are closed that are attached

TODO: Before this merges need to create a BZ robot account

### DIFF
--- a/.github/workflows/bz-pr-action.yml
+++ b/.github/workflows/bz-pr-action.yml
@@ -1,0 +1,38 @@
+# This is a basic workflow to help you get started with Actions
+
+name: BZ PR Creation
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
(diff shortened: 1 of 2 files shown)


shawn-hurley:
Going to use my token for now so I will be updating a lot of BZ's

Going to try and get a robot created

alaypatel07 (approved):
ACK

PR #913, opened 2021-01-22
Bug 1919366: Using non-root rsyncd conf for direct volume migration

shawn-hurley:
* Fixing a inefficent call for image streams and pvcs that was causing slowdown
* using fake super user to perserve file attributes
* using no chroot to prevent chroot to mount which we don't have permissions for

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -80,8 +80,8 @@ data:
     [{{ $pvc.Name }}]
         comment = archive for {{ $pvc.Name }}
         path = /mnt/{{ $.Namespace }}/{{ $pvc.Name }}
-        uid = root
-        gid = root
+        use chroot = no
+        fake super = yes
         list = yes
         hosts allow = ::1, 127.0.0.1, localhost
         auth users = {{ $.SshUser }}
--- a/pkg/controller/migmigration/task.go
+++ b/pkg/controller/migmigration/task.go
@@ -1031,20 +1031,22 @@ func (t *Task) init() error {
 
 	t.initPipeline()
 
-	hasImageStreams, err := t.hasImageStreams()
-	if err != nil {
-		return err
-	}
+	if t.stage() && !t.Owner.Status.HasCondition(StageNoOp) {
+		hasImageStreams, err := t.hasImageStreams()
+		if err != nil {
+			return err
+		}
 
-	anyPVs, _ := t.hasPVs()
-	if t.stage() && (!anyPVs && !hasImageStreams) {
-		t.Owner.Status.SetCondition(migapi.Condition{
-			Type:     StageNoOp,
-			Status:   True,
-			Category: migapi.Warn,
-			Message:  "Stage migration was run without any PVs or ImageStreams in source cluster. No Velero operations were initiated.",
-			Durable:  true,
-		})
+		anyPVs, _ := t.hasPVs()
+		if !anyPVs && !hasImageStreams {
+			t.Owner.Status.SetCondition(migapi.Condition{
+				Type:     StageNoOp,
+				Status:   True,
+				Category: migapi.Warn,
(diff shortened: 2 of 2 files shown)


pkg/controller/directvolumemigration/rsync.go
```diff
@@ -80,8 +80,8 @@ data:
     [{{ $pvc.Name }}]
         comment = archive for {{ $pvc.Name }}
         path = /mnt/{{ $.Namespace }}/{{ $pvc.Name }}
-        uid = root
-        gid = root
+        use chroot = no
```
djwhatle, line 83: I don't know enough about rsync to make a meaningful ACK here but I guess it worked?

pkg/controller/migmigration/task.go
```diff
@@ -1031,20 +1031,22 @@ func (t *Task) init() error {
 
 	t.initPipeline()
 
-	hasImageStreams, err := t.hasImageStreams()
-	if err != nil {
-		return err
-	}
+	if t.stage() {
+		hasImageStreams, err := t.hasImageStreams()
```
pranavgaikwad, line 1035: good catch!

pkg/controller/migmigration/task.go
```diff
@@ -1031,20 +1031,22 @@ func (t *Task) init() error {
...
-	anyPVs, _ := t.hasPVs()
-	if t.stage() && (!anyPVs && !hasImageStreams) {
-		t.Owner.Status.SetCondition(migapi.Condition{
-			Type:     StageNoOp,
-			Status:   True,
-			Category: migapi.Warn,
-			Message:  "Stage migration was run without any PVs or ImageStreams in source cluster. No Velero operations were initiated.",
-			Durable:  true,
-		})
+		anyPVs, _ := t.hasPVs()
```
djwhatle, line 1040: @shawn-hurley do you think it would be feasible in this PR to add one additional check?

```
if !t.Owner.Status.HasCondition(StageNoOp)
   // only run hasPVs() and hasImageStreams() if this condition hasn't already been checked
```

pkg/controller/directvolumemigration/rsync.go
shawn-hurley: It did:

> fake super
> Setting lqfake super = yesrq for a module causes the daemon side to behave as if the --fake-user command-line option had been specified. This allows the full attributes of a file to be stored without having to have the daemon actually running as root.

> The daemon must run with root privileges if you wish to use chroot, to bind to a port numbered under 1024 (as is the default 873), or to set file ownership. Otherwise, it must just have permission to read and write the appropriate data, log, and lock files.

> use chroot
> If lquse chrootrq is true, the rsync daemon will chroot to the lqpathrq before starting the file transfer with the client.

from https://linux.die.net/man/5/rsyncd.conf

pkg/controller/migmigration/task.go
shawn-hurley: yes I do give me a second

pranavgaikwad (approved):
Same thoughts as Derek mentioned about the Rsync options. Not familiar with the new options. The old options were taken from pvc-migrate as the pod was running as privileged. ACK.

pkg/controller/migmigration/task.go
shawn-hurley: Added let me know if it looks right

pkg/controller/directvolumemigration/rsync.go
djwhatle: @dymurray @alaypatel07 for awareness when you guys get back from PTO

pkg/controller/migmigration/task.go
djwhatle: LGTM

djwhatle (approved):
ACK on the part related to StageNoOp

eriknelson (approved):

PR #920, opened 2021-01-26
Bug 1920548: Adding extra info to explain to user how to fix the problem when value is missing

shawn-hurley:
* And info in the error around how to fix


pranavgaikwad (changes requested):
I ran into this. I think this is a nice solution. However, taking a step further, should we default such undefined variables to a sane value, False in this case? I dont know the consequences, but I am thinking from an user experience point of view.

alaypatel07 (approved):
LGTM

shawn-hurley:
@pranavgaikwad That is what I initially did, Alay convinced me that we should error out, because if the operator is not updated then the DVM will fail still because it does not have the SCC's to create the rsync pods. Basically we are catching the failure earlier and implicitly based on the src/dest cluster's not having an updated operator that would add this value in the config map. 

WDYT?

pranavgaikwad (approved):
Reasonable

PR #932, opened 2021-01-28
Bug 1907828: Adding pvc health checking when pod is pending

shawn-hurley:
* not just init container

### DIFF
--- a/pkg/controller/migmigration/stage.go
+++ b/pkg/controller/migmigration/stage.go
@@ -534,6 +534,12 @@ func (t *Task) stagePodReport(client k8sclient.Client) (report PodStartReport, e
 
 		// handle pod pending status
 		if pod.Status.Phase == corev1.PodPending {
+			// If the pod has unhealthy claims, we will fail the migration
+			// So the user can fix the plan/pvc.
+			if !hasHealthyClaims(&pod) {
+				report.failed = true
+				return
+			}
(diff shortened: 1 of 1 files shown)


jmontleon (approved):
ACK

eriknelson (approved):

alaypatel07 (approved):
ack

pkg/controller/migmigration/stage.go
```diff
@@ -534,6 +534,14 @@ func (t *Task) stagePodReport(client k8sclient.Client) (report PodStartReport, e
 
 		// handle pod pending status
 		if pod.Status.Phase == corev1.PodPending {
+			// If the pod has unhealthy claims, we will fail the migration
+			// So the user can fix the plan/pvc.
+			// pod Spec having the node name, means the pod has been scheduled
+			// becuase the pod has been scheduled we know that the PVC should be bound
+			if pod.Spec.NodeName != "" && !hasHealthyClaims(&pod) {
```
alaypatel07, line 541: understanding here is that if NodeName == "", we wait for the pod to be scheduled. +1 coming from this test case https://github.com/kubernetes/kubernetes/blob/e1c1fd9edf9bd4d9d0e749b16fb33a54399a1ef8/test/integration/volumescheduling/volume_binding_test.go#L831

alaypatel07 (approved):
re-ack

PR #936, opened 2021-01-28
Bug 1920911: Do not include DVM volumes in stage backups

dymurray:


pkg/controller/migmigration/task.go
```diff
@@ -1337,12 +1337,19 @@ func (t *Task) getDestinationClient() (compat.Client, error) {
 }
 
 // Get the persistent volumes included in the plan which are not skipped.
-func (t *Task) getPVs() migapi.PersistentVolumes {
+// This function will only return PVs that are being copied via restic or snapshot
+func (t *Task) getStagePVs() migapi.PersistentVolumes {
```
alaypatel07, line 1341: can we name this `getResticOrSnapshotPVs`? sorry but took me a while to figure out `Stage` means `ResticOrSnapshot`and then saw it in the comments

alaypatel07 (approved):
I wanted to understand how bad writing unit tests would be and it was not that bad

https://github.com/alaypatel07/mig-controller/commit/ab7359b648ce084f578f3c64f8d7b42c5ee4d866 if you think it makes sense please open the PR to your branch and merge it to get the unit tests here. After writing them it makes me far more confident in approving this PR :)

pkg/controller/migmigration/task.go
dymurray: @alaypatel07 this includes `Move` PVs as well which is why I included `Stage` in the name. I.e. any PV that needs to be included in a stage backup.

What do you think?

dymurray: I'll add a comment

PR #939, opened 2021-02-01
Adding comments for each resource spec fields

JaydipGabani:
This PR is adding comments for each spec field for all resources to add descriptions in CRDs for the respective spec fields for which the comment is added. 

Also updating `MakeFile` to generate CRDs with `preserveUnknownField` set to `false` to enable the use of `oc explain` command.

### DIFF
--- a/config/crds/migration.openshift.io_directvolumemigrationprogresses.yaml
+++ b/config/crds/migration.openshift.io_directvolumemigrationprogresses.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.0
+    controller-gen.kubebuilder.io/version: v0.2.5
   creationTimestamp: null
   name: directvolumemigrationprogresses.migration.openshift.io
 spec:
@@ -35,6 +35,7 @@ spec:
     shortNames:
     - dvmp
     singular: directvolumemigrationprogress
+  preserveUnknownFields: false
   scope: Namespaced
   subresources: {}
   validation:
--- a/config/crds/migration.openshift.io_directvolumemigrationprogresses.yaml
+++ b/config/crds/migration.openshift.io_directvolumemigrationprogresses.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.5
+    controller-gen.kubebuilder.io/version: v0.3.0
   creationTimestamp: null
   name: directvolumemigrationprogresses.migration.openshift.io
 spec:
--- a/config/crds/migration.openshift.io_mighooks.yaml
+++ b/config/crds/migration.openshift.io_mighooks.yaml
@@ -56,8 +56,7 @@ spec:
               type: integer
             custom:
               description: Specifies whether the hook is a custom Ansible playbook
-                or a pre-built image. The value of this field is required. This is
-                a required field.
+                or a pre-built image. This is a required field.
               type: boolean
             image:
               description: Specifies the image of the hook to be executed. This is
--- a/config/crds/migration.openshift.io_mighooks.yaml
+++ b/config/crds/migration.openshift.io_mighooks.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.5
+    controller-gen.kubebuilder.io/version: v0.3.0
   creationTimestamp: null
   name: mighooks.migration.openshift.io
 spec:
@@ -55,9 +55,9 @@ spec:
               format: int64
               type: integer
             custom:
-              description: Implies whether the hook is a custom Ansible playbook or
-                a pre-built image. The value of this field is required. This is a
-                required field.
+              description: Specifies whether the hook is a custom Ansible playbook
+                or a pre-built image. The value of this field is required. This is
+                a required field.
               type: boolean
             image:
               description: Specifies the image of the hook to be executed. This is
--- a/config/crds/migration.openshift.io_mighooks.yaml
+++ b/config/crds/migration.openshift.io_mighooks.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.0
+    controller-gen.kubebuilder.io/version: v0.2.5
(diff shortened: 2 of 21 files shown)


alaypatel07 (commented):
@JaydipGabani thanks for looking into this. It is generally a good idea to have separate commits, i.e. one commit for the changes in types.go and the other for `make manifest` or similar automation command you run. The idea is you want to keep the manual changes separate from automation so that we know what manual changes produced the results with automation. Can sync up if you have more questions regarding this

JaydipGabani:
@alaypatel07 That makes more sense. WIll keep this in mind going ahead and making changes in this PR accordingly as well.

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -33,8 +33,14 @@ type PVCToMigrate struct {
 type DirectVolumeMigrationSpec struct {
 	SrcMigClusterRef            *kapi.ObjectReference `json:"srcMigClusterRef,omitempty"`
 	DestMigClusterRef           *kapi.ObjectReference `json:"destMigClusterRef,omitempty"`
+
+	//  A list of all the PVCs that are to be migrated with direct volume migration
```
shawn-hurley, line 37: The type is already displayed and re-stating it in the comment might not be necessary.

example:
```
   paused       <boolean>
   Indicates that the deployment is paused.
```

shawn-hurley (changes requested):
I think we probably don't need to state the type in the comment.

config/crds/migration.openshift.io_mighooks.yaml
```diff
@@ -49,15 +50,26 @@ spec:
           description: MigHookSpec defines the desired state of MigHook
           properties:
             activeDeadlineSeconds:
+              description: Specifies the highest amount of time for which the hook
+                will run.
               format: int64
               type: integer
             custom:
+              description: Implies whether the hook is a custom Ansible playbook or
```
djwhatle, line 58: I wouldn't say this `Implies` since it is not a suggestion, it's a setting.

djwhatle: `Specifies` would be a good substitute.

config/crds/migration.openshift.io_migmigrations.yaml
```diff
@@ -99,12 +104,20 @@ spec:
                   type: string
               type: object
             quiescePods:
+              description: Specifies whether to quiesce the application pods before
```
djwhatle, line 107: I think this should read

```
Specifies whether to quiesce the application Pods before migrating Persistent Volume data.
```

config/crds/migration.openshift.io_migmigrations.yaml
```diff
@@ -99,12 +104,20 @@ spec:
...
+                migration or not.
               type: boolean
             rollback:
+              description: Invokes the rollback migration operation, when set to true
```
djwhatle, line 111: This isn't quite accurate. While `spec.canceled` causes an on-demand switchover to the `Cancel` itinerary, `spec.rollback` is only meant to be set prior to creation of a MigMigration, since a rollback is a type of migration (stage, migrate, rollback are the 3 types of migrations that can be run)

config/crds/migration.openshift.io_migplans.yaml
```diff
@@ -91,13 +95,18 @@ spec:
...
               items:
                 description: MigPlanHook hold a reference to a MigHook along with
                   the desired phase to run it in
                 properties:
                   executionNamespace:
+                    description: Holds the name of the namespace where hooks should
+                      be implemented.
                     type: string
                   phase:
+                    description: Indicates the phase when the hooks will be executed.
```
djwhatle, line 109: It might be useful to supply the list of available phases here.

```
	PreBackupHookPhase   = "PreBackup"
	PostBackupHookPhase  = "PostBackup"
	PreRestoreHookPhase  = "PreRestore"
	PostRestoreHookPhase = "PostRestore"
```

config/crds/migration.openshift.io_migplans.yaml
```diff
@@ -263,6 +277,8 @@ spec:
                 type: object
               type: array
             refresh:
+              description: If set True, the controller is forces to check if the migplan
```
djwhatle, line 280: `is forces` -> `is forced`

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -62,14 +62,30 @@ const (
...
 type MigClusterSpec struct {
+	// Specifies if the cluster is host (where the controller is installed) or not. This is a required field.
 	IsHostCluster           bool                  `json:"isHostCluster"`
+
+	// Stores the url of the remote cluster. The field is only required for the source cluster object.
 	URL                     string                `json:"url,omitempty"`
+
 	ServiceAccountSecretRef *kapi.ObjectReference `json:"serviceAccountSecretRef,omitempty"`
+
+	// If the migcluster needs SSL verification for connections a user can supply a custom CA bundle.
```
djwhatle, line 73: We should specify that the only time this may be required is when `spec.Insecure = false`

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -62,14 +62,30 @@ const (
...
 	CABundle                []byte                `json:"caBundle,omitempty"`
+
+	// For azure clusters -- it's the resource group that in-cluster volumes use.
 	AzureResourceGroup      string                `json:"azureResourceGroup,omitempty"`
+
+	// If set false, user will need to provide CA bundle for TSL connection to the remote cluster.
```
djwhatle, line 79: TSL -> TLS

djwhatle (changes requested):
See comments

config/crds/migration.openshift.io_directvolumemigrationprogresses.yaml
```diff
@@ -35,6 +35,7 @@ spec:
     shortNames:
     - dvmp
     singular: directvolumemigrationprogress
+  preserveUnknownFields: false
```
shawn-hurley, line 38: Worried about this, if we have fields not defined here this will break things. Have we run a smoke test w/ these CRD changes?

JaydipGabani: @shawn-hurley I tested the migrations manually, was working fine and not breaking anything, I pinged forks around with the same concern but the opinion was that it shouldn't cause trouble.

shawn-hurley: +1

shawn-hurley (approved):

config/crds/migration.openshift.io_directvolumemigrationprogresses.yaml
djwhatle: Jaydip just mentioned to me that this is needed because the `description` field is apparently unknown, and this is how `oc explain` is able to work.

djwhatle (approved):

PR #940, opened 2021-02-01
Bug 1859998: Use and propagate direct_image_copy flag

seshapad:
This PR is part of a three-part fix. Other two PRs are listed below:
1. [velero-plugin fix](https://github.com/konveyor/openshift-velero-plugin/pull/69)
2. [mig-operator fix](https://github.com/konveyor/mig-operator/pull/554)

### Description of changes: 
**disable_image_copy** flag is read from the environment variable and used to:
1. Annotate the backup based on whether the image copy is required or not. [Indirect migration case]
2. Ensure mig-analytics does not include images in the count when `disable_image_copy=true`. 
3. Inhibit image copy [direct image migration case].

Since, `disable_image_copy` is related to `disable_image_migration`, following discussion with team, sanity test was performed for following tests:
image

As per the bug 1859998, the buggy output is:
```$ oc get builds
NAME        TYPE     FROM      STATUS                         STARTED   DURATION
jws-app-1   Source   Git@1.2   New (InvalidOutputReference)   
```
Above output is avoided by setting `disable_image_copy=true` and `disable_image_migration=false` and we now get the following expected output is obtained by avoiding only image copy, but migrating image-stream and image-stream-tags, 
```$ oc get builds
NAME        TYPE     FROM          STATUS    STARTED          DURATION
jws-app-1   Source   Git@caec202   Running   48 seconds ago   
```
Signed-off-by: Padmanabha Venkatagiri Seshadri <seshapad@in.ibm.com>

### DIFF
--- a/pkg/controller/migmigration/annotation.go
+++ b/pkg/controller/migmigration/annotation.go
@@ -3,6 +3,8 @@ package migmigration
 import (
 	"context"
 	"fmt"
+	"strings"
+
 	liberr "github.com/konveyor/controller/pkg/error"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	"github.com/konveyor/mig-controller/pkg/compat"
@@ -10,7 +12,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
-	"strings"
 )
 
 // Velero Plugin Annotations
@@ -34,6 +35,11 @@ const (
 const (
 	ResticPvBackupAnnotation = "backup.velero.io/backup-volumes" // comma-separated list of volume names
 	ResticPvVerifyAnnotation = "backup.velero.io/verify-volumes" // comma-separated list of volume names
+)
+
+// Disables the internal image copy
+const (
+	DisableImageCopy = "migration.openshift.io/disable-image-copy"
 )
 
 // Labels.
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strconv"
 	"time"
 
 	mapset "github.com/deckarep/golang-set"
@@ -18,6 +19,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+var Settings = &settings.Settings
 
 // Ensure the initial backup on the source cluster has been created
 // and has the proper settings.
@@ -38,6 +41,7 @@ func (t *Task) ensureInitialBackup() (*velero.Backup, error) {
 	if err != nil {
 		return nil, liberr.Wrap(err)
 	}
+
 	newBackup.Labels[InitialBackupLabel] = t.UID()
 	newBackup.Labels[MigMigrationDebugLabel] = t.Owner.Name
 	newBackup.Labels[MigPlanDebugLabel] = t.Owner.Spec.MigPlanRef.Name
@@ -46,6 +50,14 @@ func (t *Task) ensureInitialBackup() (*velero.Backup, error) {
 	newBackup.Spec.IncludedResources = toStringSlice(settings.IncludedInitialResources.Difference(toSet(t.PlanResources.MigPlan.Status.ExcludedResources)))
 	newBackup.Spec.ExcludedResources = toStringSlice(settings.ExcludedInitialResources.Union(toSet(t.PlanResources.MigPlan.Status.ExcludedResources)))
 	delete(newBackup.Annotations, QuiesceAnnotation)
+
+	if Settings.DisImgCopy {
+		if newBackup.Annotations == nil {
+			newBackup.Annotations = map[string]string{}
+		}
+		newBackup.Annotations[DisableImageCopy] = strconv.FormatBool(Settings.DisImgCopy)
+	}
+
 	err = client.Create(context.TODO(), newBackup)
 	if err != nil {
 		return nil, liberr.Wrap(err)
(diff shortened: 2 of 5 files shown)


jmontleon:
Operator PR is ready: https://github.com/konveyor/mig-operator/pull/554

pkg/controller/migmigration/backup.go
```diff
@@ -114,6 +137,14 @@ func (t *Task) ensureStageBackup() (*velero.Backup, error) {
 	newBackup.Spec.IncludedResources = toStringSlice(includedResources.Difference(toSet(t.PlanResources.MigPlan.Status.ExcludedResources)))
 	newBackup.Spec.ExcludedResources = toStringSlice(settings.ExcludedStageResources.Union(toSet(t.PlanResources.MigPlan.Status.ExcludedResources)))
 	newBackup.Spec.LabelSelector = &labelSelector
+
```
sseago, line 140: A few lines up, there is code which excludes ImageStreams from stage backup if we're doing direct image migration. We actually need to *include* these in the backup if DIM is configured *and* DISABLE_IMAGE_COPY is set (otherwise if the DIM flag is set on the MigPlan, we won't create the placeholder. Basically the below:
```
	if t.indirectImageMigration() {
		includedResources = settings.IncludedStageResources
	} else {
		includedResources = settings.IncludedStageResources.Difference(mapset.NewSetFromSlice([]interface{}{settings.ISResource}))
	}
￼```
Should probably be replaced with something like this:
```
	if t.indirectImageMigration() || imageCopyFlag {
		includedResources = settings.IncludedStageResources
	} else {
		includedResources = settings.IncludedStageResources.Difference(mapset.NewSetFromSlice([]interface{}{settings.ISResource}))
```

**Review by sseago — COMMENTED** [review]
One other point. See my comments on the openshift-velero-plugin PR, but we may need to add the skip-image-copy annotation to the restore as well as the backup, depending on whether we need to handle the non-MTC skip images use case differently from the MTC one.

File: pkg/apis/migration/v1alpha1/model.go [hunk 2]
```diff
@@ -23,6 +26,11 @@ const (
 	MigplanMigrationFailed  = "migplan.migration.openshift.io/failed"
 )
 
+// Environment variable for disable_image_copy flag
+const (
```
djwhatle, line 30: This should probably be loaded in through the `settings` pkg.

seshapad: Moved the function to settings pkg.

pkg/controller/migmigration/backup.go
seshapad: @sseago This change is done.

sseago (approved):
So I think this looks good overall, although I have one question for @djwhatle as to whether it's OK to query the migrationcontroller CR to determine whether image copy is disabled or whether we should be setting an envioronment variable and referencing that instead, so consider this a conditional ack awaiting feedback from Derek.

seshapad:
@sseago @djwhatle I did try the env variable approach. But the issue I faced is that the env appears to be loaded when controller pod starts. If the user changes the migcontroller cr, while the controller pod is active, the env variables don't get updated with the new values given to the flag. That's why I retained the existing approach.

sseago:
@seshapad Hmm. I would think that updating the migrationcontroller CR would cause the operator to deploy a new pod anyway, so the env var should still work. In any case, I think the env var approach is how we're handling this in most other cases of controller-wide global settings.

seshapad:
> @seshapad Hmm. I would think that updating the migrationcontroller CR would cause the operator to deploy a new pod anyway, so the env var should still work. In any case, I think the env var approach is how we're handling this in most other cases of controller-wide global settings.

Got it @sseago. @djwhatle could you please guide me where I could be faltering w.r.t the env variable approach I described above?

djwhatle:
@seshapad if you rebase this on master, you should see that a restart of the mig-controller pod whenever the configmap changes, meaning we can read from the env var in mig-controller PR.

seshapad:
> @seshapad if you rebase the operator PR on master, you should see that a restart of the mig-controller pod whenever the configmap changes, meaning we can read this settings from the env var in mig-controller PR.
> 
> In the future I'd still like to read all vars from kube api to further an equivalent local dev workflow, but not for this PR.

@djwhatle Sure. So, for this one, we will be going with reading the env in the settings, right?

djwhatle:
Yes 👍

seshapad:
> > @seshapad if you rebase the operator PR on master, you should see that a restart of the mig-controller pod whenever the configmap changes, meaning we can read this settings from the env var in mig-controller PR.
> > In the future I'd still like to read all vars from kube api to further an equivalent local dev workflow, but not for this PR.
> 
> @djwhatle Sure. So, for this PR, we will be going with reading the env in the settings, right?
@djwhatle @sseago 
The PR is rebased and updated to read the env from settings. I have provided the details of the pod restart issue in the slack thread.

seshapad:
> > > @seshapad if you rebase the operator PR on master, you should see that a restart of the mig-controller pod whenever the configmap changes, meaning we can read this settings from the env var in mig-controller PR.
> > > In the future I'd still like to read all vars from kube api to further an equivalent local dev workflow, but not for this PR.
> > 
> > 
> > @djwhatle Sure. So, for this PR, we will be going with reading the env in the settings, right?
> > @djwhatle @sseago
> > The PR is rebased and updated to read the env from settings. I have provided the details of the pod restart issue in the slack thread.

Restart issue was resolved by updating the operator image.

sseago (approved):

pkg/controller/migmigration/annotation.go
```diff
@@ -36,6 +37,11 @@ const (
 	ResticPvVerifyAnnotation = "backup.velero.io/verify-volumes" // comma-separated list of volume names
 )
 
+// Disables the internal image copy
+const (
+	DisableImageCopy = "openshift.io/disable-image-copy"
```
djwhatle, line 42: For future reference, we should be namespacing our annotations with `migration.openshift.io`

djwhatle (approved):
ACK, if you could switch the annotations in this PR + the plugin PR to be `migration.openshift.io/[...]` that would be appreciated, but other than that LGTM.

sseago:
@seshapad FYI, the mig-operator PR fix link is wrong in the description. It should be https://github.com/konveyor/mig-operator/pull/554

sseago:
OK I fixed the mig-operator link in the description.

PR #941, opened 2021-02-01
Add exit codes and runtime to DVMP CR

alaypatel07:
This is a part of a larger effort to detect a system-wide failure on the source side in establishing the network connection to the destination. 

Currently, a follow-up PR is anticipated that detects one bucket of egressnetwork policy set up on the source namespaces to be migrated.


config/crds/migration.openshift.io_directvolumemigrationprogresses.yaml
```diff
@@ -180,6 +183,8 @@ spec:
               description: PodPhase is a label for the condition of a pod at the current
                 time.
               type: string
+            runtime:
```
pranavgaikwad, line 186: `runtime` has a conflicting meaning in this context. I'd prefer `duration` or `elapsedTime` instead

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -267,22 +268,35 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
 		if transferRate != "" {
 			pvProgress.Status.LastObservedTransferRate = transferRate
 		}
+		pvProgress.Status.Runtime = nil
 	case !containerStatus.Ready && containerStatus.LastTerminationState.Terminated != nil && containerStatus.LastTerminationState.Terminated.ExitCode != 0:
 		// pod has a failure, report last failure reason
 		pvProgress.Status.PodPhase = kapi.PodFailed
 		pvProgress.Status.LogMessage = containerStatus.LastTerminationState.Terminated.Message
+		exitCode := containerStatus.LastTerminationState.Terminated.ExitCode
```
pranavgaikwad, line 276: Does container return the same return code as returned by the startup command?

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -267,22 +268,35 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
...
+		pvProgress.Status.ExitCode = &exitCode
+		pvProgress.Status.Runtime = &metav1.Duration{Duration: containerStatus.LastTerminationState.Terminated.FinishedAt.Sub(containerStatus.LastTerminationState.Terminated.StartedAt.Time)}
```
pranavgaikwad, line 278: Prefer using `.Round(time.Second)` to round off duration to nearest second value to avoid fractional duration

pranavgaikwad (changes requested):
Added some questions.

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
alaypatel07: yes

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
alaypatel07: @pranavgaikwad I think the `metav1.Duration` will do the pretty formatting for us

config/crds/migration.openshift.io_directvolumemigrationprogresses.yaml
alaypatel07: +1, `duration` can be an overloaded term as well, `elapsedTime`, +1

dymurray: +1 on `elapsedTime`

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
dymurray: +1 `metav1.Duration` should handle this

config/crds/migration.openshift.io_directvolumemigrationprogresses.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.0
```
alaypatel07, line 7: needs to be dropped

config/crds/migration.openshift.io_directvolumemigrationprogresses.yaml
alaypatel07: updated

pranavgaikwad (approved):
LGTM

dymurray (approved):

alaypatel07:
Corresponding mig-operator PR https://github.com/konveyor/mig-operator/pull/569

PR #942, opened 2021-02-02
Optimize flag evaluation for image streams

pranavgaikwad:
This PR optimizes call to `t.hasImageStreams()` function in flag evaluation for a phase. This call was being made every time a phase transitioned to next regardless of which phase it is. This PR fixes it by making it run only once for particular phase.


djwhatle (approved):
Change makes sense to me, +1 on catching this. Assuming you're able to confirm that the IS check is still working, merge away 👍

pranavgaikwad:
I have confirmed this change doesnt break image migrations.

PR #945, opened 2021-02-03
Add explicit error message on DirectVolumeMigration network misconfiguration

### LINKED ISSUE #912: [MIG-451] Add direct migration network failure warning
The key dependency of Direct Volume Migration (DVM) is that the destination cluster should have functional ingress controllers(routes) and the source namespaces should be able to access those routes in the destination namespace. I would like to propose we spend time thinking about how to accurately catch a misconfiguration and develop a more sophisticated message to the user than just the rsync pods displaying "connection reset by peer" 

To be brainstormed

-

**Additional context**
-

alaypatel07:
When running DVM with egressnetworkpolicy setup on the source side,
the client stunnel times out after 20 seconds, generating an error with
exit code 12 and connection reset by peer.

This PR checks if all the client pods have timed out in 20 seconds, and 
raises a network misconfiguraiton error condition on the dvm CR.

Closes #912

### DIFF
--- a/pkg/controller/directvolumemigration/task.go
+++ b/pkg/controller/directvolumemigration/task.go
@@ -436,11 +436,12 @@ func (t *Task) Run() error {
 					Reason:   SourceToDestinationNetworkError,
 					Category: migapi.Error,
 					Message: "All the rsync client pods on source are timing out at 20 seconds, " +
-						"please check your network configuration (like egressnetwork) that would block traffic from " +
+						"please check your network configuration (like egressnetworkpolicy) that would block traffic from " +
 						"source namespace to destination",
 					Durable: true,
 				})
 				t.fail(MigrationFailed, []string{"all the source rsync pods have timed out, look at error condition for more details"})
+				t.Requeue = NoReQ
 				return nil
 			}
 			if failed {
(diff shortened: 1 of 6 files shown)


dymurray (approved):

pkg/controller/directvolumemigration/task.go
```diff
@@ -425,6 +425,25 @@ func (t *Task) Run() error {
 			return liberr.Wrap(err)
 		}
 		if completed {
+			isEgressNetworkSetup, err := hasAllRsyncClientPodsTimedOut(t.getPVCNamespaceMap(), t.Client, t.Owner.Name)
```
pranavgaikwad, line 428: NIT: `isEgressNetworkSetup` is kind of confusing. Feels like its conveying something is "right" when its set to True.  while it's actually the opposite. I'd like `isNetworkBusted` or something similar :)

pranavgaikwad (approved):
small nit. But I understand the logic and seems reasonable to me.

pkg/controller/directvolumemigration/task.go
alaypatel07: @pranavgaikwad I took care of that in a follow-up PR, renamed this to `isStunnelTimeout` to more accurately reflect the cause of the error. I am merging this

PR #946, opened 2021-02-04
Bug 1807112: Check node-selectors in source pod list.

seshapad:
### Summary of changes:
 Check for any node-selectors in pod list from the source namespaces, and raise a warning if *any* such pod exists.

Signed-off-by: Padmanabha Venkatagiri Seshadri <seshapad@in.ibm.com>

### DIFF
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -43,6 +43,7 @@ const (
 	NsNotFoundOnDestinationCluster             = "NamespaceNotFoundOnDestinationCluster"
 	NsLimitExceeded                            = "NamespaceLimitExceeded"
 	NsLengthExceeded                           = "NamespaceLengthExceeded"
+	NsHaveNodeSelectors                        = "NamespacesHaveNodeSelectors"
 	PodLimitExceeded                           = "PodLimitExceeded"
 	SourceClusterProxySecretMisconfigured      = "SourceClusterProxySecretMisconfigured"
 	DestinationClusterProxySecretMisconfigured = "DestinationClusterProxySecretMisconfigured"
@@ -87,16 +88,17 @@ const (
 
 // Reasons
 const (
-	NotSet         = "NotSet"
-	NotFound       = "NotFound"
-	KeyNotFound    = "KeyNotFound"
-	NotDistinct    = "NotDistinct"
-	LimitExceeded  = "LimitExceeded"
-	LengthExceeded = "LengthExceeded"
-	NotDone        = "NotDone"
-	Done           = "Done"
-	Conflict       = "Conflict"
-	NotHealthy     = "NotHealthy"
+	NotSet                = "NotSet"
+	NotFound              = "NotFound"
+	KeyNotFound           = "KeyNotFound"
+	NotDistinct           = "NotDistinct"
+	LimitExceeded         = "LimitExceeded"
+	LengthExceeded        = "LengthExceeded"
+	NotDone               = "NotDone"
+	Done                  = "Done"
+	Conflict              = "Conflict"
+	NotHealthy            = "NotHealthy"
+	NodeSelectorsDetected = "NodeSelectorsDetected"
 )
 
 // Statuses
@@ -134,8 +136,9 @@ func (r ReconcileMigPlan) validate(plan *migapi.MigPlan) error {
 		return liberr.Wrap(err)
 	}
 
-	// Pod limit within each namespace.
-	err = r.validatePodLimit(plan)
+	// Validates pod properties (e.g. limit of number of active pods, presence of node-selectors)
+	// within each namespace.
+	err = r.validatePodProperties(plan)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
@@ -158,8 +161,8 @@ func (r ReconcileMigPlan) validate(plan *migapi.MigPlan) error {
 		return liberr.Wrap(err)
 	}
 
-	// Pods
-	err = r.validatePods(plan)
+	// Validate health of Pods
+	err = r.validatePodHealth(plan)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
@@ -291,8 +294,11 @@ func (r ReconcileMigPlan) validateNamespaceLengthForDVM(plan *migapi.MigPlan) []
 	return items
 }
 
-// Validate the total number of running pods (limit) across namespaces.
-func (r ReconcileMigPlan) validatePodLimit(plan *migapi.MigPlan) error {
+// Validates the following properties of pods across namespaces:
+// 1. Validate the total number of running pods (limit) across namespaces.
+// 2. Whether any pods have node-selectors or node names associated with it. If so, a warning is raised to indicate the
+// list of namespaces associated with that pod.
+func (r ReconcileMigPlan) validatePodProperties(plan *migapi.MigPlan) error {
 	if plan.Status.HasAnyCondition(Suspended, NsLimitExceeded) {
 		plan.Status.StageCondition(PodLimitExceeded)
 		return nil
@@ -310,6 +316,8 @@ func (r ReconcileMigPlan) validatePodLimit(plan *migapi.MigPlan) error {
 	}
 	count := 0
 	limit := Settings.Plan.PodLimit
+
+	nsWithNodeSelectors := make([]string, 0)
 	for _, name := range plan.GetSourceNamespaces() {
 		list := kapi.PodList{}
 		options := k8sclient.ListOptions{
@@ -324,6 +332,9 @@ func (r ReconcileMigPlan) validatePodLimit(plan *migapi.MigPlan) error {
 			return liberr.Wrap(err)
 		}
 		count += len(list.Items)
+		if r.hasNodeSelectors(list.Items) {
+			nsWithNodeSelectors = append(nsWithNodeSelectors, name)
+		}
 	}
 	if count > limit {
 		plan.Status.SetCondition(migapi.Condition{
@@ -335,7 +346,30 @@ func (r ReconcileMigPlan) validatePodLimit(plan *migapi.MigPlan) error {
 		})
 	}
 
+	if len(nsWithNodeSelectors) > 0 {
+		msgFormat := "Found Pods with `Spec.NodeSelector` or `Spec.NodeName` set in namespaces: [%s]. " +
+			"These fields will be cleared on Pods restored into the target cluster."
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     NsHaveNodeSelectors,
+			Status:   True,
+			Reason:   NodeSelectorsDetected,
+			Category: Warn,
+			Message: fmt.Sprintf(msgFormat,
+				strings.Join(nsWithNodeSelectors, ", ")),
+		})
+	}
+
 	return nil
+}
+
+func (r ReconcileMigPlan) hasNodeSelectors(podList []kapi.Pod) bool {
+	for _, pod := range podList {
+		if len(pod.Spec.NodeSelector) > 0 || len(pod.Spec.NodeName) > 0 {
+			return true
+		}
+	}
+
+	return false
 }
 
 // Validate the referenced source cluster.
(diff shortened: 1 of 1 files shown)


djwhatle:
Hi @seshapad, thanks for the pull request.

I want to double-check our approach here, because I'm concerned that raising a "StageNoOp" warning while the migration is running won't give the user enough advance notice to be able to adequately handle this problem, and their migration will probably fail.

@dymurray or @sseago do you have thoughts? Dylan, I saw you responded to the original bug.

seshapad:
@djwhatle shall I move the morning to init() method of the pipeline? that will mean raising the warning before the migration happens. cc: @dymurray @sseago

djwhatle:
@seshapad no, I think what I'm trying to get at is that this warning probably needs to surface to the user before they even click the migrate button (or create the migmigration resource) because once the migration is running, it's unlikely the user is going to be closely watching for this non critical condition and will be able to intervene in time to cancel the migration.

If there is information that a user needs to know that would affect their decision to run a migration or not, we typically provide that info on the migplan object where warnings that are raised surface in the final page of the plan creation wizard. 

I need to learn more about this node selector issue to know what's right. I will get in contact with Dylan and Scott to understand the issue better.

vconzola:
@djwhatle I don't understand this issue completely, but it seems this is another example of a place where we need a pre-migration checklist before starting a migration. In lieu of that, can we put the plan in a "Not ready" state if this condition exists? I'm not sure currently if "Not ready" blocks a migration or is just a warning, or which is appropriate in this case.

djwhatle:
> starting a migration. In lieu of that, can we put the plan in a "Not ready" state if this condition exists? I'm not sure currently if "Not ready" blocks a migration or is just a warning, or which is appropriate in this case.

@vconzola I think this is definitely a case where we want to warn rather than block, because it's going to be very common but not fatal to the migration for node selectors to exist on resources that we'll be migrating, since we clear the nodeselectors automatically during Velero restore. 

After talking it over with Scott and Dylan, I think our ideal solution would allow the user an opportunity to map src cluster nodeselectors onto target cluster nodeselectors, but that is not feasible to implement for 1.4.1. This would be possible for 1.5.0 if is determined to be a priority, would also need UI support.

@seshapad after talking it over with the team, the consensus is that this check should be part of the MigPlan validations, and if nodeselectors are found on Pods, we should raise a warning condition on the MigPlan so that the user sees the warning on the final page of MigPlan wizard. 

In the warning condition we set, I would mention the namespaces that contain pods with nodeselectors so that the user can go investigate.

vconzola:
@djwhatle The intent of the plan validations should be to surface issues that can be resolved by going back through the plan wizard and making changes (or at least acknowledging) things in the plan itself. Once the plan is saved these issues are lost. So if the user doesn't need to be aware of this warning after the plan is created that's fine. But if it's something the user may want to revisit after plan creation the warning(s) should be surfaced elsewhere.

sseago:
So the problem with putting it on migmigration validation is that this is something we can check before they even start a migration, and we'd probably want to warn the user then rather than later. There are 2 ways a user could deal with this. First, if they don't care about the issue (because stripping the node selectors won't break them), they can ignore it and migrate anyway. The other possibility is that they do care about this, in which case until we have the node selector mapping they may have to manually update their deployments post-migration and bounce the pods. Hopefully that's all that will be required, but it's hard to say in the abstract.

djwhatle:
> Once the plan is saved these issues are lost. So if the user doesn't need to be aware of this warning after the plan is created that's fine. But if it's something the user may want to revisit after plan creation the warning(s) should be surfaced elsewhere.

@vconzola From a UI perspective, I agree that the Warning Conditions aren't visible anymore after you close out the wizard, but just to be clear the Warning Conditions are never _cleared_ from the MigPlan resource, they remain until the issue is resolved. I assume they would all be visible again if you re-opened the wizard and made it to the final page.

The action a user could take to resolve this problem from the MigPlan wizard would be to reverse back to the namespace selection screen and remove namespaces containing nodeselectors if desired. This is similar to what the user is able to do if their namespaces contain incompatible GVKs.

Node selectors are a feature of OpenShift, and their presence will likely be common, so I think it will be important that the user is well-informed that we will strip nodeselectors if they proceed.

sseago:
@vconzola @djwhatle Another UX-related question. Once the plan is created, on the plans list (at /plans ), where the user would select migrate/stage/edit/etc., would it make sense to have some sort of indication if there are warnings on the plan (either in a new column, or as some sort of visual indicator in one of the existing columns, wherever it makes the most sense). That way if something were to happen post-plan-creation that triggers a warning condition (or if there's a leftover warning condition from before), there's a way to see it without going through the full set of wizard steps (which the user would have no reason to do at this point)?

djwhatle:
@sseago I think it would be useful to let the user know if a MigPlan has any warning conditions without digging into the wizard. Thoughts @vconzola ?

vconzola:
@djwhatle @sseago That's what i meant when I suggested putting a plan into a "not ready" state. That's probably not exactly what we want, but "not ready" is a way to alert the user that there's something to be aware of and possibly take action before a migration is started. I don't know how the ready/not ready status is determined for a plan but that's the place I'd like to see this information surfaced. Maybe we need another pre-migration plan state - "Ready with warnings"? 

Also, given what Derek said above it makes total sense to also include it in the plan validation warnings so user can go back into the wizard and remove namespaces containing nodeselectors if desired.

djwhatle:
> Maybe we need another pre-migration plan state - "Ready with warnings"?

@vconzola this is already how the back-end works, but maybe you're thinking we should have this type of language in the UI to match. 

Any warning condition will not block a MigPlan (or any other of our resources) from being ready. Only critical conditions will block readiness.

djwhatle:
@seshapad some information to help you go forward: I think we are in agreement now that this should surface as a warning condition on the migration plan. The place to implement the check should be in warning conditions are validated by functions called from https://github.com/konveyor/mig-controller/blob/master/pkg/controller/migplan/validation.go#L110

seshapad:
Thanks @djwhatle ... I will make appropriate changes.

vconzola:
@djwhatle Yes, I think we should have a "Ready with warnings" plan state in the UI. Is a GitHub issue the way to get this started? What other conditions in the backend should result in this plan state?

seshapad:
@djwhatle The warning is now raised during the plan phase as shown in the figure. 
image

djwhatle:
Good thinking to combine with an existing pull of pod data @seshapad. 

I think the warning message needs a little bit of work so that it's clear to the user what they can do to resolve the problem / what the effects of node selectors being present will be.

Here's the current warn message:
```
fmt.Sprintf("Node selector or name defined for namespaces: [%s].", strings.Join(nsWithNodeSelectors, ",")),
```

@sseago could you provide some links to where in the plugin we are clearing out nodeselector / nodename? I found [this](https://github.com/konveyor/openshift-velero-plugin/blob/cadc03a0c7874736b8725d0b5b819dc3b21a9199/velero-plugins/pod/restore.go#L40-L43), but I'm guessing that's not the extent of it.

seshapad:
> Good thinking to combine with an existing pull of pod data @seshapad. 
> 
> I think the warning message needs a little bit of work so that it's clear to the user what they can do to resolve the problem / what the effects of node selectors being cleared will be.
> 
> Here's the current warn message:
> ```
> fmt.Sprintf("Node selector or name defined for namespaces: [%s].", strings.Join(nsWithNodeSelectors, ",")),
> ```
> 
> First I'd suggest being a little more explicit by modifying as such:
> ```
> fmt.Sprintf("Node selector or node name defined for namespaces: [%s].", strings.Join(nsWithNodeSelectors, ",")),
> ```
> 
> @sseago could you provide some links to where in the plugin we are clearing out nodeselector / nodename? I found [this](https://github.com/konveyor/openshift-velero-plugin/blob/cadc03a0c7874736b8725d0b5b819dc3b21a9199/velero-plugins/pod/restore.go#L40-L43), but I'm guessing that's not the extent of it.
> 
> 
> ---
> 
> Also, now that the `validatePodLimit` validation is serving a dual purpose, I think we should change the function name to reflect  that @seshapad .
> 

@djwhatle to summarize:
1. Warning msg needs modification. Will do that. 
2. ValidatePodLimit function name needs to be changed. Shall I make it `ValidatePods`?

djwhatle:
> * Warning msg needs modification. Will do that.
Yes, and I think we need more info from @sseago on what should be included in the message. Upon closer inspection seems like we are just looking at `pod.Spec.NodeSelector` and `pod.Spec.NodeName` / nodename and not looking at namespaces themselves. In this case, I think the message might be modified to read like:

```
Found Pods with `Spec.NodeSelector` or `Spec.NodeName` set in namespaces: [%s]. 
These fields will be cleared on Pods restored into the target cluster.
```

@sseago thoughts on above?

> * ValidatePodLimit function name needs to be changed. Shall I make it `ValidatePods`?
Yeah that sounds like a good name. Would you update the comment above that function to specify both of the things that it validates?

sseago:
@djwhatle "@sseago could you provide some links to where in the plugin we are clearing out nodeselector / nodename? I found this, but I'm guessing that's not the extent of it."

That's exactly it. For pods we restore, we remove the NodeSelector. Actually, there's also a plugin that's part of velero which removes NodeName for pods on restore.

Note that when there are owner references, we don't restore the pod, so this is only really relevant for stage pods and for standalone pods.

sseago:
@djwhatle I'm wondering whether we should avoid warning for pods with owner references. I wonder, though, about what would happen if the Deployment's pod template had a node selector specified. I don't see any code which would strip that, so it's possible that this would still break restore if the node selector wasn't useful in the dest cluster.

djwhatle:
> That's exactly it. For pods we restore, we remove the NodeSelector. Actually, there's also a plugin that's part of velero which removes NodeName for pods on restore.

Ok, so the message is good as-is.


> Note that when there are owner references, we don't restore the pod, so this is only really relevant for stage pods and for standalone pods.

In that case , the node selector will have propagated down from the owning resource thought right? So it would still be valid to warn the user, or no?

I looked through the [velero plugins](https://github.com/konveyor/openshift-velero-plugin/tree/master/velero-plugins) and didn't see signs that we clear nodeselector / nodename on anything other  than Pod resources.

seshapad:
@djwhatle Following from our discussion, here is the modified warning msg:
image
Also, there was already a `validatePods()` function which was looking into pod health. So, I renamed `validatePodLimit()` to `validatePodProperties()` (this looks at pod limits and also node-selectors) and `validatePods()` to `validatePodHealth()` (this looks at pod health). Please let me know if you have any comments.

pkg/controller/migplan/validation.go
```diff
@@ -43,6 +43,7 @@ const (
 	NsNotFoundOnDestinationCluster             = "NamespaceNotFoundOnDestinationCluster"
 	NsLimitExceeded                            = "NamespaceLimitExceeded"
 	NsLengthExceeded                           = "NamespaceLengthExceeded"
+	NsHaveNodeSelectors                        = "NameSpacesHaveNodeSelectors"
```
djwhatle, line 46: `NameSpacesHaveNodeSelectors ` should be `NamespacesHaveNodeSelectors`

pkg/controller/migplan/validation.go
```diff
@@ -335,9 +346,32 @@ func (r ReconcileMigPlan) validatePodLimit(plan *migapi.MigPlan) error {
...
+	if len(nsWithNodeSelectors) > 0 {
+		msgFormat := "Found Pods with `Spec.NodeSelector` or `Spec.NodeName` set in namespaces: [%s]. " +
+			"These fields will be cleared on Pods restored into the target cluster."
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     NsHaveNodeSelectors,
+			Status:   True,
+			Reason:   NodeSelectorsDetected,
+			Category: Warn,
+			Message: fmt.Sprintf(msgFormat,
+				strings.Join(nsWithNodeSelectors, ",")),
```
djwhatle, line 358: We should separate the list of namespace with a comma and a space ", "

seshapad: Done!

pkg/controller/migplan/validation.go
seshapad: Done!

djwhatle (approved):

PR #950, opened 2021-02-09
MigAnalytic: Initial API changes for PV resizing

pranavgaikwad:

This PR:
- Introduces new fields in MigAnalytic API required for PV resizing feature
- Adds a new boolean `Refresh` on MigAnalytic which allows refreshing a MigAnalytic on demand
- Adds unit tests to test behavior of Refresh boolean

https://github.com/konveyor/mig-operator/pull/573


pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -174,6 +176,8 @@ func (r *ReconcileMigAnalytic) Reconcile(request reconcile.Request) (reconcile.R
 		!analytic.Status.HasBlockerCondition(),
 		"The analytic is ready.")
 
+	analytic.Spec.Refresh = false
```
alaypatel07, line 179: Maybe I have missed the conversation, but I am wondering why do we need this spec field? Is there a way we can refresh the mic-analytics CR at a particular interval even without the refresh?

pkg/controller/miganalytic/miganalytics_controller_test.go
```diff
@@ -78,3 +88,254 @@ func TestReconcile(t *testing.T) {
...
+			got, err := r.Reconcile(tt.args.request)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("ReconcileMigAnalytic.Reconcile() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("ReconcileMigAnalytic.Reconcile() = %v, want %v", got, tt.want)
+			}
+			obj := migapi.MigAnalytic{}
+			if err := r.Client.Get(context.TODO(), types.NamespacedName{
```
alaypatel07, line 326: nit: the obj being looked up can be part of the test's struct `tt.migAnalytics`. It keeps the input and output of the test in the same region and declaritive.

pkg/controller/miganalytic/miganalytics_controller.go
pranavgaikwad: @alaypatel07 there is no need to keep refreshing at particular interval. Before running a migration, we can never be sure whether the data currently shown in MigAnalytic is latest or not. We need a way to refresh MigAnalytic just-in-time.

alaypatel07: @pranavgaikwad if that is the case can we make the migmigration create the miganalytic CR? I do not like the  pattern of creating the `refresh` field unless absolutely needed on our CRs

pranavgaikwad: @alaypatel07 MigPlan will create MigAnalytic CR

djwhatle: We definitely need a way to trigger analytics gathering on demand and only when needed in the PV migration configuration process. I think this accomplishes that. 

This explicit refresh trigger seems to give most precise control over when the trigger is sent. I guess we could always do the UI process of delete/create a new CR if that seems preferable, but this seems like less work for the APIserver.

One concern that comes to mind: we need to watch out for problems where the general-purpose MigAnalytics created by the UI today blocks the MigAnalytics controller reconcile for a while, and our "mini" MigAnalytics for PV size analysis is stuck in the queue behind it. This could cause the MigPlan creation wizard to spin for a while.

pkg/controller/miganalytic/miganalytics_controller_test.go
```diff
@@ -78,3 +88,254 @@ func TestReconcile(t *testing.T) {
 	// g.Expect(c.Delete(context.TODO(), deploy)).To(gomega.Succeed())
 
 }
+
+func TestReconcileMigAnalytic_analyzeExtendedPVCapacity(t *testing.T) {
```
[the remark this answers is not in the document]
pranavgaikwad: @djwhatle this is an auto generated test for a placeholder function. I created the placeholder as a cue to other implementation @shubham-pampattiwar and @JaydipGabani are working on. When we actually write this function, we will populate the test as well. +1 for comment.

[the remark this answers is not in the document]
djwhatle: +1

djwhatle (approved):
LGTM

jmontleon (approved):

PR #951, opened 2021-02-10
Implementing migplan changes for intelligent pv resizing

JaydipGabani:
This PR checks the existence and state of `migAnalytics` of a `migplan`, and retrieves the analyzes volume data.

### DIFF
--- a/pkg/controller/miganalytic/validation.go
+++ b/pkg/controller/miganalytic/validation.go
@@ -30,6 +30,7 @@ const (
 	NotFound     = "NotFound"
 	KeyError     = "KeyError"
 	TestFailed   = "TestFailed"
+	NotReady     = "NotReady"
 )
 
 // Statuses
--- a/pkg/controller/miganalytic/validation.go
+++ b/pkg/controller/miganalytic/validation.go
@@ -30,7 +30,6 @@ const (
 	NotFound     = "NotFound"
 	KeyError     = "KeyError"
 	TestFailed   = "TestFailed"
-	NotReady     = "NotReady"
 )
 
 // Statuses
--- a/pkg/controller/migplan/migplan_controller.go
+++ b/pkg/controller/migplan/migplan_controller.go
@@ -18,12 +18,11 @@ package migplan
 
 import (
 	"context"
-	"fmt"
+	e "errors"
 	"github.com/konveyor/mig-controller/pkg/errorutil"
 	"sort"
 	"strconv"
 	"time"
-	e"errors"
 
 	liberr "github.com/konveyor/controller/pkg/error"
 	"github.com/konveyor/controller/pkg/logging"
@@ -46,8 +45,9 @@ import (
 )
 
 const (
-	MigPlan         = "migplan"
+	MigPlan = "migplan"
 )
+
 var log = logging.WithName("plan")
 
 // Application settings.
@@ -281,7 +281,7 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
 	// Loading PV statistics
 	pvcMap := make(map[string][]migapi.MigAnalyticPersistentVolumeClaim)
 	for _, mapvc := range migAnalytic.Status.Analytics.Namespaces {
-		pvcMap[mapvc.Namespace] =  mapvc.PersistentVolumes
+		pvcMap[mapvc.Namespace] = mapvc.PersistentVolumes
 	}
 
 	// Ready
@@ -475,16 +475,18 @@ func (r ReconcileMigPlan) ensureMigAnalytics(plan *migapi.MigPlan) error {
 	migAnalytics := &migapi.MigAnalyticList{}
 	err := r.List(context.TODO(), k8sclient.MatchingLabels(map[string]string{MigPlan: plan.Name}), migAnalytics)
 	if err != nil {
-		if !errors.IsNotFound(err){
+		if !errors.IsNotFound(err) {
 			return err
 		}
 	}
 	for _, migAnalytic := range migAnalytics.Items {
-		if migAnalytic.Spec.AnalyzeExntendedPVCapacity && plan.Spec.Refresh{
-			migAnalytic.Spec.Refresh = true
-			err := r.Update(context.TODO(), &migAnalytic)
-			if err != nil{
-				return err
+		if migAnalytic.Spec.AnalyzeExntendedPVCapacity {
+			if plan.Spec.Refresh {
+				migAnalytic.Spec.Refresh = true
+				err := r.Update(context.TODO(), &migAnalytic)
+				if err != nil {
+					return err
+				}
 			}
 			return nil
 		}
@@ -494,6 +496,8 @@ func (r ReconcileMigPlan) ensureMigAnalytics(plan *migapi.MigPlan) error {
 	pvMigAnalytics.GenerateName = plan.Name + "-"
 	pvMigAnalytics.Namespace = plan.Namespace
 	pvMigAnalytics.Spec.AnalyzeExntendedPVCapacity = true
+	pvMigAnalytics.Annotations = map[string]string{MigPlan: plan.Name}
+	pvMigAnalytics.Labels = map[string]string{MigPlan: plan.Name}
 	pvMigAnalytics.OwnerReferences = append(pvMigAnalytics.OwnerReferences, metav1.OwnerReference{
 		APIVersion: plan.APIVersion,
 		Kind:       plan.Kind,
@@ -502,7 +506,7 @@ func (r ReconcileMigPlan) ensureMigAnalytics(plan *migapi.MigPlan) error {
 	})
 	pvMigAnalytics.Spec.MigPlanRef = &kapi.ObjectReference{Namespace: plan.Namespace, Name: plan.Name}
 	err = r.Create(context.TODO(), pvMigAnalytics)
-	if err != nil{
+	if err != nil {
 		return err
 	}
 	return nil
--- a/pkg/controller/migplan/migplan_controller.go
+++ b/pkg/controller/migplan/migplan_controller.go
@@ -18,7 +18,6 @@ package migplan
 
 import (
 	"context"
-	"github.com/konveyor/mig-controller/pkg/controller/miganalytic"
 	"github.com/konveyor/mig-controller/pkg/errorutil"
 	"sort"
 	"strconv"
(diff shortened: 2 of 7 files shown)


pkg/controller/migplan/migplan_controller.go
```diff
@@ -216,6 +218,13 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	// Check if migAnalytics exists
+	err = r.validateMigAnalytics(plan)
```
pranavgaikwad, line 222: `ensureMigAnalytics` would be a better choice of word here

pkg/controller/migplan/migplan_controller.go
```diff
@@ -445,3 +467,62 @@ func (r ReconcileMigPlan) deleteImageRegistryDeploymentForClient(client k8sclien
...
+	migAnalytics := &migapi.MigAnalyticList{}
+	err := r.List(context.TODO(), k8sclient.MatchingLabels(map[string]string{"migplan": plan.Name}), migAnalytics)
+	if err != nil {
+		if !errors.IsNotFound(err){
+			return err
+		}
+	}
+	for _, migAnalytic := range migAnalytics.Items {
+		if migAnalytic.Spec.AnalyzeExntendedPVCapacity && plan.Spec.Refresh{
+			log.Info("refreshing miganalytics")
```
pranavgaikwad, line 482: please remove logs unless there is strong reason to keep them

pkg/controller/migplan/migplan_controller.go
```diff
@@ -445,3 +467,62 @@ func (r ReconcileMigPlan) deleteImageRegistryDeploymentForClient(client k8sclien
 	}
 	return nil
 }
+
+func (r ReconcileMigPlan) validateMigAnalytics(plan *migapi.MigPlan) error {
+	log.Info(fmt.Sprintf("Checking if migpanalytics exists or not"))
+	migAnalytics := &migapi.MigAnalyticList{}
+	err := r.List(context.TODO(), k8sclient.MatchingLabels(map[string]string{"migplan": plan.Name}), migAnalytics)
```
pranavgaikwad, line 474: can we use constants for labels here?

pkg/controller/migplan/migplan_controller.go
```diff
@@ -445,3 +467,62 @@ func (r ReconcileMigPlan) deleteImageRegistryDeploymentForClient(client k8sclien
...
+				return err
+			}
+			return nil
+		}
+	}
+
+	log.Info(fmt.Sprintf("Creating new analytics"))
+	pvMigAnalytics := &migapi.MigAnalytic{}
+	pvMigAnalytics.GenerateName = plan.Name + "-"
+	pvMigAnalytics.Namespace = plan.Namespace
```
pranavgaikwad, line 495: If we are listing MigAnalytics by labels, we need to make sure we add those labels on the new ones we create here

pranavgaikwad (changes requested):

pkg/controller/migplan/migplan_controller.go
JaydipGabani: The labels are added by the `MigAnalytic` controller I think. I ran this code and labels were present in the yaml of the resource.

pkg/controller/migplan/migplan_controller.go
JaydipGabani: +1, it was in my TODO

pkg/controller/migplan/migplan_controller.go
```diff
@@ -43,6 +45,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+const (
+	MigPlan         = "migplan"
```
pranavgaikwad, line 49:
```suggestion
	MigPlanLabel         = "migplan"
```

pkg/controller/migplan/migplan_controller.go
pranavgaikwad: @JaydipGabani best to make sure migplan adds them as there is no coupling between the two, duplicate labels get overwritten anyway

pranavgaikwad (changes requested):

pkg/controller/migplan/migplan_controller.go
JaydipGabani: Added label for migplan

pkg/controller/migplan/migplan_controller.go
```diff
@@ -445,3 +470,62 @@ func (r ReconcileMigPlan) deleteImageRegistryDeploymentForClient(client k8sclien
...
 	return nil
 }
+
+func (r ReconcileMigPlan) ensureMigAnalytics(plan *migapi.MigPlan) error {
+	migAnalytics := &migapi.MigAnalyticList{}
+	err := r.List(context.TODO(), k8sclient.MatchingLabels(map[string]string{MigPlan: plan.Name}), migAnalytics)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
```
shubham-pampattiwar, line 480: qq: Should we not return `err` over here, like if the error is other that the `not found` one ?

shubham-pampattiwar (commented):
Added one question @JaydipGabani

pkg/controller/migplan/migplan_controller.go
JaydipGabani: We are returning `err` in any other case except `notFound` here, and if the error is `Object not found` then we are creating that object

pkg/controller/migplan/migplan_controller.go
```diff
@@ -216,6 +221,15 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	// If interlligent pv resizing is enabled, Check if migAnalytics exists
```
pranavgaikwad, line 224: +1 on feature flag. good stuff!

pranavgaikwad (approved):
LGTM, let's wait for review from Derek/Alay/Scott since they are most familiar with MigPlan controller

pkg/controller/miganalytic/validation.go
```diff
@@ -30,6 +30,7 @@ const (
 	NotFound     = "NotFound"
 	KeyError     = "KeyError"
 	TestFailed   = "TestFailed"
+	NotReady     = "NotReady"
```
pranavgaikwad, line 33: Doesn't look like this is being consumed by MigAnalytic. The consumer is actually MigPlan. Therefore, this should be removed.

pkg/controller/migplan/migplan_controller.go
```diff
@@ -445,3 +481,69 @@ func (r ReconcileMigPlan) deleteImageRegistryDeploymentForClient(client k8sclien
...
+		return nil, err
+	}
+	for _, migAnalytic := range migAnalytics.Items {
+		if migAnalytic.Spec.AnalyzeExtendedPVCapacity == true {
+			if migAnalytic.Status.IsReady() {
+				return &migAnalytic, nil
+			}
+			if time.Now().Sub(migAnalytic.CreationTimestamp.Time) > 3 * time.Minute {
+				plan.Status.SetCondition(migapi.Condition{
+					Type:               miganalytic.NotReady,
```
pranavgaikwad, line 539: I would prefer the condition to be defined in MigPlan instead of MigAnalytic keeping concerns separate. This condition should be called something like `IntelligentPVResizingDisabled` or something similar that conveys the actual user impact.

pkg/controller/migplan/migplan_controller.go
```diff
@@ -445,3 +481,69 @@ func (r ReconcileMigPlan) deleteImageRegistryDeploymentForClient(client k8sclien
...
+					Status:             True,
+					Category:           Warn,
+					Message:            "MigAnalytics is still not ready after 3 minutes of its creation",
```
pranavgaikwad, line 542: I'd prefer a Message that conveys what it means for the user instead of telling the cause.

pranavgaikwad (changes requested):
Requesting very minor changes. Feature flag, safeguard of 3 mins, everything is very helpful for this feature. Good stuff!

pkg/controller/migplan/migplan_controller.go
```diff
@@ -445,3 +481,69 @@ func (r ReconcileMigPlan) deleteImageRegistryDeploymentForClient(client k8sclien
...
+	err := r.List(context.TODO(), k8sclient.MatchingLabels(map[string]string{MigPlan: plan.Name}), migAnalytics)
+	if err != nil {
+		return nil, err
+	}
+	for _, migAnalytic := range migAnalytics.Items {
+		if migAnalytic.Spec.AnalyzeExtendedPVCapacity == true {
+			if migAnalytic.Status.IsReady() {
+				return &migAnalytic, nil
+			}
+			if time.Now().Sub(migAnalytic.CreationTimestamp.Time) > 3 * time.Minute {
```
pranavgaikwad, line 537: Lets also use a constant/variable to allow setting a custom timeout value if 3 mins is deemed low.

JaydipGabani: +1

pkg/controller/migplan/migplan_controller.go
JaydipGabani: +1

pkg/controller/migplan/migplan_controller.go
JaydipGabani: +1, Good catch!

pkg/controller/migplan/migplan_controller.go
```diff
@@ -43,6 +43,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+const (
+	MigPlan   = "migplan"
+	CreatedBy = "CreatedBy"
+)
+
+// define maximum waiting time for mig analytic to be ready
+var timeLimit = 3 * time.Minute
```
pranavgaikwad, line 52: please make this variable more specific: `migAnalyticsTimeout`

JaydipGabani: Done

pranavgaikwad (approved):
ACK

shubham-pampattiwar (approved):
ACK

PR #952, opened 2021-02-10
Discovery and Grouping of PVs for Intelligent PV resizing

shubham-pampattiwar:
This PR does the following:
- Introduces a new struct `MigAnalyticPersistentVolumeDetails` to hold the PV details needed for resizing 
- Adds a new function `getNodeToPVMapForNS` which builds a map of `NodeName : []MigAnalyticPersistentVolumeDetails`which will be later consumed by `analyzeExtendedPVCapacity`  for PV resizing

### DIFF
--- a/pkg/controller/miganalytic/miganalytics_controller.go
+++ b/pkg/controller/miganalytic/miganalytics_controller.go
@@ -50,7 +50,6 @@ import (
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	imagev1 "github.com/openshift/api/image/v1"
 	corev1 "k8s.io/api/core/v1"
-	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -320,7 +319,7 @@ func (r *ReconcileMigAnalytic) getNodeToPVMapForNS(ns *migapi.MigAnalyticNamespa
 		if pod.Status.Phase == migapi.Running {
 			for _, vol := range pod.Spec.Volumes {
 				if vol.PersistentVolumeClaim != nil {
-					pvcObject := kapi.PersistentVolumeClaim{}
+					pvcObject := corev1.PersistentVolumeClaim{}
 					err := client.Get(
 						context.TODO(),
 						types.NamespacedName{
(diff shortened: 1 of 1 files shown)


pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -275,6 +287,75 @@ func (r *ReconcileMigAnalytic) analyze(analytic *migapi.MigAnalytic) error {
...
+	err := r.Client.List(
+		context.TODO(),
+		k8sclient.InNamespace(ns.Namespace),
+		&podList)
+
+	if err != nil {
+		return nil, liberr.Wrap(err)
+	}
+
+	if len(podList.Items) > 0 {
```
pranavgaikwad, line 308: unneeded...`range` takes care of this

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -275,6 +287,75 @@ func (r *ReconcileMigAnalytic) analyze(analytic *migapi.MigAnalytic) error {
...
+	podList := corev1.PodList{}
+	nodeToPVDetails := map[string][]persistentVolumeDetails{}
+
+	err := r.Client.List(
+		context.TODO(),
+		k8sclient.InNamespace(ns.Namespace),
+		&podList)
+
+	if err != nil {
+		return nil, liberr.Wrap(err)
```
pranavgaikwad, line 305:
```suggestion
		return nodeToPVDetails, liberr.Wrap(err)
```
Better to return empty map since this is not a reference

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -275,6 +287,75 @@ func (r *ReconcileMigAnalytic) analyze(analytic *migapi.MigAnalytic) error {
...
+						}
+
+					}
+				}
+			}
+		}
+		return nodeToPVDetails, nil
+	}
+
+	return nil, nil
```
pranavgaikwad, line 356: Same as above, better to return a non-nil empty map

pranavgaikwad (changes requested):
requesting small changes, rest all looks good

pkg/controller/miganalytic/miganalytics_controller.go
shubham-pampattiwar: Done, updated the PR

pkg/controller/miganalytic/miganalytics_controller.go
shubham-pampattiwar: Done

pkg/controller/miganalytic/miganalytics_controller.go
shubham-pampattiwar: Done

pranavgaikwad (approved):
Thanks for addressing the feedback! LGTM

Let's wait for the API PR to merge, would need a rebase after that

PR #953, opened 2021-02-11
Catch no route to host rsync error in dvm controller

alaypatel07:
Depends on #945 to merge

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -1237,6 +1237,28 @@ func hasAllRsyncClientPodsTimedOut(pvcMap map[string][]pvcMapElement, client k8s
 	return true, nil
 }
 
+func isAllRsyncClientPodsNoRouteToHost(pvcMap map[string][]pvcMapElement, client k8sclient.Client, dvmName string) (bool, error) {
+	for ns, vols := range pvcMap {
+		for _, vol := range vols {
+			dvmp := migapi.DirectVolumeMigrationProgress{}
+			err := client.Get(context.TODO(), types.NamespacedName{
+				Name:      getMD5Hash(dvmName + vol.Name + ns),
+				Namespace: migapi.OpenshiftMigrationNamespace,
+			}, &dvmp)
+			if err != nil {
+				return false, err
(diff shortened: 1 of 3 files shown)


pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1237,6 +1237,28 @@ func hasAllRsyncClientPodsTimedOut(pvcMap map[string][]pvcMapElement, client k8s
 	return true, nil
 }
 
+func isAllRsyncClientPodsNoRouteToHost(pvcMap map[string][]pvcMapElement, client k8sclient.Client, dvmName string) (bool, error) {
+	for ns, vols := range pvcMap {
+		for _, vol := range vols {
+			dvmp := migapi.DirectVolumeMigrationProgress{}
+			err := client.Get(context.TODO(), types.NamespacedName{
```
pranavgaikwad, line 1244: I'd like to replace this with a `List()` call instead and then looking up the DVMP you want from the list to reduce total number of n/w calls.

pranavgaikwad (approved):
I have been using this strategy of using List() instead of Get() to reduce n/w calls. Totally up to you whether though.

pkg/controller/directvolumemigration/rsync.go
alaypatel07: There are two points to be considered here:

1. the get calls are served by local cache, so it is not a network call to apiserver like the remote clusters
2. the dvmp CRs are not deleted out of the box in MTC. So if we list all dvmp, chances are we will get more CRs than we need for all the migrations. This could lead to a situation where more time is spent in encoding/decoding  the CRs in list call for ones we dont need anyway.

I do not think that "List" is going to be a blanket win in all the cases. Maybe I am missing something, open for thoughts.

pranavgaikwad: I see. I was unaware of the differences in both. Ignore

dymurray: Yeah if get call is served by local cache I don't see a huge benefit of the list call here

dymurray (approved):

PR #954, opened 2021-02-12
Add operator version check to cluster 

### LINKED ISSUE #633: Check for matching MTC version at beginning of MigCluster reconcile
- Help user avoid inadvertently running a migration with mismatched versions of MTC on source/target cluster
- Help user realize they need to upgrade MTC on both src/target cluster, not just on target

If user runs with mismatched MTC versions, they could experience undefined behavior. We should raise a condition on the MigCluster if this is detected.

Suggested by @sseago 

Related:  https://issues.redhat.com/browse/MIG-309

ibolton336:
Closes #633 

Matching operator PR: https://github.com/konveyor/mig-operator/pull/598

### DIFF
--- a/pkg/apis/migration/v1alpha1/directimagemigration_types.go
+++ b/pkg/apis/migration/v1alpha1/directimagemigration_types.go
@@ -33,7 +33,7 @@ type DirectImageMigrationSpec struct {
 	DestMigClusterRef *kapi.ObjectReference `json:"destMigClusterRef,omitempty"`
 
 	// Holds names of all namespaces to run DIM to get all the imagestreams in these namespaces.
-	Namespaces        []string              `json:"namespaces,omitempty"`
+	Namespaces []string `json:"namespaces,omitempty"`
 }
 
 // DirectImageMigrationStatus defines the observed state of DirectImageMigration
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -675,7 +675,7 @@ func (m *MigCluster) SetOperatorVersion(c k8sclient.Client) error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	newOperatorVersion, _, _, err := m.GetOperatorVersion(clusterClient)
+	newOperatorVersion, err := m.GetOperatorVersion(clusterClient)
 	if err != nil {
 		return err
 	}
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -220,19 +220,15 @@ func (m *MigCluster) GetClusterSubdomain(c k8sclient.Client) (string, error) {
 	return clusterSubdomain, nil
 }
 
-// GetRegistryImage gets a MigCluster specific registry image from ConfigMap
+// GetOperatorVersion retrieves the operator version from the respective controllers ConfigMap
 func (m *MigCluster) GetOperatorVersion(c k8sclient.Client) (string, error) {
-	client, err := m.GetClient(c)
-	if err != nil {
-		return "", err
-	}
-	clusterConfig, err := m.GetClusterConfigMap(client)
+	clusterConfig, err := m.GetClusterConfigMap(c)
 	if err != nil {
 		return "", liberr.Wrap(err)
 	}
 	operatorVersion, ok := clusterConfig.Data[OperatorVersionKey]
 	if !ok {
-		return "", liberr.Wrap(errors.Errorf("configmap key not found: %v", OperatorVersionKey))
+		return "", liberr.Wrap(err)
 	}
 	return operatorVersion, nil
 }
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -684,22 +684,22 @@ func (m *MigCluster) GetRegistryPath(c k8sclient.Client) (string, error) {
 	return m.GetInternalRegistryPath(c)
 }
 
+// Pulls the operatorVersion from the migration-cluster-config configmap and
+// loads it into MigCluster.Status.OperatorVersion
 func (m *MigCluster) SetOperatorVersion(c k8sclient.Client) error {
 	oldOperatorVersion := m.Status.OperatorVersion
 	clusterClient, err := m.GetClient(c)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	newOperatorVersion, _ := m.GetOperatorVersion(clusterClient)
 	// Ignore error here. Missing configmap/key is already raised to user in validation,
 	// we don't want reconcile to exit w/ error on MTC < 1.4.2. GetOperatorVersion will
 	// return "" on error which is usable below.
-	if err != nil {
-		return nil
-	}
+	newOperatorVersion, _ := m.GetOperatorVersion(clusterClient)
+
 	// When operator version changes, all other MigClusters will be updated
-	// NOTE: cleaner way to do this (that would support concurrent reconciles)
-	//       would be to write a watch with a predicate that would enqueue
+	// NOTE: In future if we need to support concurrent reconciles for MigCluster,
+	//       we should remove this and write a watch with a predicate that would enqueue
 	//       reconciles when other MigClusters change versions. This is simpler
 	//       for now, but will break if MaxConcurrentReconciles is turned up,
 	//       since other clusters would start reconciling before our change
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -229,7 +229,8 @@ func (m *MigCluster) GetOperatorVersion(c k8sclient.Client) (string, error) {
 
 	operatorVersion, ok := clusterConfig.Data[OperatorVersionKey]
 	if !ok {
-		return "", liberr.Wrap(err)
+		return "", liberr.Wrap(errors.Errorf("configmap key %v not found in configmap %v/%v for migcluster %v/%v",
+			OperatorVersionKey, clusterConfig.Namespace, clusterConfig.Name, m.Namespace, m.Name))
 	}
 
 	return operatorVersion, nil
@@ -690,9 +691,6 @@ func (m *MigCluster) SetOperatorVersion(c k8sclient.Client) error {
 		return liberr.Wrap(err)
 	}
 	newOperatorVersion, err := m.GetOperatorVersion(clusterClient)
-	if err != nil {
-		return liberr.Wrap(err)
-	}
 
 	// When operator version changes, all other MigClusters will be updated
 	// NOTE: cleaner way to do this (that would support concurrent reconciles)
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -691,7 +691,13 @@ func (m *MigCluster) SetOperatorVersion(c k8sclient.Client) error {
 		return liberr.Wrap(err)
 	}
 	newOperatorVersion, err := m.GetOperatorVersion(clusterClient)
-
+	// Ignore error here. Missing configmap/key is already raised to user in validation,
+	// we don't want reconcile to exit w/ error on MTC < 1.4.2. We also don't want to
+	// accidentally trigger the expensive propagation below if there's an error
+	// getting operator version.
+	if err != nil {
+		return nil
+	}
 	// When operator version changes, all other MigClusters will be updated
 	// NOTE: cleaner way to do this (that would support concurrent reconciles)
 	//       would be to write a watch with a predicate that would enqueue
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -282,6 +282,19 @@ func (m *MigCluster) BuildRestConfig(c k8sclient.Client) (*rest.Config, error) {
 	return restConfig, nil
 }
 
+// Test whether OPERATOR_VERSION in configmap on MigCluster matches status.OperatorVersion
+func (m *MigCluster) OperatorVersionMatchesConfigmap(c k8sclient.Client) (bool, error) {
+	clusterClient, err := m.GetClient(c)
+	if err != nil {
+		return false, liberr.Wrap(err)
+	}
+	operatorVersion, err := m.GetOperatorVersion(clusterClient)
+	if operatorVersion == m.Status.OperatorVersion {
+		return true, nil
+	}
+	return false, nil
+}
+
 // Delete resources on the cluster by label.
 func (m *MigCluster) DeleteResources(client k8sclient.Client, labels map[string]string) error {
 	client, err := m.GetClient(client)
@@ -671,13 +684,41 @@ func (m *MigCluster) GetRegistryPath(c k8sclient.Client) (string, error) {
 }
 
 func (m *MigCluster) SetOperatorVersion(c k8sclient.Client) error {
+	oldOperatorVersion := m.Status.OperatorVersion
 	clusterClient, err := m.GetClient(c)
 	if err != nil {
 		return liberr.Wrap(err)
 	}
 	newOperatorVersion, err := m.GetOperatorVersion(clusterClient)
 	if err != nil {
-		return err
+		return liberr.Wrap(err)
+	}
+
+	// When operator version changes, all other MigClusters will be updated
+	// NOTE: cleaner way to do this (that would support concurrent reconciles)
+	//       would be to write a watch with a predicate that would enqueue
+	//       reconciles when other MigClusters change versions. This is simpler
+	//       for now, but will break if MaxConcurrentReconciles is turned up,
+	//       since other clusters would start reconciling before our change
+	//       is written.
+	if oldOperatorVersion != newOperatorVersion {
+		clusterList, err := ListClusters(c)
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		// Update all other MigClusters to propagate version checks
+		for _, cluster := range clusterList {
+			if cluster.UID == m.UID {
+				continue
+			}
+			if !cluster.Spec.Refresh {
+				cluster.Spec.Refresh = true
+				err := c.Update(context.Background(), &cluster)
+				if err != nil {
+					return liberr.Wrap(err)
+				}
+			}
+		}
 	}
 
 	if newOperatorVersion != "" {
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -690,11 +690,10 @@ func (m *MigCluster) SetOperatorVersion(c k8sclient.Client) error {
 	if err != nil {
 		return liberr.Wrap(err)
 	}
-	newOperatorVersion, err := m.GetOperatorVersion(clusterClient)
+	newOperatorVersion, _ := m.GetOperatorVersion(clusterClient)
 	// Ignore error here. Missing configmap/key is already raised to user in validation,
-	// we don't want reconcile to exit w/ error on MTC < 1.4.2. We also don't want to
-	// accidentally trigger the expensive propagation below if there's an error
-	// getting operator version.
+	// we don't want reconcile to exit w/ error on MTC < 1.4.2. GetOperatorVersion will
+	// return "" on error which is usable below.
 	if err != nil {
 		return nil
 	}
@@ -725,9 +724,7 @@ func (m *MigCluster) SetOperatorVersion(c k8sclient.Client) error {
 		}
 	}
 
-	if newOperatorVersion != "" {
-		m.Status.OperatorVersion = newOperatorVersion
-	}
+	m.Status.OperatorVersion = newOperatorVersion
 
 	return nil
 }
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -222,7 +222,11 @@ func (m *MigCluster) GetClusterSubdomain(c k8sclient.Client) (string, error) {
 
 // GetRegistryImage gets a MigCluster specific registry image from ConfigMap
 func (m *MigCluster) GetOperatorVersion(c k8sclient.Client) (string, error) {
-	clusterConfig, err := m.GetClusterConfigMap(c)
+	client, err := m.GetClient(c)
+	if err != nil {
+		return "", err
+	}
+	clusterConfig, err := m.GetClusterConfigMap(client)
 	if err != nil {
 		return "", liberr.Wrap(err)
 	}
@@ -669,6 +673,9 @@ func (m *MigCluster) GetRegistryPath(c k8sclient.Client) (string, error) {
 }
 
 func (m *MigCluster) SetOperatorVersion(c k8sclient.Client) error {
+	fmt.Printf("m is: %+v\n", m)
+	fmt.Printf("c is: %+v\n", c)
+
 	newOperatorVersion, err := m.GetOperatorVersion(c)
 	if err != nil {
 		return err
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -64,37 +64,37 @@ const (
 // MigClusterSpec defines the desired state of MigCluster
 type MigClusterSpec struct {
 	// Specifies if the cluster is host (where the controller is installed) or not. This is a required field.
-	IsHostCluster           bool                  `json:"isHostCluster"`
+	IsHostCluster bool `json:"isHostCluster"`
 
 	// Stores the url of the remote cluster. The field is only required for the source cluster object.
-	URL                     string                `json:"url,omitempty"`
+	URL string `json:"url,omitempty"`
 
 	ServiceAccountSecretRef *kapi.ObjectReference `json:"serviceAccountSecretRef,omitempty"`
 
 	// If the migcluster needs SSL verification for connections a user can supply a custom CA bundle. This field is required only when spec.Insecure is set false
-	CABundle                []byte                `json:"caBundle,omitempty"`
+	CABundle []byte `json:"caBundle,omitempty"`
 
 	// For azure clusters -- it's the resource group that in-cluster volumes use.
-	AzureResourceGroup      string                `json:"azureResourceGroup,omitempty"`
+	AzureResourceGroup string `json:"azureResourceGroup,omitempty"`
 
 	// If set false, user will need to provide CA bundle for TLS connection to the remote cluster.
-	Insecure                bool                  `json:"insecure,omitempty"`
+	Insecure bool `json:"insecure,omitempty"`
 
 	// An override setting to tell the controller that the source cluster restic needs to be restarted after stage pod creation.
-	RestartRestic           *bool                 `json:"restartRestic,omitempty"`
+	RestartRestic *bool `json:"restartRestic,omitempty"`
 
 	// If set True, forces the controller to run a full suite of validations on migcluster.
-	Refresh                 bool                  `json:"refresh,omitempty"`
+	Refresh bool `json:"refresh,omitempty"`
 
 	// Stores the path of registry route when using direct migration.
-	ExposedRegistryPath     string                `json:"exposedRegistryPath,omitempty"`
+	ExposedRegistryPath string `json:"exposedRegistryPath,omitempty"`
 }
 
 // MigClusterStatus defines the observed state of MigCluster
 type MigClusterStatus struct {
-	Conditions     `json:","`
-	ObservedDigest string `json:"observedDigest,omitempty"`
-	RegistryPath   string `json:"registryPath,omitempty"`
+	Conditions      `json:","`
+	ObservedDigest  string `json:"observedDigest,omitempty"`
+	RegistryPath    string `json:"registryPath,omitempty"`
 	OperatorVersion string `json:"operatorVersion,omitempty"`
 }
 
--- a/pkg/apis/migration/v1alpha1/migcluster_types.go
+++ b/pkg/apis/migration/v1alpha1/migcluster_types.go
@@ -221,16 +221,20 @@ func (m *MigCluster) GetClusterSubdomain(c k8sclient.Client) (string, error) {
 }
 
 // GetOperatorVersion retrieves the operator version from the respective controllers ConfigMap
-func (m *MigCluster) GetOperatorVersion(c k8sclient.Client) (string, error) {
+func (m *MigCluster) GetOperatorVersion(c k8sclient.Client) (string, string, string, error) {
 	clusterConfig, err := m.GetClusterConfigMap(c)
 	if err != nil {
-		return "", liberr.Wrap(err)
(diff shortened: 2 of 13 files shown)


jwmatthews:
@ibolton336 awesome to see you getting your hands into some of the controller side!

pkg/apis/migration/v1alpha1/directimagemigration_types.go
```diff
@@ -33,7 +33,7 @@ type DirectImageMigrationSpec struct {
 	DestMigClusterRef *kapi.ObjectReference `json:"destMigClusterRef,omitempty"`
 
 	// Holds names of all namespaces to run DIM to get all the imagestreams in these namespaces.
-	Namespaces        []string              `json:"namespaces,omitempty"`
```
djwhatle, line 36: @ibolton336 seems like there may be an editor config issue here. We have a mix of vs code, goland, and probably vim with controller contributors. 

I can tell you that with VS code if you just installed go language support and use default settings it shouldn't be deleting all of this whitespace. I think this needs to be fixed otherwise we'll have thrashing in whitespace settings on every PR.

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -218,6 +220,23 @@ func (m *MigCluster) GetClusterSubdomain(c k8sclient.Client) (string, error) {
 	return clusterSubdomain, nil
 }
 
+// GetRegistryImage gets a MigCluster specific registry image from ConfigMap
+func (m *MigCluster) GetOperatorVersion(c k8sclient.Client) (string, error) {
+	client, err := m.GetClient(c)
+	if err != nil {
+		return "", err
+	}
+	clusterConfig, err := m.GetClusterConfigMap(client)
```
djwhatle, line 229: In this situation, you need to grab the client for this MigCluster first, since the `ClusterConfigMap` is located in etcd on the MigCluster rather than on the controller cluster (these are equivalent when MigCluster == host).

See `MigCluster.GetClient()`

```
# Pass the reconciler client to MigCluster.GetClient
clusterClient, err := m.GetClient(c)
# Use the clusterClient to get the configmap from the (possibly remote) MigCluster
clusterConfig, err := m.GetClusterConfigMap(clusterClient)
```

pkg/apis/migration/v1alpha1/directimagemigration_types.go
ibolton336: @djwhatle This was actually happening when running 'make run' on my machine. I will take another look but I set up my vscode settings to turn all formatting off unless a config file was present for prettier.

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -218,6 +220,23 @@ func (m *MigCluster) GetClusterSubdomain(c k8sclient.Client) (string, error) {
 	return clusterSubdomain, nil
 }
 
+// GetRegistryImage gets a MigCluster specific registry image from ConfigMap
+func (m *MigCluster) GetOperatorVersion(c k8sclient.Client) (string, error) {
+	client, err := m.GetClient(c)
+	if err != nil {
+		return "", err
```
djwhatle, line 227: Replace err returns with 

```
return "", liberr.Wrap(err)
```

This provides stack trace from full call stack when error returned down the chain is finally displayed

pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -152,6 +152,13 @@ func (r *ReconcileMigCluster) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	// Set Status.OperatorVersion
+	err = cluster.SetOperatorVersion(r)
```
djwhatle, line 156: +1

pkg/controller/migcluster/validation.go
```diff
@@ -15,32 +15,36 @@ import (
 	liberr "github.com/konveyor/controller/pkg/error"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	migref "github.com/konveyor/mig-controller/pkg/reference"
+	"github.com/pkg/errors"
```
djwhatle, line 18: Looking at other places in the codebase, I think we just want the pkg "errors", not github.com/pkg/errors.

pkg/controller/migcluster/validation.go
```diff
@@ -335,3 +345,44 @@ func (r *ReconcileMigCluster) validateSaTokenPrivileges(cluster *migapi.MigClust
...
+// validate operator version.
+func (r ReconcileMigCluster) validateOperatorVersionMatchesHost(cluster *migapi.MigCluster) error {
+	if cluster.Spec.IsHostCluster {
+		return nil
+	}
+	if cluster.Status.HasCriticalCondition() {
+		return nil
+	}
+
+	client, err := cluster.GetClient(r)
```
djwhatle, line 358: I think naming this `clusterClient` may be helpful here to differentiate the reconciler client (r) from a MigCluster client.

pkg/controller/migcluster/validation.go
```diff
@@ -335,3 +345,44 @@ func (r *ReconcileMigCluster) validateSaTokenPrivileges(cluster *migapi.MigClust
...
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	clusterConfig, err := cluster.GetClusterConfigMap(client)
```
djwhatle, line 363: I think you could just use your own func `GetOperatorVersion()` here, maybe that's the plan.

pkg/controller/migcluster/validation.go
```diff
@@ -335,3 +345,44 @@ func (r *ReconcileMigCluster) validateSaTokenPrivileges(cluster *migapi.MigClust
...
+
+	operatorVersionMatchesHost := operatorVersion == hostOperatorVersion
+
+	if !operatorVersionMatchesHost {
+		cluster.Status.SetCondition(migapi.Condition{
+			Type:     OperatorVersionMismatch,
+			Status:   True,
+			Reason:   VersionCheckFailed,
+			Category: Warn,
+			Message:  fmt.Sprintf("This cluster is running a different version of the Migration Toolkit for Containers operator than the host cluster. Migrating to or from this cluster might result in a failed migration and data loss. Make sure all clusters are running the same version of the operator before attempting a migration."),
```
djwhatle, line 383: +1, good error message.

pkg/controller/migplan/validation.go
```diff
@@ -474,6 +481,40 @@ func (r ReconcileMigPlan) validateDestinationCluster(plan *migapi.MigPlan) error
...
+	}
+
+	destCluster, err := migapi.GetCluster(r, destRef)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+
+	srcClusterOperatorVersion, err := srcCluster.GetOperatorVersion(r)
+	destClusterOperatorVersion, err := destCluster.GetOperatorVersion(r)
+	//Not sure how to access the host operator version from here?
```
djwhatle, line 500: I think you could just read the MigCluster conditions here and avoid doing any remote API calls.

Since we already compare the version with host in the MigCluster validation, we could just do a lookup on MigCluster conditions with:

```
SomeMigCluster.HasCondition(OperatorVersionMismatch)
```

djwhatle (changes requested):

djwhatle:
Nice work on this so far @ibolton336 , added some comments to try to help tie up the last pieces.

pkg/apis/migration/v1alpha1/directimagemigration_types.go
pranavgaikwad: @djwhatle This actually happens when there's a comment on api fields. If there is no comment, all api fields are right aligned automatically by adding whitespaces, if there's a comment, the whitespaces are removed. I think go fmt is at fault here.

pkg/apis/migration/v1alpha1/migcluster_types.go
ibolton336: @djwhatle I thought I was grabbing the client on line 225 just above the error check?

pkg/controller/migcluster/validation.go
ibolton336: I was referencing the dependency in migcluster_types.go on line 30 .. seems that I need that specific package to access errors.Errorf

ibolton336: removing in favor of using GetOperatorVersion function like you suggested below :+1:

pkg/controller/migcluster/validation.go
ibolton336: Good call

pkg/apis/migration/v1alpha1/directimagemigration_types.go
djwhatle: Oh weird, thanks for pointing that out @pranavgaikwad , so it sounds like maybe after this change goes in we won't see things shift around anymore, just probably never ran `go fmt` after the field documentation that Shubham and Jaydip worked on.

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -218,6 +220,23 @@ func (m *MigCluster) GetClusterSubdomain(c k8sclient.Client) (string, error) {
 	return clusterSubdomain, nil
 }
 
+// GetRegistryImage gets a MigCluster specific registry image from ConfigMap
+func (m *MigCluster) GetOperatorVersion(c k8sclient.Client) (string, error) {
+	clusterClient, err := m.GetClient(c)
+	if err != nil {
+		return "", liberr.Wrap(err)
+	}
+	clusterConfig, err := m.GetClusterConfigMap(clusterClient)
```
djwhatle, line 229: This looks right

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -218,6 +220,23 @@ func (m *MigCluster) GetClusterSubdomain(c k8sclient.Client) (string, error) {
...
+	if err != nil {
+		return "", liberr.Wrap(err)
+	}
+	operatorVersion, ok := clusterConfig.Data[OperatorVersionKey]
+	if !ok {
+		return "", liberr.Wrap(errors.Errorf("configmap key not found: %v on MigCluster %v. Ensure that all MigClusters are running the same version of MTC Operator.", OperatorVersionKey, m.ClusterName))
```
djwhatle, line 235: I actually wonder if we should raise a critical condition `ClusterConfigmapKeyNotFound` when this occurs so that it's visible in the UI when this key can't be found, since it's going to be the case that everyone who gets autoupgraded to 1.4.2 or whatever will have an older MTC version on their 3.x cluster and will hit this error and need to dig through logs to fix.

pkg/controller/migcluster/validation.go
djwhatle: @ibolton336 is a `Warn` condition the right thing here?  I kind of expected `Critical` since migration will likely fail with mismatch, but don't recall what group consensus was.

pkg/controller/migplan/validation.go
```diff
@@ -474,6 +483,36 @@ func (r ReconcileMigPlan) validateDestinationCluster(plan *migapi.MigPlan) error
...
+	}
+
+	destCluster, err := migapi.GetCluster(r, destRef)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+	srcClusterCondition := srcCluster.Status.HasCondition(migcluster.OperatorVersionMismatch)
+	destClusterCondition := destCluster.Status.HasCondition(migcluster.OperatorVersionMismatch)
+
+	if !srcClusterCondition && !destClusterCondition {
```
djwhatle, line 502: Shouldn't the logic be flipped here? Seems like this is issuing a warning when neither cluster is mismatched.

djwhatle (changes requested):

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -653,6 +668,18 @@ func (m *MigCluster) GetRegistryPath(c k8sclient.Client) (string, error) {
 	return m.GetInternalRegistryPath(c)
 }
 
+func (m *MigCluster) SetOperatorVersion(c k8sclient.Client) error {
+	fmt.Printf("m is: %+v\n", m)
```
djwhatle, line 672: Need to remove these before merge

pkg/controller/migcluster/validation.go
```diff
@@ -346,3 +356,46 @@ func (r *ReconcileMigCluster) validateSaTokenPrivileges(cluster *migapi.MigClust
...
+	}
+	clusterOperatorVersion, err := cluster.GetOperatorVersion(clusterClient)
+
+	if err != nil {
+		cluster.Status.SetCondition(migapi.Condition{
+			Type:     ClusterOperatorVersionNotFound,
+			Status:   True,
+			Reason:   VersionNotFound,
+			Category: Critical,
+			Message:  fmt.Sprintf("The configmap key could not be found. See logs for details."),
```
djwhatle, line 381: Conditions are the first-class way we try to get information to users, so where possible it's preferable to put the information directly in the message instead of asking user to go to logs. In this case, I think we could make this something like:
```
MTC Operator version lookup failed. Key %key_name  not found in 
configmap %configmap_namespace/%configmap_name.
Ensure that all clusters are running the same MTC Operator version.
```

djwhatle: As discussed, for 1.4.2 (upcoming z-stream) we should set all of these conditions to be `Warning` instead of `Critical`. Sorry for the flip-flopping.

pkg/controller/migcluster/validation.go
```diff
@@ -346,3 +356,46 @@ func (r *ReconcileMigCluster) validateSaTokenPrivileges(cluster *migapi.MigClust
...
+
+	hostOperatorVersion, err := cluster.GetOperatorVersion(r)
+	operatorVersionMatchesHost := clusterOperatorVersion == hostOperatorVersion
+
+	if !operatorVersionMatchesHost {
+		cluster.Status.SetCondition(migapi.Condition{
+			Type:     OperatorVersionMismatch,
+			Status:   True,
+			Reason:   VersionCheckFailed,
+			Category: Critical,
```
djwhatle, line 395: Another place we need to switch to warn for the upcoming release instead of block.

pkg/controller/migplan/validation.go
```diff
@@ -474,6 +482,35 @@ func (r ReconcileMigPlan) validateDestinationCluster(plan *migapi.MigPlan) error
...
+		return liberr.Wrap(err)
+	}
+
+	destCluster, err := migapi.GetCluster(r, destRef)
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+	srcHasMismatch := srcCluster.Status.HasCondition(migcluster.OperatorVersionMismatch)
+	destHasMismatch := destCluster.Status.HasCondition(migcluster.OperatorVersionMismatch)
+	if srcHasMismatch || destHasMismatch {
```
djwhatle, line 500: Looks good

pkg/controller/migplan/validation.go
```diff
@@ -474,6 +482,35 @@ func (r ReconcileMigPlan) validateDestinationCluster(plan *migapi.MigPlan) error
...
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     ClusterVersionMismatch,
+			Status:   True,
+			Category: Critical,
```
djwhatle, line 504: -> warning

pkg/controller/migplan/validation.go
```diff
@@ -474,6 +482,35 @@ func (r ReconcileMigPlan) validateDestinationCluster(plan *migapi.MigPlan) error
...
+			Reason:   Conflict,
+			Message:  fmt.Sprintf("Cluster operator versions do not match. Source and destination clusters must have the same operator version."),
```
djwhatle, line 506: same operator version -> same MTC operator version

djwhatle (changes requested):

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -218,6 +220,23 @@ func (m *MigCluster) GetClusterSubdomain(c k8sclient.Client) (string, error) {
...
+	clusterConfig, err := m.GetClusterConfigMap(c)
+	if err != nil {
+		return "", "", "", liberr.Wrap(err)
+	}
+	configMapNamespace := clusterConfig.Namespace
+	configMapName := clusterConfig.Name
+
+	operatorVersion, ok := clusterConfig.Data[OperatorVersionKey]
+	if !ok {
+		return "", configMapNamespace, configMapName, liberr.Wrap(err)
```
djwhatle, line 234: I think we could avoid passing so many values back, because the source of truth for the `configmapNamespace` and `configMapName` are defined in consts `ClusterConfigMapName`, `VeleroNamespace`

pkg/controller/migcluster/validation.go
```diff
@@ -346,3 +357,65 @@ func (r *ReconcileMigCluster) validateSaTokenPrivileges(cluster *migapi.MigClust
...
+		return liberr.Wrap(err)
+	}
+	if clusterOperatorVersion == "" {
+
+		cluster.Status.SetCondition(migapi.Condition{
+			Type:     ClusterOperatorVersionNotFound,
+			Status:   True,
+			Reason:   VersionNotFound,
+			Category: Warn,
+			Message:  fmt.Sprintf(`MTC operator version lookup failed. Key %s  not found in configmap %s/%s. Ensure that all clusters are running the same MTC Operator version.`, migapi.OperatorVersionKey, clusterConfigMapNamespace, clusterConfigMapName),
```
djwhatle, line 385: I think we could just use the consts directly here as mentioned above.

djwhatle (approved):
Two final small tweaks suggested to re-use consts instead of passing configmap ns/name around. Other than that, LGTM! Is operator PR next?

djwhatle:
PR is looking great, thanks for working with me on this @ibolton336 !

ibolton336:
> Two final small tweaks suggested to re-use consts instead of passing configmap ns/name around. Other than that, LGTM! Is operator PR next?

Thanks for all of the reviews & helping me get up to speed @djwhatle 
Operator PR is all done - https://github.com/konveyor/mig-operator/pull/568
UI PR is almost done - need to handle the key not found warning condition. https://github.com/konveyor/mig-ui/pull/1122

djwhatle:
@ibolton336 I think we have a problem: when the non-host cluster doesn't have the OPERATOR_VERSION key, other clusters will show up with warnings, making it look like the configmap is missing there.

See screenshot. My OCP 3.11 cluster has latest operator and has `OPERATOR_VERSION: latest, but still raises a warning that makes it unclear what to do.

image

djwhatle:
We also have the problem of: when the host migcluster updates, the other migcluster won't realize it needs to update as well to check again if versions are aligned.

image

djwhatle:
Some possible solutions to the above:

 - Piggyback on the `connection check` that [migcluster controller does every 1 minute](https://github.com/konveyor/mig-controller/blob/c1162b5b814ad99933e5a69e8034f614db79d0cd/pkg/controller/migcluster/remotecluster.go#L40-L69), add some logic that will:
   - For each cluster, compare the configmap with `OPERATOR_VERSION` in it, see if matches that clusters `status.operatorVersion`
   - If the two don't match, enqueue the cluster for reconcile which will update the `status.version` and re-run validations
   - In MigCluster version validation code (already included in this PR), when we detect `OPERATOR_VERSION` has changed, pull all known MigClusters and set `refresh.true` on their specs so that they will re-check compatibility with the host cluster.
   - _At this point_, all version match validations should be correct

djwhatle:
I added a commit, check it out.

pkg/controller/migcluster/validation.go
```diff
@@ -346,3 +357,65 @@ func (r *ReconcileMigCluster) validateSaTokenPrivileges(cluster *migapi.MigClust
...
+
+		cluster.Status.SetCondition(migapi.Condition{
+			Type:     ClusterOperatorVersionNotFound,
+			Status:   True,
+			Reason:   VersionNotFound,
+			Category: Warn,
+			Message:  fmt.Sprintf(`MTC operator version lookup failed. Key %s not found in configmap %s/%s. Ensure that all clusters are running the same MTC Operator version.`, migapi.OperatorVersionKey, migapi.VeleroNamespace, migapi.ClusterConfigMapName),
+		})
+
+		return liberr.Wrap(err)
```
pranavgaikwad, line 388: @djwhatle Wouldn't error be nil here?

pkg/controller/migcluster/validation.go
```diff
@@ -346,3 +357,65 @@ func (r *ReconcileMigCluster) validateSaTokenPrivileges(cluster *migapi.MigClust
...
```
pranavgaikwad, line 405: same as above

pkg/controller/migcluster/validation.go
```diff
@@ -346,3 +357,65 @@ func (r *ReconcileMigCluster) validateSaTokenPrivileges(cluster *migapi.MigClust
...
+	}
+
+	operatorVersionMatchesHost := clusterOperatorVersion == hostOperatorVersion
+
+	if !operatorVersionMatchesHost {
+		cluster.Status.SetCondition(migapi.Condition{
+			Type:     OperatorVersionMismatch,
+			Status:   True,
+			Reason:   VersionCheckFailed,
+			Category: Warn,
```
pranavgaikwad, line 415: @djwhatle Ignore if you have already thought about this. It feels like this should be a Critical condition instead of Warn

pkg/controller/migplan/validation.go
```diff
@@ -474,6 +482,35 @@ func (r ReconcileMigPlan) validateDestinationCluster(plan *migapi.MigPlan) error
...
+	}
+	srcHasMismatch := srcCluster.Status.HasCondition(migcluster.OperatorVersionMismatch)
+	destHasMismatch := destCluster.Status.HasCondition(migcluster.OperatorVersionMismatch)
+	if srcHasMismatch || destHasMismatch {
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     ClusterVersionMismatch,
+			Status:   True,
+			Category: Warn,
+			Reason:   Conflict,
+			Message:  fmt.Sprintf("Cluster operator versions do not match. Source and destination clusters must have the same MTC operator version."),
```
pranavgaikwad, line 506: Looks like we are matching the operator version with the host cluster. Should the message be worded something like: `Source and destination clusters must have the same MTC operator version as the host`?

pranavgaikwad (changes requested):
Please see comments

pkg/controller/migcluster/validation.go
djwhatle: We decided `Warning` was appropriate here for now. I think when we get a bit of mileage on this system and know it won't malfunction I'd be more willing to make it a hard blocker.

pkg/controller/migplan/validation.go
djwhatle: There are two different validations:
 - MigCluster validation looks at `host` vs. `whatever` migcluster
 - MigPlan validation looks at `src` vs `dest` migcluster.

djwhatle: I misunderstood your comment at first but on second reading I think I get what you're saying. Since we are checking that both clusters are equivalent to the host cluster, the warning message should reflect if either one doesn't match. Subtle thing but makes sense to change I think.

sseago:
Does it make sense to separate out the formatting fixes added by `make run/go fmt` into a separate PR, or does that not matter so much at this point -- either way we need to get them in soon, as we're all seeing these local formatting changes every time we run the controller locally due to the discrepancy.

pkg/controller/migplan/validation.go
ibolton336: Message:  fmt.Sprintf("Cluster operator versions do not match. Source, destination, and host clusters must all have the same MTC operator version."),
 What do we think about this @vconzola

sseago: @ibolton336 How do we handle cases where the API warning/error message doesn't match what the UI would want to display? In this case, the terminology at the controller/API/CR level is different from what the UI presents. "Destination" is exactly the word to use for the cluster we restore into in a controller/API context, but the UI presents that as the "target" cluster. While it would be incorrect in an API-level message to say "target cluster", it might be considered equally incorrect to display an error message to the user referencing "destination cluster".

ibolton336:
> Does it make sense to separate out the formatting fixes added by `make run/go fmt` into a separate PR, or does that not matter so much at this point -- either way we need to get them in soon, as we're all seeing these local formatting changes every time we run the controller locally due to the discrepancy.

Opened a PR https://github.com/konveyor/mig-controller/pull/977

pkg/controller/migplan/validation.go
ibolton336: @sseago Good point. Would it be fair to say that we would like to prioritize the UI/UX & tailor to whatever the UXD team suggests to be the best fit?

vconzola:
@sseago @ibolton336  The warning text should use the same terminology as the UI if possible.

pkg/controller/migplan/validation.go
sseago: @ibolton336 @vconzola The problem is that this text is explicitly coming from the API. The term "target cluster" exists nowhere in the mig-controller API and controller context and it would actually be incorrect and confusing to use that term in an API-specific message. When we made the decision to use a different term in the UI context, it seems like that decision brought with it the responsibility on the UI side to do the string substitution (replacing "destination" with "target") when the UI is taking an API-specific message and presenting it in the UI.

ibolton336: @vconzola Was it a conscious decision along the way to use "target" in the UI? If so we should be ok to write the string replace functions on UI side for display. Just want to make sure that is what we want first.

pkg/controller/migcluster/validation.go
```diff
@@ -82,6 +87,12 @@ func (r ReconcileMigCluster) validate(cluster *migapi.MigCluster) error {
 		return liberr.Wrap(err)
 	}
 
+	// cluster version
+	err = r.validateOperatorVersionMatchesHost(cluster)
+	if err != nil {
```
djwhatle, line 92: @ibolton336 I think we need some modification here.

`validateOperatorVersionMatchesHost` will call
```
	clusterOperatorVersion, err := cluster.GetOperatorVersion(clusterClient)
	if err != nil {
		return liberr.Wrap(err)
	}
```

`cluster.GetOperatorVersion` includes
```
	operatorVersion, ok := clusterConfig.Data[OperatorVersionKey]
	if !ok {
		return "", liberr.Wrap(err)
	}
```

So I think `validateOperatorVersionMatchesHost` will return an error down the chain, which will cause the MigCluster to requeue. One of the ways an error gets generated will be if the configmap key isn't present, which we _are expecting to happen_. 

We should not pass the error back down and exit the reconcile for an _expected_ issue like this. Exiting a reconcile with an error will lead to a requeue, and in this case it would be an infinite requeue (not desired, can cause perf problems) until the user upgrades this MigCluster MTC install to a version including the expected configmap key.

```
	// Validations.
	err = r.validate(cluster)
	if err != nil {
		log.Trace(err)
	}
```

djwhatle (changes requested):
See comment, realized we have an issue

pkg/controller/migplan/validation.go
vconzola: @ibolton336 We've been using "target" in the UI since day one of CAM. It's the term that was used in IMS, so we kept it for CAM. No one ever balked. Since the project started way back when with UI mockups, the API should have adopted the UI terminology from the beginning. I'm not sure why that didn't happen.

sseago: I wasn't on the team when the target/destination split first happened, but "destination" was already established on the mig-controller side back in March 2019. So with 2 years of history, I guess we're stuck with it unless one side or the other wants to undergo a major rewrite (breaking API compability, etc. if it happens on the controller side). I'm assuming that is unlikely at this point, so a string search-and-replace on api-generated strings before presenting them in the UI is probably the easiest way to keep each side internally consistent for the moment.

djwhatle: _Here's what I remember from history of the terms:_ 

There was a split in usage of terms early on: early versions of mig-controller used `destination` before the UI existed when we were in proof of concept phase.

In parallel to controller proof of concept work, folks on the project planning side of things (Todd, John, Vince, etc.) started using the term `target` as part of their everyday conversations. I think that's where the rift came from. No one pushed for the two terms to be reconciled before we shipped the first version of our API to users.

Making API changes would require some sort of API migration path (a migration for your migration software 😄 ) for users who have already deployed. Contents of MigPlan CRs would need to be migrated, which isn't something we have ever handled thus far (`spec.destCluster` -> `spec.targetCluster`).

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -218,6 +220,21 @@ func (m *MigCluster) GetClusterSubdomain(c k8sclient.Client) (string, error) {
...
+// GetOperatorVersion retrieves the operator version from the respective controllers ConfigMap
+func (m *MigCluster) GetOperatorVersion(c k8sclient.Client) (string, error) {
+	clusterConfig, err := m.GetClusterConfigMap(c)
+	if err != nil {
+		return "", liberr.Wrap(err)
+	}
+
+	operatorVersion, ok := clusterConfig.Data[OperatorVersionKey]
+	if !ok {
+		return "", liberr.Wrap(err)
```
djwhatle, line 232: @ibolton336 I don't think this line isn't doing what we want, since no `err` is generated it's actually going to return the err (or nil value) from line 225. 

We need to create an error to return, like 

```
return "", liberr.Wrap(errors.Errorf("configmap key %v not found in configmap %v/%v" for migcluster %v/%v, 
OperatorVersionKey,  clusterConfig.Namespace, clusterConfig.Name, m.Namespace, m.Name))
```

djwhatle:
Here's what it looks like when there's a version mismatch and it is resolved.

https://user-images.githubusercontent.com/7576968/109848450-48f7e200-7c1e-11eb-919d-89305aad0955.mov

Here's what it looks like when versions previously matched, but now are mismatched

https://user-images.githubusercontent.com/7576968/109848476-51501d00-7c1e-11eb-999c-05eb3c2aa9f3.mov

djwhatle:
@ibolton336 I think I had the wrong idea about how we should handle a situation where the configmap w/ version info can't be found. In this case we still should update the status.OperatorVersion to be a blank string. If the user downgrades the version will be stuck on what is was previously (my fault).

Submitting fix

pkg/apis/migration/v1alpha1/migcluster_types.go
```diff
@@ -690,11 +690,10 @@ func (m *MigCluster) SetOperatorVersion(c k8sclient.Client) error {
...
 	}
-	newOperatorVersion, err := m.GetOperatorVersion(clusterClient)
+	newOperatorVersion, _ := m.GetOperatorVersion(clusterClient)
 	// Ignore error here. Missing configmap/key is already raised to user in validation,
-	// we don't want reconcile to exit w/ error on MTC < 1.4.2. We also don't want to
-	// accidentally trigger the expensive propagation below if there's an error
-	// getting operator version.
+	// we don't want reconcile to exit w/ error on MTC < 1.4.2. GetOperatorVersion will
+	// return "" on error which is usable below.
 	if err != nil {
```
djwhatle, line 697: Hmm, this error check really should not be here if I'm not generating an error above.

djwhatle:
@ibolton336 I'm not sure what's causing it, but there's a bug here leading to repeated reconcile of the MigCluster when the cluster version can't be found in configmap. Working through finding the bug, need to resolve this to avoid serious perf issues.

djwhatle (approved):

PR #955, opened 2021-02-15
PV resizing logic

pranavgaikwad:

This PR: 
- Adds DF command basics
- Adds ResticDFExecutor, responsible for running df using Restic pods
- Adds VolumeAdjuster responsible for computing adjusted volume size

### DIFF
--- a/pkg/controller/miganalytic/miganalytics_controller.go
+++ b/pkg/controller/miganalytic/miganalytics_controller.go
@@ -108,9 +108,9 @@ type ReconcileMigAnalytic struct {
 // MigAnalyticPersistentVolumeDetails defines extended properties of a volume discovered by MigAnalytic
 type MigAnalyticPersistentVolumeDetails struct {
 	Name                string
-	RequestedCapacity   *resource.Quantity
+	RequestedCapacity   resource.Quantity
 	PodUID              types.UID
-	ProvisionedCapacity *resource.Quantity
+	ProvisionedCapacity resource.Quantity
 	StorageClass        *string
 	Namespace           string
 	VolumeName          string
@@ -212,7 +212,7 @@ func (r *ReconcileMigAnalytic) analyze(analytic *migapi.MigAnalytic) error {
 		return liberr.Wrap(err)
 	}
 
-	if analytic.Status.Analytics.PercentComplete == 100 {
+	if analytic.Status.Analytics.PercentComplete == 100 && !analytic.Spec.Refresh {
 		return nil
 	}
 
@@ -236,8 +236,10 @@ func (r *ReconcileMigAnalytic) analyze(analytic *migapi.MigAnalytic) error {
 		return liberr.Wrap(err)
 	}
 
-	analytic.Status.Analytics.Plan = plan.Name
+	nodeToPVMap := make(map[string][]MigAnalyticPersistentVolumeDetails)
 
+	analytic.Status.Analytics.Plan = plan.Name
+	analytic.Status.Analytics.Namespaces = make([]migapi.MigAnalyticNamespace, 0)
 	for i, namespace := range plan.Spec.Namespaces {
 		for _, ns := range analytic.Status.Analytics.Namespaces {
 			if ns.Namespace == namespace {
@@ -263,6 +265,7 @@ func (r *ReconcileMigAnalytic) analyze(analytic *migapi.MigAnalytic) error {
 				return liberr.Wrap(err)
 			}
 		}
+
 		if analytic.Spec.AnalyzePVCapacity && !(isExcluded("persistentvolumes", excludedResources) &&
 			isExcluded("persistentvolumeclaims", excludedResources)) {
 
@@ -273,7 +276,13 @@ func (r *ReconcileMigAnalytic) analyze(analytic *migapi.MigAnalytic) error {
 		}
 
 		if analytic.Spec.AnalyzeExntendedPVCapacity {
-			err := r.analyzeExtendedPVCapacity(client, &ns)
+			namespacedNodeToPVMap, err := r.getNodeToPVCMapForNS(&ns, client)
+			for node, pvcs := range namespacedNodeToPVMap {
+				if _, exists := nodeToPVMap[node]; !exists {
+					nodeToPVMap[node] = make([]MigAnalyticPersistentVolumeDetails, 0)
+				}
+				nodeToPVMap[node] = append(nodeToPVMap[node], pvcs...)
+			}
 			if err != nil {
 				return liberr.Wrap(err)
 			}
@@ -294,27 +303,49 @@ func (r *ReconcileMigAnalytic) analyze(analytic *migapi.MigAnalytic) error {
 			return liberr.Wrap(err)
 		}
 	}
+
+	if analytic.Spec.AnalyzeExntendedPVCapacity {
+		err := r.analyzeExtendedPVCapacity(client, analytic, nodeToPVMap)
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+		err = r.Update(context.TODO(), analytic)
+		if err != nil {
+			return liberr.Wrap(err)
+		}
+	}
+
 	return nil
 }
 
-func (r *ReconcileMigAnalytic) analyzeExtendedPVCapacity(client compat.Client, ns *migapi.MigAnalyticNamespace) error {
+// analyzeExtendedPVCapacity given a source client, owner cr and map of nodeNames -> pvs, run volume adjustment
+func (r *ReconcileMigAnalytic) analyzeExtendedPVCapacity(sourceClient compat.Client, analytic *migapi.MigAnalytic, nodeToPVMap map[string][]MigAnalyticPersistentVolumeDetails) error {
+	volumeAdjuster := PersistentVolumeAdjuster{
+		Owner:  analytic,
+		Client: r.Client,
+		DFExecutor: &ResticDFCommandExecutor{
+			Namespace: migapi.VeleroNamespace,
+			Client:    sourceClient,
+		},
+	}
+	err := volumeAdjuster.Run(nodeToPVMap)
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
-func (r *ReconcileMigAnalytic) getNodeToPVMapForNS(ns *migapi.MigAnalyticNamespace, client compat.Client) (map[string][]MigAnalyticPersistentVolumeDetails, error) {
-
+// getNodeToPVCMapForNS given a ns and client, returns map of nodeName -> PV Info present on that node
+func (r *ReconcileMigAnalytic) getNodeToPVCMapForNS(ns *migapi.MigAnalyticNamespace, client compat.Client) (map[string][]MigAnalyticPersistentVolumeDetails, error) {
 	podList := corev1.PodList{}
 	nodeToPVDetails := map[string][]MigAnalyticPersistentVolumeDetails{}
-
 	err := client.List(
 		context.TODO(),
 		k8sclient.InNamespace(ns.Namespace),
 		&podList)
-
 	if err != nil {
 		return nil, liberr.Wrap(err)
 	}
-
 	for _, pod := range podList.Items {
 		if pod.Status.Phase == migapi.Running {
 			for _, vol := range pod.Spec.Volumes {
@@ -339,9 +370,9 @@ func (r *ReconcileMigAnalytic) getNodeToPVMapForNS(ns *migapi.MigAnalyticNamespa
 						MigAnalyticPersistentVolumeDetails{
 							Name:                vol.PersistentVolumeClaim.ClaimName,
 							Namespace:           pvcObject.Namespace,
-							RequestedCapacity:   pvcObject.Spec.Resources.Requests.StorageEphemeral(),
+							RequestedCapacity:   pvcObject.Spec.Resources.Requests[corev1.ResourceStorage],
 							PodUID:              pod.UID,
-							ProvisionedCapacity: pvcObject.Status.Capacity.StorageEphemeral(),
+							ProvisionedCapacity: pvcObject.Status.Capacity[corev1.ResourceStorage],
 							StorageClass:        pvcObject.Spec.StorageClassName,
 							VolumeName:          pvcObject.Spec.VolumeName,
 						})
--- a/pkg/controller/miganalytic/restic_executor.go
+++ b/pkg/controller/miganalytic/restic_executor.go
@@ -1,0 +1,151 @@
+/*
+Copyright 2021 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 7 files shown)


pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -108,9 +108,9 @@ type ReconcileMigAnalytic struct {
 // MigAnalyticPersistentVolumeDetails defines extended properties of a volume discovered by MigAnalytic
 type MigAnalyticPersistentVolumeDetails struct {
 	Name                string
-	RequestedCapacity   *resource.Quantity
+	RequestedCapacity   resource.Quantity
```
shawn-hurley, line 111: Sorry, if this is a small change, but is there a distinction between no RequestedCapacity and `0` RequestedCapacity? Could there already be a stored version on a cluster with nil here and would that cause issues on upgrade?

pranavgaikwad: @shawn-hurley valid question. AFAIK, that shouldn't cause the problem, the new struct will default to `0` values. That case will be handled further in the VolumeAdjustment logic

pkg/controller/miganalytic/restic_executor.go
```diff
@@ -0,0 +1,150 @@
...
+	// dfOutputs for n nodes
+	dfOutputs := make(map[string]DFCommand, len(pvcNodeMap))
+	var waitGroup sync.WaitGroup
+	var mutex sync.Mutex
+	waitGroup.Add(len(pvcNodeMap))
+	// run df concurrently for 'n' nodes
+	for node := range pvcNodeMap {
+		resticPodRef := r.getResticPodForNode(node)
+		// if no Restic pod is found for this node, all PVCs on this node are skipped
+		if resticPodRef == nil {
```
shawn-hurley, line 119: if we hit this condition, we won't make on of the wait-groups as done and we will look I think.

pranavgaikwad: @shawn-hurley good catch!

pkg/controller/miganalytic/restic_executor.go
```diff
@@ -0,0 +1,150 @@
...
+				dfOutput := DFOutput{
+					IsError:   true,
+					Name:      pvc.Name,
+					Namespace: pvc.Namespace,
+				}
+				gatheredData = append(gatheredData, dfOutput)
+			}
+			continue
+		}
+		go func(n string, podRef *corev1.Pod, outputs map[string]DFCommand) {
```
shawn-hurley, line 130: it does not appear that there is any limit or buffer, so lets say we have 500 nodes we will have 500 threads, and then we will quickly make 500 calls and then they will block until the map is updated. 

Not only will this not really gain us as much value from the async (I think right?) it will spam the API server will a ton of requests really quickly. Assuming that each call takes some period of time, we may want to consider making a buffered channel executor that also parses the output from below in each thread and the return is pvcDFInfo?

WDYT? too much over engineering?

pkg/controller/miganalytic/volume_adjustment.go
```diff
@@ -0,0 +1,249 @@
...
+	case BinarySIGiga, BinarySIMega:
+		quantity = fmt.Sprintf("%si", quantity)
+		break
+	}
+	return resource.ParseQuantity(quantity)
+}
+
+// GetDFOutputForPV given a volume name and pod uid, returns structured df ouput for the volume
+// only works on outputs of commands created by DFCommand.PrepareDFCommand()
+func (cmd *DFCommand) GetDFOutputForPV(volName string, podUID types.UID) (pv DFOutput) {
```
shawn-hurley, line 114: I probably would like a unit test for this, I feel like this could be pretty brittle as we are parsing console out

Have you considered using some `awk` to format the output as JSON?

pkg/controller/miganalytic/volume_adjustment.go
```diff
@@ -0,0 +1,249 @@
...
+				pvNames,
+				dfOutput.Name)
+		}
+	}
+	pva.Owner.Status.Conditions.SetCondition(migapi.Condition{
+		Category: migapi.Warn,
+		Status: True,
+		Type: ExtendedPVAnalysisFailed,
+		Reason: FailedRunningDf,
+		Message: fmt.Sprintf(warningStrings, pvNames),
```
shawn-hurley, line 197: s/warningStrings/waringString?

pkg/controller/miganalytic/volume_adjustment.go
```diff
@@ -0,0 +1,249 @@
...
+	return nil
+}
+
+// generateWarningForErroredPVs given a list of dfoutputs, generates warnings for associated pvcs
+func (pva *PersistentVolumeAdjuster) generateWarningForErroredPVs(erroredPVs []*DFOutput) {
+	warningString := "Failed gathering extended PV usage information for PVs [%s]"
+	pvNames := ""
+	for _, dfOutput := range erroredPVs {
+		if dfOutput.IsError {
+			fmt.Sprintf(
```
shawn-hurley, line 187: is this supposed to saved as pvNames?

pkg/controller/miganalytic/volume_adjustment.go
```diff
@@ -0,0 +1,249 @@
...
+			if pvc.Name == pvDetailRef.Name && pvc.Namespace == pvDetailRef.Namespace {
+				return pvDetailRef
+			}
+		}
+	}
+	return nil
+}
+
+// generateWarningForErroredPVs given a list of dfoutputs, generates warnings for associated pvcs
+func (pva *PersistentVolumeAdjuster) generateWarningForErroredPVs(erroredPVs []*DFOutput) {
```
shawn-hurley, line 182: A unit test here might a good thing as well?

pkg/controller/miganalytic/volume_adjustment.go
```diff
@@ -0,0 +1,249 @@
...
+			pva.StatusRefCache[namespace] = nsRef
+			return nsRef
+		}
+	}
+	return nil
+}
+
+func (pva *PersistentVolumeAdjuster) calculateProposedVolumeSize(usagePercentage int64, actualCapacity resource.Quantity,
+	requestedCapacity resource.Quantity) (proposedSize resource.Quantity, reason string) {
+	return
```
shawn-hurley, line 221: Assume this still needs to be implemented?

pkg/controller/miganalytic/volume_adjustment_test.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+var testQuantities = []testQuantity{
+	{
+		dfQuantity:  "1090MB",
+		k8sQuantity: "1090M",
+	},
+}
+
+func TestDFCommand_convertDFQuantityToKubernetesResource(t *testing.T) {
```
shawn-hurley, line 140: I would like to see a test for GB as well

pkg/controller/miganalytic/restic_executor.go
pranavgaikwad: @shawn-hurley good point. I think it's fair to assume a scenario for 500 nodes. I can use your help implementing some sort of a buffered channel. Haven't really done that.

pkg/controller/miganalytic/volume_adjustment.go
pranavgaikwad: @shawn-hurley correct. Shubham is implementing this piece. We will get this in a different PR

pkg/controller/miganalytic/miganalytics_controller.go
shawn-hurley: Wouldn't you have a situation like this? because we don't have omit-empty?

https://play.golang.org/p/RQQFv9_mfS-

pkg/controller/miganalytic/restic_executor.go
```diff
@@ -0,0 +1,183 @@
...
+	cmdString := dfCmd.PrepareDFCommand(persistentVolumes)
+	restCfg := r.Client.RestConfig()
+	podCommand := pods.PodCommand{
+		Pod:     podRef,
+		RestCfg: restCfg,
+		Args:    cmdString,
+	}
+	err := podCommand.Run()
+	if err != nil {
+		log.Info(
```
shawn-hurley, line 66: how does the user know that we are unsuccessful in determining the current pod volumes, and why

Can't this also error for other reasons like pod does not exist or unable to connect to pod or some other reason?

pranavgaikwad: @shawn-hurley according to current implementation, if this happens, all PVCs on this node will be added to Warning indicating that we failed to gather data for these volumes. It's either we succeed or we fail. Maybe we can log the message here, but unsure how to bubble up the detailed failure information all the way up until MigPlan.

pkg/controller/miganalytic/miganalytics_controller.go
pranavgaikwad: @shawn-hurley I see your point now. This is a new change we introduced specifically for PV resize stuff in an earlier API PR. So we don't have this struct present on any MigAnalytics in previous versions. MigAnalytic controller fills this when it finds the PVs. So I think we wont come across scenario you described

pkg/controller/miganalytic/volume_adjustment.go
pranavgaikwad: @shawn-hurley since we are running DF for all volumes on a node at once, I thought it might be just easier to parse the output in golang. For instance, we will generate command like this:

```sh
df -BM /path-1 /path-2 /path-3 /path-4 /path-5 ...
```

The logic is very simple. We just look for volume names in output lines (StdErr/StdOut).

shawn-hurley: The logic may be simple, but what if we get back something we don't know how to handle? 

I think why I think a unit test is important, is AFAICT we run a bunch of regex's to find the info, let's make sure that if there is a bug in here and someone tries to fix it, that they don't regress on the output that we are expecting.

The reason for JSON or some other format is probably not needed as I doubt DF will be changing its output structure.

pranavgaikwad: @shawn-hurley sounds good. Actually there's already a unit test for this function. Looking at it, I realize that I should add more cases to it which handle unparseable inputs.

djwhatle: +1 this seems like a great place for unit tests, pretty self contained input/output

pkg/controller/miganalytic/restic_executor.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		waitGroup.Add(1)
+		go func(n string, podRef *corev1.Pod) {
+			// block until channel empty
+			bufferedExecutionChannel <- struct{}{}
+			defer waitGroup.Done()
+			output := r.DF(podRef, pvcNodeMap[n])
+			mutex.Lock()
+			dfOutputs[n] = output
+			mutex.Unlock()
```
pranavgaikwad, line 139: Use `defer`

djwhatle (commented):
I think we need to be sure that this triggers at the right times, and what we're lacking right now is a trigger on namespace change. Pranav and I discussed adding a `status.NamespaceDigest` field to MigPlan and also a `HasReconciledNamespaces()` function to help  with this. 

Down the road (or if Pranav feels up to changing this now ;) )we can also use this field to limit PV discovery to only run when 
1. MigPlan is new
2. Namespaces change 
3. Refresh is requested.

Also think we should shrink the footprint of the PV resize logic in the main reconcile func and move some of that code out to a func called from main reconcile. 

For example of deciding whether to trigger an action based on `HasReconciled()`: https://github.com/konveyor/mig-controller/blob/master/pkg/controller/migplan/gvk.go#L12-L16

Edit: There are several open PRs related to PV intelligent resize, this comment is generally applicable to these PRs but perhaps not relevant to content of this particular PR.

pkg/controller/miganalytic/volume_adjustment.go
pranavgaikwad: updated test

pranavgaikwad:
I have addressed all concerns folks had on the contents of this PR. There is one open concern raised by @shawn-hurley about error handling. Since 1.4.1 is just API support for the feature, I'd like to merge this for now without addressing error handling and come up with a holistic solution for that case when we start working on UI/UX for the feature. In effect, this PR alone is just focuses on creating a MigAnalytic flow for collecting DF information from Restic pods, and doesn't affect / break any existing features.

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -236,8 +236,10 @@ func (r *ReconcileMigAnalytic) analyze(analytic *migapi.MigAnalytic) error {
 		return liberr.Wrap(err)
 	}
 
-	analytic.Status.Analytics.Plan = plan.Name
+	nodeToPVMap := make(map[string][]MigAnalyticPersistentVolumeDetails)
 
+	analytic.Status.Analytics.Plan = plan.Name
+	analytic.Status.Analytics.Namespaces = make([]migapi.MigAnalyticNamespace, 0)
```
alaypatel07, line 242: do we know the namespaces length already? can we seen it with length that is not 0?

alaypatel07: FTR its a NIT, not a merge blocker

pkg/controller/miganalytic/restic_executor.go
alaypatel07: +1 good idea

pkg/controller/miganalytic/restic_executor.go
```diff
@@ -0,0 +1,156 @@
...
+					Namespace: pvc.Namespace,
+				}
+				gatheredData = append(gatheredData, dfOutput)
+			}
+			continue
+		}
+		waitGroup.Add(1)
+		go func(n string, podRef *corev1.Pod) {
+			// block until channel empty
+			bufferedExecutionChannel <- struct{}{}
```
djwhatle, line 134: Worth noting for any other reviewers looking at this scratching their heads, this is just stuffing a blank struct in the channel as a means of messaging. No actual data is contained.

alaypatel07:
glanced quickly, love the unit tests and buffered reader implementation, +1, will leave the ack for other folks who have followed more closely, but lgtm

pkg/controller/miganalytic/restic_executor.go
```diff
@@ -0,0 +1,156 @@
...
+	for i := range resticPodList.Items {
+		if resticPodList.Items[i].Spec.NodeName != "" {
+			r.ResticPodReferences[resticPodList.Items[i].Spec.NodeName] = &resticPodList.Items[i]
+		}
+	}
+	return nil
+}
+
+// Execute given a map node->[]pvc, runs Df command for each, returns list of structured df output per pvc
+func (r *ResticDFCommandExecutor) Execute(pvcNodeMap map[string][]MigAnalyticPersistentVolumeDetails) ([]DFOutput, error) {
```
djwhatle, line 104: I wonder if we should consider bumping `MaxConcurrentReconciles` to `10` instead of `1` so that the PV resize MigAnalytic is never stuck waiting in line behind the general purpose MigAnalytic that pulls all resource counts for plan.

https://github.com/konveyor/mig-controller/blob/master/pkg/controller/miganalytic/miganalytics_controller.go#L79

djwhatle (approved):

PR #956, opened 2021-02-15
Proposed volume size computation for PV resizing

shubham-pampattiwar:
This PR does the following:
- Adds a function `calculateProposedVolumeSize` , it accepts `usagePercentage` , `actualCapacity` and `requestedCapacity` and returns a `proposedSize` alongside a `reason` string
- Adds unit tests

Depends on #955 

Please review 955 first


pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -721,4 +721,4 @@ type routingConfig struct {
 
 type imagePolicyConfig struct {
 	InternalRegistryHostname string `json:"internalRegistryHostname"`
-}
+}
```
pranavgaikwad, line 724: please restore this change

pkg/controller/miganalytic/volume_adjustment.go
```diff
@@ -63,7 +63,14 @@ const (
 	BinarySIGiga  = DFBaseUnit("G")
 )
 
-// DFCommand represents a df command
+// Proposed volume size computation reasons
+const (
+	VolumeAdjustmentNoOp             = "PV sizes not altered. Using original size"
+	VolumeAdjustmentUsageExceeded    = "PV size was adjusted by adding extra headroom of 3%"
+	VolumeAdjustmentCapacityMismatch = "Provisioned size not equal to the requested size, using provisioned size"
```
pranavgaikwad, line 70:
```suggestion
	VolumeAdjustmentNoOp             = "No change in original PV capacity is needed."
	VolumeAdjustmentUsageExceeded    = "PV usage is close to 100%, an extra headroom is added to avoid disk capacity issue in the target cluster."
	VolumeAdjustmentCapacityMismatch = "Requested capacity of PV is not equal to its actual provisioned capacity.  Maximum of both values is used to avoid disk capacity issue in the target cluster."
```

pranavgaikwad (changes requested):

pkg/controller/miganalytic/volume_adjustment.go
shubham-pampattiwar: Done

pranavgaikwad (approved):
LGTM

PR #957, opened 2021-02-15
Proposed PV size processing and checking

shubham-pampattiwar:
This PR does the following:
- Introduces two new fields in `Migplan.Spec.PV` - `Confirmed` and `ProposedSize`
- Adds a new function `processExtendedPVCapacity` which maps PV size data of the `MigPlan` volumes to that of the volumes in `MigAnalytic` , checks if the proposed size is updated and confirmed in the `MigPlan` or not.
- Adds a new function `validatePVSizeConfirmation` which sets or deletes `UnconfirmedPV` condition on `MigPlan` based on the value of `Confirmed` flag specified per PV.
- Relays `EnxtendedPVAnalysisFailed` condition from `MigAnalytic` to `Migplan`

Depends on #955 

Please review 955 first

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -716,16 +716,16 @@ const (
 // NFS - NFS properties.
 // staged - A PV has been explicitly added/updated.
 type PV struct {
-	Name             string                `json:"name,omitempty"`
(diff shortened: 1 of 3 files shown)


pranavgaikwad (changes requested):
LGTM

pranavgaikwad (approved):

PR #958, opened 2021-02-16
Bug 1916554: failing if rsync client pods are stuck 

JaydipGabani:
Failing the migration if Rsync client pods are stuck in "container creating" phase for more than 10 mins.

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -1170,7 +1170,7 @@ func (t *Task) haveRsyncClientPodsCompletedOrFailed() (bool, bool, error) {
 	t.Owner.Status.RunningPods = []*migapi.PodProgress{}
 	t.Owner.Status.FailedPods = []*migapi.PodProgress{}
 	t.Owner.Status.SuccessfulPods = []*migapi.PodProgress{}
-
+	var pendingPods []string
 	pvcMap := t.getPVCNamespaceMap()
 	for ns, vols := range pvcMap {
 		for _, vol := range vols {
@@ -1206,13 +1206,30 @@ func (t *Task) haveRsyncClientPodsCompletedOrFailed() (bool, bool, error) {
 					LastObservedProgressPercent: dvmp.Status.LastObservedProgressPercent,
 					LastObservedTransferRate:    dvmp.Status.LastObservedTransferRate,
 				})
+			case dvmp.Status.PodPhase == corev1.PodPending:
+				t.Owner.Status.PendingPods = append(t.Owner.Status.PendingPods, &migapi.PodProgress{
+					ObjectReference:             objRef,
+					LastObservedProgressPercent: dvmp.Status.LastObservedProgressPercent,
+					LastObservedTransferRate:    dvmp.Status.LastObservedTransferRate,
+				})
+				pendingPods = append(pendingPods, fmt.Sprintf("%s/%s", objRef.Name, objRef.Namespace))
 			}
 		}
 	}
 
 	isCompleted := len(t.Owner.Status.SuccessfulPods)+len(t.Owner.Status.FailedPods) == len(t.Owner.Spec.PersistentVolumeClaims)
 	hasAnyFailed := len(t.Owner.Status.FailedPods) > 0
-
+	isAnyPending := len(t.Owner.Status.PendingPods) > 0
+	if isAnyPending {
+		pendingMessage := fmt.Sprintf("Pods %s are stuck in Pending state for more than 10 mins",strings.Join(pendingPods[:], ", "))
+		t.Owner.Status.SetCondition(migapi.Condition{
+			Type:     RsyncClientPodsPending,
+			Status:   migapi.True,
+			Reason:   "PodStuckInContainerCreating",
+			Category: migapi.Warn,
+			Message:  pendingMessage,
+		})
+	}
 	return isCompleted, hasAnyFailed, nil
 }
 
--- a/pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
+++ b/pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
@@ -46,6 +46,8 @@ import (
 
 var log = logging.WithName("pvmigrationprogress")
 
+var TimeLimit = 10*time.Minute
+
 const (
 	NotFound    = "NotFound"
 	NotSet      = "NotSet"
@@ -58,6 +60,12 @@ const (
 	ClusterNotReady   = "ClusterNotReady"
 	InvalidPodRef     = "InvalidPodRef"
 	InvalidPod        = "InvalidPod"
+	PodNotReady       = "PodNotReady"
+)
+
+const (
+	// CreatingContainer initial container state
+	ContainerCreating = "ContainerCreating"
 )
 
 // Add creates a new DirectVolumeMigrationProgress Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
@@ -276,6 +284,13 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
 		exitCode := containerStatus.LastTerminationState.Terminated.ExitCode
 		pvProgress.Status.ExitCode = &exitCode
 		pvProgress.Status.ContainerElapsedTime = &metav1.Duration{Duration: containerStatus.LastTerminationState.Terminated.FinishedAt.Sub(containerStatus.LastTerminationState.Terminated.StartedAt.Time).Round(time.Second)}
+	case !containerStatus.Ready && pod.Status.Phase == kapi.PodPending && containerStatus.State.Waiting != nil && containerStatus.State.Waiting.Reason == ContainerCreating:
+		// if pod is not in running state after 10 mins of its creation, raise a
+		if time.Now().UTC().Sub(pod.CreationTimestamp.Time.UTC()) > TimeLimit {
+			pvProgress.Status.PodPhase = kapi.PodPending
+			pvProgress.Status.LogMessage = fmt.Sprintf("Pod %s/%s is stuck in Pending state for more than 10 mins", pod.Namespace, pod.Name)
+			pvProgress.Status.ContainerElapsedTime = &metav1.Duration{Duration: time.Now().Sub(pod.CreationTimestamp.Time).Round(time.Second)}
+		}
 	case pod.Status.Phase == kapi.PodFailed:
 		// Its possible for the succeeded pod to not have containerStatuses at all
 		pvProgress.Status.PodPhase = kapi.PodFailed
--- a/pkg/apis/migration/v1alpha1/directvolumemigration_types.go
+++ b/pkg/apis/migration/v1alpha1/directvolumemigration_types.go
@@ -31,17 +31,17 @@ type PVCToMigrate struct {
 
 // DirectVolumeMigrationSpec defines the desired state of DirectVolumeMigration
 type DirectVolumeMigrationSpec struct {
-	SrcMigClusterRef            *kapi.ObjectReference `json:"srcMigClusterRef,omitempty"`
-	DestMigClusterRef           *kapi.ObjectReference `json:"destMigClusterRef,omitempty"`
+	SrcMigClusterRef  *kapi.ObjectReference `json:"srcMigClusterRef,omitempty"`
+	DestMigClusterRef *kapi.ObjectReference `json:"destMigClusterRef,omitempty"`
 
 	//  Holds all the PVCs that are to be migrated with direct volume migration
-	PersistentVolumeClaims      []PVCToMigrate        `json:"persistentVolumeClaims,omitempty"`
+	PersistentVolumeClaims []PVCToMigrate `json:"persistentVolumeClaims,omitempty"`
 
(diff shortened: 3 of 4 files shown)


alaypatel07:
I was out of office today, will take a look at this tomorrow morning. I think this bug is manifested in many forms, we need to create a generic solution and get some unit tests in. I will take a detailed look tomorrow morning

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -276,6 +281,12 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
 		exitCode := containerStatus.LastTerminationState.Terminated.ExitCode
 		pvProgress.Status.ExitCode = &exitCode
 		pvProgress.Status.ContainerElapsedTime = &metav1.Duration{Duration: containerStatus.LastTerminationState.Terminated.FinishedAt.Sub(containerStatus.LastTerminationState.Terminated.StartedAt.Time).Round(time.Second)}
+	case !containerStatus.Ready && containerStatus.State.Waiting != nil && containerStatus.State.Waiting.Reason == ContainerCreating:
+		// if pod is not in running state after 10 mins of its creation, assume the pod is in failed state
+		if time.Now().Sub(pod.CreationTimestamp.Time) > 10*time.Minute {
+			pvProgress.Status.PodPhase = kapi.PodFailed
```
alaypatel07, line 287: do we really want this? 

Ideally, we should raise a warning right? that says that a pod is in creating the state since 10 mins?

JaydipGabani: +1 to rasing warning

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -276,6 +284,20 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
 		exitCode := containerStatus.LastTerminationState.Terminated.ExitCode
 		pvProgress.Status.ExitCode = &exitCode
 		pvProgress.Status.ContainerElapsedTime = &metav1.Duration{Duration: containerStatus.LastTerminationState.Terminated.FinishedAt.Sub(containerStatus.LastTerminationState.Terminated.StartedAt.Time).Round(time.Second)}
+	case !containerStatus.Ready && pod.Status.Phase == kapi.PodPending && containerStatus.State.Waiting != nil && containerStatus.State.Waiting.Reason == ContainerCreating:
+		// if pod is not in running state after 10 mins of its creation, raise a
+		if time.Now().Sub(pod.CreationTimestamp.Time) > TimeLimit {
+			pvProgress.Status.PodPhase = kapi.PodPending
+			pvProgress.Status.LogMessage = fmt.Sprintf("Pod %s/%s is stuck in ContainerCreating for more than 10 mins", pod.Name, pod.Namespace)
+			pvProgress.Status.ContainerElapsedTime = &metav1.Duration{Duration: time.Now().Sub(pod.CreationTimestamp.Time).Round(time.Second)}
+			//pvProgress.Status.SetCondition(migapi.Condition{
```
dymurray, line 293: Any reason we are keeping this commented out?

pkg/controller/directvolumemigration/task.go
```diff
@@ -452,6 +452,15 @@ func (t *Task) Run() error {
 				return liberr.Wrap(err)
 			}
 		} else {
+			//if pending {
+			//	t.Owner.Status.SetCondition(migapi.Condition{
```
dymurray, line 456: Why is this commented out?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1206,14 +1206,31 @@ func (t *Task) haveRsyncClientPodsCompletedOrFailed() (bool, bool, error) {
 					LastObservedProgressPercent: dvmp.Status.LastObservedProgressPercent,
 					LastObservedTransferRate:    dvmp.Status.LastObservedTransferRate,
 				})
+			case dvmp.Status.PodPhase == corev1.PodPending:
+				t.Owner.Status.PendingPods = append(t.Owner.Status.PendingPods, &migapi.PodProgress{
+					ObjectReference:             objRef,
+					LastObservedProgressPercent: dvmp.Status.LastObservedProgressPercent,
+					LastObservedTransferRate:    dvmp.Status.LastObservedTransferRate,
+				})
+				pendingMessage += dvmp.Status.LogMessage + " "
```
dymurray, line 1215: So this is grabbing the logs from the status field and setting it as a warning message in the status condition? is that right?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1166,11 +1166,11 @@ func getMD5Hash(s string) string {
 	return hex.EncodeToString(hash[:])
 }
 
-func (t *Task) haveRsyncClientPodsCompletedOrFailed() (bool, bool, error) {
+func (t *Task) haveRsyncClientPodsCompletedOrFailed() (bool, bool, bool, error) {
```
dymurray, line 1169: Consider renaming function. This is now returning 3 booleans so name doesn't make much sense any longer

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
JaydipGabani: Forgot to delete it

pkg/controller/directvolumemigration/task.go
JaydipGabani: Not needed removing

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: Yes! Since there could be multiple pods (each for one dvmp), that might be in the pending state for DVM, to me aggregating logMessage from each such dvmp and raising one warning in DVM makes more sense.

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: +1, updated the name

pkg/controller/directvolumemigration/rsync.go
dymurray: Okay, my only concern is wouldn't this mean if I have multiple pods that are pending it will look like one long log message but will really be multiple pods logs?

Could we separate the logs with a line indicating its a new pod? Or am I missing something?

JaydipGabani: good catch! starting new log from new line. There is only one logMessage per pod for this situation, so starting each from new line would make it less confusing.

dymurray (approved):
LGTM

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -276,6 +284,13 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
 		exitCode := containerStatus.LastTerminationState.Terminated.ExitCode
 		pvProgress.Status.ExitCode = &exitCode
 		pvProgress.Status.ContainerElapsedTime = &metav1.Duration{Duration: containerStatus.LastTerminationState.Terminated.FinishedAt.Sub(containerStatus.LastTerminationState.Terminated.StartedAt.Time).Round(time.Second)}
+	case !containerStatus.Ready && pod.Status.Phase == kapi.PodPending && containerStatus.State.Waiting != nil && containerStatus.State.Waiting.Reason == ContainerCreating:
+		// if pod is not in running state after 10 mins of its creation, raise a
+		if time.Now().Sub(pod.CreationTimestamp.Time) > TimeLimit {
```
alaypatel07, line 289: NIT: do we want `time.Now().UTC().Sub(pod.CreationTimestamp.Time.UTC())` instead of `time.Now().Sub(pod.CreationTimestamp.Time)` just to be sure the timezone is common?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1206,14 +1206,31 @@ func (t *Task) haveRsyncClientPodsCompletedOrFailed() (bool, bool, error) {
...
-	return isCompleted, hasAnyFailed, nil
+	isAnyPending := len(t.Owner.Status.PendingPods) > 0
+
+	if isAnyPending {
+		t.Owner.Status.SetCondition(migapi.Condition{
+			Type:     RsyncClientPodsPending,
+			Status:   migapi.True,
+			Reason:   "PodStuckInContainerCreating",
+			Category: migapi.Warn,
+			Message:  pendingMessage,
```
alaypatel07, line 1230: I wonder if there is some amount of redundancy here?

`Reason: PodStuckInContainerCreating` message: `fmt.Sprintf("Pod %s/%s is stuck in ContainerCreating for more than 10 mins\n", pod.Namespace, pod.Name)` 

I expect just the list of name of the pods in the message or something like 9/15 pods are stuck, something more general, less repetitive

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
JaydipGabani: +1

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: a message with `Name/namespace` of each pod that is in the pending state is more user friendly I think

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1206,14 +1206,32 @@ func (t *Task) haveRsyncClientPodsCompletedOrFailed() (bool, bool, error) {
 					LastObservedProgressPercent: dvmp.Status.LastObservedProgressPercent,
 					LastObservedTransferRate:    dvmp.Status.LastObservedTransferRate,
 				})
+			case dvmp.Status.PodPhase == corev1.PodPending:
+				t.Owner.Status.PendingPods = append(t.Owner.Status.PendingPods, &migapi.PodProgress{
+					ObjectReference:             objRef,
+					LastObservedProgressPercent: dvmp.Status.LastObservedProgressPercent,
+					LastObservedTransferRate:    dvmp.Status.LastObservedTransferRate,
+				})
+				pendingPods = append(pendingPods, strings.Split(dvmp.Status.LogMessage, " ")[1])
```
alaypatel07, line 1215: Failing to understand why this is not as simple as `obj.Name` and `object.Namespace`? What am I missing?

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -276,6 +284,13 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
 		exitCode := containerStatus.LastTerminationState.Terminated.ExitCode
 		pvProgress.Status.ExitCode = &exitCode
 		pvProgress.Status.ContainerElapsedTime = &metav1.Duration{Duration: containerStatus.LastTerminationState.Terminated.FinishedAt.Sub(containerStatus.LastTerminationState.Terminated.StartedAt.Time).Round(time.Second)}
+	case !containerStatus.Ready && pod.Status.Phase == kapi.PodPending && containerStatus.State.Waiting != nil && containerStatus.State.Waiting.Reason == ContainerCreating:
+		// if pod is not in running state after 10 mins of its creation, raise a
+		if time.Now().UTC().Sub(pod.CreationTimestamp.Time.UTC()) > TimeLimit {
+			pvProgress.Status.PodPhase = kapi.PodPending
+			pvProgress.Status.LogMessage = fmt.Sprintf("Pod %s/%s is stuck in ContainerCreating for more than 10 mins", pod.Namespace, pod.Name)
```
alaypatel07, line 291: with the above do we really need this?

pkg/controller/directvolumemigration/rsync.go
JaydipGabani: Ahh, missed that obj has these two pieces of info.

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
JaydipGabani: We don't, but there is no harm in keeping this information in dvmp as well.

pkg/controller/directvolumemigration/task.go
```diff
@@ -420,11 +420,11 @@ func (t *Task) Run() error {
 			return liberr.Wrap(err)
 		}
 	case WaitForRsyncClientPodsCompleted:
-		completed, failed, err := t.haveRsyncClientPodsCompletedOrFailed()
+		completed, failed, pending, err := t.haveRsyncClientPodsCompletedFailedOrPending()
 		if err != nil {
 			return liberr.Wrap(err)
 		}
-		if completed {
+		if completed && !pending {
```
alaypatel07, line 427: why do you need pending here? if the pods are in a pending state doesn't that automatically mean they are not completed?

JaydipGabani: Yip You are on the right track. removed the additional variable since it is not neeed

dymurray (approved):

PR #962, opened 2021-02-17
Dynamic Pipeline: Only add steps that will be part of migration

### LINKED ISSUE #934: [MIG-531] Dynamic pipelines
We've been interested in a "Dynamic Pipeline" rather than a "Statically Defined Pipeline" that we have today. This means building the pipeline of steps that will take place at runtime as a function of the state of the migplan and migration (Does this plan have pvs? Does it have hooks, and when are those hooks configured to run?). We should include the conditional steps that will happen, including hooks, and we should exclude anything that would otherwise be skipped.

pranavgaikwad:
This PR updates existing logic of building pipeline. 

Instead of adding all steps present in the itinerary to the pipeline, this PR only adds steps that will be actually part of a Migration in the pipeline. 

Fixes #934

### DIFF
--- a/pkg/controller/migmigration/task.go
+++ b/pkg/controller/migmigration/task.go
@@ -1029,7 +1029,10 @@ func (t *Task) init() error {
 
 	t.Step = t.Itinerary.GetStepForPhase(t.Phase)
 
-	t.initPipeline()
+	err := t.initPipeline(t.Owner.Status.Itinerary)
+	if err != nil {
+		return err
+	}
 
 	if t.stage() && !t.Owner.Status.HasCondition(StageNoOp) {
 		hasImageStreams, err := t.hasImageStreams()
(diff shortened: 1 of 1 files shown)


pkg/controller/migmigration/task.go
```diff
@@ -1052,12 +1055,32 @@ func (t *Task) init() error {
...
 }
 
-func (t *Task) initPipeline() {
-	for _, phase := range t.Itinerary.Phases {
-		t.Owner.Status.AddStep(&migapi.Step{
-			Name:    phase.Step,
-			Message: "Not started",
-		})
+func (t *Task) initPipeline(prevItinerary string) error {
+	if t.Itinerary.Name != prevItinerary {
```
djwhatle, line 1059: For anyone looking: any time we switch itineraries, the list of steps will be pre-constructed from the list of phases.

djwhatle (approved):
VISACK, LGTM

PR #963, opened 2021-02-18
Bug 1929853: Don't exit annotatePVs prematurely

sseago:
This commit removes some logic which incorrectly tries to determine
when we don't need to annotate any PVs or PVCs. Instead, rely on
the returned pvs list. If there aren't any returned, then the
for loop won't iterate, and exit naturally.

This fixes a regression introduced in https://github.com/konveyor/mig-controller/pull/937


djwhatle (approved):
VISACK

PR #965, opened 2021-02-22
PV resizing final set of changes 

pranavgaikwad:
This PR: 
- Wraps errors introduced in PV resizing changes
- Bumps number of concurrent reconciles for MigAnalytic from 1 to 2
- Refactors MigPlan Conditions for standardizing
- Uses `HasReconciled()` of MigPlan to determine whether MigAnalytic should be refreshed
- Adds warnings in MigPlan for cases when DVM PV resizing is enabled/disabled
- Adds a new flag to enable/disable PV resizing in DVM https://github.com/konveyor/mig-operator/pull/585
- Fix some edge cases around waiting on MigAnalytics
- Adds missing CRD changes in one of the previous PRs https://github.com/konveyor/mig-operator/pull/586


pkg/controller/migplan/migplan_controller.go
```diff
@@ -532,87 +529,100 @@ func (r ReconcileMigPlan) waitForMigAnalyticsReady(plan *migapi.MigPlan) (*migap
...
-	for _, migAnalytic := range migAnalytics.Items {
+	for i := range migAnalytics.Items {
+		migAnalytic := &migAnalytics.Items[i]
 		if migAnalytic.Spec.AnalyzeExtendedPVCapacity == true {
-			if migAnalytic.Status.IsReady() {
-				return &migAnalytic, nil
+			if migAnalytic.Status.IsReady() && !migAnalytic.Spec.Refresh {
+				return migAnalytic, nil
 			}
 			if time.Now().Sub(migAnalytic.CreationTimestamp.Time) > migAnalyticsTimeout {
```
djwhatle, line 540: This elapsed time measurement seems like it won't work properly for migAnalytics triggered after the initial one, since it uses the creation timestamp.

pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -49,7 +51,7 @@ func (t *Task) createDestinationPVCs() error {
 
 		// Adjusting destination PVC storage size request
 		// max(requested capacity on source, capacity reported in migplan, proposed capacity in migplan)
-		if matchingMigPlanPV != nil {
+		if matchingMigPlanPV != nil && settings.Settings.DvmOpts.EnablePVResizing {
```
djwhatle, line 54: Why do we have a switch for enabling PV resizing specifically for DVM? Maybe there's a good reason, I was kind of expecting a global on/off.

pkg/controller/miganalytic/restic_executor.go
```diff
@@ -64,7 +65,7 @@ func (r *ResticDFCommandExecutor) DF(podRef *corev1.Pod, persistentVolumes []Mig
 	err := podCommand.Run()
 	if err != nil {
 		log.Info(
-			fmt.Sprintf("df command inside pod %s returned non-zero error code", podRef.Name))
+			fmt.Sprintf("Failed running df command inside pod %s", podRef.Name))
```
djwhatle, line 68: Do we have any more detailed info we could log here? Err code, stderr, stdout? Or is there somewhere else we're putting this info

pkg/controller/migplan/migplan_controller.go
```diff
@@ -289,13 +289,11 @@ func (r *ReconcileMigPlan) Reconcile(request reconcile.Request) (reconcile.Resul
 		}
 
 		if migAnalytic != nil {
-			// Process PV Capacity
-			r.processExtendedPVCapacity(plan, migAnalytic)
-			// raise condition for unconfirmed PV sizes
-			r.validatePVSizeConfirmation(plan, migAnalytic)
```
djwhatle, line 295: I guess we don't need to validate PV size confirmation anymore?

pkg/controller/migplan/migplan_controller.go
```diff
@@ -532,87 +529,100 @@ func (r ReconcileMigPlan) waitForMigAnalyticsReady(plan *migapi.MigPlan) (*migap
...
 		if migAnalytic.Spec.AnalyzeExtendedPVCapacity == true {
-			if migAnalytic.Status.IsReady() {
-				return &migAnalytic, nil
+			if migAnalytic.Status.IsReady() && !migAnalytic.Spec.Refresh {
+				return migAnalytic, nil
 			}
 			if time.Now().Sub(migAnalytic.CreationTimestamp.Time) > migAnalyticsTimeout {
 				plan.Status.SetCondition(migapi.Condition{
-					Type:     IntelligentPVResizingDisabled,
+					Type:     PVResizingDisabled,
```
djwhatle, line 542: I think `PVUsageAnalysisFailed` or `PVUsageDetectionFailed` or `PVSizeDetectionFailed` would be a more appropriate name for this condition. When I first saw it, I thought this was related to whether the feature flag was set.

pkg/controller/migplan/migplan_controller.go
```diff
@@ -532,87 +529,100 @@ func (r ReconcileMigPlan) waitForMigAnalyticsReady(plan *migapi.MigPlan) (*migap
...
-						if planVol.Capacity.Cmp(analyticNSVol.ProposedCapacity) < 0 {
-							planVol.CapacityConfirmed = false
-						}
-
-						// If new proposed capacity is greater than original capacity, set confirmed to False
+						// If new proposed capacity is greater than previous proposed capacity, set confirmed to False
 						if planVol.ProposedCapacity.Cmp(analyticNSVol.ProposedCapacity) > 0 {
 							planVol.CapacityConfirmed = false
 						}
+						if analyticNSVol.Comment != miganalytic.VolumeAdjustmentNoOp {
```
djwhatle, line 570: What is a comment in this context?

pkg/settings/dvm.go
```diff
@@ -54,3 +61,14 @@ func (r *RsyncOpts) Load() error {
 	}
 	return err
 }
+
+// Load loads DVM options
+func (r *DvmOpts) Load() error {
+	var err error
+	r.EnablePVResizing = getEnvBool(EnablePVResizing, false)
```
djwhatle, line 68: This feels like it should be a Plan option to me, just because PV discovery happens on the plan. Feel free to tell me if you think otherwise.

pkg/controller/migplan/migplan_controller.go
pranavgaikwad: good catch !!

pkg/settings/dvm.go
pranavgaikwad: @djwhatle There are two kinds of flags here:
1. Enable/Disable discovering PV usage statistics **(Enabled by default)**
    -  When Set, it allows letting users know that the PVs will result in errors if they proceed to migrate without instructing DVM to use the corrected sizes
    - When Unset, it allows us to disable the feature, this is used as an escape hatch if we see issues with this feature in customer environment. We will remove this flag in future releases. 
    - this is not advertised until users run into any issues
2. Enable/Disable PV resizing in DVM controller **(Disabled by default)**
    - When Set, DVM controller will use the proposed sizes
    - When Unset, DVM controller will take no action on proposed sizes
    - This is the flag we will let users know about for 1.4.1, in case they have volumes resulting in warnings

pkg/controller/migplan/migplan_controller.go
pranavgaikwad: `Comment` is a string set by MigAnalytic controller about the proposed size. If it takes no action on the capacities, a no-op string is set.

pkg/controller/migplan/migplan_controller.go
pranavgaikwad: @djwhatle correct. we only need to set warnings for 1.4.1

pkg/controller/directvolumemigration/pvcs.go
JaydipGabani: We decided to decouple the analysis and actual resizing feature. @pranavgaikwad feel free to add more info or correct this.

pkg/controller/migplan/migplan_controller.go
JaydipGabani: +1

pkg/controller/directvolumemigration/pvcs.go
djwhatle: @pranavgaikwad so there is a separate setting on the plan? Or is this the only option

pranavgaikwad: @djwhatle see my comment here: https://github.com/konveyor/mig-controller/pull/965#discussion_r580533976

pkg/controller/migplan/migplan_controller.go
pranavgaikwad: @djwhatle Updated to use the `LastTransitionTime` on a Condition set by MigAnalytics indicating that it has started running DF for volumes

pkg/controller/miganalytic/restic_executor.go
pranavgaikwad: @djwhatle we do not have any other error info. This is the case when the command itself fails to run for some unknown reason. In that case, all PVCs present on this node will be skipped from PV resizing. The exact error is suppressed. However, if the command runs, we process command's StdErr and StdOut both to determine the error per PV.

JaydipGabani (approved):

pkg/controller/miganalytic/restic_executor.go
djwhatle: Yeah I can see that we have the stderr and stdout, that's kind of why I'm asking. How does the user see what went wrong here? Where do they look?

shubham-pampattiwar (approved):

pkg/controller/migplan/migplan_controller.go
djwhatle: Makes sense, +1

pkg/controller/miganalytic/restic_executor.go
pranavgaikwad: @djwhatle The name of the PVs for which the PV resizing failed are shown in a Warning condition on plan validation page. But the exact errors are not shown

pkg/controller/migplan/migplan_controller.go
djwhatle: @pranavgaikwad would `ProposedSize` be a more specific name for this then? Or is this a multi-use field.

pkg/settings/dvm.go
djwhatle: Thanks for the overview. I was thinking that if we were to enable this for indirect migration down the road, it would make more sense for the resize flag to exist on plan. If we get to that point we can always move it. Seems like we may never implement for indirect migration.

pkg/controller/migplan/migplan_controller.go
pranavgaikwad: @djwhatle `ProposedCapacity` is a field we are actually using to store the 'value' of corrected size and `Comment` is a 'type' of correction applied from the criteria described here: https://github.com/pranavgaikwad/mig-operator/blob/pvResizeDocs/docs/usage/IntelligentPVResizing.md#criteria

pkg/controller/migplan/migplan_controller.go
```diff
@@ -523,96 +518,109 @@ func (r ReconcileMigPlan) ensureMigAnalytics(plan *migapi.MigPlan) error {
...
 	err = r.Create(context.TODO(), pvMigAnalytics)
 	if err != nil {
-		return err
+		return liberr.Wrap(err)
 	}
 	return nil
 }
 
-func (r ReconcileMigPlan) waitForMigAnalyticsReady(plan *migapi.MigPlan) (*migapi.MigAnalytic, error) {
+func (r *ReconcileMigPlan) waitForMigAnalyticsReady(plan *migapi.MigPlan) (*migapi.MigAnalytic, error) {
```
djwhatle, line 526: This method name should reflect that we're _checking_ if MigAnalytics is ready, we aren't waiting and blocking.

pkg/settings/dvm.go
pranavgaikwad: @djwhatle Thats the plan. When we have UI change for toggling the option from MigPlan, we are planning to move this flag to plan level.

djwhatle (changes requested):
See comments

pkg/controller/migplan/migplan_controller.go
pranavgaikwad: @djwhatle fixed

pkg/controller/miganalytic/restic_executor.go
djwhatle: Some recent discussion with @shawn-hurley and @jortel has focused on need for more informative logs when things go wrong. This feels like somewhere we may want to give info about what went wrong if we have it. I'll leave to your judgement if there's anything reasonable we can give to the user.

pkg/controller/migplan/migplan_controller.go
djwhatle: Gotcha. Sounds good. Marking as resolved.

pkg/settings/dvm.go
djwhatle: +1, marking as resolved.

djwhatle (approved):
I am still wondering if there's any more detail we could provide on the logs (e.g. the stderr + stdout) when MigAnalytics fails. Trying to increase the context we give users with our log and give them breadcrumbs on where to investigate next. 

If you see any way to add more context easy, please do. Otherwise, PR LGTM!

pranavgaikwad:
@djwhatle I totally understand the concern here. The expected behaviour when error happens during execution of `df` is that we want to move on to next set of PVs and do nothing about the error. The motivation here is that we dont want to requeue/retry entire PV discovery stuff because `df` execution failed on one of the nodes. We want to back-off gracefully, moving onto next set of PVs. As a result, I only changed the `log.Info` call to `log.Error` allowing the actual error to get logged along with the name of the pod it occurred in.

PR #967, opened 2021-02-22
Add jaeger tracing

djwhatle:
Todo

- [x] Add tracing to remaining controllers that are part of migration process (dim, dism, dvmp)
- [x] Hide tracing behind a env var flag since it's not needed all the time, and may have perf impact
- [x] Make sure that spans terminate when migration ends
- [x] Evaluate if sampling frequency is OK as-is (decided this isn't a problem for now since only use-case will be for dev perf evaluation, not enabling by default in user clusters)

Follow-on PR:
- [ ] Figure out how to deploy a jaeger collector inside OCP 4 cluster, see if there's an official way to do this without us needing to package

To test:

Start the jaeger collector:
```
docker run -d --name jaeger \
  -e COLLECTOR_ZIPKIN_HTTP_PORT=9411 \
  -p 5775:5775/udp \
  -p 6831:6831/udp \
  -p 6832:6832/udp \
  -p 5778:5778 \
  -p 16686:16686 \
  -p 14268:14268 \
  -p 14250:14250 \
  -p 9411:9411 \
  jaegertracing/all-in-one:1.22
```
Navigate to http://localhost:16686

Run mig-controller with jaeger tracing enabled
```
JAEGER_ENABLED=true make run-fast
```

## Span hierarchy overview

```
- Parent MigMigration Span
  - Child MigMigration reconcile span
    - Child MigMigration phase execution span
  - Child DVM reconcile span
    - Child DVM phase execution span
  - Child DVMP reconcile span
  - Child DIM reconcile span
    - Child DIM phase execution span
  - Child DISM reconcile span
    - Child DISM phase execution span
```

Jaeger

## Preview of Jaeger tracing showing interacting controllers
image

## Some useful phase timing statistics get for free by having Jaeger running while a migration runs

image

### DIFF
--- a/pkg/apis/migration/v1alpha1/directimagemigration_types.go
+++ b/pkg/apis/migration/v1alpha1/directimagemigration_types.go
@@ -17,13 +17,16 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"context"
 	"fmt"
 	"path"
 	"strings"
 
+	liberr "github.com/konveyor/controller/pkg/error"
 	imagev1 "github.com/openshift/api/image/v1"
 	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -87,6 +90,20 @@ func (r *DirectImageMigration) GetSourceCluster(client k8sclient.Client) (*MigCl
 
 func (r *DirectImageMigration) GetDestinationCluster(client k8sclient.Client) (*MigCluster, error) {
 	return GetCluster(client, r.Spec.DestMigClusterRef)
+}
+
+// Get the MigMigration that owns this DirectImageMigration. If not owned, return nil.
+func (r *DirectImageMigration) GetOwner(client k8sclient.Client) (*MigMigration, error) {
+	owner := &MigMigration{}
+	ownerRefs := r.GetOwnerReferences()
+	if len(ownerRefs) > 0 {
+		ownerRef := types.NamespacedName{Name: ownerRefs[0].Name, Namespace: r.Namespace}
+		err := client.Get(context.TODO(), ownerRef, owner)
+		if err != nil {
+			return nil, liberr.Wrap(err)
+		}
+	}
+	return owner, nil
 }
 
 // GetSourceNamespaces get source namespaces without mapping
--- a/pkg/apis/migration/v1alpha1/directimagemigration_types.go
+++ b/pkg/apis/migration/v1alpha1/directimagemigration_types.go
@@ -93,17 +93,21 @@ func (r *DirectImageMigration) GetDestinationCluster(client k8sclient.Client) (*
 }
 
 // Get the MigMigration that owns this DirectImageMigration. If not owned, return nil.
-func (r *DirectImageMigration) GetOwner(client k8sclient.Client) (*MigMigration, error) {
+func (r *DirectImageMigration) GetMigrationForDIM(client k8sclient.Client) (*MigMigration, error) {
 	owner := &MigMigration{}
 	ownerRefs := r.GetOwnerReferences()
-	if len(ownerRefs) > 0 {
-		ownerRef := types.NamespacedName{Name: ownerRefs[0].Name, Namespace: r.Namespace}
+	for _, ownerRef := range ownerRefs {
+		if ownerRef.Kind != "MigMigration" {
+			continue
+		}
+		ownerRef := types.NamespacedName{Name: ownerRef.Name, Namespace: r.Namespace}
 		err := client.Get(context.TODO(), ownerRef, owner)
 		if err != nil {
 			return nil, liberr.Wrap(err)
 		}
+		return owner, nil
 	}
-	return owner, nil
+	return nil, nil
 }
 
 // GetSourceNamespaces get source namespaces without mapping
--- a/pkg/apis/migration/v1alpha1/directimagestreammigration_types.go
+++ b/pkg/apis/migration/v1alpha1/directimagestreammigration_types.go
@@ -17,11 +17,14 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"context"
 	"errors"
 
+	liberr "github.com/konveyor/controller/pkg/error"
 	imagev1 "github.com/openshift/api/image/v1"
 	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -74,6 +77,20 @@ func (r *DirectImageStreamMigration) GetSourceCluster(client k8sclient.Client) (
 
 func (r *DirectImageStreamMigration) GetDestinationCluster(client k8sclient.Client) (*MigCluster, error) {
 	return GetCluster(client, r.Spec.DestMigClusterRef)
+}
+
+// Get the DirectImageMigration that owns this DirectImageStreamMigration. If not owned, return nil.
+func (r *DirectImageStreamMigration) GetOwner(client k8sclient.Client) (*DirectImageMigration, error) {
+	owner := &DirectImageMigration{}
+	ownerRefs := r.GetOwnerReferences()
+	if len(ownerRefs) > 0 {
+		ownerRef := types.NamespacedName{Name: ownerRefs[0].Name, Namespace: r.Namespace}
+		err := client.Get(context.TODO(), ownerRef, owner)
+		if err != nil {
+			return nil, liberr.Wrap(err)
+		}
+	}
+	return owner, nil
 }
 
 func (r *DirectImageStreamMigration) GetImageStream(c k8sclient.Client) (*imagev1.ImageStream, error) {
--- a/pkg/apis/migration/v1alpha1/directimagestreammigration_types.go
+++ b/pkg/apis/migration/v1alpha1/directimagestreammigration_types.go
@@ -80,17 +80,21 @@ func (r *DirectImageStreamMigration) GetDestinationCluster(client k8sclient.Clie
 }
 
 // Get the DirectImageMigration that owns this DirectImageStreamMigration. If not owned, return nil.
-func (r *DirectImageStreamMigration) GetOwner(client k8sclient.Client) (*DirectImageMigration, error) {
+func (r *DirectImageStreamMigration) GetDIMforDISM(client k8sclient.Client) (*DirectImageMigration, error) {
 	owner := &DirectImageMigration{}
 	ownerRefs := r.GetOwnerReferences()
-	if len(ownerRefs) > 0 {
-		ownerRef := types.NamespacedName{Name: ownerRefs[0].Name, Namespace: r.Namespace}
+	for _, ownerRef := range ownerRefs {
+		if ownerRef.Kind != "DirectImageMigration" {
+			continue
+		}
+		ownerRef := types.NamespacedName{Name: ownerRef.Name, Namespace: r.Namespace}
 		err := client.Get(context.TODO(), ownerRef, owner)
 		if err != nil {
 			return nil, liberr.Wrap(err)
 		}
+		return owner, nil
 	}
-	return owner, nil
+	return nil, nil
 }
 
 func (r *DirectImageStreamMigration) GetImageStream(c k8sclient.Client) (*imagev1.ImageStream, error) {
--- a/pkg/controller/directimagemigration/trace.go
+++ b/pkg/controller/directimagemigration/trace.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2020 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package directimagemigration
+
+import (
+	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
+	"github.com/konveyor/mig-controller/pkg/settings"
+	migtrace "github.com/konveyor/mig-controller/pkg/tracing"
+	"github.com/opentracing/opentracing-go"
+)
+
+func (r *ReconcileDirectImageMigration) initTracer(dim *migapi.DirectImageMigration) opentracing.Span {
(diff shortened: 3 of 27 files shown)


djwhatle:
Squashed. Ready for review.

pkg/apis/migration/v1alpha1/directimagemigration_types.go
```diff
@@ -89,6 +92,20 @@ func (r *DirectImageMigration) GetDestinationCluster(client k8sclient.Client) (*
 	return GetCluster(client, r.Spec.DestMigClusterRef)
 }
 
+// Get the MigMigration that owns this DirectImageMigration. If not owned, return nil.
+func (r *DirectImageMigration) GetOwner(client k8sclient.Client) (*MigMigration, error) {
+	owner := &MigMigration{}
+	ownerRefs := r.GetOwnerReferences()
+	if len(ownerRefs) > 0 {
+		ownerRef := types.NamespacedName{Name: ownerRefs[0].Name, Namespace: r.Namespace}
```
shawn-hurley, line 100: why the first owner ref? are we not looking for a specific resource?

pkg/controller/directimagestreammigration/trace.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	// Set tracer on reconciler if it's not already present.
+	// We will never close this, so the 'closer' is discarded.
+	if r.tracer == nil {
+		r.tracer, _ = migtrace.InitJaeger("DirectImageStreamMigration")
+	}
+
+	// Go from dism -> dim -> migration, use migration UID to get span
+	dim, err := dism.GetOwner(r)
```
shawn-hurley, line 42: Can we not add the migration as an owner of dism? it probably makes sense in this case if we have the ability to adopt a resource.

Say a DIM is accidently deleted, the DISM's then in this case would stick around and get re-adopted when the migration controller re-creates the DIM.

shawn-hurley (changes requested):
I am wondering why we have to go through two layers to get the migration owner ref

pkg/controller/directimagestreammigration/trace.go
sseago: Currently the migration owns the dim, and the dim owns the dism. I guess we could add the migration as a second owner of the dism, although I haven't really experimented with two-owner resource behavior. If there are 2 owners and one gets deleted, what happens to the resource? Also, if the DIM gets deleted, I think we'd *want* any DISMs it creates to be deleted as well. Also, the DIM controller doesn't currently have any code which would adopt pre-existing DISM resources. If there is no DISM owned by *this* DIM, then it creates one.

pkg/apis/migration/v1alpha1/directimagemigration_types.go
sseago: I think we're doing this because we only ever create a single owner reference, so if there are more than one, then someone has manually hacked the resource into an unsupported scenario anyway. But yes, we should probably check the type of the owning resource to make sure it's what we expect. But even then, what do we do if we find more than one? Do we take the first, or to we error out? (this is a general question which might apply elsewhere. There are various cases where it is *possible* to find 2 resources in a slice but we only really expect one, so we generally take the first and ignore the rest. Perhaps we should always error out if we find more than one when we expect one?)

pkg/controller/directimagestreammigration/trace.go
sseago: One thing to keep in mind here is that it's possible for there to be a DIM that has no owner references. That just means that the user created the DIM directly rather than a migmigration creating it, which is essentially an images-only migration. Likewise with the DISM. It's possible for a single DISM to be created without an owning DIM, which will migrate a single imagestream.

pkg/apis/migration/v1alpha1/directimagemigration_types.go
```diff
@@ -89,6 +92,20 @@ func (r *DirectImageMigration) GetDestinationCluster(client k8sclient.Client) (*
...
+		err := client.Get(context.TODO(), ownerRef, owner)
+		if err != nil {
+			return nil, liberr.Wrap(err)
+		}
+	}
+	return owner, nil
```
sseago, line 106: Make sure that any code that deals with the returned owner acts sanely if the owner is an empty `MigMigration{}`, since there will be no owner for standalone DIM resoruces.

pkg/apis/migration/v1alpha1/directimagestreammigration_types.go
```diff
@@ -76,6 +79,20 @@ func (r *DirectImageStreamMigration) GetDestinationCluster(client k8sclient.Clie
...
+	owner := &DirectImageMigration{}
+	ownerRefs := r.GetOwnerReferences()
+	if len(ownerRefs) > 0 {
+		ownerRef := types.NamespacedName{Name: ownerRefs[0].Name, Namespace: r.Namespace}
+		err := client.Get(context.TODO(), ownerRef, owner)
+		if err != nil {
+			return nil, liberr.Wrap(err)
+		}
+	}
+	return owner, nil
```
sseago, line 93: Make sure that any code that deals with the returned owner acts sanely if the owner is an empty `DirectImageMigration{}`, since there will be no owner for standalone DISM resources.

pkg/controller/directimagestreammigration/trace.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		if errors.IsNotFound(errorutil.Unwrap(err)) {
+			return nil, nil
+		}
+		return nil, liberr.Wrap(err)
+	}
+	migration, err := dim.GetOwner(r)
```
sseago, line 49: DIM could be empty, if the DISM has no owner, so you should probably check that here. The GetOwner method returns an empty `DirectImageMigration{}`

pkg/controller/directimagestreammigration/trace.go
shawn-hurley: I see, we may just want to create a follow-up item to handle the re-adoption.

sseago: OK, so first, to clarify, a resource is only deleted if the *last* owner ref is deleted, not the first, right?

And, assuming this, if we adoped the approach of adding both `MigMigration` and `DIM` owner refs, then yes, when we added MigMigration owners to DISM resources, we'd need to modify the DIM controller's DISM creation code as follows:
If no DISM owned by this DIM is found, instead of immediately moving to creation of a new DISM, we'd have a second check looking for a DISM owned by the MigMIgration owner of this DIM (if it's owned by a MigMIgration), and if it's found, then we'd adopt the existing DISM by adding the current DIM as a second owner. If now found, follow through to the existing "create a new DISM" code.

pkg/controller/directimagemigration/trace.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (r *ReconcileDirectImageMigration) initTracer(dim *migapi.DirectImageMigration) opentracing.Span {
+	// Exit if tracing disabled
+	if !settings.Settings.JaegerOpts.Enabled {
+		return nil
+	}
+	// Set tracer on reconciler if it's not already present.
+	// We will never close this, so the 'closer' is discarded.
+	if r.tracer == nil {
+		r.tracer, _ = migtrace.InitJaeger("DirectImageMigration")
```
djwhatle, line 34: Looking at this again I feel like I actually should be passing the closer back so that the reconcile func can close out the tracer once the reconcile finishes.

pkg/apis/migration/v1alpha1/directimagemigration_types.go
djwhatle: +1 I'll check the type of the ownerRef. In the current system there is never more than one owner but since that could change, it would be good to support the future case.

pkg/controller/directimagestreammigration/trace.go
djwhatle: Sounds like a good follow-up item if this is something we're interested in changing. That said, it costs us next to nothing to traverse multiple levels here since all of these resources are coming from cache. The lookups are almost instant.

pkg/apis/migration/v1alpha1/directimagemigration_types.go
djwhatle: +1, will go make sure this is handled properly

pkg/controller/directimagestreammigration/trace.go
djwhatle: Good catch, will add a checks for empty resource getting returned.

djwhatle:
@shawn-hurley @sseago updated in response to PR review.

pkg/controller/directimagestreammigration/trace.go
djwhatle: Also made the GetOwner funcs actually return nil, was a mistake that I was returning an empty object.

pkg/controller/directimagemigration/trace.go
djwhatle: Looked again. I did this because we want to close out the spans, but never the overall tracer.

pkg/apis/migration/v1alpha1/directimagemigration_types.go
djwhatle: Fixed.

pkg/apis/migration/v1alpha1/directimagestreammigration_types.go
djwhatle: Fixed

djwhatle:
Hmm, latest update broke parenting of migmigration -> reconcile spans. Looking for fix.
image

djwhatle:
Fixed 
image

djwhatle:
@shawn-hurley @sseago there were some issues after I last pinged you guys, just wanted to let you know that I've resolved the issues now and this is ready for review again.

pkg/apis/migration/v1alpha1/directimagemigration_types.go
```diff
@@ -89,6 +92,24 @@ func (r *DirectImageMigration) GetDestinationCluster(client k8sclient.Client) (*
...
+	owner := &MigMigration{}
+	ownerRefs := r.GetOwnerReferences()
+	for _, ownerRef := range ownerRefs {
+		if ownerRef.Kind != "MigMigration" {
+			continue
+		}
+		ownerRef := types.NamespacedName{Name: ownerRef.Name, Namespace: r.Namespace}
+		err := client.Get(context.TODO(), ownerRef, owner)
+		if err != nil {
+			return nil, liberr.Wrap(err)
```
shawn-hurley, line 106: What happens here on a 500 error from the API server? what happens if there is a 404? 

What I would be worried about is a slow cache update I think  right?

pkg/controller/directvolumemigrationprogress/trace.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/konveyor/mig-controller/pkg/settings"
+	migtrace "github.com/konveyor/mig-controller/pkg/tracing"
+	"github.com/opentracing/opentracing-go"
+
+	liberr "github.com/konveyor/controller/pkg/error"
+	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/errors"
+)
+
+func (r *ReconcileDirectVolumeMigrationProgress) initTracer(dvmp migapi.DirectVolumeMigrationProgress) (opentracing.Span, error) {
```
shawn-hurley, line 30: I don't see where we handle the error from this function. I could just be missing it.

pkg/tracing/trace.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		panic(fmt.Sprintf("ERROR: cannot init Jaeger: %v\n", err))
+	}
+	return tracer, closer
+}
+
+// SetSpanForMigrationUID sets the parent jaeger span for a migration
+func SetSpanForMigrationUID(migrationUID string, span opentracing.Span) {
+	// Init map if needed
+	createMsmMapOnce.Do(func() {
```
shawn-hurley, line 63: Just a question,

What is the value of doing this on the first call rather than init?

pkg/tracing/trace.go
```diff
@@ -0,0 +1,118 @@
...
+// CloseMigrationSpan closes out the parent Migration Span safely
+func CloseMigrationSpan(migrationUID string) {
+	migrationSpan := GetSpanForMigrationUID(migrationUID)
+	if migrationSpan != nil {
+		// Immediately remove span from map so other controllers stop adding child spans
+		RemoveSpanForMigrationUID(migrationUID)
+		// Wait 5 minutes before terminating span, since other controllers writing to span
+		// post-close will result in undefined jaeger behavior (e.g. broken statistics page)
+		go func(migrationSpan opentracing.Span) {
+			time.Sleep(5 * time.Minute)
```
shawn-hurley, line 114: Is there a better way to note that other things are potentially writing to the span rather than relay on a timeout?

shawn-hurley (commented):
Some more questions might be good to merge with follow-ups

pkg/apis/migration/v1alpha1/directimagemigration_types.go
djwhatle: I don't believe we've had any notable problems with cache being out of date for the local cluster. The outdated cache issue primarily showed up as an issue communicating with remote clusters. All of our CRs are located on the same cluster as us.

The strategy for creating these spans is best-effort: if we can find the associated resources then we'll create a span. Otherwise we ignore it and the migration proceeds without jaeger spans. The jaeger span is not necessary for migration success, just helps us understand the system.

https://github.com/konveyor/mig-controller/pull/967/files/bc6c07c64755c415f2b1c8ae09acb4caca4e6f74#diff-c1e9e79d4cf358ba29d2e6ec3d769c82cd837a9269aae54f489504c64dbb672bR53-R58

pkg/tracing/trace.go
djwhatle: I thought about using a semaphore here or something similar, I think that would be a more robust solution. 

This was a sort of quick and easy hack that I think will be effective as long as no controller has a reconcile extending past 5 minutes. In general our reconciles are much shorter than this, with the exception of DirectImageMigration related things.

I actually wasn't sure why jaeger data corruption was happening, but I was able to confirm that this fixed the problem by making sure that the top-level migration span isn't written to post-close.

djwhatle: If you feel this is important to fix now I can definitely do it before merge.

pkg/tracing/trace.go
djwhatle: I think it could go either place. This just initializes the map if the function is ever used, and skips init if it isn't used. Think of all the memory we're saving /s

pkg/controller/directvolumemigrationprogress/trace.go
djwhatle: The error isn't handled. If this function fails then the reconcile span won't get set and no further child spans will be created under the reconcile span. This uses best-effort to create the span but ignores and continues if an error occurred.

pkg/tracing/trace.go
djwhatle: I will plan to fix this if we see any problems with the current approach in the future, or if we plan to turn Jaeger tracing on outside of dev environments. Filed an issue.

djwhatle: https://github.com/konveyor/mig-controller/issues/1023

PR #973, opened 2021-02-25
Add support for customizing moveable PVs via storageclass

dymurray:


sseago (approved):
ACK. I've tested the changes here.

PR #982, opened 2021-03-04
Bug 1925065: raise a warning when rsync pod fails and fail dvmp

JaydipGabani:
This PR raises a warning if rsync client pod are in failed state with no `containerStatus`.

### DIFF
--- a/pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
+++ b/pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
@@ -46,7 +46,7 @@ import (
 
 var log = logging.WithName("pvmigrationprogress")
 
-var TimeLimit = 10*time.Minute
+var TimeLimit = 10 * time.Minute
 
 const (
 	NotFound    = "NotFound"
@@ -247,14 +247,20 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
 	}
 
 	if containerStatus == nil {
-		pvProgress.Status.SetCondition(migapi.Condition{
-			Type:     InvalidPod,
-			Status:   migapi.True,
-			Reason:   NotFound,
-			Category: migapi.Critical,
-			Message: fmt.Sprintf("The spec.podRef %s must reference a `Pod` with container name %s",
-				path.Join(podRef.Namespace, podRef.Name), containerName),
-		})
+		if pod.Status.Phase == kapi.PodFailed {
+			pvProgress.Status.PodPhase = kapi.PodFailed
+			pvProgress.Status.LogMessage = pod.Status.Message
+			pvProgress.Status.ContainerElapsedTime = nil
+		} else {
+			pvProgress.Status.SetCondition(migapi.Condition{
+				Type:     InvalidPod,
+				Status:   migapi.True,
+				Reason:   NotFound,
+				Category: migapi.Critical,
+				Message: fmt.Sprintf("The spec.podRef %s must reference a `Pod` with container name %s",
+					path.Join(podRef.Namespace, podRef.Name), containerName),
+			})
+		}
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -247,14 +248,25 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
...
-			Status:   migapi.True,
-			Reason:   NotFound,
-			Category: migapi.Critical,
-			Message: fmt.Sprintf("The spec.podRef %s must reference a `Pod` with container name %s",
-				path.Join(podRef.Namespace, podRef.Name), containerName),
-		})
+		if pod.Status.Phase == kapi.PodFailed {
+			pvProgress.Status.PodPhase = kapi.PodFailed
+			pvProgress.Status.SetCondition(migapi.Condition{
+				Type:     PodFailed,
```
alaypatel07, line 254: why is this a condition an not in `pvProgress.Status.Phase`? 

if this is the condition, how would dvm controller pick this? and what would be the effect on migmigration?

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -247,14 +248,25 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
...
+				Status:   migapi.True,
+				Reason:   pod.Status.Reason,
+				Category: migapi.Warn,
+				Message:  fmt.Sprintf("The rsync pod %s is in %s state because of the following reason: %s", path.Join(pod.Name, pod.Namespace), pod.Status.Phase, pod.Status.Message),
```
pranavgaikwad, line 258: adding to Alay's question, since this `if` block will be only triggered when Pod phase is `kapi.PodFailed`, it doesnt make sense to me why we have a dynamic message here. This could as well be a static message saying - we found out that the pod is failed. There could also be scenarios where Status.Message is not set, in that case the message here will add to confusion

pranavgaikwad (changes requested):
Added a question

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
JaydipGabani: Nice catch! removing adding critical condition altogether as per alay's remark

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -247,14 +247,19 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
...
 	if containerStatus == nil {
-		pvProgress.Status.SetCondition(migapi.Condition{
-			Type:     InvalidPod,
-			Status:   migapi.True,
-			Reason:   NotFound,
-			Category: migapi.Critical,
-			Message: fmt.Sprintf("The spec.podRef %s must reference a `Pod` with container name %s",
-				path.Join(podRef.Namespace, podRef.Name), containerName),
-		})
+		if pod.Status.Phase == kapi.PodFailed {
```
pranavgaikwad, line 250: What happens when Pod phase is Unknown?

JaydipGabani: Not sure if that ever happens while `containerStatus` is nil, but if it does then the controller will raise critical condition.

pranavgaikwad (approved):
VISIACK

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -247,14 +247,19 @@ func (r *ReconcileDirectVolumeMigrationProgress) reportContainerStatus(pvProgres
...
+			pvProgress.Status.PodPhase = kapi.PodFailed
+			pvProgress.Status.LogMessage = pod.Status.Message
```
alaypatel07, line 252: do you want to add `pvProgress.Status.ContainerElapsedTime = nil` just to be sure that it is always nil?

alaypatel07 (commented):
Left a question, apart from there looks good to me, thanks for handling this

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
JaydipGabani: Good Idea!!

alaypatel07 (approved):
VISIACK

PR #985, opened 2021-03-04
CHERYPICK: catch no route to host as rsync failure condition #953

JaydipGabani:
carry over changes from #953


alaypatel07 (changes requested):
need the preserve the original commits to make sure it gets tracked properly. Synced up with Jaydip on slack

alaypatel07 (approved):
ACK

PR #990, opened 2021-03-08
Bug 1936549: Fix problem with dest cluster configmap missing not triggering version warning

djwhatle:
Reported by Sergio, migplan wasn't showing warning when host cluster was 3.x


pranavgaikwad (approved):
ACK, I think I understood what the issue was from this PR. Didn't look at the original BZ.

PR #992, opened 2021-03-08
Logs enhancement.

djwhatle:
Complement to https://github.com/konveyor/enhancements/pull/20, this PR fulfills one part of enhanced debug: logging what the controller is doing and why we aren't proceeding past the current point.

See below for examples of generated logs from migrations.


djwhatle:
@jortel @shawn-hurley I present to you... My proposal for logs in the form of two log files. `old-logs.txt` contains the logs from running an indirect migration on master. `new-logs.txt` contains the logs from running the same migration on this branch. 

I believe it should be pretty clear which log output is more informative for the user between the two, which I think we can all agree on. 

**Some areas of change I'd like to highlight:**
 - I customized zap logger so that our logs are finally aligned, not longer thrown off by variable precision floating point ms/ns epoch time (would like to somehow integrate this into std log library in the `konveyor/controller`)
 - In the `[RUN]` message that shows at the start of every reconcile we now show the extended phase description
 - Any time the controller stops and waits, I have provided a reason for it with as much context as is available
 - I have gone into each method called during the migration and added what I consider to be a healthy amount of logs that all contain enough breadcrumbs for the user to dig deeper into whatever is being logged about.
 - I have temporarily removed logging of conditions on each reconcile for migmigration controller, as I believe this to be a huge source of noise. My proposal is that we log any _deltas_ on conditions from reconcile start to end. Haven't implemented this yet.
 - I'm using fmt.Sprintf because I believe the log messages are more human-readable than using k/v pairs everywhere since you can inject context directly into your sentence, instead of context showing up much later on the line.

[new-logs.txt](https://github.com/konveyor/mig-controller/files/6118789/new-logs.txt)
[old-logs.txt](https://github.com/konveyor/mig-controller/files/6118790/old-logs.txt)

cmd/manager/main.go
```diff
@@ -38,7 +39,7 @@ import (
 )
 
 func main() {
-	logf.SetLogger(logf.ZapLogger(false))
+	logf.SetLogger(zapcustom.ZapLogger(false))
```
djwhatle, line 42: This is a temporary name

djwhatle:
- Structured logging conversion follow-on
 - Potentially introducing different log levels, deciding which types of conditions need to get logged.

pkg/controller/directvolumemigration/stunnel.go
```diff
@@ -578,6 +600,16 @@ func (t *Task) areStunnelClientPodsRunning() (bool, error) {
 		}
 		for _, pod := range pods.Items {
 			if pod.Status.Phase != corev1.PodRunning {
+				for _, podCond := range pod.Status.Conditions {
+					if podCond.Reason == corev1.PodReasonUnschedulable {
+						t.Log.Info(fmt.Sprintf("Found UNSCHEDULABLE Stunnel Client Pod [%v/%v] "+
```
djwhatle, line 605: Added this check that logs reason if pods are unschedulable. I ran into this a few times while testing out new logging.

djwhatle:
@jortel  Updated log samples (now includes step number in [RUN] message):

[dvm_log_latest.txt](https://github.com/konveyor/mig-controller/files/6142372/dvm_log_latest.txt)
[indirect_log_latest.txt](https://github.com/konveyor/mig-controller/files/6142373/indirect_log_latest.txt)

pkg/controller/directimagemigration/migrate.go
```diff
@@ -42,8 +44,11 @@ func (r *ReconcileDirectImageMigration) migrate(imageMigration *migapi.DirectIma
 	err := task.Run()
 	if err != nil {
```
shawn-hurley, line 45: Do we want to handle more errors here in the future?

pkg/controller/directimagemigration/task.go
```diff
@@ -277,3 +280,25 @@ func (t *Task) getDestinationClient() (compat.Client, error) {
...
 	return client, nil
 }
+
+// Get the extended phase description for a phase.
+func (t *Task) getPhaseDescription(phaseName string) string {
+	// Log the extended description of current phase
+	if phaseDescription, found := PhaseDescriptions[t.Phase]; found {
+		return phaseDescription
+	}
+	t.Log.V(4).Info("Missing phase description for phase: " + phaseName)
```
shawn-hurley, line 290: I almost want to suggest this should be an error. 

The only time I can see this happening is if 
1. unknown phase is set in the status, and this will error somewhere else as well right?
2. someone added a phase but did not add a description

for the 2nd use case making this loud by default might make some sense, as the first use case is already loud just somewhere else? WDYT?

pkg/controller/directimagemigration/task.go
```diff
@@ -277,3 +280,25 @@ func (t *Task) getDestinationClient() (compat.Client, error) {
...
+	// If no description available, just return phase name.
+	return phaseName
+}
+
+// Log the "[RUN] <Phase description>" phase kickoff string unless
+// DISMs are already logging a duplicate phase description.
+// This is meant to cut down on log noise when two controllers
+// are waiting on the same thing.
+func (t *Task) logRunHeader() {
+	if t.Phase != WaitingForDirectImageStreamMigrationsToComplete {
```
shawn-hurley, line 300: Is there some way we tell the user that we are still waiting on completion?

pkg/controller/directimagestreammigration/migrate.go
```diff
@@ -44,6 +46,8 @@ func (r *ReconcileDirectImageStreamMigration) migrate(imageStreamMigration *miga
 		if errors.IsConflict(errorutil.Unwrap(err)) {
```
shawn-hurley, line 46: Same question, should we create an issue for this?

pkg/controller/directvolumemigration/descriptions.go
```diff
@@ -1,13 +1,29 @@
+/*
+Copyright 2021 Red Hat Inc.
```
shawn-hurley, line 2: +1

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1291,11 +1350,15 @@ func (t *Task) deleteRsyncResources() error {
...
 	}
 
+	t.Log.Info(fmt.Sprintf("Checking for stale Rsync resources on source MigCluster [%v/%v]",
+		t.PlanResources.SrcMigCluster.Namespace, t.PlanResources.SrcMigCluster.Name))
 	err = t.findAndDeleteResources(srcClient)
 	if err != nil {
 		return err
 	}
 
+	t.Log.Info(fmt.Sprintf("Checking for stale Rsync resources on destination MigCluster [%v/%v]",
```
shawn-hurley, line 1360: For my own edification, what makes a resource stale?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1543,9 +1636,11 @@ func (t *Task) deleteProgressReportingCRs(client k8sclient.Client) error {
 
 	for ns, vols := range pvcMap {
 		for _, vol := range vols {
+			dvmpName := getMD5Hash(t.Owner.Name + vol.Name + ns)
+			t.Log.Info(fmt.Sprintf("DELETING stale DVMP CR [%v/%v]", dvmpName, ns))
 			err := client.Delete(context.TODO(), &migapi.DirectVolumeMigrationProgress{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      getMD5Hash(t.Owner.Name + vol.Name + ns),
+					Name:      dvmpName,
```
shawn-hurley, line 1643: Just as future implementation note, I prefer meaningful names using `generate-name` to get uniqueness. 

The reason being a user can then hopefully tell what that thing is for rather than just some random pods getting spun up.

pkg/controller/migmigration/backup.go
```diff
@@ -614,13 +640,20 @@ func (t *Task) deleteStaleBackupsOnCluster(cluster *migapi.MigCluster) (int, int
...
+				"Status.Phase=[%v] is not 'New' or 'InProgress'. Skipping deletion.",
+				backup.Namespace, backup.Name, backup.Status.Phase))
 			continue
 		}
 		// Skip if missing a migmigration correlation label (only delete our own CRs)
 		// Example 'migmigration: 4c9d317f-f410-430b-af8f-4ecd7d17a7de'
 		corrKey, _ := t.Owner.GetCorrelationLabel()
 		migMigrationUID, ok := backup.ObjectMeta.Labels[corrKey]
 		if !ok {
+			t.Log.V(4).Info(fmt.Sprintf("Backup [%v/%v] with "+
```
shawn-hurley, line 653: +1 on v4 here

pkg/controller/migmigration/quiesce.go
```diff
@@ -131,6 +134,12 @@ func (t *Task) quiesceDeploymentConfigs(client k8sclient.Client) error {
 				continue
 			}
 			dc.Annotations[ReplicasAnnotation] = strconv.FormatInt(int64(dc.Spec.Replicas), 10)
+			t.Log.Info(fmt.Sprintf("Quiescing DeploymentConfig [%v/%v]. "+
```
shawn-hurley, line 137: I think we may want to file an issue for this, but a call to the scale subresource may be better than updating each individual GVK and will capture CRD's that have the scale subresource mapped to a value so we can scale down those applications as well. 

We could use discovery API to determine if a given GVK has a scale subresource.

pkg/controller/migmigration/restore.go
```diff
@@ -523,13 +544,20 @@ func (t *Task) deleteStalePVRsOnCluster(cluster *migapi.MigCluster) (int, error)
...
+				"Status.Phase=[%v] is not 'New' or 'InProgress'. Skipping deletion.",
+				pvr.Namespace, pvr.Name, pvr.Status.Phase))
 			continue
 		}
 
 		// Skip delete if PVR is associated with running migration
 		pvrHasRunningMigration := false
 		for _, ownerRef := range pvr.OwnerReferences {
 			if ownerRef.Kind != "Restore" {
+				t.Log.V(4).Info(fmt.Sprintf("PodVolumeRestore [%v/%v] with "+
```
shawn-hurley, line 557: +1 on v4 here

pkg/controller/migplan/validation.go
```diff
@@ -280,6 +280,10 @@ func (r ReconcileMigPlan) validateNamespaces(plan *migapi.MigPlan) error {
 
 func (r ReconcileMigPlan) validateNamespaceLengthForDVM(plan *migapi.MigPlan) []string {
 	items := []string{}
+	// If validation for destination cluster ref failed, we can't check this
```
shawn-hurley, line 283: This does not appear to be a log change, but rather a bugfix is that true?

shawn-hurley (approved):
It mostly looks good, just some comments on some potential follow-up items.

I would like to see if the change to migplan validation is a bugfix or not; if it is a bugfix, I would think creating a bug and allowing QE to verify might be the better approach than adding it in the log PR WDYT?

pkg/controller/directimagemigration/task.go
djwhatle: Yeah, the DVM controller or DIM controller will be logging a message every 3 seconds stating that they're waiting. This is just meant to reduce the log clutter (otherwise both controllers write the same log message every 3 seconds)

pkg/controller/migplan/validation.go
djwhatle: Oops I forgot to remove this here. I submitted a bugfix PR that just got merged to master separately from this PR https://github.com/konveyor/mig-controller/pull/1000

pkg/controller/migmigration/quiesce.go
djwhatle: Considering we also need to issue an UPDATE to add an annotation with old replica count at the same time, is there still improvement to be had from changing this to using scale? We currently do:

1. Replica count change
2. Annotation add

In a single reconcile.

pkg/controller/directimagemigration/task.go
djwhatle: +1, My intent was to help resolve the thing you mention in `2. someone added a phase but did not add a description
`.  I found a few phases missing descriptions and wanted the controller to report that so we can fix it as you mentioned.

I don't believe there is any way for us to end up in an undefined phase (your concern in `1.`) unless someone manually sets the phase. This would be bad obviously, I don't think it's likely though given that Phase is just advanced by `t.next()` according to itinerary contents.

I'm torn on how loud to make this. My first instinct was to make it an error as well, but then I became concerned that if I don't manage to run migrations that hit every possible phase before the next z-stream release, we'll end up possibly shipping a z-stream that shoots out error messages (with stack traces etc) on every migration.

pkg/controller/directimagemigration/migrate.go
djwhatle: I think a lot of errors that are non-fatal are caught early within phase logic and not passed back to this point. If an error makes it here, it's considered critical and fails the migration. 

Did you have other types of errors in mind? Want to make sure I've got your point.

pkg/controller/directimagestreammigration/migrate.go
djwhatle: Would like to understand what other kind of errors we would want to handle here, see https://github.com/konveyor/mig-controller/pull/992#discussion_r595439193

pkg/controller/directvolumemigration/rsync.go
djwhatle: This is a sort of generic term for resources that are either:

1) Leftover from a previous migration where we exited unexpectedly and weren't able to clean up
2) Leftover from a migration that ran successfully, and the resources (rsync pods, secrets, etc) that are no longer needed are considered "stale"

pkg/controller/directvolumemigration/rsync.go
djwhatle: +1. I recall that in some places, we started with `generateName` and switched to deterministic names so that if a CREATE call failed and we had to exit the reconcile, we could re-discover anything we've successfully created so far and not end up with duplicate resources upon second reconcile attempt.

This was important when we weren't using ownerRefs, but since I think DVMP stuff uses ownerRefs we could probably switch back to using generateName in some places.

djwhatle:
Going ahead with the merge, planning to take on requests for further enhancement in future PRs since this one is already huge. 🚢

pkg/controller/directimagemigration/migrate.go
shawn-hurley: I am pretty sure that is not the case in actuality, (i.e NotFound errors or Service Errors from Kube)

pkg/controller/directimagemigration/task.go
shawn-hurley: I think I am ok with the compromise and a follow-up item to come back and make them errors in the future?

pkg/controller/directvolumemigration/rsync.go
shawn-hurley: So we can never adopt a resource that is pre-created? 

that is fine, just making sure I understand :)

pkg/controller/directvolumemigration/rsync.go
shawn-hurley: yeah, a label and an owner ref comparison might do the trick?

pkg/controller/migmigration/quiesce.go
shawn-hurley: Why do we have to set that annotation?  sorry if I am missing it.

pkg/controller/directimagemigration/task.go
djwhatle: Sounds good. I think once we shore up the gaps in missing labels it would make sense to bump this up to error.

pkg/controller/directvolumemigration/rsync.go
djwhatle: I think anything that we are considering "stale" is something we explicitly don't want to leave sitting around for some reason, but I get your point that we should re-use resources where possible if there aren't resource issues associated with doing so.

pkg/controller/migmigration/quiesce.go
djwhatle: We need to keep track of which resources we've quiesced and their previous replica counts so that we can:
1. Come back later and unquiesce to previous replica count them as part of usual migration routine, since the resources land on destination in quiesced
2. Unquiesce source cluster to appropriate replica count on rollback

PR #997, opened 2021-03-10
Make rollback deletion routine more resilient to incorrect initial deletion ordering

djwhatle:
**Changes:**
- Removed phase `DeleteMigrated`, combined into phase `EnsureMigratedDeleted`
- Phase `EnsureMigratedDeleted` now deletes **and** checks for termination of resources

`EnsureMigratedDeleted` is used in rollback and I've seen it get stuck every time I've tried to use it on the rocketchat sample app today (about 10 times in a row).  There is a stubborn PVC that gets stuck in "Terminating" due to a Pod that is "Completed", I'm not sure how this is possible but deleting the completed Pods fixes it.

Possible additions:
 - Limited Retry. I need to look and see how retry is implemented in other parts of the controller. I don't think we want this rollback routine to run forever since it's basically just hammering the API server in a loop. That said, we are doing this forever-retry routine on the "Watch for Deletion" part today, this PR just make a forever-loop of "Delete and Watch for Deletion"


djwhatle:
This didn't fix the issue.

PR #1001, opened 2021-03-16
resticting capabilities in pod scc field

JaydipGabani:
Adding in drop capabilities list of `pod.container.SCC` to make sure openshift always picks `rsync-anyuid`

Below is the result of file permissions after migration with modified `rsync-anyuid`

```
destination 

$ ls -la /data/db1
total 364
drwxrwxrwx. 5 root       root  8192 Mar 24 15:06 .
drwxr-xr-x. 3 root       root    17 Mar 24 15:06 ..
-rw-r--r--. 1 1000990002 root 18502 Mar 24 15:03 1.html
-rw-r--r--. 1 1000990004 root 18473 Mar 24 15:03 2.html
-


source 

$ ls -l data/db1
total 340
-rw-r--r--. 1 1000990002 root 25984 Mar 24 15:07 1.html
-rw-r--r--. 1 1000990004 root 25955 Mar 24 15:07 2.html
```

The uids on the files are getting preserved along with permissions.

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -600,6 +600,9 @@ func (t *Task) createRsyncTransferPods() error {
 							Privileged:             &isRsyncPrivileged,
 							RunAsUser:              &runAsUser,
 							ReadOnlyRootFilesystem: &trueBool,
+							Capabilities: &corev1.Capabilities{
+								Drop: []corev1.Capability{"MKNOD", "SETPCAP"},
+							},
 						},
 						Resources: corev1.ResourceRequirements{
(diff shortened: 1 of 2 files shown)


alaypatel07 (commented):
Working with Jaydip here, this was tested in gp2->gp2 migration, the file permissions were intact. I asked Jaydip to create a PVC in glusterfs with 2 containers having different gid/uid combinations and try to migrate that over. Once that is tested, this is ready to be merged.

alaypatel07 (approved):
Thanks for verifying on both gp2 and glusterfs with fsGroup and supplemental groups respectively. LGTM

PR #1006, opened 2021-03-17
Log phase completion times for perf profiling

djwhatle:
```
{"level":"info","ts":[...],"logger":"[...]","msg":"[RUN] (Step 9/23) Checking whether the created PVCs are bound","DVM":[...]}
{"level":"info","ts":[...],"logger":"[...]","msg":"Phase completed",[...],"Phase": "StartRefresh", "phaseElapsed":0.522302}
```

Logs formatted in this way allows us to use a nasty bash one-liner to generate a performance report of which phases are taking the longest to run.

```
make run-fast 2>&1 | tee controller.log

echo "PhaseName ElapsedSeconds PercentOfTotal" > phase_timing.csv; total_time=$(cat controller.log | grep '\"Phase\ completed\"' | jq '([inputs | .phaseElapsed] | add)'); cat controller.log | grep '\"Phase\ completed\"' | jq -r --arg total_time "$total_time" '"\(.Phase) \(.phaseElapsed) \(.Elapsed / ($total_time|tonumber))"' >> phase_timing.csv
```

```
PhaseName ElapsedTime PercentOfTotal                
StartRefresh 49.691784 14.405797870070389    
WaitForRefresh 17.651623 5.117258680362241     
CleanStaleAnnotations 2.386096 0.6917364181286685
CleanStaleResticCRs 3.936635 1.1412423449768787    
CleanStaleResticCRs 7.816086 2.265906876096197     
CleanStaleVeleroCRs 7.983351 2.314397503454984                                                             
RestartVelero 1.007199 0.2919900241242501                                                                  
CleanStaleStagePods 1.669287 0.4839313297573738   
WaitForStaleStagePodsTerminated 2.195697 0.6365391744824445
WaitForVeleroReady 0.843389 0.24450101167309263                                                            
EnsureCloudSecretPropagated 5.634364 1.6334190962111823
PreBackupHooks 0.762824 0.2211449754840473       
CreateDirectImageMigration 0.912454 0.264523163220246      
EnsureInitialBackup 3.609576 1.0464269556644856 
Started 2.006556 0.5817066288257422            
EnsureInitialBackup 4.150545 1.2032554983461916                                                            
Prepare 1.541324 0.4468344705895118                                                                        
CreateDestinationNamespaces 3.497918 1.0140569651128017  
ListImageStreams 2.87607 0.8337813566961764              
```

And from this CSV we can easily create a Google Sheet with color formatting that highlights which phases are taking the longest.

image

### DIFF
--- a/pkg/controller/directimagemigration/task.go
+++ b/pkg/controller/directimagemigration/task.go
@@ -207,6 +207,15 @@ func (t *Task) Run() error {
 
 // Advance the task to the next phase.
 func (t *Task) next() error {
+	// Write time taken to complete phase
+	t.Owner.Status.StageCondition(migapi.Running)
+	cond := t.Owner.Status.FindCondition(migapi.Running)
+	if cond != nil {
+		now := time.Now().UTC()
+		elapsed := now.Sub(cond.LastTransitionTime.Time.UTC())
+		t.Log.Info("Phase completed", "Phase_Elapsed", elapsed)
+	}
+
 	current := -1
 	for i, step := range t.Itinerary.Steps {
 		if step.phase != t.Phase {
--- a/pkg/controller/directimagemigration/task.go
+++ b/pkg/controller/directimagemigration/task.go
@@ -215,7 +215,7 @@ func (t *Task) next() error {
 	if cond != nil {
 		now := time.Now().UTC()
 		elapsed := now.Sub(cond.LastTransitionTime.Time.UTC())
-		t.Log.Info("Phase completed", "Elapsed", elapsed)
+		t.Log.Info("[NEXT]", "Elapsed", elapsed)
 	}
 
 	current := -1
--- a/pkg/controller/directimagemigration/task.go
+++ b/pkg/controller/directimagemigration/task.go
@@ -214,7 +214,7 @@ func (t *Task) next() error {
 	cond := t.Owner.Status.FindCondition(migapi.Running)
 	if cond != nil {
 		elapsed := time.Since(cond.LastTransitionTime.Time)
-		t.Log.Info("Phase completed", "Elapsed", elapsed)
+		t.Log.Info("Phase completed", "phaseElapsed", elapsed)
 	}
 
 	current := -1
--- a/pkg/controller/directimagemigration/task.go
+++ b/pkg/controller/directimagemigration/task.go
@@ -127,6 +127,8 @@ func (t *Task) init() error {
 }
 
 func (t *Task) Run() error {
+	t.Log = t.Log.WithValues("Phase", t.Phase)
+
 	// Init
 	err := t.init()
 	if err != nil {
@@ -213,7 +215,7 @@ func (t *Task) next() error {
 	if cond != nil {
 		now := time.Now().UTC()
 		elapsed := now.Sub(cond.LastTransitionTime.Time.UTC())
-		t.Log.Info("Phase completed", "Phase_Elapsed", elapsed)
+		t.Log.Info("Phase completed", "Elapsed", elapsed)
 	}
 
 	current := -1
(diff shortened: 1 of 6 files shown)


pkg/controller/directimagemigration/task.go
```diff
@@ -207,6 +209,15 @@ func (t *Task) Run() error {
 
 // Advance the task to the next phase.
 func (t *Task) next() error {
+	// Write time taken to complete phase
+	t.Owner.Status.StageCondition(migapi.Running)
+	cond := t.Owner.Status.FindCondition(migapi.Running)
+	if cond != nil {
+		now := time.Now().UTC()
+		elapsed := now.Sub(cond.LastTransitionTime.Time.UTC())
```
alaypatel07, line 217: Minor NIT: you can probably replace this with
```
elapsed := time.Since(cond.LastTransitionTime.Time)
```

From what I have read, this pattern is known to produce fewer bugs. I have been guilty of not using it as well though. Also, I am not sure how to handle UTC but just spraying thoughts if you want to play with it.

pkg/controller/directimagemigration/task.go
```diff
@@ -207,6 +209,15 @@ func (t *Task) Run() error {
...
+		t.Log.Info("[NEXT]", "Elapsed", elapsed)
```
alaypatel07, line 218: I wonder if instead of `[NEXT]` the `[COMPLETE]` would make sense?

Just trying to think of how the log would read to a normal user no aware of why this is for? [COMPLETE] with a phase name and time will give them a good idea of why this log entry exists, I guess.

pkg/controller/directimagemigration/task.go
djwhatle: ACK. I saw DVM controller was using UTC in its elapsed time calculation and just copied that assuming there was a reason for the UTC conversion.

pkg/controller/directimagemigration/task.go
djwhatle: Since `[COMPLETED]` is already used when the entire migration completes I felt it was too similar, although that was my first intuition as well. 

For some context, the task logger has the a KV `"Phase": "<PhaseName>"` attached. 

I was considering some alternatives like `[PHASECOMPLETE]` but that seemed maybe a bit long? WDYT

djwhatle: I guess I should note we could just do a message like this
```
{"level":"info","ts":[...],"logger":"[...]","msg":"Phase completed.",[...],"Phase": "StartRefresh", "Elapsed":0.522302}
```

pkg/controller/directimagemigration/task.go
djwhatle: Changed

pkg/controller/directimagemigration/task.go
djwhatle: Changed to 
```
{"level":"info","ts":[...],"logger":"[...]","msg":"Phase completed",[...],"Phase": "StartRefresh", "Elapsed":0.522302}

djwhatle:
@alaypatel07 updated, PTAL

pkg/controller/directimagemigration/task.go
alaypatel07: the latest change works with me, thanks @djwhatle

alaypatel07 (approved):
LGTM, thanks

shawn-hurley:
So this is not for Jaeger for I assume gather this output in a must gather and being able to get this performance report with that awesomeness that is a bash one liner! 

What happens if we requeue a phase?

djwhatle:
> So this is not for Jaeger for I assume gather this output in a must gather and being able to get this performance report with that awesomeness that is a bash one liner!

Yeah I was mainly intending this to be used for local dev to easily turn logs into a perf profile but since must-gather collects logs we can just as easily use this to understand perf on the user clusters.

PR #1009, opened 2021-03-18
API Changes for DVM retry mechanism

pranavgaikwad:
This PR introduces API changes in DVM and DVMP for DVM retry mechanism. 

Please review operator PR : https://github.com/konveyor/mig-operator/pull/617


pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
```diff
@@ -27,20 +27,65 @@ import (
...
-	ObservedDigest              string           `json:"observedDigest,omitempty"`
-	LastObservedProgressPercent string           `json:"lastObservedProgressPercent,omitempty"`
-	LastObservedTransferRate    string           `json:"lastObservedTransferRate,omitempty"`
+	Conditions `json:",omitempty"`
+	// RsyncPodStatus observed state of most recent Rsync attempt
+	RsyncPodStatus `json:",inline"`
+	// RsyncPodStatuses history of all Rsync attempts
+	RsyncPodStatuses []RsyncPodStatus `json:"rsyncPodStatuses,omitempty"`
+	// Succeded whether Rsync operation succeded
+	Succeded bool `json:"succeded,omitempty"`
```
djwhatle, line 49: Spelling ➡️  `succeeded`

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
```diff
@@ -27,20 +27,65 @@ import (
...
 // DirectVolumeMigrationProgressStatus defines the observed state of DirectVolumeMigrationProgress
 type DirectVolumeMigrationProgressStatus struct {
-	Conditions                  `json:",omitempty"`
-	PodPhase                    kapi.PodPhase    `json:"phase,omitempty"`
-	ExitCode                    *int32           `json:"exitCode,omitempty"`
-	ContainerElapsedTime        *metav1.Duration `json:"containerElapsedTime,omitempty"`
-	LogMessage                  string           `json:"logMessage,omitempty"`
-	ObservedDigest              string           `json:"observedDigest,omitempty"`
-	LastObservedProgressPercent string           `json:"lastObservedProgressPercent,omitempty"`
-	LastObservedTransferRate    string           `json:"lastObservedTransferRate,omitempty"`
```
djwhatle, line 43: Will mig-operator be able to gracefully handle removal of this field? I wonder if we want to avoid removing this, am not sure if it will cause problems. Thinking about clusters where DVMPs already exist.

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -17,11 +17,19 @@ limitations under the License.
...
+
 	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// labels
+const (
+	// RsyncPodIdentityLabel identifies sibling Rsync attempts/pods
+	RsyncPodIdentityLabel = "openshift.migration.io/created-for-pvc"
```
djwhatle, line 30: migration.openshift.io

pranavgaikwad: :facepalm: :facepalm:

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
pranavgaikwad: :facepalm:

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
pranavgaikwad: there is also a top level reference to this field in the same struct (used to store status of most recent Rsync attempt). Its just that I encapsulated the overall status in another struct to make it easier to keep history

shawn-hurley: Consider the API is rolled back case, I think you need to keep both for $SOME_PERIOD_OF_TIME

pranavgaikwad: @shawn-hurley I have not removed any fields. It's just that these fields will now go into a new struct and that struct is also included "inline" in the Sptatus struct. See line 45, so these new Status fields are compatible with both old and new versions

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
pranavgaikwad: @djwhatle fixed

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
pranavgaikwad: fixed

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
shawn-hurley: From a JSON perspective it might be fine, what about a user using our APIs from golang?

pranavgaikwad: Right, upon upgrade/downgrade, API users attempting to create/update a new DVMPStatus will have to create typed instances of RsyncPodStatus or remove typed status, respectively. Read operations can directly happen through inline field without any change. However, at this point, this CR is only managed by migmigration and controller will take right actions in both old and new version. I wouldn't expect anyone consuming DVM api directly as we don't support running DVM alone.

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
```diff
@@ -27,20 +27,65 @@ import (
...
+	// RsyncPodStatus observed state of most recent Rsync attempt
+	RsyncPodStatus `json:",inline"`
+	// RsyncPodStatuses history of all Rsync attempts
+	RsyncPodStatuses []RsyncPodStatus `json:"rsyncPodStatuses,omitempty"`
+	// Succeeded whether Rsync operation succeded
+	Succeeded bool `json:"succeeded,omitempty"`
+	// Failed whether Rsync operation failed
+	Failed bool `json:"failed,omitempty"`
+	// Completed wether Rsync operation completed
+	Completed bool `json:"completed,omitempty"`
```
alaypatel07, line 53: It looks like this field is a derived field right? you can have competed as `Failed | Succeeded` ?

Is this intended to show whether it is running, if so would it make sense to have `Running` instead of `Completed`?

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
```diff
@@ -27,20 +27,65 @@ import (
...
+	// RsyncElapsedTime total elapsed time of Rsync operation
+	RsyncElapsedTime *metav1.Duration `json:"rsyncElapsedTime,omitempty"`
```
alaypatel07, line 55: Will this be the addition of all the container's duration? or will it be the time elapsed since the start of this CR?

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -34,6 +40,12 @@ type DirectVolumeMigrationSpec struct {
 	SrcMigClusterRef  *kapi.ObjectReference `json:"srcMigClusterRef,omitempty"`
 	DestMigClusterRef *kapi.ObjectReference `json:"destMigClusterRef,omitempty"`
 
+	// BackOffLimit retry limit on Rsync pods
+	BackOffLimit int `json:"backOffLimit,omitempty"`
+
+	// RsyncPodGarbageCollection whether to clean up failed pods during retry
+	RsyncPodGarbageCollection bool `json:"rsyncPodGarbageCollection,omitempty"`
```
alaypatel07, line 47: I wonder if this needs to be exposed in the API?

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
pranavgaikwad: @alaypatel07 that's a good point. Earlier when I worked on API, I didn't realize that these fields were creating an unnecessary contract between DVM -> DVMP, where DVM would require DVMP to know the information. However, after discussing with you, it was clear that we should keep DVMP dumb enough and only provide "current" view of Rsync pods in the source cluster. I ended up removing these fields altogether.

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
pranavgaikwad: @alaypatel07 It will be addition of all container durations

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
pranavgaikwad: @alaypatel07 Keeping in the api in an effort to reduce the dependencies (thinking about state only transfers), open to moving this around or moving to MigrationController spec instead

alaypatel07: @pranavgaikwad is there any value in allowing users to configure this? Can we just decide for the user that we will garbage collect pods by default or may be have a global setting for this at migrationcontroller CR level? It feels to me this is increasing the fields in API which we don't necessarily absolutely need.

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
alaypatel07: can you please update the comment in the previous line to include this?

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -17,11 +17,19 @@ limitations under the License.
...
+
 	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// labels
+const (
+	// RsyncPodIdentityLabel identifies sibling Rsync attempts/pods
+	RsyncPodIdentityLabel = "migration.openshift.io/created-for-pvc"
```
alaypatel07, line 30: I am not sure how you intended to use it..  but it looks to me that these are for rsync-client pods on the source side. If that is the case do you think the following will be more readable

```
RsyncClientPodLabel = "migration.openshift.io/rsync-client-pod"
```

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
```diff
@@ -27,20 +27,58 @@ import (
 
 // DirectVolumeMigrationProgressSpec defines the desired state of DirectVolumeMigrationProgress
 type DirectVolumeMigrationProgressSpec struct {
-	ClusterRef *kapi.ObjectReference `json:"clusterRef,omitempty"`
-	PodRef     *kapi.ObjectReference `json:"podRef,omitempty"`
+	ClusterRef  *kapi.ObjectReference `json:"clusterRef,omitempty"`
+	PodRef      *kapi.ObjectReference `json:"podRef,omitempty"`
+	PodSelector map[string]string     `json:"podSelector,omitempty"`
```
alaypatel07, line 32: I think you will need a `Namespace` field here right? in order to list using this selector?

alaypatel07 (approved):
did a second pass. Added some questions, apart from that LGTM

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
pranavgaikwad: @alaypatel07 this label actually holds the name of the PVC this Pod is created for. Since we are generating names instead of having a fixed name for Rsync pods, I think `created-for-pvc` is more specific.

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
pranavgaikwad: fixed

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
djwhatle: ACK. This field is not going away, just moved to a new struct that is inlined. Users will access the information in the same way. No breaking change from user perspective.

djwhatle (approved):

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
pranavgaikwad: @alaypatel07 agreed. I think we should remove the API field and move it to MigrationController. Removed

PR #1013, opened 2021-03-18
Fix DVM/DVMP hotlooping, remove remote watch TouchAnnotation, fix requeues so they used values from r.Migrate()

### LINKED ISSUE #1011: DVM, DVMP (possibly DIM, DISM) sometimes hotloop
**Describe the bug**
I discovered that the controllers added in release-1.4.0 are relying on TouchAnnotation for requeues in some places, which is causing excessive reconciles in DVMP / DVM. I don't believe the function or existence of TouchAnnotation was known when writing the controllers.

After removing TouchAnnotation (a relic of remote watch that should not be used without predicates, which are also not present) the controllers get stuck in various places and do not complete the migration. 

Solution is two things:
 - Remove TouchAnnotation to eliminate echo reconciles
 - Respect RequeueAfter times returned from `r.migrate()`

We need to fix these issues so that DVM and DVMP won't hotloop waiting for Pods to progress.

djwhatle:
Closes #1011

### DIFF
--- a/pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
+++ b/pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
@@ -19,7 +19,6 @@ package v1alpha1
 import (
 	"context"
 
-	"github.com/google/uuid"
 	liberr "github.com/konveyor/controller/pkg/error"
 	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -108,11 +107,9 @@ type DirectVolumeMigrationProgress struct {
 }
 
 func (d *DirectVolumeMigrationProgress) MarkReconciled() {
-	u, _ := uuid.NewUUID()
 	if d.Annotations == nil {
 		d.Annotations = map[string]string{}
 	}
-	d.Annotations[TouchAnnotation] = u.String()
 	d.Status.ObservedDigest = digest(d.Spec)
 }
 
--- a/pkg/controller/directimagemigration/directimagemigration_controller.go
+++ b/pkg/controller/directimagemigration/directimagemigration_controller.go
@@ -18,6 +18,7 @@ package directimagemigration
 
 import (
 	"context"
+	"time"
 
 	"github.com/konveyor/controller/pkg/logging"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
@@ -122,7 +123,7 @@ func (r *ReconcileDirectImageMigration) Reconcile(request reconcile.Request) (re
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
-		return reconcile.Result{}, err
+		return reconcile.Result{Requeue: true}, err
 	}
 
 	// Set up jaeger tracing
@@ -146,8 +147,11 @@ func (r *ReconcileDirectImageMigration) Reconcile(request reconcile.Request) (re
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	// Default to PollReQ, can be overridden by r.migrate phase-specific ReQ interval
+	requeueAfter := time.Duration(PollReQ)
+
 	if !imageMigration.Status.HasBlockerCondition() {
-		_, err = r.migrate(imageMigration, reconcileSpan)
+		requeueAfter, err = r.migrate(imageMigration, reconcileSpan)
 		if err != nil {
 			log.Trace(err)
 			return reconcile.Result{Requeue: true}, nil
@@ -171,6 +175,10 @@ func (r *ReconcileDirectImageMigration) Reconcile(request reconcile.Request) (re
 		return reconcile.Result{Requeue: true}, nil
 	}
 
-	// Done
+	// Requeue
+	if requeueAfter > 0 {
+		return reconcile.Result{RequeueAfter: requeueAfter}, nil
+	}
+
 	return reconcile.Result{}, nil
 }
--- a/pkg/controller/directimagemigration/directimagemigration_controller.go
+++ b/pkg/controller/directimagemigration/directimagemigration_controller.go
@@ -120,10 +120,10 @@ func (r *ReconcileDirectImageMigration) Reconcile(request reconcile.Request) (re
 		if errors.IsNotFound(err) {
 			// Object not found, return.  Created objects are automatically garbage collected.
 			// For additional cleanup logic use finalizers.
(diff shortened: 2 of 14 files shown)


pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
```diff
@@ -64,11 +63,9 @@ type DirectVolumeMigrationProgress struct {
 }
 
 func (d *DirectVolumeMigrationProgress) MarkReconciled() {
-	u, _ := uuid.NewUUID()
 	if d.Annotations == nil {
 		d.Annotations = map[string]string{}
 	}
-	d.Annotations[TouchAnnotation] = u.String()
```
djwhatle, line 71: This TouchAnnotation is an unintended leftover from the remote watch system. This was likely causing requeues where they shouldn't have been happening in DVM/DVMP.

pkg/controller/directimagemigration/directimagemigration_controller.go
```diff
@@ -117,17 +118,20 @@ func (r *ReconcileDirectImageMigration) Reconcile(request reconcile.Request) (re
 		if errors.IsNotFound(err) {
 			// Object not found, return.  Created objects are automatically garbage collected.
 			// For additional cleanup logic use finalizers.
-			return reconcile.Result{}, nil
+			return reconcile.Result{Requeue: false}, nil
```
djwhatle, line 121: I think it's more obvious what's happening if we explicitly use the requeue flag

pkg/controller/directimagemigration/directimagemigration_controller.go
```diff
@@ -117,17 +118,20 @@ func (r *ReconcileDirectImageMigration) Reconcile(request reconcile.Request) (re
...
 	}
 
 	// Completed.
 	if imageMigration.Status.Phase == Completed {
-		return reconcile.Result{}, nil
+		return reconcile.Result{Requeue: false}, nil
 	}
 
+	// Default re-queue (after) in seconds.
+	requeueAfter := time.Duration(PollReQ)
```
djwhatle, line 133: PollReQ is the default, and it is overridden by the output of `r.migrate` below. This is the combination that has been proven to work well, tested heavily during removal of remote watch. 

After removing the TouchAnnotation (which causes constant echo reconciles), I saw migrations halt in the middle because they weren't using the output of r.Migrate in DIM, DISM, DVM, DVMP.

pkg/controller/directimagemigration/directimagemigration_controller.go
```diff
@@ -139,7 +143,7 @@ func (r *ReconcileDirectImageMigration) Reconcile(request reconcile.Request) (re
 	}
 
 	if !imageMigration.Status.HasBlockerCondition() {
-		_, err = r.migrate(imageMigration)
+		requeueAfter, err = r.migrate(imageMigration)
```
djwhatle, line 146: No longer throwing away the requeueAfter, so now if a requeue duration is sleected in phase logic it actually gets used.

pkg/apis/migration/v1alpha1/directvolumemigrationprogress_types.go
alaypatel07: +1000, thanks for removing this

pkg/controller/directimagemigration/directimagemigration_controller.go
alaypatel07: +1. thanks for digging into this, makes a lot of sense

alaypatel07: for locality, do you think moving this to just before the if block at line 145 makes sense? would make the code much more readable when coming back to this in the future.

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -171,7 +171,7 @@ func (r *ReconcileDirectVolumeMigrationProgress) Reconcile(request reconcile.Req
 	}
 
 	if !pvProgress.Status.IsReady() {
-		return reconcile.Result{Requeue: true}, nil
+		return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 5}, nil
```
alaypatel07, line 174: I wonder if this if block is necessary? looks like the intent is to requeue at 5 seconds irrespective of the status

alaypatel07 (approved):
Just small re-org comments and minor nit on if block otherwise looks good to me

VISIACK

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
djwhatle: Yeah maybe not, let me take a look and see if it's safe to remove

pkg/controller/directimagemigration/directimagemigration_controller.go
djwhatle: Yeah I think that makes sense, will change 👍

pkg/controller/directvolumemigration/task.go
```diff
@@ -512,7 +512,6 @@ func (t *Task) Run() error {
 
 	if t.Phase == Completed {
 		t.Requeue = NoReQ
-		t.Log.Info("[COMPLETED]")
```
djwhatle, line 515: No idea why this got removed, I am adding it back

djwhatle:
I verified there is no observable perf change from this PR, which is what I expected. The goal of this PR is to reduce pointless API requests without reducing our migration performance.

### Migration time for rocket-chat (this PR):
2 minutes, 55 seconds

### Migration time for rocket-chat (master)
2 minutes, 56 seconds

alaypatel07 (approved):
thanks for the changes, added a small nit, feel free to merge after that from my end.

VISIACK

sseago (approved):

pkg/controller/directimagestreammigration/directimagestreammigration_controller.go
```diff
@@ -154,6 +158,10 @@ func (r *ReconcileDirectImageStreamMigration) Reconcile(request reconcile.Reques
...
 	}
 
-	// Done
-	return reconcile.Result{}, nil
+	// Requeue
+	if requeueAfter > 0 {
+		return reconcile.Result{RequeueAfter: requeueAfter}, nil
+	}
+
+	return reconcile.Result{Requeue: false}, nil
```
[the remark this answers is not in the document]
djwhatle: Good point, will update.

PR #1014, opened 2021-03-18
Add plan check before PV resize 

shubham-pampattiwar:
This PR adds a `nil`  check for `MigPlan` before performing PV resize during the creation of destination PVs


pranavgaikwad (approved):
LGTM, thanks for the fix!

PR #1017, opened 2021-03-19
DVM & DVMP Rsync retry logic

pranavgaikwad:
This PR introduces a notion of retrying Rsync on failure in DVM controller and updates logic in DVMP to watch all pods associated with 1 PVC. This could have been 2 PRs but I intentionally kept both together to provide a cogent view of interactions between DVM <-> DVMP. Open to moving around if that helps you review. 

The changes in this PR are based on API changes submitted earlier here: https://github.com/konveyor/mig-controller/pull/1009

Summary of changes : 
[1] DVM
- Removes existing phases `CreateRsyncClientPods` and `WaitForRsyncClientPodsCompleted` and replaces both with a new phase called `RunRsyncOperations`. This phase will be responsible for handling all attempts of Rsync. 
- Cleans up and re-organizes business logic in itinerary and moves different decision making logic into their own functions.
- Introduces `runRsyncOperations` function to orchestrate Rsync pods, gather progress and analyze failures/errors (all testeable separately)

[2] DVMP
- Re-organizes `reconcile()` function, moves validation bits to `validation.go` for consistency.
- Adds new logic to look for multiple Rsync Pods instead of watching just one Pod. 

Additionally, this PR also introduces a new fake compat client implementation to assist testing across the project.


pkg/compat/fake/client.go
```diff
@@ -0,0 +1,59 @@
...
+}
+func (fake fakeCompatClient) Get(ctx context.Context, key k8sclient.ObjectKey, obj runtime.Object) error {
+	return fake.Client.Get(ctx, key, obj)
+}
+func (fake fakeCompatClient) List(ctx context.Context, opt *k8sclient.ListOptions, obj runtime.Object) error {
+	return fake.Client.List(ctx, opt, obj)
+}
+func (fake fakeCompatClient) Create(ctx context.Context, obj runtime.Object) error {
+	switch obj.(type) {
+	case *corev1.Pod:
```
pranavgaikwad, line 39: by default, k8s fake client implementation cannot generate names on resources through `generateName` which is required for Rsync pod creation logic. This is a rudimentary way of achieving that

pkg/controller/directvolumemigrationprogress/validation_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"testing"
+
+	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
+	"github.com/konveyor/mig-controller/pkg/compat"
+	fakecompat "github.com/konveyor/mig-controller/pkg/compat/fake"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestReconcileDirectVolumeMigrationProgress_validateSpec(t *testing.T) {
```
pranavgaikwad, line 13: this test is added to make sure that pre MTC 1.4.3 CRs and new CRs both can co-exist without breaking the controller

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1651,3 +1678,591 @@ func (t *Task) deleteProgressReportingCRs(client k8sclient.Client) error {
...
+		}
+	}
+	return isComplete, anyFailed, failureReasons, nil
+}
+
+// reportAdvancedErrorHeuristics processes DVMP CRs for all PVCs,
+// for all errored pods, attempts to determine whether the errors fall into any
+// of the special categories we can identify and reports them as conditions
+// returns reasons and error for reconcile decisions
+func (t *Task) reportAdvancedErrorHeuristics() ([]string, error) {
```
pranavgaikwad, line 1922: these functions were previously called directly in the switch case for itinerary, moved them outside here to keep itinerary  logic easy and readable

pkg/controller/directvolumemigration/task.go
```diff
@@ -409,77 +411,31 @@ func (t *Task) Run() error {
...
 		if err = t.next(); err != nil {
 			return liberr.Wrap(err)
 		}
-	case WaitForRsyncClientPodsCompleted:
-		completed, failed, err := t.haveRsyncClientPodsCompletedOrFailed()
-		if err != nil {
-			return liberr.Wrap(err)
-		}
-		if completed {
-			isStunnelTimeout, err := hasAllRsyncClientPodsTimedOut(t.getPVCNamespaceMap(), t.Client, t.Owner.Name)
```
pranavgaikwad, line 436: These functions are now called directly in the RunRsyncOperations phase to keep this switch case minimal

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1651,3 +1678,591 @@ func (t *Task) deleteProgressReportingCRs(client k8sclient.Client) error {
...
+	waitGroup.Wait()
+	close(rateLimiter)
+	close(outputChan)
+	<-finishChan
+	return statusList, nil
+}
+
+// garbageCollectRsyncPods garbage collection routine
+// will run in background, returns list of failures when encountered
+func (t *Task) garbageCollectPodsForRequirements(podRequirements []rsyncClientPodRequirements) []error {
```
pranavgaikwad, line 2118: not implemented yet. Two concerns to consider: 
1. Where should this happen? Want to make sure the reconcile is fast enough
2. If we fail to delete a pod, do we just report errors as they happen and continue deleting next pod making it at best effort?

pkg/compat/fake/client.go
djwhatle: +1

pkg/controller/directvolumemigration/task.go
```diff
@@ -123,8 +124,9 @@ var VolumeMigration = Itinerary{
 		{phase: WaitForRsyncTransferPodsRunning},
 		{phase: CreateStunnelClientPods},
 		{phase: WaitForStunnelClientPodsRunning},
-		{phase: CreateRsyncClientPods},
-		{phase: WaitForRsyncClientPodsCompleted},
+		// {phase: CreateRsyncClientPods},
```
djwhatle, line 127: Guessing you didn't mean to leave this commented

pkg/controller/miganalytic/miganalytics_controller_test.go
```diff
@@ -277,7 +277,7 @@ func TestReconcileMigAnalytic_Reconcile(t *testing.T) {
 					},
 				}, &testMigPlan, &testMigStorage, &testSourceMigCluster, &testTargetMigCluster, &testSourceNamespace),
 			},
-			want:      reconcile.Result{},
+			want:      reconcile.Result{RequeueAfter: time.Second * RequeueInterval},
```
pranavgaikwad, line 280: this comes from the latest fixes that went into MigAnalytics

djwhatle (changes requested):
Added a few comments. I think we need to figure out what we want to do about the garbage collection routine before merged, as discussed.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1651,3 +1678,625 @@ func (t *Task) deleteProgressReportingCRs(client k8sclient.Client) error {
...
+			Containers:    containers,
+			NodeName:      req.nodeName,
+			SecurityContext: &corev1.PodSecurityContext{
+				SupplementalGroups: req.pvInfo.supplementalGroups,
+				FSGroup:            req.pvInfo.fsGroup,
+				SELinuxOptions:     req.pvInfo.seLinuxOptions,
+			},
+		},
+	}
+	return clientPod
```
alaypatel07, line 1776: this looks like a perfect use case for templates, wonder if we should define a YAML file/variable, read and parse the variables through the template?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1651,3 +1678,625 @@ func (t *Task) deleteProgressReportingCRs(client k8sclient.Client) error {
...
+			PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+				ClaimName: req.pvInfo.name,
+			},
+		},
+	})
+	rsyncCommand := []string{"rsync"}
+	rsyncCommand = append(rsyncCommand, req.rsyncOptions...)
+	rsyncCommand = append(rsyncCommand, fmt.Sprintf("/mnt/%s/%s/", req.namespace, req.pvInfo.name))
+	rsyncCommand = append(rsyncCommand, fmt.Sprintf("rsync://root@%s/%s", req.destIP, req.pvInfo.name))
+	labels := map[string]string{
```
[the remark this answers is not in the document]
pranavgaikwad: Fixed

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1651,3 +1678,625 @@ func (t *Task) deleteProgressReportingCRs(client k8sclient.Client) error {
...
+				SupplementalGroups: req.pvInfo.supplementalGroups,
+				FSGroup:            req.pvInfo.fsGroup,
+				SELinuxOptions:     req.pvInfo.seLinuxOptions,
+			},
+		},
+	}
+	return clientPod
+}
+
+func (t *Task) prepareRsyncPodRequirements(srcClient compat.Client) ([]rsyncClientPodRequirements, error) {
```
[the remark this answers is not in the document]
pranavgaikwad: added back the logs

pkg/controller/directvolumemigration/rsync.go
pranavgaikwad: @alaypatel07 I investigated whether we can use template here...we can. I feel like the gain of readability doesnt justify in this case because we need templates for containers, volumes, and the pods at the very least since they are all dynamically created. Maintaining three templates, executing them and unmarshling them back into the Pod object seems unnecessary.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1660,3 +1690,683 @@ func (t *Task) deleteProgressReportingCRs(client k8sclient.Client) error {
...
+	waitGroup := &sync.WaitGroup{}
+	mutex := &sync.Mutex{}
+	for i := range podRequirements {
+		req := &podRequirements[i]
+		lastObservedOperationStatus := t.Owner.Status.GetRsyncOperationStatusForPVC(&corev1.ObjectReference{
+			Name:      req.pvInfo.name,
+			Namespace: req.namespace,
+		})
+		// from this point onwards, do not mutate the original reference, create a copy and use it
+		threadSafeOperationStatus := *lastObservedOperationStatus.DeepCopy()
```
pranavgaikwad, line 2126: @djwhatle I investigated the wrapper around api fields and usage for thread-safe access methods. I ran into some problems with the encoder not being able to reflect into fields. Therefore, I decided to make it outside of API. The above deep copy here makes sure that every goroutine gets their own copy of Observed Status. I changed the functions to use copy of the variable instead of taking references, which makes sure there will never be concurrent reads/writes on the data structures used as they are no longer shared.

pkg/controller/directvolumemigration/rsync.go
alaypatel07: @pranavgaikwad I would expect only 1 pod template and all the variable parts will be filled as variables from the struct. That gets rid of all the verbosity of the existing code.

alaypatel07: If you still feel it's not something we want I'm fine.

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -116,8 +133,10 @@ type DirectVolumeMigrationList struct {
 
 type PodProgress struct {
 	*kapi.ObjectReference       `json:",inline"`
-	LastObservedProgressPercent string `json:"lastObservedProgressPercent,omitempty"`
-	LastObservedTransferRate    string `json:"lastObservedTransferRate,omitempty"`
+	PVCReference                *kapi.ObjectReference `json:"pvcRef,omitempty"`
```
pranavgaikwad, line 136: @alaypatel07 @djwhatle Just wanted to bring to your attention this API change. This is required for the progress reporting system to show PVC name & total elapsed time in the pipeline progress messages.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1651,3 +1678,625 @@ func (t *Task) deleteProgressReportingCRs(client k8sclient.Client) error {
...
+				}
+			}
+		}
+	}
+	return errors
+}
+
+// reconcileRsyncOperationState reconciles observed state of 1 Rsync operation
+// takes relevant actions wherever required such as launching a new pod when previous one fails
+func (t *Task) reconcileRsyncOperationState(client compat.Client, req *rsyncClientPodRequirements,
```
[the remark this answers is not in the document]
pranavgaikwad: @djwhatle The upper bound is set by DefaultOperationConcurrency variable which is set to 5.

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -88,6 +88,23 @@ func (ds *DirectVolumeMigrationStatus) GetRsyncOperationStatusForPVC(pvcRef *kap
 	return newStatus
 }
 
+// AddRsyncOperation adds a new RsyncOperation to list, updates an existing one if found
+func (ds *DirectVolumeMigrationStatus) AddRsyncOperation(podStatus *RsyncOperation) {
+	if podStatus == nil {
```
alaypatel07, line 93: out of curiosity do you need to remove the rsync operation from the list here if found?

alaypatel07: actually nevermind, I was misreading.

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -132,6 +151,24 @@ type RsyncOperation struct {
 	Failed bool `json:"failed,omitempty"`
 }
 
+func (r *RsyncOperation) Equal(other *RsyncOperation) bool {
```
alaypatel07, line 154: Wonder if we can refactor this function to 

```
func (x *RsyncOperation) Equal(y *RsyncOperation) bool {
```
When reading equality x on the left and y on the right side of equality makes it easier to read.

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
alaypatel07: @pranavgaikwad thanks for the explanation. I am not super proud of introducing the PodProgress type. Retrospecting, it feels like had it been `PVCProgress` we would have had a much cleaner API.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1027,7 +1034,7 @@ func (t *Task) createRsyncClientPods() error {
 	// Get client for destination
 	srcClient, err := t.getSourceClient()
 	if err != nil {
-		return err
+		return liberr.Wrap(err)
```
alaypatel07, line 1037: Why is this an exception? Other errors later in the function are returned without Wraps?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1211,21 +1217,44 @@ func (t *Task) createPVProgressCR() error {
...
+			existingDvmp := migapi.DirectVolumeMigrationProgress{}
+			shouldDelete := false
+			// Make sure existing DVMP CRs which don't have required fields are deleted
+			err := t.Client.Get(context.TODO(), types.NamespacedName{Name: dvmp.Name, Namespace: dvmp.Namespace}, &existingDvmp)
+			if err != nil {
+				if !k8serror.IsNotFound(err) {
+					return err
+				}
+			}
+			if existingDvmp.Name != "" && existingDvmp.Namespace != "" {
```
alaypatel07, line 1234: put this in a separate function like `deleteDVMPForSmoothUpgrade` to make it explicit and not pollute this function?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1244,62 +1273,57 @@ func getMD5Hash(s string) string {
 	return hex.EncodeToString(hash[:])
 }
 
-func (t *Task) haveRsyncClientPodsCompletedOrFailed() (bool, bool, error) {
+// reportRsyncPodProgress reads DVMP CR present for all PVCs and generates progress information in CR status
+func (t *Task) reportRsyncPodProgress() (bool, error) {
```
alaypatel07, line 1277: the function name does not suggest we are returning a bool. rename to `haveAllRsyncOperationsCompleted` instead?

pkg/controller/directvolumemigration/rsync.go
pranavgaikwad: @alaypatel07 this is certainly not an exception. I just copied this block from somewhere and changed the error randomly. I will make sure I make the change everywhere. Thanks for pointing out!

pkg/controller/directvolumemigration/rsync.go
pranavgaikwad: +1, will do

pkg/controller/directvolumemigration/rsync.go
pranavgaikwad: +1, updating

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
pranavgaikwad: +1, updating

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1651,3 +1678,625 @@ func (t *Task) deleteProgressReportingCRs(client k8sclient.Client) error {
...
+	if err != nil {
+		return reasons, liberr.Wrap(err)
+	}
+	if isStunnelTimeout {
+		t.Owner.Status.SetCondition(migapi.Condition{
+			Type:     migapi.ReconcileFailed,
+			Status:   True,
+			Reason:   SourceToDestinationNetworkError,
+			Category: migapi.Error,
+			Message: "All the rsync client pods on source are timing out at 20 seconds, " +
```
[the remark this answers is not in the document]
pranavgaikwad: @djwhatle updated the condition Reason / Type. The intent here is to not reconcile Volume Migration. It is still achieved through `Critical` condition through `HasBlockerCondition()` check. Added log.

djwhatle (approved):
LGTM. I'd ask that we run a test to make sure that existing DVMPs and DVMs will upgrade successfully to this new version.

In the future, I question if we'll need the concurrent approach if we have a cached client again, but for now I think it will be good.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1244,62 +1287,57 @@ func getMD5Hash(s string) string {
...
-				})
-			case dvmp.Status.PodPhase == corev1.PodSucceeded:
-				t.Owner.Status.SuccessfulPods = append(t.Owner.Status.SuccessfulPods, &migapi.PodProgress{
-					ObjectReference:             objRef,
-					LastObservedProgressPercent: dvmp.Status.LastObservedProgressPercent,
-					LastObservedTransferRate:    dvmp.Status.LastObservedTransferRate,
-				})
+			case dvmp.Status.PodPhase == corev1.PodRunning ||
+				// when last pod failed, but whole operation is not failed, show as running until next pod is created to avoid flicker
+				(dvmp.Status.PodPhase == corev1.PodFailed && !operation.Failed):
```
alaypatel07, line 1328: this flickering is dubious. A smart user sees in the dvm CR that the pod is running but they see in dvmp CR that pod is failing? isn't that contradictory?

pranavgaikwad: @alaypatel07 it happens for a very small period of time but the end effect of allowing that behaviour is bad in the UI. Like the pod goes completely missing from the details page and then suddenly shows up with a new Pod name because it got retried. I wouldn't worry about the CLI user as this happens for a very small amount of time (time between a failed pod being registered as failed to the new one being created, a few ms)

alaypatel07: @pranavgaikwad why dont we have `backoffLimit` on `dvmp.spec` and let it declare failure after it sees n pods instead of polluting the logic here?

pranavgaikwad: @alaypatel07 I don't think it's DVMP's concern to even know about backOffLimit, if you put backOffLimit in DVMP it will also have to run to Completion. You cannot re-use the same CR over and over again.

alaypatel07: > @alaypatel07 it happens for a very small period of time but the end effect of allowing that behaviour is bad in the UI. Like the pod goes completely missing from the details page and then suddenly shows up with a new Pod name because it got retried. I wouldn't worry about the CLI user as this happens for a very small amount of time (time between a failed pod being registered as failed to the new one being created, a few ms) But that flicker is pretty evident in the UI if youre continuously looking at the screen. I hope I explained it right, but I can show you the video if not clear.

Ack, if we are lucky it will not be visible to the user. But I still feel this switch cases are little over loaded. We can delegate some of the work from here to dvmp controller and have this cleanly depend on the dvmp status purely. If dvmp status says we succeeded, failed or are running, dvm blindly copies it. its a good way to hide the complexity underneath the API status IMO

alaypatel07: Synced with Pranav over a call, this is not an issue as of now, we will resolve and re-visit when the time is right

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1690,3 +1724,710 @@ func (t *Task) deleteProgressReportingCRs(client k8sclient.Client) error {
...
+				destIP:       ip,
+				rsyncOptions: rsyncOptions,
+			}
+			req = append(req, podRequirements)
+		}
+	}
+	return req, nil
+}
+
+func (t *Task) getRsyncOperationsContext() (compat.Client, []rsyncClientPodRequirements, error) {
```
alaypatel07, line 1895: wonder if we can rename `getRsyncOperationsContext` to getRsyncOperationsClientAndRequirements?

*Context in the name can be understand as returning context.Context

alaypatel07 (approved):
Great job @pranavgaikwad, VISIACK

FTR: Had some questions around time.Now() and going back 5 mins for warnings, but those were determined as extreme corner cases.

PR #1037, opened 2021-03-28
Introduce cached client behind feature flag

djwhatle:
**Summary**

- Makes mig-controller go _BRRR_
- Reintroduces remote cluster cache behind feature flag `ENABLE_CACHED_CLIENT`
  - Each remote cluster gets its own manager with cache 
  - On `Create` and `Update`, blocks until the cache catches up 

Complement to operator PR: https://github.com/konveyor/mig-operator/pull/675

**To test**

```
# Local
ENABLE_CACHED_CLIENT=true make run-fast
```

**Open (Potential) Problems:**

* [x] Need to add safe map access to the stored list of clients. 

* [x] Need to properly handle teardown and recreation of client in shared map when a migcluster credential set changes
  - Need a graceful strategy for dealing with bad creds being given to the cache `manager` due to bad MigCluster creds (this may be handled by current logic)

* [x] Need to properly destroy manager which holds the cache when a migcluster is removed. 
  - This removal must be safe such that we don't crash due to someone else using the manager client as we destroy it

* [x] Need to make sure that `update` calls are blocking until receipt of update is present in our cache (e.g. a GET shows the expected resourceVersion)

* [x] Need to add a while polling for cache to populate after a `create` or `update` (for a rare race condition I'm imagining where we create something but something else deletes, we wouldn't want to hang forever waiting, maybe 10 second max)

### DIFF
--- a/pkg/compat/client.go
+++ b/pkg/compat/client.go
@@ -3,21 +3,29 @@ package compat
 import (
 	"context"
 	"errors"
+	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/konveyor/mig-controller/pkg/settings"
 	appsv1 "k8s.io/api/apps/v1"
 	appsv1beta1 "k8s.io/api/apps/v1beta1"
 	batchv1beta "k8s.io/api/batch/v1beta1"
 	batchv2alpha "k8s.io/api/batch/v2alpha1"
 	extv1beta1 "k8s.io/api/extensions/v1beta1"
+	k8serror "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	dapi "k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+var Settings = &settings.Settings
 
 //
 // A smart client.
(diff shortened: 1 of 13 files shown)


pkg/compat/client.go
```diff
@@ -225,7 +223,47 @@ func (c client) Create(ctx context.Context, in runtime.Object) error {
...
+			if err != nil || !found {
+				golog.Printf("[kind=%v] meta.namespace not found", kind)
+				break
+			}
+			uName, found, err := unstructured.NestedString(u, "metadata", "name")
+			if err != nil || !found {
+				golog.Printf("[kind=%v] meta.name not found", kind)
+				break
+			}
+			golog.Print(fmt.Sprintf("\n\n\n\n\n\n<<<[CREATE OBJ]>>> kind=[%v] key=[%v/%v]\n\n\n\n\n\n", kind, uNs, uName))
```
djwhatle, line 247: Ignore this logger and log statements, will be removed but helped to understand while debugging.

pranavgaikwad:
I used this PR and I would just like to post the results here.

1. Before this PR, `RunRsyncOperations` phase in DVM took about 10-15 seconds to run once: 

Screenshot from 2021-03-31 13-05-20

2. After this PR, the same phase takes a few seconds (worst case) to run once:

Screenshot from 2021-03-31 13-06-04

pkg/controller/directvolumemigration/stunnel.go
```diff
@@ -419,3 +424,235 @@ func (t *Task) setupCerts() error {
 	}
 	return nil
 }
+
```
djwhatle, line 427: Whoah, didn't mean to pick this up. Will need to grab this file off master.

pranavgaikwad (changes requested):
Overall looks good to me. I will wait for operator PR for the feature flag.

djwhatle:
Need to rebase. Doing it now.

djwhatle:
Squashed to make the rebase easier.

djwhatle:
Just got the rebase finished. Need to do a bit of testing to verify that everything is working as expected.

pranavgaikwad (approved):
LGTM

sseago (approved):
The manager stop changes look sane to me.

PR #1044, opened 2021-03-31
No traces on DVMP validation fail

djwhatle:
log.V(4).error will override the loglevel and print the error regardless. I changed these to info messages with the error as a k/v.

### DIFF
--- a/pkg/controller/directvolumemigrationprogress/validation.go
+++ b/pkg/controller/directvolumemigrationprogress/validation.go
@@ -34,7 +34,7 @@ func (r *ReconcileDirectVolumeMigrationProgress) validate(pvProgress *migapi.Dir
 	cluster *migapi.MigCluster, client compat.Client, err error) {
 	cluster, err = r.validateCluster(pvProgress)
 	if err != nil {
-		log.V(4).Error(err, "Validation check for referenced MigCluster failed")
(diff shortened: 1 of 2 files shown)


pkg/controller/directvolumemigrationprogress/validation.go
```diff
@@ -34,7 +34,7 @@ func (r *ReconcileDirectVolumeMigrationProgress) validate(pvProgress *migapi.Dir
 	cluster *migapi.MigCluster, client compat.Client, err error) {
 	cluster, err = r.validateCluster(pvProgress)
 	if err != nil {
-		log.V(4).Error(err, "Validation check for referenced MigCluster failed")
+		log.V(4).Info("Validation check for referenced MigCluster failed", err)
```
pranavgaikwad, line 37:
```suggestion
		log.V(4).Info("Validation check for referenced MigCluster failed", "error", err)
```

pranavgaikwad (changes requested):
1 small change and its ready to go

pkg/controller/directvolumemigrationprogress/validation.go
djwhatle: Oops, good catch

djwhatle:
Good catch. Fixed. Merging.

PR #1045, opened 2021-03-31
Add common migMigration key to all controller involved in migration

djwhatle:
@eriknelson requested we get this one into 1.4.3 since it is an easy add-on.

This change makes DVM, DVMP, DIM, DISM attach the migMigration name to their log messages if they are associated with a migmigration, otherwise they just leave it out.

```json
{"level":"info","ts":1617223088168,
"logger":"directvolume|78jbc",
"msg":"Getting Rsync Password from Secret on host MigCluster",
"dvm":"56dcf8d0-9260-11eb-b77e-f5309eab2e0b-tjx7r",
"migMigration":"56dcf8d0-9260-11eb-b77e-f5309eab2e0b",
"phase":"RunRsyncOperations",
"secret":""}    
```

The idea here is that we will be able to easily grep for all log messages associated with a particular migration.

This matches the key already being used in the migmigration controller.

### DIFF
--- a/pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
+++ b/pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
@@ -119,7 +119,7 @@ func (r *ReconcileDirectVolumeMigrationProgress) Reconcile(request reconcile.Req
 	}
 
 	// Set MigMigration name key on logger
-	migration, err := pvProgress.GetDVMforDVMP(r)
+	migration, err := pvProgress.GetMigrationforDVMP(r)
 	if migration != nil {
 		log.SetValues("migMigration", migration.Name)
 	}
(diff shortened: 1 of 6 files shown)


pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller.go
```diff
@@ -118,6 +118,12 @@ func (r *ReconcileDirectVolumeMigrationProgress) Reconcile(request reconcile.Req
 		return reconcile.Result{Requeue: true}, err
 	}
 
+	// Set MigMigration name key on logger
+	migration, err := pvProgress.GetDVMforDVMP(r)
```
djwhatle, line 122: Oops

djwhatle: Fixed.

alaypatel07 (approved):
ACK

I had a small hesitation with `migMigration` camel case is odd. But we can get it and convert it into `mm` when we have a short name for migmigraiton.

djwhatle:
> ACK
> 
> I had a small hesitation with `migMigration` camel case is odd. But we can get it and convert it into `mm` when we have a short name for migmigraiton.

I agree it is a bit awkward, +1

PR #1046, opened 2021-03-31
Fix missing nil check in dvm progress function

pranavgaikwad:


djwhatle (approved):

djwhatle:
It works, I guarantee it.

PR #1058, opened 2021-04-06
Use server-side filtering of events when looking for abnormalities

djwhatle:
We may want to get this into 1.4.3 if possible.

I verified this is working as expected in a debugger.

The changes in this PR are equivalent to doing this from the `oc` cli tool:

```
oc get event -n mssql-persistent --field-selector 'involvedObject.kind=Pod,involvedObject.name=directvolumemigration-rsync-transfer' -v=6
```

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -2,8 +2,8 @@ package event
 
 import (
 	"context"
+	"fmt"
 	"path"
-	"strings"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


pkg/event/event.go
```diff
@@ -15,29 +15,24 @@ import (
...
 // related to a resource we're waiting on.
+// Make sure to pass in the kind exactly as it is capitalized on the event, e.g. "Pod"
 func GetAbnormalEventsForResource(client client.Client,
 	nsName types.NamespacedName, resourceKind string) ([]corev1.Event, error) {
 	uniqueEventMap := make(map[string]corev1.Event)
 
 	eList := corev1.EventList{}
 	options := k8sclient.InNamespace(nsName.Namespace)
-	err := client.List(context.TODO(), options, &eList)
+	fieldSelector := fmt.Sprintf("involvedObject.name=%s,involvedObject.kind=%s,type!=Normal",
```
djwhatle, line 25: Using a field selector here so that we don't have to pull back the full event list. Luckily, events are one of the fields it's possible to use field selectors on. See [this list](https://github.com/flant/shell-operator/blob/8fa3c3b8cfeb1ddb37b070b7a871561fdffe788b/HOOKS.md#fieldselector)

alaypatel07: NIT: Wonder if you can release the lines 24-30 with this:
```
		&k8sclient.ListOptions{
			Namespace:     nsName.Namespace,
			FieldSelector: fields.SelectorFromSet(fields.Set{"involvedObject.name": nsName.Name, "involvedObject.namespace": nsName.Namespace},
		}
```

alaypatel07 (approved):
VISIACK.

Added a code simplification nit, apart from that LGTM

pkg/event/event.go
djwhatle: I don't think the field selector from set supports the `!=` operator, otherwise I would have used it.

alaypatel07: oh I totally missed the `!` part, +1

PR #1066, opened 2021-04-09
Make ownerReferences on DVM optional

### LINKED ISSUE #1053: [MIG-559] Make MigMigration ownerReferences in DVM optional
**Describe the bug**
DVM requires a valid _MigMigration_ ownerReference set. Without it, DVM doesn't reconcile the CR. OwnerReference on MigMigration should be made optional as we move towards making DVM capable of State Only transfer. 

**To Reproduce**
Create DVM using `oc` command which doesn't have `metadata.ownerReferences` set. 

**Expected behavior**
The DVM should be reconciled assuming that it was created stand-alone and it's not part of a _MigMigration_.

**Screenshots & Snippets**

**Additional context**

pranavgaikwad:
Fixes #1053

### DIFF
--- a/pkg/controller/directvolumemigration/migrate.go
+++ b/pkg/controller/directvolumemigration/migrate.go
@@ -2,7 +2,6 @@ package directvolumemigration
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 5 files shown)


djwhatle:
@pranavgaikwad I think this issue exists in a few other funcs. Let me find them to make sure we knock out all of the potential issues in one PR.

djwhatle (approved):

pkg/controller/directvolumemigration/migrate.go
```diff
@@ -38,9 +37,8 @@ func (r *ReconcileDirectVolumeMigration) migrate(ctx context.Context, direct *mi
 		Phase:            direct.Status.Phase,
 		PhaseDescription: direct.Status.PhaseDescription,
 		PlanResources:    planResources,
-		MigrationUID:     string(migration.UID),
-
-		Tracer: r.tracer,
+		MigrationUID:     migrationUID,
```
djwhatle, line 40: If the MigrationUID is only used in one place, it feels like maybe we should just use `GetMigrationForDVM()` to grab it the Migration later on instead of putting the UID in this struct. No strong feelings either way though.

pranavgaikwad: @djwhatle Fixed

PR #1069, opened 2021-04-13
Bug 1949238: No warning on NodeName. Warn if NodeSelector is non-default

djwhatle:
This PR:
 - stops MigPlan Wizard from warning the user about NodeSelectors when they are just using the default one
 - stops MigPlan wizard from warning the user if NodeName is set, since NodeName appears to be _always_ set on OCP 3.11.
 - adds TCs for above
 - makes the warning condition durable so it doesn't disappear when the migration starts

### DIFF
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -394,7 +394,8 @@ func (r ReconcileMigPlan) validatePodProperties(ctx context.Context, plan *migap
 	return nil
 }
 
-// Checks the list of Pods for any non-default nodeselectors. Returns list of custom nodeselectors.
+// Checks the list of Pods for any non-default nodeselectors.
+// Returns true if custom nodeselectors found on any Pod.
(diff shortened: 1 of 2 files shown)


pkg/controller/migplan/validation.go
```diff
@@ -394,10 +394,21 @@ func (r ReconcileMigPlan) validatePodProperties(ctx context.Context, plan *migap
...
+	// Known default node selector values. Ignore these if we spot them on Pods
+	defaultNodeSelectors := map[string]string{
+		"node-role.kubernetes.io/compute": "true",
+	}
 	for _, pod := range podList {
-		if len(pod.Spec.NodeSelector) > 0 || len(pod.Spec.NodeName) > 0 {
-			return true
+		// Return true if pod has node selectors not in the default list
+		if pod.Spec.NodeSelector != nil {
+			for defaultSelector, _ := range defaultNodeSelectors {
```
djwhatle, line 407: This has a bug. Fixing.

pranavgaikwad (approved):
LGTM

PR #1078, opened 2021-04-16
Add parsing of log messages for failed pods in DVMP

pranavgaikwad:
DVMP didn't parse the log messages when Pod went to failed state. This made DVMP lose some of the progress values even when they are clearly present in the LogMessage of the exited container.


alaypatel07 (approved):
ACK

PR #1083, opened 2021-04-21
Move repetitive log messages to V4 in rsync reconcile

djwhatle:
Since we modified this phase to take on more responsibilities in the DVM (and to run repeatedly), I think the original log messages I added have become a large source of noise since they run on every reconcile and probably drown out the signal that the user is looking for about what the running Rsync operations are doing.

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -1697,33 +1697,33 @@ func (t *Task) prepareRsyncPodRequirements(srcClient compat.Client) ([]rsyncClie
 	if err != nil {
 		return req, liberr.Wrap(err)
 	}
-	t.Log.Info("Getting image for Rsync client Pods that will be created on source MigCluster")
+	t.Log.V(4).Info("Getting image for Rsync client Pods that will be created on source MigCluster")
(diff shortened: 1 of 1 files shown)


pkg/controller/directvolumemigration/rsync.go
```diff
@@ -2162,14 +2162,19 @@ func (t *Task) reconcileRsyncOperationState(client compat.Client, req *rsyncClie
...
+				t.Log.Info("Previous attempt of Rsync failed, created a new Rsync Pod", "pvc", operation, "attempt", operation.CurrentAttempt)
 			} else {
 				operation.Failed = currentStatus.failed
 				operation.Succeeded = currentStatus.succeeded
 				if operation.IsComplete() {
 					t.Log.Info(
 						fmt.Sprintf("Rsync operation completed after %d attempts", operation.CurrentAttempt),
 						"pvc", operation, "failed", operation.Failed, "succeded", operation.Succeeded)
+				} else {
+					t.Log.Info("Rsync operation is still running. Waiting for completion",
```
djwhatle, line 2174: I also added this log message to give the user an idea of why the controller isn't proceeding.

pranavgaikwad (approved):

PR #1084, opened 2021-04-21
Add events to discovery service, queryable by involvedObject UID

djwhatle:
Collecting events in the discovery service allows the tree view to become a lot more useful, we'll be able to tally up abnormal events related to migration resources and show them overlaid on each line in the tree view.

**New API**
To query all events for a particular resource, you can use its UID.
E.g.
```
# Get events for object with UID
curl http://localhost:8080/namespaces/:ns1/events/:involvedobjectuid

# Get events for the object with UID dc46e3b1-b290-11eb-8a86-16b833b160e9 (on any cluster, any namespace)
curl http://localhost:8080/namespaces/openshift-migration/events/dc46e3b1-b290-11eb-8a86-16b833b160e9
```

Todo:

- [x] Add events to discovery
- [x] Remove events via watch mechanism according to event-ttl policy (default 3 hrs)
- [x] Add a config switch to disable event collection if this turns out to be problematic on large clusters

### DIFF
--- a/pkg/controller/discovery/model/event.go
+++ b/pkg/controller/discovery/model/event.go
@@ -3,7 +3,7 @@ package model
 import (
 	"encoding/json"
 
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 //
@@ -14,7 +14,7 @@ type Event struct {
 
 //
 // Update the model `with` a k8s Event.
-func (m *Event) With(object *v1.Event) {
+func (m *Event) With(object *corev1.Event) {
 	m.UID = string(object.UID)
 	m.Version = object.ResourceVersion
 	m.Namespace = object.Namespace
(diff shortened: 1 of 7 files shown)


pkg/controller/discovery/model/event.go
```diff
@@ -0,0 +1,93 @@
...
+
+//
+// Update the model `with` a k8s Event.
+func (m *Event) With(object *v1.Event) {
+	m.UID = string(object.UID)
+	m.Version = object.ResourceVersion
+	m.Namespace = object.Namespace
+	m.Name = object.Name
+	m.labels = Labels{
+		"involvedObjectKind":      object.InvolvedObject.Kind,
```
djwhatle, line 23: Events don't have labels, so I'm deriving some based on the `involvedObject` field on the event to allow us to query for events related to a particular resource.

ibolton336 (approved):
LGTM

djwhatle:
@ibolton336 I re-requested review, just pushed changes that make queries work differently. PR comment updated with new instructions.

ibolton336 (approved):
Working great. Tested with the UI today.

PR #1085, opened 2021-04-22
[MIG-627] Add configureable threshold value for pv usage percentage

pranavgaikwad:
This PR adds ability to configure a custom threshold value for PV resizing.

### DIFF
--- a/pkg/controller/miganalytic/volume_adjustment.go
+++ b/pkg/controller/miganalytic/volume_adjustment.go
@@ -30,6 +30,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const DEFAULT_PV_USAGE_THRESHOLD = 3
+
 // DFCommandExecutor defines an executor responsible for running DF
 type DFCommandExecutor interface {
 	Execute(map[string][]MigAnalyticPersistentVolumeDetails) ([]DFOutput, error)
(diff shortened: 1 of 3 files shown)


pkg/controller/miganalytic/volume_adjustment.go
```diff
@@ -30,6 +30,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const DEFAULT_PV_USAGE_THRESHOLD = 3
```
djwhatle, line 33: With this 3% threshold, is it possible that folks will end up with a new PV getting provisioned that is smaller than the one they had before?

djwhatle (approved):
VISACK, just a question about how PV resize works.

I think users would find it useful to have the freedom to resize individual PVs however they want and not have to choose a system-wide setting, don't remember if that's possible with the system today.

pkg/controller/miganalytic/volume_adjustment.go
pranavgaikwad: @djwhatle the users will get `max(93 gigs, 10 gigs, 100 gigs) = 100 gigs`. `93 gigs` comes from `90 gigs + 3%*100 gigs`. This `3%` becomes configureable with this PR.

djwhatle: +1, that makes sense.

PR #1089, opened 2021-04-23
Bug 1952903: fix rsync pods (daemon and client) to use dns safe volume name

alaypatel07:
This change makes sure:

1. Rsyncd configmap has DNS safe volume module name
2. Rsyncd pod has DNS safe volume name and mount path
3. Rsync client pod has DNS safe volume name and mount path

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -16,6 +16,7 @@ import (
 	"sync"
 	"text/template"
 	"time"
+	"unicode/utf8"
 
 	liberr "github.com/konveyor/controller/pkg/error"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
@@ -226,7 +227,11 @@ func (t *Task) createRsyncConfig() error {
 	for ns, vols := range pvcMap {
 		pvcList := []pvc{}
 		for _, vol := range vols {
-			pvcList = append(pvcList, pvc{Name: vol.Name})
+			dnsSafeName, err := getDNSSafeName(vol.Name)
+			if err != nil {
+				return err
+			}
+			pvcList = append(pvcList, pvc{Name: dnsSafeName})
 		}
 		// Generate template
 		rsyncConf := rsyncConfig{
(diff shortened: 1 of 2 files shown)


alaypatel07:
will post a branch for this after testing and merging this.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -226,7 +227,11 @@ func (t *Task) createRsyncConfig() error {
 	for ns, vols := range pvcMap {
 		pvcList := []pvc{}
 		for _, vol := range vols {
-			pvcList = append(pvcList, pvc{Name: vol.Name})
+			dnsSafeName, err := getDNSSafeName(vol.Name)
```
djwhatle, line 230: Are there different rules for naming Pods and Volumes?

alaypatel07: yeah sorry I buried the link in the BZ, will put it in the PR description: https://github.com/kubernetes/kubernetes/blob/9cdd673a8b784f22709b133dde1da16332b14889/staging/src/k8s.io/api/core/v1/types.go#L39, it seems like the volume name rule is special

pkg/controller/directvolumemigration/rsync_test.go
```diff
@@ -1150,3 +1150,43 @@ func TestTask_processRsyncOperationStatus(t *testing.T) {
...
+		{
+			name:       "longer than 63 chars",
+			volumeName: strings.Repeat("1", 64),
+			want:       strings.Repeat("1", 63),
+			wantErr:    false,
+		},
+		{
+			name:       "valid length but invalid characters",
+			volumeName: strings.Repeat("1", 10) + "." + strings.Repeat("1", 10),
+			want:       strings.Repeat("1", 10) + "-" + strings.Repeat("1", 10),
```
pranavgaikwad, line 1176: Would you mind if I ask you to add tcs for all special chars you added in the regex? Just so we dont break the behavior mistakenly :)

pranavgaikwad (changes requested):
Requesting additional TCs

pkg/controller/directvolumemigration/rsync.go
djwhatle: **Most resource types require a name that can be used as a DNS subdomain name as defined in RFC 1123. This means the name must:**
- contain no more than 253 characters
- contain only lowercase alphanumeric characters, '-' or '.'
- start with an alphanumeric character
- end with an alphanumeric character

**Some resource types require their names to follow the DNS label standard as defined in RFC 1123. This means the name must:**
- contain at most 63 characters
- contain only lowercase alphanumeric characters or '-'
- start with an alphanumeric character
- end with an alphanumeric character

pkg/controller/directvolumemigration/rsync_test.go
alaypatel07: updated

djwhatle (approved):
VISACK

pranavgaikwad (approved):

PR #1091, opened 2021-04-28
Add namespace duplication validation to src/dest ns validate methods

sseago:


djwhatle:
Here's what this looks like on the API:
image

And here's what it looks like in the UI
image

pkg/controller/migplan/validation.go
```diff
@@ -47,6 +47,8 @@ const (
 	NsLimitExceeded                            = "NamespaceLimitExceeded"
 	NsLengthExceeded                           = "NamespaceLengthExceeded"
 	NsHaveNodeSelectors                        = "NamespacesHaveNodeSelectors"
+	DuplicateNsOnSourceCluster                 = "NamespaceNotFoundOnSourceCluster"
```
djwhatle: Is this a copy paste error? The content of the string doesn't seem to match the const name.

pkg/controller/migplan/validation.go
```diff
@@ -683,10 +694,25 @@ func (r ReconcileMigPlan) validateSourceNamespaces(plan *migapi.MigPlan) error {
...
+
+	namespaces := plan.GetDestinationNamespaces()
+	duplicates := listDuplicateNamespaces(namespaces)
+	if len(duplicates) > 0 {
+		plan.Status.SetCondition(migapi.Condition{
+			Type:     DuplicateNsOnDestinationCluster,
+			Status:   True,
+			Reason:   DuplicateNs,
+			Category: Critical,
+			Message:  "Duplicate destination cluster namespaces [] in migplan.",
```
djwhatle, line 709: could we change "migplan" here to migration plan?

djwhatle (changes requested):

pkg/controller/migplan/validation.go
sseago: Yes. I've already fixed it in the commit.

pkg/controller/migplan/validation.go
sseago: Sure. I was thinking that API-level errors were better using the actual api resource names, but if we're using "migration plan" elsewhere I can do that here.

sseago: Ahh. We're inconsistent here. It looks liike we *do* use `migstorage` and `migcluster` in other condition messages, but we also use `migration plan` elsewhere.

djwhatle: Yeah, that was my conclusion as well. It could probably go either way to be honest.

djwhatle (approved):

djwhatle:
@sseago ACK, merge whenever you're ready.

PR #1092, opened 2021-04-28
Add change namespace support to DVM

sseago:


pkg/controller/directvolumemigration/namespace.go
```diff
@@ -37,7 +39,7 @@ func (t *Task) ensureDestinationNamespaces() error {
 		// Create namespace on destination with same annotations
 		destNs := corev1.Namespace{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:        ns,
+				Name:        nsNames[1],
```
djwhatle, line 42: Seems like we should verify the length of this array before accessing numeric keys.

djwhatle:
@sseago can you share an example migplan that shows how to test this PR?

sseago:
@djwhatle So you could look at my sprint demo video to see how I did it, but basically what I did was create a migplan via the UI for a single namespace with PVCs and configured it for filesystem copy via direct migration. Then `oc edit` the migplan and change the namespace list item from `mynamespace` to `mynamespace:destnamespace` and then go back to the UI and migrate.

djwhatle:
@sseago I am seeing that when I retarget the `rocket-chat` namespace to `rocketchat-2`, I end up with both the original and new namespace getting created on my target cluster. I'm guessing this may be due to lack of CRD changes on my end. Looks like this PR is missing a corresponding operator PR.

image


image

djwhatle (changes requested):
Needs linked operator PR for CRD changes.

sseago:
@devarshshah15 I hadn't created the operator PR yet, mainly because I wasn't sure whether someone with more DVM experience was going to suggest a different approach. But I can create the operator PR from the CRD changes here shortly. In the meantime, you could probably just `oc apply` the PR changes from here. In any case, what you're seeing in your cluster (both namespaces created) is what I observed without this code. DVM migrates to the src namespace, and everything else migrates to the dest namespace.

djwhatle:
This PR made me realize that if we're going to support change of namespace properly, several changes will be needed to the tree view to support the `srcNS:targetNS` mapping. Right now it's not working as expected, resources are missing in the tree.

sseago:
@djwhatle operator changes are here: https://github.com/konveyor/mig-operator/pull/668

djwhatle:
I was able to verify that namespace change worked for DVM. All content was moved from `sock-shop` to `sock-shop-2` by setting the entry in `MigPlan.Spec.Namespaces[]` to `sock-shop:sock-shop-2`.

djwhatle (changes requested):

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -27,7 +27,8 @@ import (
 type PVCToMigrate struct {
 	*kapi.ObjectReference `json:",inline"`
 	TargetStorageClass    string                            `json:"targetStorageClass"`
-	TargetAccessModes     []kapi.PersistentVolumeAccessMode `json:"targetAccessModes,omitEmpty"`
+	TargetAccessModes     []kapi.PersistentVolumeAccessMode `json:"targetAccessModes,omitempty"`
+	TargetNamespace       string                            `json:"targetNamespace,omitempty"`
 	Verify                bool                              `json:"verify,omitEmpty"`
```
pranavgaikwad, line 32:
```suggestion
	Verify                bool                              `json:"verify,omitempty"`
```

pkg/controller/directvolumemigration/namespace.go
```diff
@@ -22,13 +23,14 @@ func (t *Task) ensureDestinationNamespaces() error {
 	}
 
 	// Get list namespaces to iterate over
-	nsMap := t.getPVCNamespaceMap()
-	for ns, _ := range nsMap {
+	nsMap := t.getBothPVCNamespaceMap()
+	for bothNs, _ := range nsMap {
+		nsNames := strings.Split(bothNs, ":")
```
pranavgaikwad, line 28: If I understand this correctly, `bothNs` is a key in the map which is stored after splittting the string with `:` delimeter. Looks like this operation will be redundant. For clarity, I suggest iterating over the map as:

```golang
for originalNs, mappedNs := range nsMap {
```

pranavgaikwad: Sorry, I got confused between the two different methods. The above isn't true. Please ignore.

config/crds/migration.openshift.io_directvolumemigrations.yaml
```diff
@@ -131,7 +133,6 @@ spec:
                   verify:
                     type: boolean
                 required:
-                - targetAccessModes
```
pranavgaikwad, line 134: I know that this is a required value on a PVC object. Any reason why we would like to remove this as required from the API?

pkg/controller/directvolumemigration/namespace.go
```diff
@@ -22,13 +23,14 @@ func (t *Task) ensureDestinationNamespaces() error {
...
 		// Get namespace definition from source cluster
 		// This is done to get the needed security context bits
 
 		srcNS := corev1.Namespace{}
-		key := types.NamespacedName{Name: ns}
+		key := types.NamespacedName{Name: nsNames[0]}
```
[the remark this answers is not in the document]
sseago: Sure. The way we call this basically guarantees we'll have "src:target", even if they're the same in this context, but it can't hurt to add the additional check. If len(nsNames) is 1 we can assume dest name == nsnames[0]. I'll update it.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -742,6 +745,42 @@ func (t *Task) getPVCNamespaceMap() map[string][]pvcMapElement {
...
+
+func (t *Task) getBothPVCNamespaceMap() map[string][]pvcMapElement {
+	nsMap := map[string][]pvcMapElement{}
+	for _, pvc := range t.Owner.Spec.PersistentVolumeClaims {
+		srcNs := pvc.Namespace
+		destNs := srcNs
+		if pvc.TargetNamespace != "" {
+			destNs = pvc.TargetNamespace
+		}
+		bothNs := srcNs + ":" + destNs
```
pranavgaikwad, line 773: Don't feel strongly about it, but something to consider, the only difference between the two functions `get<>PVCMap()` is the above 1 line. How about we keep only the `getBothPVCNamespaceMap()` function which will contain mapping by default, and a set of helper functions to simply split the key in the map, something like `getSourceNs()` & `getDestNs()`. The helper functions will abstract the mapping logic. So anyone else who isn't familar with the change can simply re-use the helpers.

pranavgaikwad: Also, @djwhatle's suggestion about length of array can go into the helper.

pranavgaikwad (changes requested):
Added some comments, functionality/approch LGTM

pkg/apis/migration/v1alpha1/directvolumemigration_types.go
sseago: Ahh, good catch. I noticed the problem with AccessModes but it's here for Verify as well.

pkg/controller/directvolumemigration/namespace.go
sseago: Yeah, unfortunately golang doesn't allow slices as array keys. I want to index this by (srcNS, destNS) pairs, but the only way to do that is embedding it in a string.

config/crds/migration.openshift.io_directvolumemigrations.yaml
sseago: It shouldn't be required. This is only used when access mode is being changed. The only reason it was required before was the "omitEmpty" typo (instead of "omitempty"). Let me verify that where it's used in the code it's not expected, though.

sseago: Actually, maybe the way DVM is written they're always set here. Hmm. Maybe omitEmpty shouldn't have been set on the CRD?

sseago: Looks like while MigPlan selection makes AccessModes optional, DVM copies the PVC's access modes if the migplan doesn't override them.

djwhatle:
@sseago I attached an issue with the Jira ticket name in the title. This automatically creates a Jira subtask for your work tied to the ticket. In the future, doing this helps Erik and John to know where the actual work is being done on Jira items.

pkg/controller/directvolumemigration/rsync.go
sseago: @pranavgaikwad So if I"m understanding correctly, you're saying that of the 3 get...PVCMap functions, we only retain the getBoth() one and that the individual callers will call getSourceNs() or getDestNS() on each key element returned in their respective loops? If that's correct, then yes, that makes sense and I can make that change.

djwhatle (approved):

PR #1094, opened 2021-04-29
Use meaningful names for velero backups and restores

djwhatle:
image

### DIFF
--- a/pkg/controller/migmigration/backup.go
+++ b/pkg/controller/migmigration/backup.go
@@ -459,7 +459,7 @@ func (t *Task) buildBackup(client k8sclient.Client, backupTypePrefix string) (*v
 
 	// Construct a bakcup name like "$migrationname-initial-" or "$migrationname-stage-".
 	// Truncate the name to 57 chars to leave room for 5 char generateName suffix. Limit is 63 chars.
-	truncatedGenerateName := fmt.Sprintf("%.57s", fmt.Sprintf("%s-%s-", t.Owner.GetName(), backupTypePrefix))
+	truncatedGenerateName := fmt.Sprintf("%.57s-", fmt.Sprintf("%s-%s", t.Owner.GetName(), backupTypePrefix))
 
 	backup := &velero.Backup{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


pkg/controller/migmigration/backup.go
```diff
@@ -456,10 +456,15 @@ func (t *Task) buildBackup(client k8sclient.Client) (*velero.Backup, error) {
 	if err != nil {
 		return nil, liberr.Wrap(err)
 	}
+
+	// Construct a bakcup name like "$migrationname-initial-" or "$migrationname-stage-".
+	// Truncate the name to 57 chars to leave room for 5 char generateName suffix. Limit is 63 chars.
+	truncatedGenerateName := fmt.Sprintf("%.57s-", fmt.Sprintf("%s-%s", t.Owner.GetName(), backupTypePrefix))
```
pranavgaikwad, line 462: It looks like we are truncating the end product of string concatenation. Can we truncate the `t.Owner.GetName()` alone so we never truncate the suffix `stage` or `initial` since we know the length of these strings beforehand, we can truncate the migration name to 50 chars `57 - max('stage', 'initial')`

pranavgaikwad (changes requested):
Overall looks good, a question for truncating strings.

pkg/controller/migmigration/backup.go
djwhatle: Good idea, will change.

djwhatle: @pranavgaikwad changed. This generates a 57 char string now. 

If you'd like to try it out with some examples. https://play.golang.org/p/coW_c-fUnJ0

Also note that generateName gracefully handles names > 63 chars, it just truncates on its own. I verified against my  cluster.

pkg/controller/migmigration/backup.go
```diff
@@ -456,10 +456,17 @@ func (t *Task) buildBackup(client k8sclient.Client) (*velero.Backup, error) {
 	if err != nil {
 		return nil, liberr.Wrap(err)
 	}
+
+	// Construct a restore name like "$migrationname-54823-initial" or "$migrationname-54823-stage".
+	// This will produce a 63 character string max. Note that generateName gracefully handles strings >63 char.
+	fmtString := fmt.Sprintf("%%.%ds", 55-len(backupTypePrefix))
```
pranavgaikwad, line 462: OMG! :rocket:

pranavgaikwad (approved):

PR #1096, opened 2021-04-30
 Bug 1952903: fix rsync pods (daemon and client) to use dns safe volu…

alaypatel07:
…me name

This is the cherry-pick from #1089


djwhatle (approved):
VISACK

PR #1097, opened 2021-04-30
Bug 1934877: Add configuration for liveness and readiness timeout values for mig registry

pranavgaikwad:


pranavgaikwad:
Please also review  https://github.com/konveyor/mig-operator/pull/664

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -47,6 +47,12 @@ const (
 	ClosedIndexField = "closed"
 )
 
+// Registry deployment constants
+const (
+	MigRegistryDefaultLivenessTimeout  = 3
+	MigRegistryDefaultReadinessTimeout = 3
```
sseago, line 53: The above two values don't seem to be used anywhere. It looks like `RegistryDefaultProbeTimeout` is used instead. Should these be removed?

sseago (commented):
Looks good overall, other than the fact that MigRegistryDefaultLivenessTimeout and MigRegistryDefaultReadinessTimeout are defined and never used, while RegistryDefaultProbeTimeout seems to be used instead. We shoud probably either remove them or use them somewhere instead of the general default value.

pkg/apis/migration/v1alpha1/migplan_types.go
pranavgaikwad: @sseago yeah forgot to remove them. will do now

sseago (approved):

Issue #1080, opened 2021-04-19
[MIG-597] Discovery service tree view should show PVCs, PVs, Routes associated with migration


djwhatle:
SSIA

PR #1098, opened 2021-04-30
Add stage pod associated PVCs and PVs to tree view

djwhatle:
I missed this in an earlier pass. Related to #1080

### DIFF
--- a/pkg/controller/discovery/web/plan.go
+++ b/pkg/controller/discovery/web/plan.go
@@ -561,7 +561,7 @@ func (t *PlanTree) addRestores(migration *model.Migration, parent *TreeNode) err
 	return nil
 }
 
-// Add direct volume pods
+// Add migration associated pods (stage, registry, rsync, stunnel)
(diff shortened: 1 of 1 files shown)


pkg/controller/discovery/web/plan.go
```diff
@@ -599,16 +599,72 @@ func (t *PlanTree) addMigrationPodsForCluster(cluster model.Cluster, migration *
...
+
+		parent.Children = append(parent.Children, node)
+	}
+	return nil
+}
+
+//
+// Add PVCs related to a pod
+func (t *PlanTree) addPVCsForPod(pod *model.Pod, parent *TreeNode) error {
+	cluster := t.cluster.source
```
djwhatle, line 624: Hmm, just realizing this is an odd assumption. We could also have stage pods on the target cluster. Will fix.

pranavgaikwad: @djwhatle Are you done changing this?

djwhatle: Not yet, sorry I realized the problem right after I requested review. Will ping you again later.

PR #1099, opened 2021-04-30
Provide call to action in stalled message, reduce time to 3 minutes

djwhatle:
- Call to action helps the user know what to do next
 - Warning showing up sooner means that inexperienced user of the tool doesn't waste time waiting for something to happen

image

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -112,7 +112,7 @@ data:
    {{ end }}
 `
 
-func (t *Task) areRsyncTransferPodsRunning() (arePodsRunning bool, nonRunningPod *corev1.Pod, e error) {
+func (t *Task) areRsyncTransferPodsRunning() (arePodsRunning bool, nonRunningPods []*corev1.Pod, e error) {
 	// Get client for destination
 	destClient, err := t.getDestinationClient()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/directvolumemigration/rsync.go
```diff
@@ -155,17 +155,17 @@ func (t *Task) areRsyncTransferPodsRunning() (bool, error) {
 							"pod", path.Join(pod.Namespace, pod.Name),
 							"podPhase", pod.Status.Phase,
 							"podConditionMessage", podCond.Message)
-						return false, nil
+						return false, &pod, nil
```
pranavgaikwad, line 158: This is a good improvement. I think we should collect _all_ non-running Rsync Pods instead of returning on the first one. What do you think?

pranavgaikwad (changes requested):

pkg/controller/directvolumemigration/rsync.go
djwhatle: +1, I will adapt the PR.

djwhatle: @pranavgaikwad were you thinking we would also provide all of the commands to check on individual Pods, or just one?

pranavgaikwad: @djwhatle the typical pattern I have followed elsewhere (PV resizing/Retry) is to raise a warning with names of problematic resources. I will defer to you on this one. Maybe one command with placeholders and names of the resources followed by that, or perhaps, just text explaining how to use debug view to get to the root of the problem should be fine I think.

pranavgaikwad: Now that I am thinking about it, I think redirecting users to the debug view should be the default go-to pattern for such warnings. Open to discuss pros/cons of that, but seems like a good way to get users used to the debug tree.

djwhatle: > Now that I am thinking about it, I think redirecting users to the debug view should be the default go-to pattern for such warnings. Open to discuss pros/cons of that, but seems like a good way to get users used to the debug tree.

I feel this pointer should exist in the UI, but not on the API messaging. It would be weird for a user who is doing a purely CLI driven migration to get directed to the UI, and might not even be an option for them. We can provide this indicator anytime a warning shows up during a migration (see debug tree for more info).

djwhatle:
@pranavgaikwad updated in response to your review.

This command is now printed since it could get out of hand to have individual commands for each pod:

```
Rsync Transfer Pod(s) [ns1/rsync-pod-1, ns2/rsync-pod-2, ns3/rsync-pod-3] on destination cluster have not started Running within 3 minutes.
Run this command on the destination cluster and check the Pod events.  oc describe pods --selector purpose=RsyncTransferPod --all-namespaces

djwhatle:
This needs testing still.

pranavgaikwad (approved):
LGTM

djwhatle:
When we include all of the Pods, I think the message is getting unwieldy in the UI, it looks kind of bad IMO, and I also realized that the label selector I was providing isn't specific enough:

image

I modified it to look like so:

image

djwhatle:
image

PR #1100, opened 2021-05-03
Add index() tag to support non-unique indexes.

jortel:
Add index() tag to support non-unique indexes.
Example:
```
type Person struct {
    First string `sql:"index(name)`
    Last string  `sql:"index(name)`
}
```


djwhatle (approved):
LGTM. Tested rebased on master and everything is still working as expected.

@jortel I'll work with you on how to apply this to the event filtering use case. Any examples you have handy?

PR #1111, opened 2021-05-13
Bug 1960410: use appsv1 instead of old extensions/v1beta1, rely on compat client

alaypatel07:


djwhatle (approved):
VISACK, I wonder if we should consider changing the compat client to handle both upgrades and downgrades.

alaypatel07:
> VISACK, I wonder if we should consider changing the compat client to handle both upgrades and downgrades.

I actually like the concept of just handling downgrades. That way devs can "blindly" use the latest version(v1) in most API and not worry about the version unavailability. v2 does not look to be on the horizon for a long time so I am not sure if upgrade concerns are too big a deal as of now.

PR #1112, opened 2021-05-17
PV Resizing: Give precedance to usage percentage

### LINKED ISSUE #1074: PV Resizing: Actual used percentage should be given precedance over actual volume size
**Describe the bug**
In some cases, it is possible that the provisioned volume is relatively big as compared to the requested size. However, if the usage of the volume remains low, we should give precedance to used volume size. 

**To Reproduce**
1. Create a NFS volume by requesting 1GB size in mig-agnosticd 3.x cluster. The actual disk size will probably be a lot more than 1 GB
2. Create a MigPlan for the namespace of this volume. 
3. Look at MigAnalytic created for the MigPlan and see that the proposed size is huge than the actual usage percentage.

**Expected behavior**
The usage percentage should be given precedance at all times to avoid provisioning volumes when they are not been used.

pranavgaikwad:
Fixes #1074


pkg/controller/miganalytic/volume_adjustment_test.go
```diff
@@ -270,11 +270,11 @@ func TestPersistentVolumeAdjuster_calculateProposedVolumeSize(t *testing.T) {
 				actualCapacity:    resource.MustParse("200"),
 				requestedCapacity: resource.MustParse("20"),
 			},
-			wantProposedSize: resource.MustParse("200"),
+			wantProposedSize: resource.MustParse("106"),
```
djwhatle, line 273: I see that 103% is desired but `wantProposedSize: "106"`, is this a typo?

pkg/controller/miganalytic/volume_adjustment.go
```diff
@@ -265,13 +265,14 @@ func (pva *PersistentVolumeAdjuster) calculateProposedVolumeSize(usagePercentage
...
 	if actualCapacity.Cmp(maxSize) == 1 {
-		maxSize = actualCapacity
 		reason = migapi.VolumeAdjustmentCapacityMismatch
 	}
 
 	if volumeSizeWithThreshold.Cmp(maxSize) == 1 {
 		maxSize = volumeSizeWithThreshold
-		reason = migapi.VolumeAdjustmentUsageExceeded
+		if usagePercentage+int64(pva.getVolumeUsagePercentageThreshold()) > 100 {
+			reason = migapi.VolumeAdjustmentUsageExceeded
```
djwhatle, line 274: Question: do we let users know that we are planning to downsize their volumes from the requested capacity?

pranavgaikwad: @djwhatle Good point. No we do not.

pkg/controller/miganalytic/volume_adjustment_test.go
pranavgaikwad: Yes, a typo and a logic flaw :)

pranavgaikwad: Ohh now that I realized, this is not a logic issue. 106 is the correct value, the reference for usage percentage is the actual volume size which is 200 in this case, which makes the 3% = 6Gi buffer.

PR #1113, opened 2021-05-17
dep update for velero 1.6 and openshift-velero-plugin

sseago:


pkg/compat/client.go
```diff
@@ -109,31 +108,59 @@ func (c client) MinorVersion() int {
...
 // supportedVersion will determine correct version of the object provided, based on cluster version
-func (c client) supportedVersion(obj runtime.Object) runtime.Object {
+func (c client) supportedVersion(obj k8sclient.Object) k8sclient.Object {
 	if c.Minor < 16 {
 		switch obj.(type) {
 
 		// Deployment
 		case *appsv1.Deployment:
 			return &appsv1beta1.Deployment{}
-		case *appsv1.DeploymentList:
```
shawn-hurley, line 119: Why did re move the list types here?

pkg/compat/client.go
```diff
@@ -13,7 +13,6 @@ import (
 	batchv1beta "k8s.io/api/batch/v1beta1"
 	batchv2alpha "k8s.io/api/batch/v2alpha1"
 	extv1beta1 "k8s.io/api/extensions/v1beta1"
-	"k8s.io/apimachinery/pkg/runtime"
```
shawn-hurley, line 16: I think that we can keep this?? 

https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#Object

sseago: So the issue seems to be that pretty much across the board, this version of controller-runtime has replaced basically any use of runtime.Object with either client.Object or client.ObjectList.

pkg/controller/directimagemigration/directimagemigration_controller.go
```diff
@@ -110,11 +106,9 @@ type ReconcileDirectImageMigration struct {
 // +kubebuilder:rbac:groups=apps,resources=deployments/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=migration.openshift.io,resources=directimagemigrations,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=migration.openshift.io,resources=directimagemigrations/status,verbs=get;update;patch
-func (r *ReconcileDirectImageMigration) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	ctx := context.Background()
+func (r *ReconcileDirectImageMigration) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
 	// Fetch the DirectImageMigration instance
-	log.Reset()
```
shawn-hurley, line 116: @djwhatle I don't remember why we need this, but is this going to be a problem? seems we might have switched loggers

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1044,7 +1044,7 @@ func (t *Task) getfsGroupMapForNamespace() (map[string][]PVCWithSecurityContext,
 			return nil, err
 		}
 		podList := &corev1.PodList{}
-		err = srcClient.List(context.TODO(), &k8sclient.ListOptions{Namespace: ns}, podList)
+		err = srcClient.List(context.TODO(), podList, &k8sclient.ListOptions{Namespace: ns})
```
shawn-hurley, line 1047: if we are here, can move to the k8sclient.InNamespace(ns) helper func?

shawn-hurley: Or at least standardize on one or the other 

cc @djwhatle @jortel

pkg/controller/directvolumemigrationprogress/directpvmigrationprogress_controller_suite_test.go
```diff
@@ -65,11 +66,14 @@ func SetupTestReconcile(inner reconcile.Reconciler) (reconcile.Reconciler, chan
...
+	ctx := context.Background()
+	ctx, stopFunc := context.WithCancel(ctx)
 	wg := &sync.WaitGroup{}
 	go func() {
 		wg.Add(1)
-		g.Expect(mgr.Start(stop)).NotTo(gomega.HaveOccurred())
+		g.Expect(mgr.Start(ctx)).NotTo(gomega.HaveOccurred())
 		wg.Done()
 	}()
+	stopFunc() // this is probably the wrong thing
```
shawn-hurley, line 77: I think the right thing here is to update this to return the context and have the caller handle contexts. 

Are there too many callers for that?

pkg/controller/discovery/controller.go
```diff
@@ -40,16 +40,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-var log logging.Logger
+var log *logging.Logger
```
shawn-hurley, line 43: Why do we need this to be a pointer?

pkg/controller/discovery/web/web.go
```diff
@@ -359,7 +359,7 @@ func (h *BaseHandler) allow(sar auth.SelfSubjectAccessReview) int {
 	post := restClient.Post()
 	post.Resource("selfsubjectaccessreviews")
 	post.Body(&sar)
-	reply := post.Do()
+	reply := post.Do(context.Background())
```
shawn-hurley, line 362: context.TODO() might the more appropriate call here because we have no intention of doing anything with that context at the moment? Thoughts?

pkg/controller/migmigration/migmigration_controller.go
```diff
@@ -107,9 +104,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 	// Plan
 	err = indexer.IndexField(
+		context.Background(),
```
shawn-hurley, line 107: `context.TODO()` for the same reason as above

shawn-hurley (changes requested):
Let's discuss the following changes:

1. Standardize how to add options to the list calls either structs or using the vardic options.
2. Discuss the removal of log.Resets and changes to the logger
3. I think we need to update the caller of StartTestManagerXXXX to handle a context or a context.StopFunc instead of channel.


Other than that, changes look sane to me!

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -129,11 +129,11 @@ func (t *Task) areRsyncTransferPodsRunning() (arePodsRunning bool, nonRunningPod
 		pods := corev1.PodList{}
 		err = destClient.List(
 			context.TODO(),
+			&pods,
 			&k8sclient.ListOptions{
```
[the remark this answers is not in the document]
sseago: So, basically across-the-board I just swapped object and options in the arg order. Is there a good reason to take each substition one-by-one and figure out whether a varadic approach might be better, or would it be less error-prone to keep the single ListOptions arg? I wasn't completely clear on how multiple options, passed separately, would interact, but if there are cases where splitting it makes things clearer, I'd be fine with that too.

pkg/controller/directvolumemigration/rsync.go
sseago: In this particular case, the InNamespace seems like it would be fine. In several cases in this PR I've gone away from using InNamesapce as the newer version of InNamespace doesn't seem to lend itself to cascading calls like the old one did, but in this one case it would be fine.

pkg/controller/directvolumemigrationprogress/directpvmigrationprogress_controller_suite_test.go
sseago: Yeah, to be honest I wasn't really clear on how to fix these tests, hence this comment. The above got the compile errors to disappear, but I know the tests need fixing around this point.

pkg/controller/discovery/web/web.go
sseago: I can do that. Previously, I got comments when I'd used TODO() saying that Background() was more appropriate when we had no intention of changing it later, and TODO() was more appropriate when we weren't sure. @alaypatel07 -- amy thoughts here?

alaypatel07: hmm yeah, I think the way I learnt to use contexts is, if the context will ever be used with something higher level, I use TODO, and if I know I will never touch the context again, I use Background. That way all the TODOs in the codebase refer to contexts that will eventually get meaning and the rest will be Background.  If that is the wrong way of thinking please I know I am curious to learn. 

and yeah I was the one to share and tell this in the team in reviews :) sorry if it was mid-directed.

shawn-hurley: So the question here is what comes next @alaypatel07 is correct:

https://golang.org/pkg/context/#TODO
> Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

and https://golang.org/pkg/context/#Background

> It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.

My thinking is that eventually, we would want to handle contexts in this code path but it still does not feel like a top level context to me so IDK. 

I can be persuaded, but I think we probably want to eventually have some context to the HTTP call?

sseago:
@shawn-hurley OK I've updated the PR with logging fixes and changes to the List options. Still remaining are the fixes to the tests to handle stopping the managers properly.

pkg/compat/client.go
djwhatle: @sseago I would like to know this as well.

djwhatle: Ah, I see it's implemented in 
```
supportedVersionList(obj k8sclient.ObjectList) k8sclient.ObjectList
```

djwhatle: Why the change?

pkg/compat/client.go
```diff
@@ -171,7 +196,7 @@ func (c client) downConvert(ctx context.Context, obj runtime.Object) (runtime.Ob
 
 //
 // upConvert will convert src resource to dst as needed based on cluster version
-func (c client) upConvert(ctx context.Context, src runtime.Object, dst runtime.Object) error {
+func (c client) upConvert(ctx context.Context, src k8sclient.Object, dst k8sclient.Object) error {
```
djwhatle, line 199: @sseago Is `k8sclient.Object` a good substitute for `runtime.Object`?

pkg/controller/directimagemigration/directimagemigration_controller.go
```diff
@@ -69,13 +69,9 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 	err = c.Watch(
 		&source.Kind{Type: &migapi.MigCluster{}},
-		&handler.EnqueueRequestsFromMapFunc{
-			ToRequests: handler.ToRequestsFunc(
-				func(a handler.MapObject) []reconcile.Request {
-					return migref.GetRequests(a, migapi.DirectImageMigration{})
-				}),
-		},
-		//		&ClusterPredicate{}
+		handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
```
djwhatle, line 72: @sseago I guess this is equivalent to the previous content here? It looks different but maybe equivalent.

pkg/controller/directimagemigration/directimagemigration_controller.go
```diff
@@ -130,7 +124,7 @@ func (r *ReconcileDirectImageMigration) Reconcile(request reconcile.Request) (re
 	// Set MigMigration name key on logger
 	migration, err := imageMigration.GetMigrationForDIM(r)
 	if migration != nil {
-		log.SetValues("migMigration", migration.Name)
+		log.Real = log.WithValues("migMigration", migration.Name)
```
djwhatle, line 127: @sseago what is log.Real? Is this the logger we're supposed to use throughout the codebase to make sure we get these keys?

pkg/controller/directvolumemigration/directvolumemigration_controller_test.go
```diff
@@ -17,17 +17,17 @@ limitations under the License.
...
 	"time"
 
-	migrationv1alpha1 "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
-	"github.com/onsi/gomega"
-	"golang.org/x/net/context"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	//migrationv1alpha1 "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
+	//"github.com/onsi/gomega"
+	//"golang.org/x/net/context"
```
djwhatle, line 25: I'm thinking we don't want to leave all of this commented out, we should remove it.

pkg/controller/directvolumemigrationprogress/directvolumemigrationprogress_controller_test.go
```diff
@@ -31,12 +31,12 @@ import (
 
 	"github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
-	"github.com/onsi/gomega"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	//"github.com/onsi/gomega"
```
djwhatle, line 34: Same thing here, anywhere we are doing mass commenting out I'd suggest just removing those lines.

pkg/controller/discovery/web/pod.go
```diff
@@ -194,7 +194,7 @@ func (h *LogHandler) getLog(ctx *gin.Context, pod *model.Pod) {
 		return
 	}
 	request := podClient.GetLogs(pod.Name, options)
-	stream, err := request.Stream()
+	stream, err := request.Stream(ctx)
```
djwhatle, line 197: Is this the right kind of context? passing a gin context to a request.Stream?

pkg/controller/discovery/web/web.go
```diff
@@ -350,7 +350,7 @@ func (h *BaseHandler) allow(sar auth.SelfSubjectAccessReview) int {
 		Log.Trace(err)
 		return http.StatusInternalServerError
 	}
-	restClient, err := apiutil.RESTClientForGVK(gvk, restCfg, codec)
+	restClient, err := apiutil.RESTClientForGVK(gvk, false, restCfg, codec)
```
djwhatle, line 353: What is this 'false'?

djwhatle (changes requested):
See comments

pkg/compat/client.go
sseago: @djwhatle the old API just used `runtime.Object` for everything. WIth the new API, it's using `client.Object` and `client.ObjectList`, so we're dealing with single objects and lists under different API methods. That required splitting this out in the same way. Note the return type differs between the two methods split out as well.

pkg/compat/client.go
sseago: It's "good" in the sense that all of the controller-runtime API calls have replaced one with the other, so we need to do the same. Basically client.Object includes both an inline reference to runtime.Object *and* an online meta reference, so it contains more information than what was passed in the older version.

pkg/controller/directimagemigration/directimagemigration_controller.go
sseago: To the best of my knowledge it is equivalent. As with many other things, the API signatures have changed quite a bit between 0.1.z and 0.7.z. Where I've changed things like this it's because what we used to call doesn't exist any more, and this seems to have replaced it.

pkg/controller/directimagemigration/directimagemigration_controller.go
sseago: log.Real is the internal logger that log deals with. In the latest version of the konveyor/controller log package, `SetValues` doesn't exist anymore. `WithValues` does the same thing, but unfortunately the return type requires me to assign it to `log.Real` instead of `log`. In consulting with @jortel though, this is equivalent to what the old code did.

pkg/controller/directvolumemigration/directvolumemigration_controller_test.go
sseago: I've temporarily disabled the tests until I can get them working. I don't think we want to remove the unit tests but to fix them instead, but I was trying to get a version of the PR running so we can merge it for QE. It didn't make musch sense for me to remove (vs. comment out) since the hope is to put it back basically immediately after merging this.

pkg/controller/discovery/web/pod.go
sseago: @djwhatle I'm actually not 100% sure. I know we need a context now. Would Context.TODO() be more appropriate? WIth most of the updated changes, things that previously didn't require a context that now do we've been using Context.TODO() if we didn't already have a context, but if there was a current one I tried to use it, but if this particular context makes no sense here, then we should probably use TODO() instead.

pkg/controller/discovery/web/web.go
sseago: @djwhatle `isUnstructured` is the newly-added param. I'm assuming we want it to be false unless we're explicitly dealing with Unstructured content.

djwhatle:
Haven't run any migrations with this so I'm expecting it's possible for lots of things to break, and possibly many things silently, but since we're up against feature freeze I'm adding my ACK.

pkg/controller/directvolumemigration/directvolumemigration_controller_test.go
djwhatle: I see. Makes sense.

pkg/controller/discovery/web/pod.go
djwhatle: I can't think of what request.Stream would do with a gin ctx, maybe passing it would mean support for cancellation of the gin request cascading through to the stream call.

pkg/controller/discovery/web/web.go
djwhatle: ACK

djwhatle (approved):
VISACK. I feel doing some basic testing of migrations with this PR to validate that things aren't badly broken would be good before sending to QE, but given we are out of time, ACK.

sseago:
@djwhatle Thanks. And yes, we'll need to get going on that basic testing while QE is looking at it, keeping them in the loop when we push changes that fix what we find.

PR #1116, opened 2021-05-26
Bug 1965065: Use DNS safe name for Rsync Pod labels

pranavgaikwad:


alaypatel07:
we would need a cherry-pick for this on release 1.4.5 branch, right?

alaypatel07 (approved):
ACK

PR #1119, opened 2021-06-03
fix adding warnings to dvm if labels are greater than 63 chars

alaypatel07:
- stage the condition before finding it, fixes warn condition for DVM
- return nil pointer if no pod is found for hasAllRsyncClientPodsTimedOut
- if dvmp.status.podPhase is empty, add it to unknown pods

### DIFF
--- a/pkg/controller/directvolumemigration/rsync.go
+++ b/pkg/controller/directvolumemigration/rsync.go
@@ -1217,6 +1217,7 @@ func (t *Task) hasAllProgressReportingCompleted() (bool, error) {
 	t.Owner.Status.FailedPods = []*migapi.PodProgress{}
 	t.Owner.Status.SuccessfulPods = []*migapi.PodProgress{}
 	t.Owner.Status.PendingPods = []*migapi.PodProgress{}
+	unknownPods := []*migapi.PodProgress{}
 	var pendingSinceTimeLimitPods []string
 	pvcMap := t.getPVCNamespaceMap()
 	for bothNs, vols := range pvcMap {
@@ -1261,6 +1262,8 @@ func (t *Task) hasAllProgressReportingCompleted() (bool, error) {
 						pendingSinceTimeLimitPods = append(pendingSinceTimeLimitPods, fmt.Sprintf("%s/%s", podProgress.Namespace, podProgress.Name))
 					}
 				}
+			case dvmp.Status.PodPhase == "":
+				unknownPods = append(unknownPods, podProgress)
 			case !operation.Failed:
 				t.Owner.Status.RunningPods = append(t.Owner.Status.RunningPods, podProgress)
 			}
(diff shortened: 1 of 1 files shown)


pkg/controller/directvolumemigration/rsync.go
```diff
@@ -2450,11 +2456,11 @@ func (t *Task) getLatestPodForOperation(client compat.Client, operation migapi.R
 		} else if _, err := strconv.Atoi(val); err != nil {
 			continue
 		}
-		if pod.CreationTimestamp.After(mostRecentPod.CreationTimestamp.Time) {
-			mostRecentPod = pod
+		if mostRecentPod != nil && pod.CreationTimestamp.After(mostRecentPod.CreationTimestamp.Time) {
```
pranavgaikwad, line 2459: This branch will never be reached because, mostRecentPod is always nil. 
```suggestion
                if mostRecentPod == nil { 
                      mostRecentPod = &pod
                } else if pod.CreationTimestamp.After(mostRecentPod.CreationTimestamp.Time) {
```

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1261,6 +1262,8 @@ func (t *Task) hasAllProgressReportingCompleted() (bool, error) {
 						pendingSinceTimeLimitPods = append(pendingSinceTimeLimitPods, fmt.Sprintf("%s/%s", podProgress.Namespace, podProgress.Name))
 					}
 				}
+			case dvmp.Status.PodPhase == "":
```
pranavgaikwad, line 1265: This is not needed if we fix the wrong nil pointer down below right?

pranavgaikwad (changes requested):

pkg/controller/directvolumemigration/rsync.go
alaypatel07: not sure, when you see `dvmp.Status.PodPhase == ""` what is the right thing to do?

pranavgaikwad: no worries. we will just let it be here I guess for safety. cant think of a scenario atm. doesnt hurt to keep

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -1281,7 +1285,7 @@ func (t *Task) hasAllProgressReportingCompleted() (bool, error) {
 			Message:  pendingMessage,
 		})
 	}
-	return !isAnyRunning && !isAnyPending && isCompleted, nil
+	return !isAnyRunning && !isAnyPending && isCompleted && !isAnyUnknown, nil
```
shawn-hurley, line 1288: nit: switch isAnyUnknown and isCompleted so all the negatives are before the one that is not?

Might not be a real thing just a thought

alaypatel07: +1

alaypatel07: Updated

pranavgaikwad (approved):

PR #1124, opened 2021-06-07
Bug 1966433: Use helper to retrieve source namespaces in MigAnalytics

pranavgaikwad:
`migplan.Spec.Namespaces` contains mapping if target namespace is different.


jmontleon (approved):
ACK

PR #1127, opened 2021-06-16
Rollback: delete leftover DeploymentConfig Pods so that PVCs can terminate

djwhatle:
From the code:
```
func (t *Task) deleteDeploymentConfigLeftoverPods() error {
	// DeploymentConfigs are an exception to the general policy of rollback deleting everything with
	// the label "migrated-by-migplan: migplan-uid" because DCs spawn additional Pods without
	// ownerRefs that will mount PVCs. When we delete the DCs, it doesn't cascade to the
	// deployer and hooks Pods that the DC created, and those Pods sometimes stop PVCs from termianting.
	// This custom deletion routine for DCs is needed to avoid rollback hanging waiting on PVCs
	// to finish terminating.
```


pkg/controller/migmigration/restore.go
```diff
@@ -644,6 +649,77 @@ func (t *Task) deleteMigrated() error {
 	return nil
 }
 
+func (t *Task) deleteDeploymentConfigLeftoverPods() error {
+	// DeploymentConfigs are an exception to the general policy of rollback deleting everything with
+	// the label "migrated-by-migplan: migplan-uid" because DCs spawn additional Pods without
+	// ownerRefs that will mount PVCs. When we delete the DCs, it doesn't cascade to the
+	// deployer and hooks Pods that the DC created, and those Pods sometimes stop PVCs from termianting.
```
sseago, line 656: If explicit hook pod deletion isn't needed here, it might be better to remove the comment about it.

pkg/controller/migmigration/restore.go
```diff
@@ -629,9 +630,13 @@ func (t *Task) deleteStalePVRsOnCluster(cluster *migapi.MigCluster) (int, error)
 
 // Delete namespace and cluster-scoped resources on dest cluster
 func (t *Task) deleteMigrated() error {
-	// TODO: delete 'deployer' Pods associated with DCs even if phase=completed
+	// Delete 'deployer' and 'hooks' Pods that DeploymentConfig leaves behind upon DC deletion.
```
sseago, line 633: If explicit hook pod deletion isn't needed here, it might be better to remove the comment about it.

djwhatle: Will update, thanks.

PR #1133, opened 2021-06-18
Bug 1972903: Add missing return statement

### LINKED ISSUE #1126: DirectVolumeMigration goes to next step even though namespace isn't created
**Describe the bug**
This is created to fix a bug when attempting staging step from source to destination. It was observed that the during CreateDestinationPVCs step , the logs reported namespaces "example" not found.

Release version : 1.4.4

**Proposed Fix**
Adding return err when an error occurs creating the namespace : https://github.com/konveyor/mig-controller/blob/release-1.4.4/pkg/controller/directvolumemigration/task.go#L232

pranavgaikwad:
Fixes #1126

### DIFF
--- a/pkg/controller/directvolumemigration/task.go
+++ b/pkg/controller/directvolumemigration/task.go
@@ -222,7 +222,7 @@ func (t *Task) Run(ctx context.Context) error {
 		// destination
 		err := t.ensureDestinationNamespaces()
 		if err != nil {
-			liberr.Wrap(err)
+			return liberr.Wrap(err)
 		}
 		t.Requeue = NoReQ
 		if err = t.next(); err != nil {
--- a/pkg/controller/directimagemigration/namespace.go
+++ b/pkg/controller/directimagemigration/namespace.go
@@ -65,13 +65,15 @@ func (t *Task) ensureDestinationNamespaces() error {
 				Annotations: srcNS.Annotations,
 			},
 		}
-		existingNamespace := corev1.Namespace{}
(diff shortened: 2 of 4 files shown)


pkg/controller/directvolumemigration/task.go
```diff
@@ -222,7 +222,7 @@ func (t *Task) Run(ctx context.Context) error {
 		// destination
 		err := t.ensureDestinationNamespaces()
 		if err != nil {
-			liberr.Wrap(err)
+			return liberr.Wrap(err)
```
alaypatel07, line 225: will this not fail when the namespace already exists?

pranavgaikwad: @alaypatel07 right, I missed that. adding an error check in the function

pranavgaikwad: @alaypatel07 added the check

pranavgaikwad:
GH Action failure is a known issue today. It should be fixed once GH resolves the Ubuntu archive issue. I ran `make test` on my machine and didn't see any failures.

pkg/controller/directvolumemigration/task.go
alaypatel07: @pranavgaikwad +1, I think there is another check to be made here, i.e. if namespace.finalizer is set. If that is set the namespace is in delete mode i.e. it will be deleted soon when the itinerary is in the next steps

alaypatel07 (approved):
VISIACK

PR #1134, opened 2021-06-21
Bug 1948778: Handle carriage return in Rsync logs for better accuracy

pranavgaikwad:


alaypatel07 (approved):
ACK

pranavgaikwad:
The GH action issue is causing the CI to fail even today. I have opened a ticket and would like to wait for the reply from support folks before merging this PR.

PR #1144, opened 2021-07-01
Bug 1978561: fix pod manager bug for verify, wrap errors

alaypatel07:
The verify logic checks for various pod controllers to see if the number of replicas is as expected. This fixes a bug where the  GVKs for the pod manager were not up-to-date. This also adds a check for NoKindFound errors and ignores them so that the migration does not fail. 

Additionally, this wraps all the errors so that the source of error is easier to find with a trace.


pranavgaikwad (approved):
ACK

shawn-hurley (approved):

PR #1146, opened 2021-07-01
Bug 1972903: Fix FailedItinerary for DIM and DISM

pranavgaikwad:


alaypatel07 (approved):
ACK

rayfordj:
> Bug was not moved to POST but was valid. Something went wrong

Appears BZ did actually move to POST successfully.

pranavgaikwad:
> > Bug was not moved to POST but was valid. Something went wrong
> 
> Appears BZ did actually move to POST successfully.

Actually I re-ran the automation manually, that may have done it

PR #1153, opened 2021-07-12
Initial E2e test for bug #196521

JaydipGabani:
To test this PR, these are the environment variable needed.

```
	EXPOSEDREGISTRYPATH   - path to the exposed registry for the source cluster
	SOURCEURL             - source cluster URL
	SOURCECONFIG          - path to kubeconfig of the source cluster
	HOSTCONFIG            - path to kubeconfig of the host cluster
	AWSBUCKETNAME         - aws bucket name
	AWSSECRETKEY          - aws secret key
	AWSACCESSKEY          - aws access key
        AWSREGION             - aws bucket region
	BACKUPSTORAGEPROVIDER - backup storage provider name
```

This PR assumes that MTC is installed and creates  `controller` (if not present), `migStorage`, `migCluster`, `migPlan` and runs a migration.

Results:
Test fails on 1.4.4
Screenshot from 2021-07-20 15-15-22

Test passes on latest
Screenshot from 2021-07-20 14-15-52


tests/tests_suite_test.go
```diff
@@ -0,0 +1,99 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "E2E Suite")
+}
+
+var sourceClient *kubernetes.Clientset
+var k8sClient client.Client
+var cfg *rest.Config
+
+var _ = BeforeSuite(func() {
+	cfg, err := clientcmd.BuildConfigFromFlags("", filepath.Join("/home/jgabani/Downloads/jgabani0518/ocp4-workshop_jgabani0518_kubeconfig"))
```
alaypatel07, line 31: this should be env var?

JaydipGabani: Yes! There is a lot of refactoring remaining.

djwhatle:
> To test this PR, these are the environment variable needed.
> 
> ```
> 	EXPOSEDREGISTRYPATH - path to the exposed registry for the source cluster
> 	SOURCEURL           - source cluster URL
> 	SOURCECONFIG        - path to kubeconfig of the source cluster
> 	HOSTCONFIG          - path to kubeconfig of the host cluster
> ```

Hey @JaydipGabani , just a thought on these env vars from an ease of use perspective, couldn't we extract some of this information without requiring the developer to set these manually?

- `EXPOSEDREGISTRYPATH` - this is available on the source [migcluster.status.registryPath](https://github.com/konveyor/mig-controller/blob/master/pkg/apis/migration/v1alpha1/migcluster_types.go#L107)
- `SOURCEURL` - this is available on the source migcluster.spec.url
- `SOURCECONFIG` - this could be derived from the source migcluster right?
- `HOSTCONFIG` - this could be derived from the host migcluster right?

It seems like we _could_ extract the needed info by just requiring `SOURCE_MIGCLUSTER_NAME` and `DEST_MIGCLUSTER_NAME`.

I always like to think about how we can reduce burden on users/developers to set a bunch of vars, but if the current approach makes things significantly better/easier in certain ways then that's OK too.

Here's a script that turns an SA into a kubeconfig file using the oc CLI for reference: https://github.com/konveyor/mig-log-reader/blob/master/sa2kubeconfig.sh

JaydipGabani:
> > To test this PR, these are the environment variable needed.
> > ```
> > 	EXPOSEDREGISTRYPATH - path to the exposed registry for the source cluster
> > 	SOURCEURL           - source cluster URL
> > 	SOURCECONFIG        - path to kubeconfig of the source cluster
> > 	HOSTCONFIG          - path to kubeconfig of the host cluster
> > ```
> 
> Hey @JaydipGabani , just a thought on these env vars from an ease of use perspective, couldn't we extract some of this information without requiring the developer to set these manually?
> 
> * `EXPOSEDREGISTRYPATH` - this is available on the source [migcluster.status.registryPath](https://github.com/konveyor/mig-controller/blob/master/pkg/apis/migration/v1alpha1/migcluster_types.go#L107)
> * `SOURCEURL` - this is available on the source migcluster.spec.url
> * `SOURCECONFIG` - this could be derived from the source migcluster right?
> * `HOSTCONFIG` - this could be derived from the host migcluster right?
> 
> It seems like we _could_ extract the needed info by just requiring `SOURCE_MIGCLUSTER_NAME` and `DEST_MIGCLUSTER_NAME`.
> 
> I always like to think about how we can reduce burden on users/developers to set a bunch of vars, but if the current approach makes things significantly better/easier in certain ways then that's OK too.
> 
> Here's a script that turns an SA into a kubeconfig file using the oc CLI for reference: https://github.com/konveyor/mig-log-reader/blob/master/sa2kubeconfig.sh

@djwhatle We are not just running a migration, but we are creating all the dependencies for migration as well, which includes creating `migPlan`, `migCluster` for source and a `migStorage`. To create these objects we need the above information right? or I am going in another direction than initially intended.

djwhatle:
> @djwhatle We are not just running a migration, but we are creating all the dependencies for migration as well, which includes creating `migPlan`, `migCluster` for source and a `migStorage`. To create these objects we need the above information right? or I am going in another direction than initially intended.

Ah I see. Ok, yes in that case what you're asking for makes sense, so please disregard above comment. I wasn't aware that we had settled on creating all of those. Thought we would be leveraging existing MigClusters.

tests/test_helper.go
```diff
@@ -0,0 +1,212 @@
...
+	"k8s.io/client-go/kubernetes"
+	"os"
+	"strings"
+)
+
+// common constants
+const (
+	TestObjectName         = "foo"
+	TestNamespace          = "openshift-migration"
+	TestMigrationNamespace = "ocp-41583-longpvcname"
```
alaypatel07, line 18: do these need to be constants? I would imaging for each name we would have to find a way to get the names of such objects derived from the test name?

tests/test_helper.go
```diff
@@ -0,0 +1,212 @@
...
+
+// migrationcontroller constants
+const (
+	MigrationController = "migration-controller"
+)
+
+func NewMigMigration() *migapi.MigMigration {
+	return &migapi.MigMigration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      TestObjectName,
```
alaypatel07, line 53: the same comment as above, does the name have to be a constant?

tests/test_helper.go
```diff
@@ -0,0 +1,212 @@
...
+		Object: map[string]interface{}{
+			"apiVersion": "migration.openshift.io/v1alpha1",
+			"kind":       "MigrationController",
+			"metadata": map[string]interface{}{
+				"name":      MigrationController,
+				"namespace": TestNamespace,
+			},
+			// TODO: take variables from env
+			"spec": map[string]interface{}{
+				"velero_plugin_fqin":        "quay.io/konveyor/openshift-velero-plugin:latest",
```
alaypatel07, line 168: you'd need a way to override this?

tests/test_helper.go
```diff
@@ -0,0 +1,212 @@
...
+			Name: ns,
+		},
+	}
+}
+
+func GetMigSaToken(sourceClient *kubernetes.Clientset) []byte {
+	ctx := context.TODO()
+	sa, err := sourceClient.CoreV1().ServiceAccounts(TestNamespace).Get(ctx, MigrationController, metav1.GetOptions{})
+	if err != nil {
+		panic(err)
```
alaypatel07, line 200: remove all panics?

tests/test_helper.go
JaydipGabani: Do you mean take the values of variables from user/env right?

JaydipGabani:
> > @djwhatle We are not just running a migration, but we are creating all the dependencies for migration as well, which includes creating `migPlan`, `migCluster` for source and a `migStorage`. To create these objects we need the above information right? or I am going in another direction than initially intended.
> 
> Ah I see. Ok, yes in that case what you're asking for makes sense, so please disregard above comment. I wasn't aware that we had settled on creating all of those. Thought we would be leveraging existing MigClusters.
> 
> For the `host` migcluster, if that is being used in the e2e test, that will always exist by default so you probably shouldn't re-create it.

@djwhatle I am not creating `migCluster` for host, just getting the kubconfig file to interact with the cluster with client-go

tests/test_helper.go
JaydipGabani: Agreed, removed the constants, for now we can pass it to function and get the `migmigartion` and `migplan` with passed name

tests/test_helper.go
JaydipGabani: +1

tests/test_helper.go
JaydipGabani: +1, now taking it from the parameters passed

tests/test_helper.go
alaypatel07: yes

tests/tests_test.go
```diff
@@ -0,0 +1,271 @@
...
+	)
+	BeforeEach(func() {
+		namespaces = []string{"ocp-41583-longpvcname"}
+		migPlan = NewMigPlan(namespaces, migPlanName)
+	})
+
+	JustBeforeEach(func() {
+		Expect(hostClient.Create(ctx, migPlan)).Should(Succeed())
+		Eventually(func() string {
+			fmt.Println("waiting for migPlan to be ready")
```
shawn-hurley, line 37: Consider using a real logger as a follow up :)

tests/tests_test.go
```diff
@@ -0,0 +1,271 @@
...
+		})
+		if err != nil {
+			fmt.Println(err)
+		}
+	})
+
+	Describe("Testing BZ #1965421, PVC name is more than 63 char long", func() {
+		Context("Running a migration", func() {
+
+			BeforeEach(func() {
```
shawn-hurley, line 62: What is the diff between this BeforeEach and the one above? 

https://github.com/konveyor/mig-controller/pull/1153/files#diff-e6060120fd60e25841122a6c7839035ef0b38ee3c856142ae3a864918b666b51R29

tests/tests_test.go
JaydipGabani: +1

tests/tests_test.go
JaydipGabani: The above one was a global one, but yeah we can get rid of the above one!

tests/tests_test.go
```diff
@@ -0,0 +1,267 @@
...
+		namespaces  []string
+	)
+
+	JustBeforeEach(func() {
+		Expect(hostClient.Create(ctx, migPlan)).Should(Succeed())
+		Eventually(func() string {
+			log.Println("waiting for migPlan to be ready")
+			hostClient.Get(ctx, client.ObjectKey{Name: migPlanName, Namespace: MigrationNamespace}, migPlan)
+			l := len(migPlan.Status.Conditions.List)
+			if l > 0 {
```
djwhatle, line 36: `IsReady()`

tests/tests_test.go
```diff
@@ -0,0 +1,267 @@
...
+		})
+		if err != nil {
+			log.Println(err)
+		}
+	})
+
+	Context("Testing BZ #1965421, PVC name is more than 63 char long", func() {
+
+		BeforeEach(func() {
+			migPlanName = "foo"
```
djwhatle, line 58: I think you declared a constant for this above, but then aren't using it here.

tests/tests_test.go
```diff
@@ -0,0 +1,267 @@
...
+					},
+					),
+					Namespace: "ocp-41583-longpvcname",
+				})
+				destinationFile := []string{}
+				for _, p := range podList.Items {
+					podCmd := pods.PodCommand{
+						Pod:     &p,
+						RestCfg: hostCfg,
+						Args:    []string{"sh", "-c", "ls -l data/test"},
```
djwhatle, line 230: @alaypatel07 @pranavgaikwad should we be checksumming the file to make sure it matches what was transferred from the source cluster?

djwhatle: Also, just want to confirm: this file is one that is actually transferred via rsync, and not one that will be re-generated after the YAML is moved around, right?

tests/tests_test.go
```diff
@@ -0,0 +1,267 @@
...
+					}
+					err = podCmd.Run()
+					if err != nil {
+						log.Println(err, "Failed running ls command inside source Pod",
+							"pod", path.Join(p.Namespace, p.Name),
+							"command", "ls data/test")
+					}
+					sourceFile := strings.Split(podCmd.Out.String(), " ")
+					if sourceFile[0] == destinationFile[0] && sourceFile[len(sourceFile)-1] == destinationFile[len(destinationFile)-1] {
```
djwhatle, line 259: Before we start indexing into this, would be good to check length unless it's guaranteed.

djwhatle (changes requested):

tests/tests_test.go
JaydipGabani: `migPlanName` should be different to make sure we don't come across an error of creating `migPlan` with the same name again. That is why I am not using a constant there.

tests/tests_test.go
JaydipGabani: +1

tests/test_helper.go
```diff
@@ -0,0 +1,248 @@
...
+	"k8s.io/client-go/kubernetes"
+	"log"
+	"os"
+	"strconv"
+	"strings"
+)
+
+// common constants
+const (
+	TestObjectName     = "foo"
```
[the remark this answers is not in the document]
JaydipGabani: +1

tests/tests_test.go
JaydipGabani: @djwhatle Nice catch! I am now creating file on source cluster's pod with exec to make sure the file won't be regenerated after YAML is migrated.

tests/tests_suite_test.go
```diff
@@ -0,0 +1,173 @@
...
+	}, time.Minute*5, time.Second).Should(Equal("Ready"))
+
+	migStorage, secret := NewMigStorage()
+	Expect(hostClient.Create(ctx, secret)).Should(Succeed())
+	Expect(hostClient.Create(ctx, migStorage)).Should(Succeed())
+	Eventually(func() string {
+		hostClient.Get(ctx, client.ObjectKey{Name: TestObjectName, Namespace: MigrationNamespace}, migStorage)
+		l := len(migStorage.Status.Conditions.List)
+		if l > 0 {
+			return migStorage.Status.Conditions.List[l-1].Type
```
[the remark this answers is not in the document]
JaydipGabani: +1

tests/tests_test.go
JaydipGabani: +1

tests/test_helper.go
```diff
@@ -0,0 +1,248 @@
...
+			Data: map[string][]byte{
+				// TODO: Get the token with kubecofig of source
+				"saToken": saToken,
+			},
+			Type: TestSecretType,
+		}
+}
+
+// We are assuming that the controller CR is created and controller is running
+func NewMigrationController() *unstructured.Unstructured {
```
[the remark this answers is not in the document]
alaypatel07: @JaydipGabani this controller instance has UI and controller set to true, can we either put this in the name

`NewMigrationControllerForControlCluster`, OR
add params to the function signatures that makes it possible for you to use the name method for both control cluster and remote cluster. I am fine with either.

tests/tests_suite_test.go
```diff
@@ -0,0 +1,165 @@
...
+var _ = BeforeSuite(func() {
+	hostCfg, err = clientcmd.BuildConfigFromFlags("", filepath.Join(os.Getenv(HOSTCONFIG)))
+	if err != nil {
+		log.Println(err)
+	}
+
+	err = migapi.AddToScheme(scheme.Scheme)
+	hostClient, err = client.New(hostCfg, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		log.Println(err)
```
alaypatel07, line 50: same comment as above regarding failing vs logging

tests/tests_suite_test.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	controllerCR = schema.GroupVersionResource{
+		Group:    "migration.openshift.io",
+		Version:  "v1alpha1",
+		Resource: "migrationcontrollers",
+	}
+
+	controller, err := dynamicClient.Resource(controllerCR).Namespace(MigrationNamespace).Get(context.TODO(), MigrationController, metav1.GetOptions{})
+	if err != nil {
+		log.Println(err)
```
alaypatel07, line 73: `errors.IsNotFound(err)`

using the above error check will be a more definitive case of when to create the controller and when to fail because the error is more fatal IMO

tests/tests_suite_test.go
```diff
@@ -0,0 +1,165 @@
...
+			log.Println(err)
+		}
+		phase, found, err := unstructured.NestedString(controller.Object, "status", "phase")
+		if err != nil || !found {
+			log.Println("phase not found or error in status", err)
+		}
+		return phase
+	}, time.Minute*5, time.Second).Should(Equal("Reconciled"))
+
+	migCluster, secret := NewMigCluster(GetMigSaToken(sourceClient))
```
alaypatel07, line 98: I don't think this is valid, what if the `GetMigSaToken()` returns an empty secret? it will fail in the `Eventually` block below because migcluster will never go to a valid state, but that means 5 mins of user time will be spent in a scenerio where we know it will fail beforehand. Can we handle this situation in code?

tests/tests_suite_test.go
```diff
@@ -0,0 +1,165 @@
...
+			Name:      TestClusterSecret,
+			Namespace: ConfigNamespace,
+		},
+	})
+	if err != nil {
+		log.Println(err)
+	}
+
+	if shouldCreate {
+		err = dynamicClient.Resource(controllerCR).Namespace(MigrationNamespace).Delete(ctx, MigrationController, metav1.DeleteOptions{})
```
alaypatel07, line 160: question: does deleting the namespace return immediately? 

I know that deletions of namespace sometimes take a lot of time because it triggers deletion of all the objects in the namespace. It might be good to explore the deletion options like Foreground and Background etc to see what is behaviour we want.

tests/tests_test.go
```diff
@@ -0,0 +1,278 @@
...
+						if err != nil {
+							log.Println(err, "Failed running ls command inside destination Pod",
+								"pod", path.Join(p.Namespace, p.Name),
+								"command", "dd if=/dev/zero of=/data/test/file_1 bs=60M count=1;")
+						}
+					}
+					return string(p.Status.Phase)
+				}
+				return ""
+			}, time.Minute*5, time.Second).Should(Equal("Running"))
```
alaypatel07, line 165: NIT: there is a constant `PodRunning` for this phase in the core/v1 package, you could use that here

tests/tests_test.go
```diff
@@ -0,0 +1,278 @@
...
+					log.Println(err)
+				}
+				for _, p := range podList.Items {
+					if p.Status.Phase == "Running" {
+						podCmd := pods.PodCommand{
+							Pod:     &p,
+							RestCfg: sourceCfg,
+							Args:    []string{"sh", "-c", "dd if=/dev/zero of=/data/test/file_1 bs=60M count=1"},
+						}
+						err = podCmd.Run()
```
alaypatel07, line 155: I am failing to understand what is the point of running this command here, 

1. the output is not saved,
2. the output does not determine the result of this Eventually block condition

What would happen if we removed this?

alaypatel07: actually never mind, you are generating the data, with this, I misread!!

tests/tests_test.go
```diff
@@ -0,0 +1,278 @@
...
+					podCmd := pods.PodCommand{
+						Pod:     &p,
+						RestCfg: sourceCfg,
+						Args:    []string{"sh", "-c", "ls -l data/test"},
+					}
+					err = podCmd.Run()
+					if err != nil {
+						log.Println(err, "Failed running ls command inside source Pod",
+							"pod", path.Join(p.Namespace, p.Name),
+							"command", "ls data/test")
```
alaypatel07, line 265: NIT: "data/test" is being used multiple times, could use an appropriate constant

tests/tests_test.go
```diff
@@ -0,0 +1,278 @@
...
+				hostClient.List(ctx, podList, &client.ListOptions{
+					LabelSelector: labels.SelectorFromSet(map[string]string{
+						"app": "longpvc-test",
+					},
+					),
+					Namespace: "ocp-41583-longpvcname",
+				})
+				destinationFile := []string{}
+				for _, p := range podList.Items {
+					podCmd := pods.PodCommand{
```
alaypatel07, line 236: question: do we need to run this command on a pod list? Should we be running on this on the first pod returned from the list? Same for source below

JaydipGabani: @alaypatel07 There is always gonna be one pod in the list since we are fetching it with labels. So what difference would it make? If any of the pods (source or destination) is not present then the check will fail. I don't know what the standard practice is though, just curious. We can break the loop after the first iteration if needed.

tests/tests_suite_test.go
JaydipGabani: @alaypatel07 Yes! It does not wait while the ns is being deleted. Move ahead with the rest of the clean-up.

tests/test_helper.go
```diff
@@ -0,0 +1,248 @@
...
+			Name: ns,
+		},
+	}
+}
+
+func GetMigSaToken(sourceClient *kubernetes.Clientset) []byte {
+	ctx := context.TODO()
+	sa, err := sourceClient.CoreV1().ServiceAccounts(MigrationNamespace).Get(ctx, MigrationController, metav1.GetOptions{})
+	if err != nil {
+		log.Print(err)
```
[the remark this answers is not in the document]
JaydipGabani: +1 to failing! Updated [here](https://github.com/konveyor/mig-controller/pull/1153/files#diff-0a9a24f44dfc4029d6c2465ca6246bf66829e97d236da6ff80172a8a9c73866aR108)

tests/tests_suite_test.go
JaydipGabani: +1

tests/tests_suite_test.go
JaydipGabani: +1

tests/tests_suite_test.go
```diff
@@ -0,0 +1,173 @@
...
+		controller, err := dynamicCLient.Resource(controllerCR).Namespace(MigrationNamespace).Get(context.TODO(), MigrationController, metav1.GetOptions{})
+		if err != nil {
+			log.Println(err)
+		}
+		phase, found, err := unstructured.NestedString(controller.Object, "status", "phase")
+		if err != nil || !found {
+			log.Println("phase not found or error in status", err)
+		}
+		return phase
+	}, time.Minute*5, time.Second).Should(Equal("Reconciled"))
```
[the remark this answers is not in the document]
JaydipGabani: We can increase the wait time to 10 mins! or whatever we find appropriate

shawn-hurley:
Just one question have we run this e2e test on a controller that does not have the bugfix and notice that this test will fail?

tests/tests_test.go
alaypatel07: so think about this, Kube is eventually consistent, i.e. there can be situations where there are two replicas temporarily. If that happens the best-case scenario would be, your code will overwrite the result of second with first. The worst-case scenario would be, if one of the two pods are in a terminating state (because kube terminates and is trying to move to a consistent state, this will error out.)

tests/test_helper.go
```diff
@@ -0,0 +1,249 @@
...
+			},
+			Data: map[string][]byte{
+				"saToken": saToken,
+			},
+			Type: TestSecretType,
+		}
+}
+
+// We are assuming that the controller CR is created and controller is running
+func NewMigrationController(install_ui bool, install_controller bool) *unstructured.Unstructured {
```
alaypatel07, line 169: camel case, please?

Go programmer don't like snake case :P haha

JaydipGabani:
> Just one question have we run this e2e test on a controller that does not have the bugfix and notice that this test will fail?

Not Yet! But I have tested that it will fail if the data is not present on the destination!

tests/test_helper.go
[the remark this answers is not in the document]
alaypatel07: This is not the fix I expected, the code two line below will panic if the error is not nil, right? i.e. sa will be nil in that case and sa.Secrets will panic.

tests/tests_test.go
JaydipGabani: Cool! Thanks for clearing out the doubt I had!

tests/test_helper.go
JaydipGabani: Oh, Okay! Didn't know this :rofl:

tests/tests_suite_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+	err = migapi.AddToScheme(scheme.Scheme)
+	hostClient, err = client.New(hostCfg, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		log.Println(err)
+		Expect(err).Should(BeNil())
+	}
+	sourceCfg, err = clientcmd.BuildConfigFromFlags("", filepath.Join(os.Getenv(SOURCECONFIG)))
+	if err != nil {
+		log.Println(err)
```
alaypatel07, line 57: NIT: missing Expect(err) block?

tests/tests_suite_test.go
```diff
@@ -0,0 +1,187 @@
...
+	controllerCR = schema.GroupVersionResource{
+		Group:    "migration.openshift.io",
+		Version:  "v1alpha1",
+		Resource: "migrationcontrollers",
+	}
+
+	controller, err := dynamicClient.Resource(controllerCR).Namespace(MigrationNamespace).Get(context.TODO(), MigrationController, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		log.Println(err)
+		Expect(err).Should(BeNil())
```
alaypatel07, line 79: this does not make sense, this block will only be reached when the error is not nil?

tests/test_helper.go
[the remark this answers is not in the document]
JaydipGabani: How? if this errors out then the list will be empty! and the function would `return nil` and the case would fail.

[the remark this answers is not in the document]
JaydipGabani: Ohh, I see your point!

tests/tests_suite_test.go
JaydipGabani: Since we are expecting to be the error `nil`, if it isn't the test will fail right and we have a log as well to show the error!

tests/tests_suite_test.go
```diff
@@ -0,0 +1,169 @@
...
+	Expect(token).ShouldNot(BeNil())
+
+	migCluster, secret := NewMigCluster(token)
+	Expect(hostClient.Create(ctx, secret)).Should(Succeed())
+	Expect(hostClient.Create(ctx, migCluster)).Should(Succeed())
+
+	Eventually(func() bool {
+		log.Println(fmt.Sprintf("Waiting for %s migCluster to be ready", E2ETestObjectName))
+		Expect(hostClient.Get(ctx, client.ObjectKey{Name: E2ETestObjectName, Namespace: MigrationNamespace}, migCluster)).Should(Succeed())
+		return migCluster.Status.IsReady()
```
alaypatel07, line 103: This is failing in my env 

```
  name: e2e-object
  resourceVersion: "184989160"
  selfLink: /apis/migration.openshift.io/v1alpha1/namespaces/openshift-migration/migclusters/e2e-object
  uid: e15be8b0-0de1-4366-a860-98acc0fdd9ca
spec:
  insecure: true
  isHostCluster: false
  serviceAccountSecretRef:
    name: e2eclustersecret
    namespace: openshift-config
  url: https://master1.alpatel-image-3110.mg.dog8code.com:443
status:
  conditions:
  - category: Warn
    lastTransitionTime: "2021-07-19T21:47:38Z"
    message: This cluster is running a different version of the Migration Toolkit for Containers operator than the host cluster. Migrating to or from this cluster might result in a failed migration and data loss. Make sure all clusters are running the same version of the operator before attempting a migration.
    reason: VersionCheckFailed
    status: "True"
    type: OperatorVersionMismatch
  - category: Required
    lastTransitionTime: "2021-07-19T21:47:38Z"
    message: The cluster is ready.
    status: "True"
    type: Ready
  observedDigest: 70011a2fb8155535aad3ca6f89ee8641c0f77349c48ac17c93ab63ce29401011
  operatorVersion: 1.4.4
```

tests/test_helper.go
```diff
@@ -0,0 +1,323 @@
...
+				Name:      E2ETestObjectName,
+				Namespace: MigrationNamespace,
+			},
+			Spec: migapi.MigStorageSpec{
+				BackupStorageConfig: migapi.BackupStorageConfig{
+					CredsSecretRef: &v1.ObjectReference{
+						Namespace: ConfigNamespace,
+						Name:      TestStorageSecret,
+					},
+					AwsBucketName: os.Getenv(AWSBUCKETNAME),
```
alaypatel07, line 118: I needed a zone to make the migstorage to ready, can we add `AWSREGION` and configure it here for completeness? 

NIT: if your bucket is provisioned in global mode, zone is not required, but in my case it was not.

tests/tests_suite_test.go
alaypatel07: nevermind it was misconfiguration on my side, fixed this!

tests/test_helper.go
JaydipGabani: Ahh, Okay! You got it! Adding the variable

alaypatel07 (approved):
I think this largely looks good to me.

I think I would see a follow up where kicking this e2e test is documented in HACKING.md or something

tests/tests_test.go
```diff
@@ -0,0 +1,218 @@
...
+						if dvm.Status.Itinerary == "VolumeMigration" {
+							dvmps := &migapi.DirectVolumeMigrationProgressList{}
+							Expect(hostClient.List(ctx, dvmps, client.MatchingLabels{"directvolumemigration": string(dvm.UID)})).Should(Succeed())
+							for _, dvmp := range dvmps.Items {
+								if dvmp.Status.TotalProgressPercentage == "100%" {
+									return dvm.Status.Phase
+								}
+							}
+						} else {
+							log.Println("Direct volume migration Failed")
```
alaypatel07, line 145:
```
  name: foo-6rddt
  ownerReferences:
  - apiVersion: migration.openshift.io/v1alpha1
    controller: true
    kind: MigMigration
    name: foo
    uid: bc59b4ef-3ecb-4727-b70c-9767de9eaf76
  resourceVersion: "185152626"
  selfLink: /apis/migration.openshift.io/v1alpha1/namespaces/openshift-migration/directvolumemigrations/foo-6rddt
  uid: 8a5bcdfd-b4fd-4fcb-b638-eacf35658c88
spec:
  createDestinationNamespaces: true
  destMigClusterRef:
    name: host
  persistentVolumeClaims:
  - name: long-name-123456789011121314151617181920212223242526272829303132
    namespace: ocp-41583-longpvcname
    targetAccessModes:
    - ReadWriteOnce
    targetStorageClass: ocs-storagecluster-ceph-rbd
    verify: false
  srcMigClusterRef:
    name: e2e-object
status:
  observedDigest: ""
  phaseDescription: ""
```

When I tried the dvm is failing here. My assumption of what is happening:
1. migmigration controller creates DVM
2. the very first reconcile cycle, dvm phase  is ""
3. the test case catches that and declares it failure

Happened this twice. More logs:

```
2021/07/19 22:32:23 waiting for dvm to complete
2021/07/19 22:32:24 waiting for dvm to complete
2021/07/19 22:32:25 Direct volume migration Failed
STEP: Creating a new migmigration
• Failure [253.857 seconds]
Running migmigration when storage, cluster and plan are ready
/home/alpatel/go/src/github.com/konveyor/mig-controller/tests/tests_test.go:24
  Testing BZ #1965421, PVC name is more than 63 char long
  /home/alpatel/go/src/github.com/konveyor/mig-controller/tests/tests_test.go:54
    Should create a new migration [It]
    /home/alpatel/go/src/github.com/konveyor/mig-controller/tests/tests_test.go:121

    Expected
        <string>:
    to be equivalent to
        <string>: VolumeMigration

    /home/alpatel/go/src/github.com/konveyor/mig-controller/tests/tests_test.go:146
------------------------------
2021/07/19 22:32:25 Cleaning up is beginning


Summarizing 1 Failure:

[Fail] Running migmigration when storage, cluster and plan are ready Testing BZ #1965421, PVC name is more than 63 char long [It] Should create a new migration
/home/alpatel/go/src/github.com/konveyor/mig-controller/tests/tests_test.go:146

Ran 1 of 1 Specs in 273.447 seconds
FAIL! -- 0 Passed | 1 Failed | 0 Pending | 0 Skipped
--- FAIL: TestMigmigration (273.45s)
FAIL

Ginkgo ran 1 suite in 4m35.45837171s
Test Suite Failed
make: *** [Makefile:132: e2e-test] Error 1
```

alaypatel07: convert this else into else if?

```
                                               } else if dvm.Status.Itinerary == "MigrationFailed" {
                                                        log.Println("Direct volume migration Failed")
                                                        Expect(dvm.Status.Itinerary).ShouldNot(BeEquivalentTo("MigrationFailed"))
                                               }
```

tests/tests_test.go
```diff
@@ -0,0 +1,218 @@
...
+					}
+				}
+				return ""
+			}, time.Minute*5, time.Second).Should(Equal("Completed"))
+			Eventually(func() string {
+				Expect(hostClient.Get(ctx, client.ObjectKey{Name: migrationName, Namespace: MigrationNamespace}, fooMigration)).Should(Succeed())
+				Expect(fooMigration.Status.Itinerary).Should(Equal("Final"))
+				log.Println(fmt.Sprintf("Running migration, currenlty in phase - %s", fooMigration.Status.Phase))
+				return fooMigration.Status.Phase
+			}, time.Minute*5, time.Second).Should(Equal("Completed"))
```
alaypatel07, line 157: make this 15 minute timeout?

alaypatel07 (changes requested):
requested a couple of nits.

alaypatel07 (approved):
ACK

alaypatel07:
@JaydipGabani some ideas after this get merged to look into:

1. The list of env variables set is non-trivial, it's not hard, but could be easier. What if we add a config file (YAML) to load those variables? Then we will read the path of the config file in the default place in the project. This will allow for more ease of use IMO
2. Can we write a doc to run this? for example we need to install ginkgo, gomega, add config etc etc (being repetitive here but just added to consolidate the list, I will keep adding more things as they occur to me)

JaydipGabani:
@alaypatel07 I had the doc file in the list already! 

+1 to using config YML for the variables

tests/tests_test.go
pranavgaikwad: @JaydipGabani @djwhatle Is it possible to populate a volume using a temporary pod as a preparation step of the test, and then attaching that same volume in a dummy Pod? so that the Pod being migrated doesn't have to create its own data. We can verify whether we find the same data in the migrated Pod that we created in the prepare step using checksum.

tests/tests_test.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+			Eventually(func() v1.PodPhase {
+				podList, err := sourceClient.CoreV1().Pods("ocp-41583-longpvcname").List(ctx, metav1.ListOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				for _, p := range podList.Items {
+					if p.Status.Phase == "Running" {
+						podCmd := pods.PodCommand{
+							Pod:     &p,
+							RestCfg: sourceCfg,
+							Args:    []string{"sh", "-c", "dd if=/dev/zero of=/" + testFilePath + "file_1 bs=60M count=1"},
```
djwhatle, line 91: @JaydipGabani I recall from testing pvc-migrate that certain classes of generated data lead to a no-op during rsync transfer. For this test to be valid, it may be worth generating random data (rather than a file of 0s), checksumming that, then verifying the checksum matches on the destination.

Take a look at this DataGenerator for an example of generating data that _won't_ result in an Rsync no-op: https://github.com/konveyor/pvc-migrate/tree/master/benchmarks/DataGenerator

@pranavgaikwad does this match your memory?

pranavgaikwad: @djwhatle yeah, zero data is not reliable test. Can replace with urandom() instead. Also, I suggested to use a pre-populated volume in previous comment
> Is it possible to populate a volume using a temporary pod as a preparation step of the test, and then attaching that same volume in a dummy Pod? so that the Pod being migrated doesn't have to create its own data. We can verify whether we find the same data in the migrated Pod that we created in the prepare step using checksum.

JaydipGabani: @djwhatle @pranavgaikwad I have added verification for checksum for migrated files and now using dataGenerator instead of `dd` command.

tests/test_helper.go
```diff
@@ -0,0 +1,337 @@
...
+					Labels: map[string]string{
+						"app": "longpvc-test",
+					},
+				},
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name:            "pod-test",
+							ImagePullPolicy: "Always",
+							Image:           "quay.io/jgabani/dg:latest",
```
djwhatle, line 300: This should be somewhere other than your personal org, probably in Konveyor auto-building from a Quay build job.

djwhatle:
@JaydipGabani some comments

- The DataGenerator image should be coming from quay.io/konveyor instead of your personal org
- We should document how to run this E2E test somewhere in the repo, I'd suggest adding a section to [HACKING.MD](https://github.com/konveyor/mig-controller/blob/master/HACKING.md)

tests/test_helper.go
JaydipGabani: Replaced with the image `pvc-benchmark-helper` from `konvoyer`

djwhatle (approved):
VISACK. I don't have clusters up to test this right now but thanks for working with us on these changes @JaydipGabani 

Sounds like a good follow-on might be the switch over to using YAML for test config.

tests/test_helper.go
```diff
@@ -0,0 +1,337 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes"
+)
+
+// common constants
+const (
+	E2ETestObjectName  = "e2e-object"
+	MigrationNamespace = "openshift-migration"
```
pranavgaikwad, line 22: Please use the constant values provided by the API wherever possible https://github.com/konveyor/mig-controller/blob/f97cd1e001b122d6fbc73b95620b6a27f64bacb1/pkg/apis/migration/v1alpha1/resource.go#L18-L19

pranavgaikwad, line 22: Please use the constant values provided by the API wherever possible https://github.com/konveyor/mig-controller/blob/f97cd1e001b122d6fbc73b95620b6a27f64bacb1/pkg/apis/migration/v1alpha1/resource.go#L18-L19

pranavgaikwad (changes requested):
Nit: use constants provided by Migration API wherever possible

pranavgaikwad (approved):

tests/tests_suite_test.go
shawn-hurley: if we are planning on creating the same namespace, we will have to wait for deletion for the next test. If we are not then this is fine with me.

shawn-hurley (approved):

PR #1155, opened 2021-07-21
Velero restore progress reporting changes

shubham-pampattiwar:
This PR does the following:
- Adds `getRestoreStats` function which reports `itemsRestored` as well as the `totalItems`.
- Adds `getRestoreDuration` function which returns the duration string.
- Updates the restore progress information per phase based on the progress data available on the Velero Restore CR.

### DIFF
--- a/pkg/controller/migmigration/restore.go
+++ b/pkg/controller/migmigration/restore.go
@@ -226,7 +226,7 @@ func getPodVolumeRestoresProgress(pvrList *velero.PodVolumeRestoreList) (progres
 
 // Returns the restore progress statistics
 func getRestoreStats(restore *velero.Restore) (itemsRestored int, totalItems int) {
-	if restore == nil {
+	if restore == nil || restore.Status.Progress == nil {
 		return
 	}
 
--- a/pkg/controller/migmigration/restore.go
+++ b/pkg/controller/migmigration/restore.go
@@ -226,7 +226,7 @@ func getPodVolumeRestoresProgress(pvrList *velero.PodVolumeRestoreList) (progres
 
 // Returns the restore progress statistics
 func getRestoreStats(restore *velero.Restore) (itemsRestored int, totalItems int) {
-	if restore.Status.Progress == nil {
+	if restore == nil {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/migmigration/restore.go
```diff
@@ -258,13 +285,16 @@ func (t *Task) hasRestoreCompleted(restore *velero.Restore) (bool, []string) {
 		progress = append(progress, stageReport...)
 	case velero.RestorePhaseCompleted:
 		completed = true
+		itemsRestored, totalItems := getRestoreStats(restore)
 		progress = append(
 			progress,
 			fmt.Sprintf(
-				"Restore %s/%s: %s",
+				"Restore %s/%s: %d out of estimated total of %d objects restored%s",
```
sseago, line 292: A minor readability point: It might be better to remove the initial space from the duration string and put it between "restored" and "%s" above.

pkg/controller/migmigration/restore.go
```diff
@@ -280,17 +310,27 @@ func (t *Task) hasRestoreCompleted(restore *velero.Restore) (bool, []string) {
 			restore.Namespace,
 			restore.Name)
 		reasons = append(reasons, message)
+		itemsRestored, totalItems := getRestoreStats(restore)
+		message = fmt.Sprintf(
+			"%s %d out of estimated total of %d objects restored%s",
```
sseago, line 315: If you make the whitespace changes suggested above, the space after "restored" is needed here too.

pkg/controller/migmigration/restore.go
```diff
@@ -280,17 +310,27 @@ func (t *Task) hasRestoreCompleted(restore *velero.Restore) (bool, []string) {
...
 			getPodVolumeRestoresProgress(pvrs)...)
 	case velero.RestorePhasePartiallyFailed:
 		completed = true
+		itemsRestored, totalItems := getRestoreStats(restore)
 		message := fmt.Sprintf(
-			fmt.Sprintf(
-				"Restore %s/%s: partially failed.",
-				restore.Namespace,
-				restore.Name))
+			"Restore %s/%s: partially failed. %d out of estimated total of %d objects restored%s",
```
sseago, line 328: If you make the whitespace changes suggested above, the space after "restored" is needed here too.

sseago (approved):
The whitespace/readability suggestion isn't required for merge, but I think it would make the code slightly more readable around the message generation.

pkg/controller/migmigration/restore.go
```diff
@@ -224,6 +224,30 @@ func getPodVolumeRestoresProgress(pvrList *velero.PodVolumeRestoreList) (progres
 	return
 }
 
+// Returns the restore progress statistics
+func getRestoreStats(restore *velero.Restore) (itemsRestored int, totalItems int) {
+	if restore.Status.Progress == nil {
```
pranavgaikwad, line 229: Please add a nil check for restore object itself

pranavgaikwad (changes requested):
Please add a nil check for Restore object apart from Scott's comments

pkg/controller/migmigration/restore.go
shubham-pampattiwar: Updated the PR.

pkg/controller/migmigration/restore.go
shubham-pampattiwar: Done, updated the PR.

pranavgaikwad (approved):

PR #1157, opened 2021-07-23
Integrate crane-lib with DVM

pranavgaikwad:


pkg/controller/directvolumemigration/rsync.go
```diff
@@ -79,623 +66,436 @@ const (
...
+			return liberr.Wrap(err)
+		}
+	}
+	return nil
+}
 
-		// Before starting rsync transfer pod, must generate rsync password in a
-		// secret and pass it into the transfer pod
+func (t *Task) createRsyncTransferClients(srcClient compat.Client,
+	destClient compat.Client, nsMap map[string][]transfer.PVCPair) (rsyncClientOperationStatusList, error) {
```
alaypatel07, line 265: NIT: consider using pointers to reduce variable copies? `(*rsyncClientOperationStatusList, error)`

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -79,623 +66,436 @@ const (
...
+				if currentStatus.failed && currentStatus.operation.CurrentAttempt < GetRsyncPodBackOffLimit(*t.Owner) {
+					labels[RsyncAttemptLabel] = fmt.Sprintf("%d", currentStatus.operation.CurrentAttempt+1)
+					rsyncOptions = append(rsyncOptions, rsync_transfer.WithSourcePodLabels(labels))
+					transfer, err := rsync_transfer.NewTransfer(
+						stunnelTransport, endpoint, srcClient.RestConfig(), destClient.RestConfig(), pvcList, rsyncOptions...)
+					if err != nil {
+						currentStatus.AddError(err)
+						continue
+					}
+					err = transfer.CreateClient(srcClient)
```
alaypatel07, line 377: this could panic, if the transfer is nil, probably want to handle those cases

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -79,623 +66,436 @@ const (
...
+				newOperation.CurrentAttempt = 0
+				labels[RsyncAttemptLabel] = fmt.Sprintf("%d", currentStatus.operation.CurrentAttempt+1)
+				rsyncOptions = append(rsyncOptions, rsync_transfer.WithSourcePodLabels(labels))
+				transfer, err := rsync_transfer.NewTransfer(
+					stunnelTransport, endpoint, srcClient.RestConfig(), destClient.RestConfig(), pvcList, rsyncOptions...)
+				if err != nil {
+					currentStatus.AddError(err)
+					continue
+				}
+				err = transfer.CreateClient(srcClient)
```
alaypatel07, line 409: nil check

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -79,623 +66,436 @@ const (
...
+			statusList.Add(currentStatus)
+			t.Log.Info("adding status of pvc", "pvc", currentStatus.operation, "errors", currentStatus.errors)
+		}
+	}
+	return statusList, nil
 }
 
-// Create rsync transfer route
-func (t *Task) createRsyncTransferRoute() error {
+func (t *Task) areRsyncTransferPodsRunning() (arePodsRunning bool, nonRunningPods []*corev1.Pod, e error) {
```
alaypatel07, line 422: NIT: take this opportunity to rename this to `areRsyncServerPodsRunning` since Transfer is now a crane-lib concept?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -79,623 +66,436 @@ const (
...
-		t.Log.Info("Creating Rsync Transfer Service for Stunnel connection "+
-			"on destination MigCluster ",
-			"service", path.Join(svc.Namespace, svc.Name))
-		err = destClient.Create(context.TODO(), &svc)
-		if k8serror.IsAlreadyExists(err) {
-			t.Log.Info("Rsync transfer svc already exists on destination",
-				"service", path.Join(svc.Namespace, svc.Name))
-		} else if err != nil {
-			return err
+		pods := corev1.PodList{}
```
alaypatel07, line 436: probably worth using this? https://github.com/konveyor/crane-lib/blob/e077c6bf9b666c3aaa85946e8e924ebf0c632dc7/state_transfer/transfer/transfer.go#L33

alaypatel07: actually I think there are log message that we need for MTC, if you also determine that IsServerHealthy is not usable probably worth filing an issue we can circle back later

jmontleon (approved):

pkg/controller/directvolumemigration/rsync.go
pranavgaikwad: @alaypatel07 I thought about using `IsHealthy` function. However, we are losing some advanced error heuristics by doing that. I propose to keep this function as-is for 1.6. WDYT?

alaypatel07: yes, works with me.

pkg/controller/migplan/pvlist.go
```diff
@@ -280,12 +280,14 @@ func (r *ReconcileMigPlan) getSupportedActions(pv core.PersistentVolume, claim m
 	if pv.Spec.NFS != nil ||
 		pv.Spec.AWSElasticBlockStore != nil {
 		return append(supportedActions,
+			migapi.PvSkipAction,
```
sseago, line 283: These two additions only add `PvSkipAction` to cases where move is allowed -- i.e. EBS and NFS volumes, or in cases where mig-operator is configured to allow move for the storage class. NOte the default return below does not add skip. Also, skip is added above in cases where `!claim.HasReference`, so it may be added twice in some cases. I think the right thing to do here, assuming you always want to allow skip, is actually to simply remove the `if !claim.HasReference` check and always include skip.

sseago (changes requested):

sseago (approved):

pkg/controller/directvolumemigration/task.go
```diff
@@ -53,8 +53,8 @@ const (
 	DirectVolumeMigrationRsyncCreds         = "directvolumemigration-rsync-creds"
 	DirectVolumeMigrationRsyncTransferSvc   = "directvolumemigration-rsync-transfer-svc"
 	DirectVolumeMigrationRsyncTransferRoute = "dvm"
-	DirectVolumeMigrationStunnelConfig      = "directvolumemigration-stunnel-config"
-	DirectVolumeMigrationStunnelCerts       = "directvolumemigration-stunnel-certs"
+	DirectVolumeMigrationStunnelConfig      = "crane2-stunnel-config"
```
alaypatel07, line 56: do we keep these label values?

alaypatel07 (approved):
This looks in great shape, thanks!!!!

some nits added, not at all blocker for merge

pkg/controller/directvolumemigration/task.go
pranavgaikwad: @alaypatel07 yes, these label values are still passed to Crane lib

PR #1158, opened 2021-07-23
remove any '/' trailer from the ExposedRegistryPath (konveyor#1154)

vlours:


sseago (approved):
@djwhatle Since we're already silently removing invalid content from the beginning of the path ("https://", etc.), doing the same with a trailing "/" seems to be consistent with this behavior. I don't know that we need validation here unless we also want to validate the whole path -- essentially requiring that this consists of *only* a valid host/domain name with an optional port.

pranavgaikwad (approved):

PR #1159, opened 2021-07-27
State Transfer API Changes

pranavgaikwad:
This PR introduces two API changes: 

1. MigPlan - Introduces new fields to provide list of included resources in the Backup and an optional Label Selector

2. DirectVolumeMigration - Introduces new fields for TargetName on each PVC

### DIFF
--- a/pkg/apis/migration/v1alpha1/directvolumemigration_types.go
+++ b/pkg/apis/migration/v1alpha1/directvolumemigration_types.go
@@ -26,11 +26,17 @@ import (
 
 type PVCToMigrate struct {
 	*kapi.ObjectReference `json:",inline"`
-	TargetStorageClass    string                            `json:"targetStorageClass"`
-	TargetAccessModes     []kapi.PersistentVolumeAccessMode `json:"targetAccessModes"`
-	TargetNamespace       string                            `json:"targetNamespace,omitempty"`
-	TargetName            string                            `json:"targetName,omitempty"`
-	Verify                bool                              `json:"verify,omitempty"`
+	// TargetStorageClass storage class of the migrated PVC in the target cluster
+	TargetStorageClass string `json:"targetStorageClass"`
+	// TargetAccessModes access modes of the migrated PVC in the target cluster
+	TargetAccessModes []kapi.PersistentVolumeAccessMode `json:"targetAccessModes"`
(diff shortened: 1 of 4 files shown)


pkg/apis/migration/v1alpha1/directvolumemigration_types.go
```diff
@@ -29,6 +29,7 @@ type PVCToMigrate struct {
 	TargetStorageClass    string                            `json:"targetStorageClass"`
 	TargetAccessModes     []kapi.PersistentVolumeAccessMode `json:"targetAccessModes"`
 	TargetNamespace       string                            `json:"targetNamespace,omitempty"`
+	TargetName            string                            `json:"targetName,omitempty"`
```
shawn-hurley, line 32: I would like to see a comment or something here that explains usage, as it appears we have some helper.

It also appears that this is not optional, I believe that we must mark it as such.

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -90,6 +90,12 @@ type MigPlanSpec struct {
 
 	// If set True, disables direct volume migrations.
 	IndirectVolumeMigration bool `json:"indirectVolumeMigration,omitempty"`
+
+	// IncludedResources optional list of included resources in Velero Backup
+	IncludedResources []*metav1.GroupKind `json:"includedResources,omitempty"`
```
shawn-hurley, line 95: These must be optional as well, can we document what the default values are for this?

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -90,6 +90,12 @@ type MigPlanSpec struct {
...
+
+	// LabelSelector optional label selector on the included resources in Velero Backup
+	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
```
shawn-hurley, line 98: These must be optional as well, I would also document what happens when this is not added.

shawn-hurley (approved):
I believe that all of these fields must be marked optional, and I believe that more documentation on usage will be helpful

shawn-hurley (changes requested):
wrong selection sorry

alaypatel07 (approved):
VISIACK

shawn-hurley (approved):
LGTM

PR #1163, opened 2021-08-02
Swap v1beta1 for v1 crds in preparation for OCP 4.9 release

djwhatle:
Implementation of [dual-stream v1beta1 removal approach ](https://github.com/konveyor/enhancements/tree/master/enhancements/crane-1.x/crd-v1beta1-removal/crd-v1beta1-removal-dual-stream) 

Adds v1 CRDs, replacing v1beta1 CRDs.

Once merged, mig-controller will no longer be runnable on OCP version 4.5-, but will work on OCP 4.9.

Related mig-operator PR: https://github.com/konveyor/mig-operator/pull/722

PR #1167, opened 2021-08-06
Version match warning: allow for mixed versions of 1.5.1 and 1.6.x for dual-version scheme

djwhatle:
**Changes**:
- Don't warn the user of mismatched versions if they're using a supported combo like 1.5.1 and 1.6.x.
- Continue to not warn users when they have an exact version match

**Example of 1.6.1 (mocked) and 1.5.1 (mocked) being allowed without warning.**

image

**Example of 1.4.3 (mocked) and 1.6.1 (mocked) triggering a warning, as user should upgrade their older 1.4.3 install**

image

This is related to work being done in #1163

### DIFF
--- a/pkg/controller/migcluster/validation.go
+++ b/pkg/controller/migcluster/validation.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strconv"
 	"strings"
 	"time"
 
@@ -439,9 +440,12 @@ func (r ReconcileMigCluster) validateOperatorVersionMatchesHost(ctx context.Cont
 		return nil
 	}
 
+	// Check for exact operator version match
 	operatorVersionMatchesHost := clusterOperatorVersion == hostOperatorVersion
+	// Check for known supported inexact version mix, e.g. 1.5.1+ and 1.6.x
+	isCrossCompatible, _ := r.isCrossVersionCompatible(clusterOperatorVersion, hostOperatorVersion)
 
-	if !operatorVersionMatchesHost {
+	if !operatorVersionMatchesHost && !isCrossCompatible {
 		cluster.Status.SetCondition(migapi.Condition{
 			Type:     OperatorVersionMismatch,
 			Status:   True,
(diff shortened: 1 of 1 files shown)


ibolton336 (approved):
Tested & looks good to me!

sseago (commented):
So I had a couple questions here -- do we really allow any 1.6 with any 1.5.1+ -- i.e. is a 1.5.1 control cluster OK with a 1.6 src or dest cluster, or does the control cluster have to be at least as new as source or destination clusters? Do we care whether the destination cluster is 1.5.z, or is it only source clusters that can be on legacy versions? Also, if both clusters are 1.6 or later, do we require clusters to match, or is it OK for 1.6.0 and 1.9.z (in any combination_ to interact?

sseago:
@djwhatle Ahh, yes. I did misread the conditionals here. I see that the only mismatching allowed is when one cluster is 1.5.z and the other is 1.6 or newer. But currently we don't care whether it's the control cluster that's newer or the remote cluster. I don't know whether we need to tighten this up or not. But yeah, on the migrating from new to old velero problem, that would probably need to be a migplan validation, since on the migcluster list we don't know what cluster is expected to be used as a src vs. destination, etc. So I'm assuming that in the 1.6.0 time frame we don't really expect much velero version skew. 1.5.1+1.6.0 would be 1.6.2 and 1.6.3. Maybe 1.5.2 moves up to 1.6.3, but I don't imagine that we'll get velero 1.7 in time to put that into MTC 1.6. Since the difference between Velero 1.6.2 and 1.6.3 is really just the availability of v1 CRDs, migrating either direction shouldn't really be a problem here.

djwhatle:
> But currently we don't care whether it's the control cluster that's newer or the remote cluster. I don't know whether we need to tighten this up or not. 

There's one scenario I can think of where a user might want to do this: if they must migrate from OCP 3.11 to 4.9, and for network reasons they have to use OCP 3.11 as their control cluster. 

> on the migrating from new to old velero problem, that would probably need to be a migplan validation, since on the migcluster list we don't know what cluster is expected to be used as a src vs. destination, etc.

+1, maybe we can add this if we see breaking changes coming down the pipe in Velero Backup format?

sseago:
@djwhatle If we expect 1.5.1 control clusters with 1.6 remote clusters, we'll want to cherry-pick this change to release-1.5.1 right? (or would this go into 1.5.2 in parallel with the 1.6.0 release?) Either way, by the time a user has 1.6 remote clusters, if we intend to support 1.5 control clusters with them, we'll probably want the validation relaxed there as well.

PR #1171, opened 2021-08-18
when finding default SC, take into account beta annotations

sseago:
If the regular is-default-class annotation is not found,
check for storageclass.beta.kubernetes.io/is-default-class


jmontleon (approved):
ACK

JaydipGabani (approved):

PR #1173, opened 2021-08-19
Velero resource inclusion for State Transfer

pranavgaikwad:


pkg/controller/migmigration/backup.go
```diff
@@ -42,14 +44,20 @@ func (t *Task) ensureInitialBackup() (*velero.Backup, error) {
...
+		return nil, liberr.Wrap(err)
+	}
 
 	newBackup.Labels[migapi.InitialBackupLabel] = t.UID()
 	newBackup.Labels[migapi.MigMigrationDebugLabel] = t.Owner.Name
 	newBackup.Labels[migapi.MigPlanDebugLabel] = t.Owner.Spec.MigPlanRef.Name
 	newBackup.Labels[migapi.MigMigrationLabel] = string(t.Owner.UID)
 	newBackup.Labels[migapi.MigPlanLabel] = string(t.PlanResources.MigPlan.UID)
 	newBackup.Spec.IncludedResources = toStringSlice(settings.IncludedInitialResources.Difference(toSet(t.PlanResources.MigPlan.Status.ExcludedResources)))
+	newBackup.Spec.IncludedResources = append(newBackup.Spec.IncludedResources, userIncludedResources...)
```
sseago, line 58: Note that the default value for IncludedInitialResources is an empty list, meaning all resources. If you add something here, it's essentially a subtraction, saying "everything not returned by `getUserIncludedResourceList` is excluded.

pkg/controller/migmigration/backup.go
```diff
@@ -121,7 +142,7 @@ func (t *Task) ensureStageBackup() (*velero.Backup, error) {
 	newBackup.Labels[migapi.MigPlanLabel] = string(t.PlanResources.MigPlan.UID)
 	var includedResources mapset.Set
 
-	if t.indirectImageMigration() || Settings.DisImgCopy {
+	if t.indirectImageMigration() || Settings.DisImgCopy || !t.Owner.IsStateMigration() {
```
sseago, line 145: This logic seems wrong. Before the change, we exclude imagestreams in stage backupif we're doing DIM *or* we're disabling image copy -- we don't want imagestreams for DIM because they're handled by the DIM controller, unless we're disabling image copy, in which case we need them to create the placeholder (empty) imagestream. With this change we'll always include imagestreams in stage backup for non-state-migration migrations, including regular DIM migrations where they should be included (or we'll be doing both direct and indirect image copy).

pranavgaikwad: @sseago Thanks for pointing that out... looks like we should be ANDing instead of ORing

sseago: @pranavgaikwad what is the intended behavior here -- i.e. how should a stage backup for state transfer differ from a stage backup for a regular migration? how should an initial backup for state transfer differ from an initial backup for state transfer?

pranavgaikwad: @sseago we just want to exclude images from the Stage Backup

pkg/controller/migmigration/backup.go
```diff
@@ -89,6 +97,19 @@ func (t *Task) getInitialBackup() (*velero.Backup, error) {
 	return t.getBackup(labels)
 }
 
+func (t *Task) getUserIncludedResourceList(srcClient compat.Client) ([]string, error) {
+	resources := []string{}
+	includedResources := t.PlanResources.MigPlan.Spec.IncludedResources
+	for _, res := range includedResources {
+		resMapper, err := srcClient.RESTMapper().RESTMapping(schema.GroupKind{Group: res.Group, Kind: res.Kind})
+		if err != nil {
+			return resources, err
```
alaypatel07, line 106: you probably want to aggregate errors here?

One GVK erroring out will stop the processing of all the GVK's below

alaypatel07 (approved):
Added a comment, LGTM otherwise

pkg/controller/migmigration/backup.go
pranavgaikwad: @alaypatel07 This is intentional. Since affects backup, we want to fail as soon as there is an error found.

pranavgaikwad: @alaypatel07 as we disussed offline, we will be aggregating errors here so that all GVKs will be processed at once.

sseago (approved):

alaypatel07 (approved):
ACK

PR #1174, opened 2021-08-20
Remove distinct cluster validation to allow intra-cluster state migrations

pranavgaikwad:
This PR removes the same cluster validation to allow intra-cluster migrations. We will be adding more validations to add warnings on corner cases such as Namespace collision, resource collision within the same namespace etc.

### DIFF
--- a/pkg/controller/directvolumemigration/validation.go
+++ b/pkg/controller/directvolumemigration/validation.go
@@ -2,7 +2,6 @@ package directvolumemigration
 
 import (
 	"context"
-	"reflect"
 
 	liberr "github.com/konveyor/controller/pkg/error"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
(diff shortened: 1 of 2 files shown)


alaypatel07 (approved):
ACK

sseago (approved):
This looks fine, but as mentioned in the comments, with this change, we absolutely need additional namespace validation when src and dest cluster are the same.

Basically, when src and dest cluster are the same, `MigPlan.GetSourceNamespaces()` must have no overlap with `MigPlan.GetDestinationNamespaces()`, but this comparsion must only be done when src and dest clusters are identical. When they are different, we expect to see the same namespaces in both.

pranavgaikwad:
@sseago Thanks for the inputs. Yes, we are currently relaxing this for the sake of state migrations. I am speaking with Erik to see whether we can add validations in 1.6 itself.

PR #1176, opened 2021-08-23
Bug 1936894: Reconcile the plan when a hook change is made

jmontleon:


pkg/controller/migplan/predicate.go
```diff
@@ -85,6 +85,19 @@ func (r PlanPredicate) mapRefs(plan *migapi.MigPlan) {
 			Name:      ref.Name,
 		})
 	}
+
+	// hooks
+	for _, hook := range plan.Spec.Hooks {
```
djwhatle, line 90: Needs a matching section in the `unmapRefs` function to handle the case where reference to a hook is removed: the MigPlan should no longer reconcile in response to MigHook change.

jmontleon: Thank you. I'll fix it.

jmontleon: Should be fixed, I believe.

djwhatle (approved):
VISACK

PR #1177, opened 2021-08-23
Bug 1951869: Plan never updates when srcMigClusterRef becomes invalid

jmontleon:


pranavgaikwad (approved):
Thanks for the fix, LGTM

djwhatle (approved):
VISACK

PR #1178, opened 2021-08-24
Bug 1968621 Unpause deployment(configs) for quiesce

jmontleon:


jmontleon:
As far as I can see only Deployments and DeploymentConfigs have Spec.Paused

jmontleon:
Will test tomorrow and make sure it actually works as expected. Need to make sure updating replicas and pause state at the same time in unquiesce actually unquiesces.

jmontleon:
Quiesces and Unquiesces as expected now.

jmontleon:
There's a second element to this in that the deployment is restored paused on the destination and needs to be unpaused to start. @djwhatle @dymurray any thoughts on how we would/could handle this?

djwhatle:
@jmontleon unless I'm missing something, we would just want to add the unpausing pieces to the appropriate unquiesce methods e.g. https://github.com/konveyor/mig-controller/blob/09c0a4e0df55d92c5622c6c525c8395e853be233/pkg/controller/migmigration/quiesce.go#L158-L159

jmontleon:
@djwhatle it works fine on a rollback, but this is just on the source side. Destination doesn't unpause/pause when velero restores it.

If nothing else I'm working on a post restore hook example that could do this.
https://github.com/konveyor/mig-operator/pull/741

PR #1179, opened 2021-08-24
Implement additional validations for intra-cluster migrations

pranavgaikwad:
This PR adds some more validations to address edge cases in intra-cluster migrations


pkg/controller/migmigration/validation.go
```diff
@@ -93,6 +95,61 @@ func (r ReconcileMigMigration) validate(ctx context.Context, migration *migapi.M
...
+	if err != nil {
+		return liberr.Wrap(err)
+	}
+	if !isIntraCluster {
+		return nil
+	}
+	conflictingNamespaces := []string{}
+	for srcNs, destNs := range plan.GetNamespaceMapping() {
+		if srcNs == destNs {
+			conflictingNamespaces = append(conflictingNamespaces, srcNs)
```
sseago, line 123: This doesn't account for namespaces in conflict via explicit mapping to other namespaces. It would catch the following:
```
- foo:foo
- bar:baz
```
and it would catch:
```
- foo
- bar
```
But it would not catch:
```
- foo:bar
- bar:foo
```
Instead, I think we need to make sure that there is no intersection in the sets returned by `plan.GetSourceNamespaces()` and `plan.GetDestinationNamespaces()`.

sseago (changes requested):

pkg/controller/migmigration/validation.go
pranavgaikwad: @sseago good catch. Thanks for pointing that out

pranavgaikwad: @sseago Note that this validation is on MigMigration. I added a new validation to avoid the scenario you mentioned in the Migration Plan instead. the new vlaidation raises a Critical condition on the migration plan making the user correct the mistake right away.

PR #1182, opened 2021-08-30
Bug 1997127: Fix rsync retry bug causing wrong status to be reported

pranavgaikwad:
Rsync network warning "must" be triggered when all of the Rsync attempts are exhausted. Status.Failed() check returns when any of the Rsync operations have completely failed (all retries exhausted). In 1.6, we changed the retry logic (removed threads) and removed a function that handled the resetting of Rsync Operation Status. This PR fixes that issue. The Rsync failed status needs to be set only when all retries are exhausted and not before that.


alaypatel07 (approved):
ACK

PR #1184, opened 2021-09-01
Bug 1951869: Fix nil ptr dereference when srcMigCluster is invalid

jmontleon:
In the course of testing https://bugzilla.redhat.com/show_bug.cgi?id=1951869 QE bumped into a nil pointer dereference coming from https://github.com/konveyor/mig-controller/blame/master/pkg/controller/migplan/validation.go#L299 which was merged about the same time my PR was.

This will allow the plan to go to Not Ready instead of crashing.

### DIFF
--- a/pkg/controller/migplan/validation.go
+++ b/pkg/controller/migplan/validation.go
@@ -296,6 +296,9 @@ func (r ReconcileMigPlan) getPotentialFilePermissionConflictNamespaces(plan *mig
 	if err != nil {
 		return nil, liberr.Wrap(err)
 	}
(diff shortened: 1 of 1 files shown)


pranavgaikwad (changes requested):
Should we also do it for the destination cluster?

jmontleon:
> Should we also do it for the destination cluster?

Seems reasonable. I've updated the PR.

pranavgaikwad (approved):

djwhatle (approved):

PR #1190, opened 2021-09-02
Bug 1970338: PollReQ if backup/restores are not immediately found aft…

dymurray:
…er creation

### DIFF
--- a/pkg/controller/migmigration/task.go
+++ b/pkg/controller/migmigration/task.go
@@ -535,7 +535,18 @@ func (t *Task) Run(ctx context.Context) error {
 			return liberr.Wrap(err)
 		}
 		if backup == nil {
-			return errors.New("Backup not found")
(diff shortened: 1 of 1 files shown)


pkg/controller/migmigration/task.go
```diff
@@ -535,7 +535,18 @@ func (t *Task) Run(ctx context.Context) error {
 			return liberr.Wrap(err)
 		}
 		if backup == nil {
-			return errors.New("Backup not found")
+			t.Owner.Status.StageCondition(migapi.Running)
```
alaypatel07, line 538: probably want to create a function for this code since it is almost similar at 4 places

alaypatel07 (commented):
Added a small nit, otherwise lgtm

dymurray:
This is not an actual fix for the problem, closing.

PR #1191, opened 2021-09-02
Bug 1970338: Fix lookup for migmigration by UID so it returns true when exists

djwhatle:
**Background info on fix:**

At the beginning of the migration process we try to clean up old Velero CRs to unblock the system.
We use various tests to determine if a backup is "stale" and worthy of deletion. One of those tests is to check if there is another running migration linked with an in-progress Backup. If we find this, we won't delete the Backup.

The function responsible for checking whether a migration is associated with the backup was always returning false because it was doing a lookup on migmigrations with an invalid labelSelector that would never return any results.
This means that when two migrations are running, one of them in the BackupCreated phase, one of them in the CleanStaleBackups phase, the second migration will clobber the first migrations active  backup

To resolve this, I added a usable label to the migmigration so we can lookup by UID, then used that label in the faulty function. I am no longer seeing failures when running parallel migrations.

### DIFF
--- a/pkg/apis/migration/v1alpha1/labels.go
+++ b/pkg/apis/migration/v1alpha1/labels.go
@@ -48,6 +48,11 @@ const (
 	// for easy search or application rollback.
 	// The value is the Task.UID().
 	MigMigrationLabel = "migration.openshift.io/migrated-by-migmigration" // (migmigration UID)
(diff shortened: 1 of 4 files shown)


dymurray (approved):
Great catch!!

PR #1196, opened 2021-09-07
Bug 1998062: update stunnel options to set image correctly using the getter functi…

pranavgaikwad:
…on (#1195)


jmontleon (approved):
ACK

PR #1197, opened 2021-09-07
Bug 1997173: Add CRDs to the GVR incompatibility list

sseago:
Add CRDs to the GVR incompatibility list. Rather than warning
on the incompatibility of the resource apiversion itself (as we do
for non-CRDs), we warn for CRDs in use in the namespaces *if*
1) the CRD apiextensions apiversion is incompatible between clusters, and
2) the CRD in use doesn't exist in the destination cluster

### DIFF
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -21,12 +21,6 @@ import (
 
 var log = logging.WithName("gvk")
 
-var crdGVR = schema.GroupVersionResource{
-	Group:    "apiextensions.k8s.io",
-	Version:  "v1beta1", // Should become v1 after 1.17, needs downscaling
-	Resource: "customresourcedefinitions",
-}
-
 type CohabitatingResource struct {
 	resource       string
 	groupResource1 schema.GroupResource
@@ -59,7 +53,21 @@ type Compare struct {
 	SrcDiscovery          discovery.DiscoveryInterface
 	DstDiscovery          discovery.DiscoveryInterface
 	SrcClient             dynamic.Interface
+	DstClient             dynamic.Interface
 	CohabitatingResources map[string]*CohabitatingResource
+}
+
+// Merge the namespace/gvr mappings from the built-in and CRD validation results
+func MergeGVRMaps(map1, map2 map[string][]schema.GroupVersionResource) map[string][]schema.GroupVersionResource {
+	for key, gvrList2 := range map2 {
+		gvrList1, ok := map1[key]
+		if !ok {
+			map1[key] = gvrList2
+		} else {
+			map1[key] = append(gvrList1, gvrList2...)
+		}
+	}
+	return map1
 }
 
 // Compare GVKs on both clusters, find incompatible GVKs
@@ -69,13 +77,17 @@ func (r *Compare) Compare() (map[string][]schema.GroupVersionResource, error) {
 	if err != nil {
 		return nil, err
 	}
+	srcCRDResource, err := collectPreferredCRDResource(r.SrcDiscovery)
+	if err != nil {
+		return nil, err
+	}
 
 	dstResourceList, err := collectNamespacedResources(r.DstDiscovery)
 	if err != nil {
 		return nil, err
 	}
 
-	preferredSrcResourceList, err = r.excludeCRDs(preferredSrcResourceList)
+	preferredSrcResourceList, err = r.excludeCRDs(preferredSrcResourceList, srcCRDResource, r.SrcClient)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


sseago:
A couple points:
1) I haven't yet been able to test this with a 4.9 dest cluster
2) I'm seeing that the CRD "compare resources" call isn't working right. It's reporting on CRDs that actually exist on dest. I need to investigate that and revise the PR.

sseago:
Figured out the issue with 2). Bitten by the absence of "version" in v1 CRDs. I need to look at the "versions" slice instead. Will update the PR.

sseago:
Updated. It's now ready for review. I've refactored the CRD list comparison to fix the version problems, as well as removed the dependency on v1beta1 availability in the source cluster for existing gvr checks. Also, I've been able to test this with a 4.9 target cluster.

pkg/gvk/gvk.go
```diff
@@ -300,14 +485,19 @@ func (r *Compare) compareResources(src []*metav1.APIResourceList, dst []*metav1.
...
 	for _, srcList := range src {
 		missing := []metav1.APIResource{}
+		matchingDstResourceList := findResourceList(srcList.GroupVersion, dst)
 		for _, resource := range srcList.APIResources {
 			if cohabitator, found := r.CohabitatingResources[resource.Name]; found {
-				if cohabitator.Seen {
-					continue
+				gv, err := schema.ParseGroupVersion(srcList.GroupVersion)
+				if err != nil &&
+					(gv.Group == cohabitator.groupResource1.Group || gv.Group != cohabitator.groupResource2.Group) {
```
alaypatel07, line 493: little confused why this is required. @sseago can you elaborate on what is happening here?

sseago: @alaypatel07 The previous code identified cohabitating resources by the two groups they cohabitated in, but the code here that identified cohabitators and passed on them didn't bother to check the group. If I were to create a CRD that I call "deployments" in my own "sseago.example.com" group, it would be flagged as cohabitating with "deployments.extensions" and "deployments.apps", which seems like a bug. This change checks the two groupresource fields (`groupResource1` and `groupResource2` ) to make sure that we only flag cohabitating resources that actually belong to the cohabitating groups. There was actually no code in mig-controller that read the values of these two fields until I added this check.

sseago: I need to check the unit tests -- looks like this change may have broken a test.

alaypatel07: ack, I think you might want

```
				if err == nil &&
					(gv.Group == cohabitator.groupResource1.Group || gv.Group == cohabitator.groupResource2.Group) 
```

alaypatel07: also, @sseago if you are fixing this bug here, likely means we need the same bug fix in velero as well.

https://github.com/vmware-tanzu/velero/blob/bd5ba66e42e48ca0338e01d7c2d4375eea68e052/pkg/backup/item_collector.go#L211

sseago: @alaypatel07 Good catch on the backwards err nil check. Yes, if we succeed in parsing GV, then we use GV to compare with cohabitator Group. And yeah, I'll create an issue and PR upstream in velero. I'd forgotten that this code was taken from Velero. They're also not using the cohabitator groupResources for anything (other than logging).

sseago: Ugh. I shouild have said good catch on *both* backwards checks. Turns out the failing unit test actually did the right thing. Fixing both of these problems made the existing unit test work succeed again.

sseago:
PR updated again. Fix from review applied. Also, I realized this morning that I didn't update the PR with the more significant refactoring work from yesterday, so that's now added to the commit as well.

pkg/gvk/gvk_test.go
```diff
@@ -149,7 +149,7 @@ func Test_compareResources(t *testing.T) {
 							{
 								Name:       "deployments",
 								Namespaced: true,
-								Group:      "apps",
+								Group:      "extensions",
```
djwhatle, line 152: @sseago why did this test need to change? Was it not doing what it claimed to do previously?

sseago: @djwhatle So I first changed it when trying to debug an issue. But looking at it again (after realizing it wasn't really needed there), it looks like an error. `APIVersion` is "extensions/v1beta1". Group and Version of items in the `APIResources` slice should match this. I think it was a copy-and-paste oversight in the test data generation which doesn't really affect test results.

djwhatle (approved):
VISACK. I haven't been able to spin up a 4.9 cluster successfully to test with. @alaypatel07 I think you remember the specifics of these code paths better than I do after the deep dive you did a while back, would like to have you take one more look before we merge.

pkg/gvk/gvk.go
```diff
@@ -153,6 +223,59 @@ func collectPreferredResources(discovery discovery.DiscoveryInterface) ([]*metav
 	return resources, nil
 }
 
+// collectPreferredResources collects all preferred namespaced scoped apiResources from the cluster
+func preferredCRDResource(discovery discovery.DiscoveryInterface) ([]*metav1.APIResourceList, error) {
```
alaypatel07, line 227: NIT: rename this to `collectPreferredCRDResources`? I think the doc comment needs an update as well

sseago: @alaypatel07 Makes sense -- although the singular "resource" was deliberate, since we will be returning exactly one resource in the slice -- the preferred GV for "customresourcedefinitions" -- should I call it `collectPreferredCRDResource` to indicate that, or should I make it plural like all the rest since it returns a slice, even though it will have a length of one?

alaypatel07 (changes requested):
I wonder if adding a test case for the specific corner case will be helpful if we ever have to revisit this. i.e a test case where there are three `deployments` each belonging to different groups, 2 of which is apps and extension and third is user defined.

pkg/gvk/gvk.go
alaypatel07: `collectPreferredCRDResource ` works

pkg/gvk/gvk.go
alaypatel07: +1, I should have pointed out both the errors. I was a little confused as to how best to communicate that so ended up pasting the exact string that worked

sseago:
@alaypatel07 It might make sense to add an entry to the compareResources  test, actually. I"m currently trying to figure out how to add a test for my velero fix upstream, but that one is more challenging since I have to somehow create an actual deployment that's not in apps/extensions. For the compareResources test, we're not attempting to use any backup logic, so it should be easy enough to add it there. I'll do that this afternoon.

sseago:
@alaypatel07 unit test case for the cohabitating fix added. I confirmed that the new case fails if I revert the cohabitating fix, and it succeeds with it.

alaypatel07 (approved):
VISIACK

PR #1198, opened 2021-09-08
Bug 2002420: making sure stage labels are added to running pods only

JaydipGabani:
Skip adding stage labels to pods that are not in running state

### DIFF
--- a/pkg/controller/migmigration/stage.go
+++ b/pkg/controller/migmigration/stage.go
@@ -97,7 +97,7 @@ func GetApplicationPodsWithStageLabels(labels map[string]string, pvcMapping map[
 		if pod.Status.Phase != corev1.PodRunning {
 			continue
 		}
-		if pod.DeletionTimestamp == nil {
+		if pod.DeletionTimestamp != nil {
 			continue
 		}
 		foundVolume := true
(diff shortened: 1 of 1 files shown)


djwhatle:
This fixed the test case mentioned in https://bugzilla.redhat.com/show_bug.cgi?id=2002420 for me. ACK

djwhatle (approved):

djwhatle:
@JaydipGabani if you have IBM ROKS clusters around, would be good to test the same TC on your end in case there is different PVC behavior there. Thinking about the scenario where a `completed` app Pod and `running` stage Pod both reference the same PVC.

pkg/controller/migmigration/stage.go
```diff
@@ -94,6 +94,12 @@ func BuildStagePods(labels map[string]string,
 func GetApplicationPodsWithStageLabels(labels map[string]string, pvcMapping map[k8sclient.ObjectKey]migapi.PV, list *[]corev1.Pod) []corev1.Pod {
 	applicationPodsWithStageLabel := []corev1.Pod{}
 	for _, pod := range *list {
+		if pod.Status.Phase != corev1.PodRunning {
+			continue
+		}
+		if pod.DeletionTimestamp == nil {
```
sseago, line 100: Isn't this the opposite of what we want? If it's deleted, we skip it. If it's not deleted, we annotate it.

JaydipGabani: Ohh! Good catch!!

sseago (approved):

pkg/controller/migmigration/stage.go
djwhatle: Good catch Scott.

djwhatle:
Re-tested after recent change set, still works 👍

PR #1201, opened 2021-09-10
Bug 1997173: Add CRDs to the GVR incompatibility list (cherry-pick to release)

sseago:
Add CRDs to the GVR incompatibility list. Rather than warning
on the incompatibility of the resource apiversion itself (as we do
for non-CRDs), we warn for CRDs in use in the namespaces *if*
1) the CRD apiextensions apiversion is incompatible between clusters, and
2) the CRD in use doesn't exist in the destination cluster


alaypatel07 (approved):
ACK

PR #1210, opened 2021-09-28
Bug 2008946: requeue cluster reconcile if registry status changes

sseago:


djwhatle (approved):
VISACK

sseago:
@djwhatle @eriknelson Do we need to find a reproducing case to test prior to merge, or does QE have a reproducer to test with? I did some minimal testing with modifying the registry path to make it invalid and then valid again, but it doesn't really test the reproducing case where the cluster goes invalid-to-valid without intervening  migcluster edits.

djwhatle:
> @djwhatle @eriknelson Do we need to find a reproducing case to test prior to merge, or does QE have a reproducer to test with? I did some minimal testing with modifying the registry path to make it invalid and then valid again, but it doesn't really test the reproducing case where the cluster goes invalid-to-valid without intervening migcluster edits.

@sseago what about a test-case where we remove the registry route and then re-introduce it? Would that make sense? I haven't read the entire BZ description so I may be missing some detail.

sseago:
@djwhatle That might verify it, although it's not an identical test case to what was reported.

sseago:
@djwhatle So your suggestion above reproduces the bug and verifies the fix quite easily. I didn't want to mess with the cluster-created default route, so I created a new test route and updated the migcluster to use it instead of the default one:
```
oc create route passthrough -n default dockertest --service=docker-registry
```
Without the fix, deleting this route for a ready cluster doesn't change its status, and adding it back for an unready cluster doesn't make it become ready on its own. With the fix, deleting the route with a ready registry causes the cluster to become unready within a minute or so (60 seconds is the requeue interval if something changed), and re-adding it causes it to return to the ready state within another minute or so.

PR #1218, opened 2021-11-03
Add CRD changes for MigMigration API

pranavgaikwad:


pranavgaikwad:
The failling Travis CI check is due to a missing dependency. The Travis CI is running the same tests that the GitHub Action is running (unit tests). I believe it would be OK to ignore them for this PR. I will discuss it offline with folks.

alaypatel07 (approved):
raised a nit if there were to much stutter, but @pranavgaikwad explained other pieces already merged.

PR #1219, opened 2021-11-05
Automatic PVC swapping for Storage Conversion migrations

pranavgaikwad:
This PR adds a new phase to Stage itinerary with a flag which evaluates only when its a storage conversion migration.

### DIFF
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -48,6 +48,10 @@ const (
 	ClosedIndexField = "closed"
 )
 
+const (
+	StorageConversionPVCNamePrefix = "new"
+)
+
 // MigPlanHook hold a reference to a MigHook along with the desired phase to run it in
 type MigPlanHook struct {
 	Reference *kapi.ObjectReference `json:"reference"`
@@ -205,13 +209,28 @@ type PlanResources struct {
 
 // GetRefResources gets referenced resources from a MigPlan.
 func (r *MigPlan) GetRefResources(client k8sclient.Client) (*PlanResources, error) {
-	// MigStorage
-	storage, err := r.GetStorage(client)
+	isIntraCluster, err := r.IsIntraCluster(client)
 	if err != nil {
 		return nil, err
 	}
-	if storage == nil {
-		return nil, errors.New("storage not found")
+	mappedNamespaces := 0
+	for srcNs, destNs := range r.GetNamespaceMapping() {
+		if srcNs != destNs {
+			mappedNamespaces += 1
+			break
+		}
+	}
+	isStorageConversionPlan := (mappedNamespaces == 0 && isIntraCluster)
+	storage := &MigStorage{}
+	if !isStorageConversionPlan {
+		// MigStorage
+		storage, err := r.GetStorage(client)
+		if err != nil {
+			return nil, err
+		}
+		if storage == nil {
+			return nil, errors.New("storage not found")
+		}
 	}
 
 	// SrcMigCluster
--- a/pkg/apis/migration/v1alpha1/migplan_types.go
+++ b/pkg/apis/migration/v1alpha1/migplan_types.go
@@ -213,18 +213,18 @@ func (r *MigPlan) GetRefResources(client k8sclient.Client) (*PlanResources, erro
 	if err != nil {
 		return nil, err
 	}
-	mappedNamespaces := 0
+	mappedNamespaces := false
 	for srcNs, destNs := range r.GetNamespaceMapping() {
 		if srcNs != destNs {
-			mappedNamespaces += 1
+			mappedNamespaces = true
 			break
 		}
 	}
-	isStorageConversionPlan := (mappedNamespaces == 0 && isIntraCluster)
+	isStorageConversionPlan := (!mappedNamespaces && isIntraCluster)
 	storage := &MigStorage{}
 	if !isStorageConversionPlan {
 		// MigStorage
-		storage, err := r.GetStorage(client)
+		storage, err = r.GetStorage(client)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/controller/migmigration/registry.go
+++ b/pkg/controller/migmigration/registry.go
@@ -62,7 +62,11 @@ func (t *Task) getAnnotations(client k8sclient.Client) (map[string]string, error
 			}
 		}
 	}
-	if t.quiesce() {
+	shouldQuiesce, err := t.isStorageConversionMigration()
+	if err != nil {
+		return nil, liberr.Wrap(err)
+	}
+	if shouldQuiesce || t.quiesce() {
 		annotations[migapi.QuiesceAnnotation] = "true"
 	}
 	return annotations, nil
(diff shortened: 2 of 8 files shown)


pkg/controller/migmigration/registry.go
```diff
@@ -62,7 +62,11 @@ func (t *Task) getAnnotations(client k8sclient.Client) (map[string]string, error
 			}
 		}
 	}
-	if t.quiesce() {
+	shouldQuiesce, err := t.isStorageConversionMigration()
```
sseago, line 65: Is the intent that storage conversions *always* quiesce regardless of whether the user specified to quiesce? That seems to be what's happening here, so I just wanted to confirm. Also, assuming that's true, I'd rename the local var `shouldQuiesce` to `shouldQuiesceForStorageConversion` just to avoid confusion, since in regular migerations where the user explicitly specified quiesce, the current `shouldQuiesce` var gets a false value, which is confusing.

pkg/controller/migmigration/task.go
```diff
@@ -167,14 +170,15 @@ var StageItinerary = Itinerary{
 		{Name: WaitForVeleroReady, Step: StepStageBackup},
 		{Name: WaitForResticReady, Step: StepStageBackup, all: HasPVs | HasStagePods},
 		{Name: WaitForRegistriesReady, Step: StepStageBackup, all: IndirectImage | EnableImage | HasISs},
-		{Name: EnsureCloudSecretPropagated, Step: StepStageBackup},
+		{Name: EnsureCloudSecretPropagated, Step: StepStageBackup, any: IndirectImage | IndirectVolume},
```
sseago, line 173: We need to be careful here. This is needed any time there's a velero restore, which means any time there's a velero backup. For final itnerary, we always have this (but that's not affected here, so that's fine). For stage itineary, maybe what we really want here is `all: HasStageBackup` instead of the indirect image/volume check?

pkg/controller/migmigration/task.go
```diff
@@ -1397,7 +1432,7 @@ func (t *Task) anyFlags(phase Phase) (bool, error) {
 	if phase.any&IndirectImage != 0 && t.indirectImageMigration() {
 		return true, nil
 	}
-	if phase.any&EnableImage != 0 && !t.PlanResources.MigPlan.IsImageMigrationDisabled() {
+	if phase.any&EnableImage != 0 && !t.PlanResources.MigPlan.IsImageMigrationDisabled() && t.migrateState() {
 		return false, nil
```
sseago, line 1436: This looks like a pre-existing bug, but the above should return `true`. The "all" test returns false when any included test fails. This is the "any" test, which returns true any time an included test succeeds. Here, the "EnableImage" test finds that image migration is *not* disabled, so it should return true. Now, for this PR, my question is whether an `EnableImage` itinerary check should also return true for stage migration? I think the answer is "no" -- so this line should be "true" and at 1435 we should *not* be adding `t.migrateState()` here.

pkg/controller/migmigration/task.go
```diff
@@ -1397,7 +1432,7 @@ func (t *Task) anyFlags(phase Phase) (bool, error) {
 	if phase.any&IndirectImage != 0 && t.indirectImageMigration() {
 		return true, nil
 	}
-	if phase.any&EnableImage != 0 && !t.PlanResources.MigPlan.IsImageMigrationDisabled() {
+	if phase.any&EnableImage != 0 && !t.PlanResources.MigPlan.IsImageMigrationDisabled() && t.migrateState() {
```
sseago, line 1435: I don't think you want this. I don't think `t.migrateState()` should be considered for the `EnableImage` test.

pkg/controller/migmigration/task.go
sseago: I think the same true/false bug exists below for EnableVolume.

sseago (changes requested):

pkg/controller/migmigration/task.go
pranavgaikwad: @sseago since we are using Stage itinerary for State Migrations & Storage Conversion, we want this. We want to disable image migration whenever any of the State Migrations or Storage Conversion is set.

shawn-hurley (approved):
mostly looks good to me

pkg/controller/migmigration/task.go
pranavgaikwad: @sseago one thing I don't understand is that how was it working until now?

pkg/apis/migration/v1alpha1/migplan_types.go
```diff
@@ -205,13 +209,28 @@ type PlanResources struct {
...
 func (r *MigPlan) GetRefResources(client k8sclient.Client) (*PlanResources, error) {
-	// MigStorage
-	storage, err := r.GetStorage(client)
+	isIntraCluster, err := r.IsIntraCluster(client)
 	if err != nil {
 		return nil, err
 	}
-	if storage == nil {
-		return nil, errors.New("storage not found")
+	mappedNamespaces := 0
```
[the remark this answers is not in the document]
pranavgaikwad: We are also using the number of mapped namespaces to run another validation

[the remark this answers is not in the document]
shawn-hurley: The don't break out of the loop right?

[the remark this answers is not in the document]
pranavgaikwad: @shawn-hurley sorry my bad. I was looking at the wrong file. Youre right

pkg/controller/migmigration/task.go
sseago: @pranavgaikwad I think the bug wasn't exposed anywhere because it looks like we're not actually using the "any" test with EnableImage anywhere. It's mainly used with "all".

pkg/controller/migmigration/task.go
```diff
@@ -155,8 +158,8 @@ var StageItinerary = Itinerary{
 		{Name: WaitForStaleStagePodsTerminated, Step: StepPrepare},
 		{Name: CreateRegistries, Step: StepPrepare, all: IndirectImage | EnableImage | HasISs},
 		{Name: CreateDirectImageMigration, Step: StepStageBackup, all: DirectImage | EnableImage},
-		{Name: QuiesceApplications, Step: StepStageBackup, all: Quiesce},
-		{Name: EnsureQuiesced, Step: StepStageBackup, all: Quiesce},
+		{Name: QuiesceApplications, Step: StepStageBackup, any: Quiesce},
+		{Name: EnsureQuiesced, Step: StepStageBackup, any: Quiesce},
 		{Name: CreateDirectVolumeMigration, Step: StepStageBackup, all: DirectVolume | EnableVolume},
```
sseago, line 163: I'm wondering why this was changed. The two should be equivalent if there's only one condition -- i.e. "any of this one thing must be true" and "all of this one thing must be true" are functionally equivalent. Currently all single-condition checks use "all".

sseago (approved):
ACK, although I'm wondering whether we should change the "any: quiesce" back to "all: quiesce" to be consistent with the other tasks.

pkg/controller/migmigration/task.go
pranavgaikwad: @sseago fixed

PR #1225, opened 2021-11-16
Bug 2022017: Restrict controller watches to watch in migration ns only

pranavgaikwad:
This fix separates our previous manager used by the controllers into two managers:

1. Manager scoped in migration ns: This manager is used to trigger watch events on Migration resources  

2. Cluster scoped manager: This manager is used in 3 places:
     a. MigCluster controller uses it to watch the Secret resources in `openshift-config` namespace
     b. MigStorage controller uses it to watch the Secret resources in `openshift-config` namespace
     c. Host MigCluster uses its cached client to list resources involved in a migration

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
@@ -35,7 +36,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 )
 
 func main() {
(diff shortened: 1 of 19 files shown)


cmd/manager/main.go
```diff
@@ -57,58 +57,81 @@ func main() {
...
 		log.Error(err, "unable to add OpenShift image APIs to scheme")
 		os.Exit(1)
 	}
-	if err := appsv1.AddToScheme(mgr.GetScheme()); err != nil {
+	if err := appsv1.AddToScheme(watchMgr.GetScheme()); err != nil {
 		log.Error(err, "unable to add OpenShift apps APIs to scheme")
 		os.Exit(1)
 	}
-	if err := routev1.AddToScheme(mgr.GetScheme()); err != nil {
+	if err := routev1.AddToScheme(watchMgr.GetScheme()); err != nil {
```
alaypatel07, line 96: will all of these add to schemes will also be required for unscopedMgr? Wondering what will happen if the dstClient tried to get routes?

pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -97,6 +101,17 @@ func add(mgr manager.Manager, r *ReconcileMigCluster) error {
 		return err
```
alaypatel07, line 101: Are lines 89 through 98 still valid?

PR #1226, opened 2021-11-17
Update predicates to filter events outside of migration namespace

pranavgaikwad:


pkg/controller/migcluster/migcluster_controller.go
```diff
@@ -90,9 +94,9 @@ func add(mgr manager.Manager, r *ReconcileMigCluster) error {
 	err = c.Watch(
 		&source.Kind{Type: &kapi.Secret{}},
 		handler.EnqueueRequestsFromMapFunc(func(a k8sclient.Object) []reconcile.Request {
-			return migref.GetRequests(a, migapi.MigCluster{})
+			return migref.GetRequests(a, migapi.OpenshiftMigrationNamespace, migapi.MigCluster{})
```
alaypatel07, line 97: will this work as expected for secrets in openshift-config namespace?

alaypatel07:
the failing tests are proof that changes intended work :) nice woork

pkg/controller/directimagemigration/directimagemigration_controller_test.go
```diff
@@ -20,9 +20,11 @@ import (
 	"testing"
 	"time"
 
+	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
```
alaypatel07, line 23: duplicate import

pkg/controller/directimagemigration/directimagemigration_controller.go
```diff
@@ -70,18 +73,20 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 	// Watch for changes to DirectImageStreamMigrations
-	err = c.Watch(&source.Kind{Type: &migapi.DirectImageStreamMigration{}}, &handler.EnqueueRequestForOwner{
-		IsController: true,
-		OwnerType:    &migapi.DirectImageMigration{},
-	})
+	err = c.Watch(&source.Kind{Type: &migapi.DirectImageStreamMigration{}},
+		&handler.EnqueueRequestForOwner{
+			IsController: true,
+			OwnerType:    &migapi.DirectImageMigration{},
+		}, &migref.MigrationNamespacePredicate{Namespace: migapi.OpenshiftMigrationNamespace})
```
alaypatel07, line 89: pointer is being used here `&migref.MigrationNamespacePredicate{Namespace: migapi.OpenshiftMigrationNamespace}` but the predicate methods are on non-pointer object `func (m MigrationNamespacePredicate) Create(evt event.CreateEvent) bool {
	if evt.Object.GetNamespace() == m.Namespace {
		return true
	}
	return false
}
`

Not sure what effect this will have

pkg/reference/handler.go
```diff
@@ -6,25 +6,27 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-func GetRequests(a client.Object, source interface{}) []reconcile.Request {
+//
+func GetRequests(a client.Object, ownerNamespace string, source interface{}) []reconcile.Request {
```
alaypatel07, line 10: NIT: s/ownerNamespace/migrationNamespace? 

technically this is not an owner namespace this is always going to be a migrationNamesace right?

pkg/controller/directvolumemigration/directvolumemigration_controller.go
```diff
@@ -64,7 +67,8 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		&handler.EnqueueRequestForOwner{
 			IsController: true,
 			OwnerType:    &migapi.DirectVolumeMigration{},
-		})
+		},
+		&migref.MigrationNamespacePredicate{Namespace: migapi.OpenshiftMigrationNamespace})
 	if err != nil {
```
alaypatel07, line 72: do you find it weird that dvm controller does not have a watch on migcluster? may be thats a bug for different time

pkg/controller/miganalytic/miganalytics_controller.go
```diff
@@ -92,7 +92,9 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	err = c.Watch(
 		&source.Kind{Type: &migapi.MigAnalytic{}},
 		&handler.EnqueueRequestForObject{},
-		&AnalyticPredicate{})
+		&AnalyticPredicate{
```
alaypatel07, line 95: pointer vs non-pointer with this one as well.

pkg/controller/migcluster/remotecluster.go
```diff
@@ -50,6 +51,9 @@ func (r *RemoteClusterSource) run() {
 		}
 
 		for _, cluster := range list {
+			if r.Namespace != "" && cluster.Namespace != r.Namespace {
```
alaypatel07, line 54: this might not be required. The list here is filtered by openshift-migration namespace

pkg/controller/migplan/migplan_controller.go
```diff
@@ -92,9 +94,11 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	err = c.Watch(
 		&source.Kind{Type: &migapi.MigCluster{}},
 		handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
-			return migref.GetRequests(a, migapi.MigPlan{})
+			return migref.GetRequests(a, migapi.OpenshiftMigrationNamespace, migapi.MigPlan{})
```
alaypatel07, line 97: the last argument should be `migapi.MigCluster{}`?

pkg/controller/migplan/migplan_controller.go
```diff
@@ -103,9 +107,11 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	err = c.Watch(
 		&source.Kind{Type: &migapi.MigStorage{}},
 		handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
-			return migref.GetRequests(a, migapi.MigPlan{})
+			return migref.GetRequests(a, migapi.OpenshiftMigrationNamespace, migapi.MigPlan{})
```
alaypatel07, line 110: last argument `migapi.MigStorage`?

pkg/controller/migplan/migplan_controller.go
```diff
@@ -114,18 +120,24 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	err = c.Watch(
 		&source.Kind{Type: &migapi.MigHook{}},
 		handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
-			return migref.GetRequests(a, migapi.MigPlan{})
+			return migref.GetRequests(a, migapi.OpenshiftMigrationNamespace, migapi.MigPlan{})
```
alaypatel07, line 123: same last argument change?

alaypatel07 (commented):
Overall looks good to me added a few nits. One other comment could be in some predicates `Namespace` is used as a field and some `InNamespace`, can we modify it to `Namespace` in all for consistency?

pkg/controller/migplan/migplan_controller.go
pranavgaikwad: @alaypatel07 Nope, the last argument is correct. the event `a` coming from MigCluster is passed to `GetRequests`. The second parameter is the kind of the Owner not the target.

pkg/controller/directimagemigration/directimagemigration_controller.go
pranavgaikwad: @alaypatel07 wont have any effect right? Because we are only consuming the `Namespace` field in the struct which itself isn't a pointer. managers only call the function on the struct and pass deps (events) from outside

alaypatel07: yeah, I am still confused about how it works, will have to dig into the type system. But it works so going to approve it.

alaypatel07 (approved):
VISIACK

PR #1231, opened 2021-11-19
Bug 2021668: Respect cluster_subdomain setting

djzager:
There was a regression where the cluster_subdomain key was not being
used when creating dvm routes. This PR allows the cluster_subdomain to
be overridden when creating routes.

Additionally, this PR grabs the subdomain from the OpenShift's cluster
ingress config (supported since 4.3
https://docs.openshift.com/container-platform/4.3/networking/ingress-operator.html),
allowing crane-lib's endpoint functionality to handle scenarios where
the route name would be too long.

PR #1237, opened 2021-11-30
Bug 2021668: Respect cluster_subdomain setting

djzager:
There was a regression where the cluster_subdomain key was not being
used when creating dvm routes. This PR allows the cluster_subdomain to
be overridden when creating routes.

Additionally, this PR grabs the subdomain from the OpenShift's cluster
ingress config (supported since 4.3
https://docs.openshift.com/container-platform/4.3/networking/ingress-operator.html),
allowing crane-lib's endpoint functionality to handle scenarios where
the route name would be too long.

Add the config API to the manager's scheme as that is automatically used
in migration cluster client schemes.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -28,6 +28,7 @@ import (
 	"github.com/konveyor/mig-controller/pkg/webhook"
 	"github.com/konveyor/mig-controller/pkg/zapmod"
 	appsv1 "github.com/openshift/api/apps/v1"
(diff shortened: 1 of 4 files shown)


djzager:
This is a cherry-pick of #1231

djzager:
cc @alaypatel07 @pranavgaikwad

alaypatel07 (approved):

pranavgaikwad (approved):

PR #1241, opened 2021-12-10
Identify sparse files and automatically set `--sparse` option if they are found 

pranavgaikwad:
[[MIG-861](https://issues.redhat.com/projects/MIG/issues/MIG-861)]

On every node where PVs are mounted, a new shell command `du` will be executed. The `--apparent-size` option of `du` identifies the actual size of files as opposed to `df` where it only shows the size of occupied blocks. If sparse files exist on a volume, the occupied blocks < apparent size. New logic is introduced to compare the two values returned by `df` and `du` commands. If the apparent size returned by `du` is more than what's returned by `df` then a special flag is set on MigAnalytic resource's status for every PV. For every PV, if the flag is set, the `--sparse` Rsync option is automatically set.


shubham-pampattiwar (approved):
VISACK, Nice work @pranavgaikwad LGTM

sseago (approved):

PR #1245, opened 2022-01-05
Bug 2036908 - Fix cloud provider secret names

dymurray:


sseago (approved):

apinnick:
Will these changes affect MTC docs?

dymurray:
@apinnick Thank you for checking, this should not affect MTC docs. I checked myself and couldn't find any reference to these secrets. These secrets are managed solely from the MTC controller. The secrets the user makes are basically copied into these secrets.

PR #1249, opened 2022-01-10
Bug 2021666: update crane-lib (#1248)

djzager:
This addresses an issue where crane-lib was not properly updating the
routePrefix when too long.


djzager:
This is a cherry-pick of #1248

pranavgaikwad (approved):

PR #1250, opened 2022-01-10
Bug 2005848 - Adding validation for DNS compliant namespace names on destination cluster in migplan

JaydipGabani:


pkg/controller/migplan/validation.go
```diff
@@ -48,6 +48,7 @@ const (
 	NsNotFoundOnDestinationCluster             = "NamespaceNotFoundOnDestinationCluster"
 	NsLimitExceeded                            = "NamespaceLimitExceeded"
 	NsLengthExceeded                           = "NamespaceLengthExceeded"
+	NotDNSCompliantNamespece                   = "NotDNSCompliantNamespece"
```
sseago, line 51: For consistency with the other errors, I'd use this instead:
```
         NsNotDNSCompliant     = "NamespaceNotDNSCompliant"
```

pkg/controller/migplan/validation.go
```diff
@@ -978,20 +980,20 @@ func (r ReconcileMigPlan) validateDestinationNamespaces(plan *migapi.MigPlan) er
 		})
 		return nil
 	}
-	longNamespaceNames := []string{}
+	nonDNSCompliantNames := []string{}
 	for _, ns := range namespaces {
-		if len(ns) > 63 {
-			longNamespaceNames = append(longNamespaceNames, ns)
+		if errs := validation.IsDNS1035Label(ns); len(errs) != 0 {
+			nonDNSCompliantNames = append(nonDNSCompliantNames, ns)
```
sseago, line 986: I'm wondering whether we should at least log these errors returned. It's probably too cumbersome to include them in the condition itself, but if we logged them, a user could look at the log if they didn't understand why they got the validation warning.

pkg/controller/migplan/validation.go
sseago: If you can confirm that `NsLengthExceeded` and `LengthExceeded` aren't used anywhere else, you might also remove them since they're being replaced by the new conditions in this PR.

sseago (changes requested):

pkg/controller/migplan/validation.go
JaydipGabani: They are getting used to checking source namespace, and the condition there is slightly different, so those variables are needed.

pkg/controller/migplan/validation.go
```diff
@@ -978,20 +980,21 @@ func (r ReconcileMigPlan) validateDestinationNamespaces(plan *migapi.MigPlan) er
 		})
 		return nil
 	}
-	longNamespaceNames := []string{}
+	nonDNSCompliantNames := []string{}
 	for _, ns := range namespaces {
-		if len(ns) > 63 {
-			longNamespaceNames = append(longNamespaceNames, ns)
+		if errs := validation.IsDNS1035Label(ns); len(errs) != 0 {
+			log.Info("Namespace validation failed!!", ns, errs)
```
pranavgaikwad, line 986:
```suggestion
			log.Info("Namespace validation failed", "namespace", ns, "errors", errs)
```

pranavgaikwad (changes requested):

pranavgaikwad (approved):

pkg/controller/migplan/validation.go
sseago: Ahh ok, makes sense.

pkg/controller/migplan/validation.go
```diff
@@ -48,6 +48,7 @@ const (
 	NsNotFoundOnDestinationCluster             = "NamespaceNotFoundOnDestinationCluster"
 	NsLimitExceeded                            = "NamespaceLimitExceeded"
 	NsLengthExceeded                           = "NamespaceLengthExceeded"
+	NsNotDNSCompliant                          = "NotDNSCompliantNamespece"
```
sseago, line 51: Nit: s/NotDNSComplianteNamespace/NamespaceNotDNSCompliant/ to be consistent with the strings above

sseago (changes requested):
One more trivial change, then I think it's good to go.

sseago (approved):

PR #1253, opened 2022-01-14
MIG-853: Apply LimitRange constraints on resource requirements of Rsync Pods

pranavgaikwad:
This PR updates the ResourceRequirements used by the Rsync Pods. It introduces new logic to look at the LimitRanges in source and destination namespaces. It constrains the final resources requested by the Rsync Pods by taking into consideration the Limits for each namespace.


pkg/controller/directvolumemigration/resource.go
```diff
@@ -0,0 +1,352 @@
...
+			//  apply Max limits on Requests and Limits, none should exceed the Max
+			if requirements.Limits.Cpu().Cmp(*limit.Max.Cpu()) > 0 {
+				requirements.Limits[corev1.ResourceCPU] = *limit.Max.Cpu()
+			}
+			if requirements.Limits.Memory().Cmp(*limit.Max.Memory()) > 0 {
+				requirements.Limits[corev1.ResourceMemory] = *limit.Max.Memory()
+			}
+		}
+	}
+	// if limits are smaller than requests, delete limit values
```
sseago, line 232: Could this result in a pod failing to start? I know there are issues when resources aren't specified but there's a resource quota in place. I would be concerned that in the case where there is a ResourceQuota with cpu/memory limits set the above may result in failed pods. Backing up a step here, the only way limits could be smaller than requests here would be if the limit ranges are doing something bizarre, though right? i.e. limit ranges with limit values lower than request values? Should we error out here, or should we just do what you're doing -- and in the edge case where irrational limit ranges intersect with resource quotas, we'll just fail to start pods?

pkg/controller/directvolumemigration/resource.go
```diff
@@ -0,0 +1,352 @@
...
+		if limit.Max != nil {
+			//  apply Max limits on Requests and Limits, none should exceed the Max
+			if requirements.Requests.Cpu().Cmp(*limit.Max.Cpu()) > 0 {
+				requirements.Requests[corev1.ResourceCPU] = *limit.Max.Cpu()
+			}
+			if requirements.Requests.Memory().Cmp(*limit.Max.Memory()) > 0 {
+				requirements.Requests[corev1.ResourceMemory] = *limit.Max.Memory()
+			}
+		}
+		if limit.Min != nil {
```
sseago, line 211: If the namespace has non-intersecting limit ranges (i.e. one has a max of "5" and another has a min of "6"), the above won't detect the problem and just respond with requirements based on the minimums. Of course if those limit ranges exist, it will be impossible to start *any* pod in that namespace (even outside of MTC), so it's not clear that we absolutely need to handle this case. Maybe we don't worry about it since it can only happen with a misconfigured cluster.

sseago (commented):
Makes sense overall -- I had a couple questions around edge cases which we may or may not want to handle. Basically, those are cases where we'd probably fail anyway -- the only question is whether we should detect them and error out, or whether we're talking about a broken cluster situation that's out of scope for us to worry about.

pkg/controller/directvolumemigration/resource.go
pranavgaikwad: @sseago Yes, if `resources.Limits` > `resources.Requests`, Pods fail to start. Note that this only happens when there are `LimitRange` present in a namespace. We are taking defensive approach here. Basically, by deleting Limits altogether, we are letting the `LimitRange` decide what the maximum should be. As long as `resources.Requests` remains under limits, the `resources.Limits` can be safely deleted such that it can be managed automatically by Kube. So this code exists such that we never fail creating Pods if Limits are insane. Also, I think this will be a corner case. 

When `LimitRange` are not present, we still fallback on the User configured values then the defaults in that order.

pkg/controller/directvolumemigration/resource.go
pranavgaikwad: @sseago As explained in the previous comment, we are taking defensive approach. We try to calculate the most restrictive values of all. There's one corner case that this code does not handle and that happens when the `LimitRanges` are configured such that it is "impossible" to run Rsync Pods with two containers and by staying in the Min and Max limits of the namespace. In that case, we will fail to create Pods and it will only succeed after the user fixes their `LimitRanges`.

sseago (approved):
Thinking about it, I think we're fine here -- the edge cases mentioned (if they can occur at all) can only happen when the user has created LimitRanges that essentially break the namespace. It's not just rsync pods that will fail to be created -- it's pods in general, so we can probably ignore that case.

JaydipGabani (approved):

PR #1260, opened 2022-01-28
Bug 2044973: Making sure that cancel condition gets added to cancelled rollback for UI to display information correctly

JaydipGabani:
Modifies intentional change made [here](https://github.com/konveyor/mig-controller/commit/9506375bba0e166bc1be43a0ab07862c75eecd7b#diff-f190212e5636f337ee1eccbf2b490647d2efda28b7f46914d24332e838b5723cR678). This modification is harmless, but a little skeptical of modifying intentional behavior.

### DIFF
--- a/pkg/controller/migmigration/task.go
+++ b/pkg/controller/migmigration/task.go
@@ -1021,9 +1021,10 @@ func (t *Task) Run(ctx context.Context) error {
 			t.Requeue = PollReQ
 		}
 	case Canceling:
-		// Skip directly to Completed if the Cancel was set on a Rollback migration.
+		// Skip directly to Cancel if the Cancel was set on a Rollback migration.
 		if t.rollback() {
-			t.Phase = Completed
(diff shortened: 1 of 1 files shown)


pkg/controller/migmigration/task.go
```diff
@@ -1023,7 +1023,8 @@ func (t *Task) Run(ctx context.Context) error {
 	case Canceling:
 		// Skip directly to Completed if the Cancel was set on a Rollback migration.
```
sseago, line 1024: If I'm understanding the change correctly, after this the migration will be at Canceled rather than Completed -- From EnsureAnnotationsDeleted to Canceled when `next()` is called just below. Maybe the above comment should be changed to reflect this as well.

sseago (changes requested):
Minor suggestion -- I think the comment should be updated to reflect the slight change in behavior.

sseago (approved):

PR #1263, opened 2022-01-31
Bug 2049062: Fix storage conversion for Job resources

pranavgaikwad:
- Completed Jobs will not be updated
- For any incomplete Jobs, a copy of existing Job with the new PVC will be created as some of the fields on the Job resource are immutable

### DIFF
--- a/pkg/controller/migmigration/storage.go
+++ b/pkg/controller/migmigration/storage.go
@@ -99,7 +99,7 @@ func (t *Task) swapPVCReferences() (reasons []string, err error) {
 	}
 	// update pvc refs on cronjobs
 	failedCronJobNames := t.swapCronJobsPVCRefs(client, mapping)
-	if len(failedJobNames) > 0 {
+	if len(failedCronJobNames) > 0 {
 		reasons = append(reasons,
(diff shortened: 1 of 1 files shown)


pkg/controller/migmigration/storage.go
```diff
@@ -354,8 +364,33 @@ func (t *Task) swapJobsPVCRefs(client k8sclient.Client, mapping pvcNameMapping)
...
+			}
 			isFailed := false
+			job := &batchv1.Job{}
+			job.Namespace = oldJob.Namespace
+			job.ObjectMeta.Labels = oldJob.Labels
+			job.ObjectMeta.Annotations = oldJob.Annotations
+			delete(job.Labels, "job-name")
+			delete(job.Labels, "controller-uid")
+			if oldJob.Spec.DeepCopy() != nil {
+				job.Spec = *oldJob.Spec.DeepCopy()
```
sseago, line 381: This is performing the DeepCopy() operation twice if its output is not nil. The first call should save the output into a local variable and use that result in the assignment here.

sseago (approved):

PR #1267, opened 2022-02-04
Bug 2004264: Making sure if the error is due to resourcequota or not

JaydipGabani:
Another approach would be to get the PVC and see if it exists in the destination, and then if it is, we assume we faced an `exceeded quota` error. Open to changing code accordingly, a little skeptical on alternative because we are making an additional call.

### DIFF
--- a/pkg/controller/directvolumemigration/pvcs.go
+++ b/pkg/controller/directvolumemigration/pvcs.go
@@ -2,8 +2,6 @@ package directvolumemigration
 
 import (
 	"context"
-	"path"
-
 	liberr "github.com/konveyor/controller/pkg/error"
 	migapi "github.com/konveyor/mig-controller/pkg/apis/migration/v1alpha1"
 	"github.com/konveyor/mig-controller/pkg/settings"
@@ -11,6 +9,7 @@ import (
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"path"
 )
 
 func (t *Task) areSourcePVCsUnattached() error {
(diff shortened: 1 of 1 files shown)


pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -119,6 +120,12 @@ func (t *Task) createDestinationPVCs() error {
 		err = destClient.Create(context.TODO(), &destPVC)
 		if k8serror.IsAlreadyExists(err) {
 			t.Log.Info("PVC already exists on destination", "name", pvc.Name)
+		} else if k8serror.IsForbidden(err) {
+			if strings.Contains(err.Error(), pvc.Name) && strings.Contains(err.Error(), "exceeded quota") {
```
pranavgaikwad, line 124: While this solves the problem, I don't like the idea of reading Error strings. We should just `Get()` the PVC and only attempt to create it when it's not present. Although it does add a call, it does not necessarily have to go to the api server.

pranavgaikwad (changes requested):

pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -116,10 +115,19 @@ func (t *Task) createDestinationPVCs() error {
...
+			Namespace: destNs,
+			Name:      destName,
+		}, &destPVCCheck)
+		if k8serror.IsNotFound(err) {
+			err = destClient.Create(context.TODO(), &destPVC)
+			if err != nil {
+				return err
+			}
+		} else if err == nil {
 			t.Log.Info("PVC already exists on destination", "name", pvc.Name)
```
pranavgaikwad, line 129:
```suggestion
			t.Log.Info("PVC already exists on destination", "namespace", pvc.Namespace, "name", pvc.Name)
```

pranavgaikwad (approved):
One small change, otherwise LGTM

PR #1278, opened 2022-04-09
Bug 2061653: Only show pods relating to the current DVMP.

ibolton336:
https://bugzilla.redhat.com/show_bug.cgi?id=2061653


pranavgaikwad (approved):
ACK

PR #1280, opened 2022-04-11
Fix ansible-runner cmd order for newer versions

jmontleon:


rayfordj (approved):
/lgtm

itewk:
tested this with MTC v1.7.0 and it worked for me testing with quay.io/konveyor/hook-runner:latest

PR #1319, opened 2022-10-19
Copy labels from source namespaces

pranavgaikwad:
This is needed to let users add labels to override PSA in case workload is running as root  on the source.


rayfordj (approved):
/lgtm

PR #1315, opened 2022-10-07
Update CronJob (un)quiesce for 4.12

jmontleon:

PR #1323, opened 2022-11-09
Bug 2132957: Update CronJob (un)quiesce for 4.12 (#1315)

rayfordj:
This was a clean cherry-pick with the exception of `go.mod` and `go.sum` which updated with `go mod tidy`.


jmontleon (approved):
ACK

PR #1324, opened 2022-11-10
Add options to use ClusterIP & NodePort services for rsync transfer

pranavgaikwad:
- Adds new options to use ClusterIP / NodePort service as Rsync endpoint
- Removes an unused function


jmontleon (approved):
ACKK

PR #1327, opened 2022-11-18
Update Dockerfile

jmontleon:


rayfordj (approved):
/lgtm

I've confirmed downstream openshift-migration-controller image already specifies `USER 65534:65534`. 

Thanks!

PR #1356, opened 2023-10-13
Bump x/net to 0.17.0

dymurray:


jmontleon:
Should we add a replace to get rid off all the other versions in go.sum?

```
replace golang.org/x/net => golang.org/x/net v0.17.0
```

will drop all of these: https://github.com/migtools/mig-controller/blob/3ca4e854ac1c6adeee0a4f3301bc663981f639d0/go.sum#L1113-L1169

jmontleon:
For anyone looking back on this, we probably started doing replaces because of dependabot upstream.
https://github.com/dependabot/dependabot-core/issues/4740

But looking at go.sum results in a lot of false positives, and with improvements in go 1.17+ to show indirect dependencies in go.mod, they stopped doing this.
https://github.com/dependabot/dependabot-core/issues/4740#issuecomment-1458532057

It is not and probably never was necessary to use replaces to fix CVEs, only to satisfy dependabot.

jmontleon (approved):
ACK / LGTM

PR #1363, opened 2024-01-04
oc sa get-token is deprecated

awels:
oc sa get-token is deprecated and no longer works on openshift 4.14 and newer, replace with current oc create token command


jmontleon:
Thank you for the PR. Do you know what the default duration oc a token is when running `oc create token` is? I believe it's 1h, but I am far from certain. In either case is 1h sufficient for what this script is doing or should we extend the life with `--duration` flag?

awels:
Yeah I am pretty sure it is an hour, and this probably not long enough. I have run into the token timing out while I was testing stuff. Let me increase the timeout to 24 hours or something like that.

jmontleon (approved):
ACK

PR #1364, opened 2024-01-08
Ensure target PVCs don't have dataSource(Ref) set

awels:
If the source PVC was created using a dataSource(Ref), then directly copying the PVC spec to the target PVC spec would also copy the dataSource(Ref). If the data source no longer exists the migration would fail because the target PVC would not become bound since the datasource was missing. This change ensures that the dataSource(Ref) is always cleared when creating the target PVC.
   
Also added a bunch of unit tests to the code that appeared to be missing.


pkg/controller/directvolumemigration/pvcs.go
```diff
@@ -114,7 +118,9 @@ func (t *Task) createDestinationPVCs() error {
 			"destPersistentVolumeClaim", path.Join(destNs, pvc.Name),
 			"pvcStorageClassName", destPVC.Spec.StorageClassName,
 			"pvcAccessModes", destPVC.Spec.AccessModes,
-			"pvcRequests", destPVC.Spec.Resources.Requests)
+			"pvcRequests", destPVC.Spec.Resources.Requests,
+			"pvcDataSource", destPVC.Spec.DataSource,
+			"pvcDataSourceRef", destPVC.Spec.DataSourceRef)
```
pranavgaikwad, line 123: nit: if it's always going to be nil, should we care logging it?

pranavgaikwad (approved):

pkg/controller/directvolumemigration/pvcs.go
awels: Its more debug information, if it is NOT nil, there is a problem. The log should probably be debug instead of info.

PR #1374, opened 2024-06-06
Add block volume support for DirectVolumeMigration

awels:
Added support to migrate block volumes. Including integration with quiesce for existing resource types. Uses the new 0.1.2 crane-lib which implements the block transfer code.

When migrating block volumes it creates an additional `blockrsync` server that contains each block volume that is being migrated. Then an additional blockrsync client per source volume, similar to how the rsync for filesystem volumes works.


config/crds/migration.openshift.io_migplans.yaml
```diff
@@ -525,16 +529,25 @@ spec:
...
-                      items:
-                        type: string
-                      type: array
                     default:
                       type: boolean
                     name:
                       type: string
                     provisioner:
                       type: string
+                    volumeAccessModes:
```
pranavgaikwad, line 538: this could be a upgrade problem...how does this affect migration plans that are already present in a cluster from an older version?

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -58,15 +61,18 @@ const (
...
+// ensureRsyncEndpoints ensures that new Endpoints are created for Rsync and Blockrsync Transfers
+func (t *Task) ensureRsyncEndpoints() error {
 	destClient, err := t.getDestinationClient()
 	if err != nil {
 		return liberr.Wrap(err)
 	}
 
 	dvmLabels := t.buildDVMLabels()
 	dvmLabels["purpose"] = DirectVolumeMigrationRsync
+	blockdvmLabels := t.buildDVMLabels()
```
pranavgaikwad, line 73: looks like this is redundant? can re-use dvm labels as-is

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -92,6 +98,15 @@ func (t *Task) ensureRsyncEndpoint() error {
 				getNodeHostnameAtRandom(hostnames),
 				t.getServiceType(),
 			)
+			blockEndpoint = svcendpoint.NewEndpoint(
```
pranavgaikwad, line 101: what if there are no block volumes in the plan? would we still create these endpoints / servers?

pranavgaikwad:
@awels if you could answer my questions, I will continue reviewing again..just need some clarification

config/crds/migration.openshift.io_migplans.yaml
awels: So the field is optional, so an existing migration plan should not have a problem. The values will just be empty.

pkg/controller/directvolumemigration/rsync.go
awels: I need 2 copies, since I am modifying it differently for block. I could make a copy and add the `app` label.

pkg/controller/directvolumemigration/rsync.go
awels: Just the endpoints, not the servers. Not sure if I can determine here if block volumes exist or not.

pkg/controller/directvolumemigration/rsync.go
```diff
@@ -574,27 +675,42 @@ func (t *Task) createRsyncTransferClients(srcClient compat.Client,
...
+							continue
+						}
 					}
-					err = transfer.CreateClient(srcClient)
-					if err != nil {
-						t.Log.Error(err, "failed creating rsync pod for pvc", "pvc", newOperation)
-						currentStatus.AddError(err)
-						statusList.Add(currentStatus)
-						continue
+					if len(blockOrVMPvcList) > 0 {
```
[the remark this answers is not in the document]
awels: That is correct, if no block volumes or no filesystem volumes, the end points for both are still created.

config/crds/migration.openshift.io_migplans.yaml
pranavgaikwad: @awels but I see that accessModes field is removed and being replaced with this

awels: Good point, I should probably preserve that, and make an internal function to simply convert that to fs since that is what it was originally.

awels: So looking at this some more, it doesn't appear to be used as a K8S resource at all. So no existing plans should be affected by it. The struct is only used when looking up information for storage classes, and then it is dynamically populated.

awels:
@pranavgaikwad I have answered your questions a while ago, did I miss anything?

config/crds/migration.openshift.io_migplans.yaml
pranavgaikwad: @awels okay so I taking your word for it on this. I don't know/remember enough to say 100% that this will work. But if you can test an upgrade scenario from a migration plan that has this field to a newer version...and it works, I am good with the change.

awels: Well there are no migration plans with this field since it is not stored at all. That is why it is not a problem.

pranavgaikwad (approved):

dymurray (approved):
LGTM

PR #1379, opened 2024-07-24
Bug 2299628: Fix watch error if kubevirt is not installed.

awels:
The watch error is caused by the fact that the
VirtualMachine CRD is not installed in the cluster. This is a common scenario when the kubevirt
operator is not installed. This change adds a
check to see if the CRD is installed before
starting the watch.


dymurray (approved):
LGTM

PR #1381, opened 2024-08-01
Don't create DVM resource if no PVCs are present

awels:
in the migration plan and IndirectVolumeMigration is set to false.

Fixed `hasDirectVolumes` to properly return false when the list is empty. It was returning a pointer to the list before, and the list object was never nil and thus would never return false. Since slices are pointers already there was no need for a pointer to the slice.


dymurray (approved):
LGTM

PR #1382, opened 2024-08-01
[Release 1.8] Don't create DVM resource if no PVCs are present

awels:
in the migration plan and IndirectVolumeMigration is set to false.


rayfordj (approved):
clean pick of https://github.com/migtools/mig-controller/pull/1381/

/lgtm

PR #1383, opened 2024-08-02
Add storage live migration support to direct DirectVolumeMigration

awels:
Add prometheus progress reporting to VM Live Migration

Validate the plan if the live migration checkbox is set, ensure that the source and destination cluster have a valid kubevirt version and configuration.

Allow canceling of migration.

This includes aborting live migrations and
cleaning up rsync pods if they are still running.
Added a finalizer to the DirectVolumeMigration
which is only removed once everything is cleaned
up properly.

Properly handle rollback of direct volume migrations that include VMs. For both running and stopped VMs as well as live migration of running VMs, or quiesce and starting them after migration.


awels:
/hold
Need to update the crane-lib after a new release is made when https://github.com/migtools/crane-lib/pull/125 is merged

pranavgaikwad (approved):

PR #1393, opened 2024-10-01
Instead of -new add -mig-xxxx to the end of the new pv names

awels:
In order to support multiple migrations of the same PV we need to add a unique suffix to the end of the PV name when creating new PVs. The new suffix is '-mig-' + 4 character random string of letters. When migrating a PV that already has the suffix, we will remove the suffix and create a new one.

In order to support legacy plans and pvs if the suffix is -new we replace -new with -mig-xxxx for the new pv name.

relaxed the validation of pvs to only include pvs that are not being skipped in the plan. This allows for the original pv to still exist in the namespace.

Goes with https://github.com/migtools/mig-ui/pull/1525 which modifies the UI to properly allow users to select the PVs and get feedback if their selection is valid or not.

### DIFF
--- a/pkg/controller/migplan/pvlist.go
+++ b/pkg/controller/migplan/pvlist.go
@@ -20,6 +20,10 @@ import (
 type PvMap map[k8sclient.ObjectKey]core.PersistentVolume
 type Claims []migapi.PVC
 
+var (
+	suffixMatcher = regexp.MustCompile("(.*)-mig-([[:alpha:]]{4})$")
+)
+
 // Update the PVs listed on the plan.
 func (r *ReconcileMigPlan) updatePvs(ctx context.Context, plan *migapi.MigPlan) error {
 	if opentracing.SpanFromContext(ctx) != nil {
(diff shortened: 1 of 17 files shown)


pkg/controller/migplan/pvlist.go
```diff
@@ -248,15 +258,17 @@ func getMappedNameForPVC(pvcRef *core.ObjectReference, podList []core.Pod, plan
 	// if plan is for storage conversion, create a new unique name for the destination PVC
 	if isStorageConversionPlan(plan) &&
 		(existingPVC == nil || existingPVC.GetSourceName() == existingPVC.GetTargetName()) {
-		// we cannot append a prefix when pvcName itself is 252 characters or more because:
+		// we cannot append a suffix when pvcName itself is 247 characters or more because:
 		// 1. total length of new pvc name cannot exceed 253 characters
-		// 2. we append a '-' char before prefix and pvc names cannot end with '-'
-		if len(pvcName) > 251 {
+		// 2. we append '-mig-' char before prefix and pvc names cannot end with '-'
+		if len(pvcName) > 247 {
```
pranavgaikwad, line 264: number 247 doesn't sound right, shouldn't it consider the variable length `plan.GetSuffix()`?

pranavgaikwad (changes requested):

pkg/controller/migplan/pvlist.go
awels: 247 ensures we get `mig-x` at worst, this is the same as before where we could get `-n` instead of `-new` if we had 251 characters.

awels: So before we added the suffix `-new`, and now the suffix is `-mig-xxxx` with xxxx being a random 4 characters. If you check the unit tests, I ensure there that we are not creating names that end with a `-`. The 247 is from checking with the unit tests that nothing illegal is generated.

pranavgaikwad (approved):

PR #1398, opened 2024-12-16
Fix quiesce not working for selected resources.

awels:
The quiesce function assumed that all resources
in a namespace would be quiesced. This is no longer true with the ability to select volumes in the UI.

This changes fixes that so only the resources that are associated with the selected volumes are quiesced.


dymurray (approved):
Tested this on friday and worked as expected

PR #1406, opened 2025-03-05
Added some new uses of VirtualMachines causing errors if virtual machine CRDs were not installed

awels:
This fixes it so that it ignores the no match for kind error and continues on.


dymurray (approved):

awels:
/hold
Still verifying this actually fixes the issue.

awels:
/hold cancel
I verified this solves the issue.

PR #1407, opened 2025-03-19
When only one migration plan exists, quiesce all the unmapped resource in the namespace

awels:
When only one migration plan exists, quiesce all the unmapped resource in the namespace as well as ones mapped to a PV.

This is to keep existing behavior of quiescing all resources in namespace when only one plan could exist per namespace.


rayfordj (approved):
visack
/lgtm

PR #1408, opened 2025-03-25
Don't error validation if kubevirt is not installed

awels:
By not checking and checking the live migration
check box in the plan, it would error out making
the plan not ready. This would only happen if kubevirt was not installed.

This PR fixes this by checking if the error is that the CRD does not exist and it will return nil instead of erroring out.


rayfordj (approved):
/lgtm

PR #1409, opened 2025-03-26
Kubevirt 1.5 and newer GAed live update features

awels:
In the plan validation we check if the kubevirt version supports storage live migration. In kubevirt 1.5 one of the feature gates was made GA, and is no longer set. This skips the check for that feature gate if kubevirt is 1.5 or newer.


rayfordj (approved):
/lgtm

PR #1410, opened 2025-03-26
Mark rsync failures as fatal errors during migration

awels:
rsync failures were not marked fatal, and not fail the migration process. Mark rsync failures as fatal which will properly indicate that the migration failed


rayfordj (approved):
/lgtm

PR #1411, opened 2025-04-01
Properly count quiesced VM pods

awels:
If multiple VMs exist in a namespace and there are multiple plans for those VMs. Running a cutover for one plan would wait for all the VMs to be quiesced, which would not happen. This fixes the check that determines which VMs should be quiesced and which ones we are waiting on.


rayfordj (approved):
/lgtm

PR #1412, opened 2025-04-01
Mark new PVs in the plan as skip by default instead of copy

awels:
When new PVs appear in the namespace any plans will automatically add them to the plan. But if the action is marked as copy it will create conflicts between plans

This defaults the action to skip instead so no conflicts happen.


rayfordj (approved):
/lgtm

PR #1414, opened 2025-04-02
Revert allow multiple plans per namespace

awels:
There are too many assumptions about a single namespace across the project. Allowing multiple
plans per namespace is generating many issues.


pranavgaikwad (approved):
LGTM, thank you!

PR #1445, opened 2025-07-15
Clean up completed virt-launcher pods after migration completes.

awels:
Fixes: https://issues.redhat.com/browse/MIG-1749

Apparently if there completed virt-launchers, deleting the migration plan will cause a rollback to happen.

### DIFF
--- a/pkg/controller/directvolumemigration/vm.go
+++ b/pkg/controller/directvolumemigration/vm.go
@@ -33,11 +33,13 @@ import (
 )
 
 const (
-	prometheusURLKey = "PROMETHEUS_URL"
-	prometheusRoute  = "prometheus-k8s"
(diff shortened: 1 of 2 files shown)


pkg/controller/directvolumemigration/vm.go
```diff
@@ -341,6 +343,32 @@ func (t *Task) verifyVMs() error {
 	return nil
 }
 
+func (t *Task) cleanupStaleVirtHandlerPods() error {
```
jpeimer, line 346:
```suggestion
func (t *Task) cleanupStaleVirtLauncherPods() error {
```

jpeimer (commented):
Thanks Alexander! 
Just a nit comment

PR #1448, opened 2025-07-16
MIG-1731 Fix limitrange panics with nil maps

dymurray:


weshayutin:
https://issues.redhat.com/browse/MIG-1731

PR #1456, opened 2025-10-19
Validate "auto" access/volume mode input string

akalenyu:
We've had consumers of the API trip over capitalization,
let's validate the options we support explicitly.


awels (approved):

awels:
@rayfordj could you approve?

PR #1458, opened 2026-01-13
Validate "auto" access/volume mode input string (#1456)

awels:
We've had consumers of the API trip over capitalization, let's validate the options we support explicitly.


rayfordj:
clean pick of PR to release branch
/lgtm
/approve

PR #1459, opened 2026-04-16
RFE-8920: Reject internal .svc S3 endpoints in MigStorage

asadawar:
## Summary

Hi, I raised RFE-8920 after working a customer case where a cross-cluster MTC migration failed with misleading S3 errors. The root cause was that the replication repository was configured with an internal `.svc` S3 endpoint (from an OBC), which isn't reachable from the source cluster.

This PR adds early validation to catch this misconfiguration before the migration starts, instead of letting it fail at runtime with confusing errors.

## What this PR does

- In `pkg/cloudprovider/aws.go`: after the existing URL scheme validation, checks if the `S3URL` or `PublicURL` hostname ends with `.svc`. If so, marks the field as `S3URL-InternalEndpoint` or `PublicURL-InternalEndpoint`.
- In `pkg/controller/migstorage/validation.go`: separates internal endpoint errors from generic field errors and sets a dedicated `InternalEndpointUsed` condition with a clear message: *"The S3 endpoint uses an internal .svc hostname that is not accessible across clusters. Use an externally reachable endpoint (e.g. an Object Gateway route) for cross-cluster migrations."*
- Added unit tests covering internal `.svc` endpoints (rejected) and external endpoints (allowed).

## Why

The customer was using ODF Ceph RGW and the OBC automatically generated an internal `.svc` hostname as the S3 endpoint. In a cross-cluster migration, the source cluster can't resolve this. The resulting errors were misleading — S3 authentication failures and registry CrashLoopBackOff — which led to a long troubleshooting cycle before identifying the actual problem.

## Test plan

- [x] `go test ./pkg/cloudprovider/` — all existing + new tests pass
- [ ] Configure MigStorage with a `.svc` S3 endpoint and verify the condition is set
- [ ] Configure MigStorage with an external route endpoint and verify it passes validation

See: https://issues.redhat.com/browse/RFE-8920

## Summary by CodeRabbit

* **New Features**
  * Detects and rejects internal `.svc`/`.svc.cluster.local` endpoints in cloud storage configs and provides clearer guidance to use cross-cluster/public endpoints.

* **User-facing behavior**
  * Validation now surfaces separate, specific messages when internal endpoints are found versus other credential/config errors.

* **Tests**
  * Added tests covering detection and reporting of internal endpoint validation.

### DIFF
--- a/pkg/cloudprovider/aws.go
+++ b/pkg/cloudprovider/aws.go
@@ -251,7 +251,7 @@ func (p *AWSProvider) Validate(secret *kapi.Secret) []string {
 			u, err := url.Parse(p.S3URL)
 			if err != nil || (u.Scheme != "http" && u.Scheme != "https") {
 				fields = append(fields, "S3URL")
-			} else if strings.HasSuffix(u.Hostname(), ".svc") || strings.Contains(u.Hostname(), ".svc.") {
+			} else if strings.HasSuffix(u.Hostname(), ".svc") {
 				fields = append(fields, "S3URL-InternalEndpoint")
 			}
 		}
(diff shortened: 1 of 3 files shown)


asadawar:
Updated based on the Slack discussion with @pranavgaikwad. Simplified the check to use only `HasSuffix(".svc")` and dropped the `Contains(".svc.")` check. Removed the  `.svc.cluster.local` test cases accordingly.

awels (approved):

dymurray (approved):

pranavgaikwad (approved):

